PR #1, opened 2015-03-22
file system problem

alonewolfx2:
if we test some other project like meteocontrol and upload http server bootstrap again , sometimes the other files missing. so we must to check all files.

### DIFF
--- a/HttpServer_Bootstrap/app/application.cpp
+++ b/HttpServer_Bootstrap/app/application.cpp
@@ -88,11 +88,21 @@ void connectOk()
 {
 	Serial.println("I'm CONNECTED");
 
-	if (!fileExist("jquery.js.gz"))
+	if (!fileExist("index.html"))
 	{
 		// Download server content at first
 		downloadTimer.initializeMs(3000, downloadContentFiles).start();
 	}
+	else if (!fileExist("bootstrap.css.gz"))
+		{
+			// Download server content at first
+			downloadTimer.initializeMs(3000, downloadContentFiles).start();
+		}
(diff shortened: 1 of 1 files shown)


anakod:
Greetings! Very glad to your participation.
I have some questions:
1) In start procedure i was checking only last file, so if it was exist, all other files should be available too. It's not enough?
2) Can you explain details of bug what you found?
3) May be, if we check all files and always start one action, will be better to combine all this cases to one branch?

alonewolfx2:
you are right. 
let me explain. i flashed httpbootstrap example and esp8266 download all files with first boot. everything ok than i flashed blink examples and than i flashed meteocontrol examples. my dsl connection down and esp8266 fail. and than i flashed httpbootstrap again. i tried webserver and tadaa, 404 not found. serial says index.html notfound. i flashed again. serial says webserver started but i am getting 404 notfound on chrome and serial says index.html notfound. module checked jquery file and jquery file exist. webserver started without index.html

anakod:
Agree, thank you!

anakod:
We should make check like this in HttpServer_Bootstrap also, yes?

alonewolfx2:
agree. and maybe can add file.list api like nodemcu.

anakod:
Yes, it should be good idea! Not sure is it implemented in spiffs?

alonewolfx2:
i dont know but is it implemented in spiffs but nodemcu using same spiffs and it can list files :)

anakod:
nodemcu using fork of original library, but it shouldn't be a problem.

PR #6, opened 2015-03-25
+ Cpu frequency get and set

alonewolfx2:
cpuUpdate(uint8 freq) and getCpuFreq() functions added. freq must be 80
or 160


alonewolfx2:
you can use something like that.

```
//update Cpu freq. to 160MHZ
	System.cpuUpdate(160);
	Serial.print("New cpu freq. is:");
	Serial.println(System.getCpuFreq());
```

PR #11, opened 2015-03-27
Semi support for esp-open-sdk. The path for esp-open-sdk is HARD-CODED, ...

ystrem:
...you have to change it!


anakod:
It will break Windows compatibility, so it's not solution of problem. I'm especially using only constants for easy changing any paths and easy movng to Linux. If you create makefile what can work on both system or two files for windows and linux it will be great.

ystrem:
Ok, I'll try.

PR #19, opened 2015-04-08
Updates in Ultrasonic lib

ploys:


anakod:
Different sample should not be changed (blink, ajax). Please resend this pull-request. Thank you!

PR #53, opened 2015-05-03
Update Timer to use Delegate

hreintke:
LS,
Updated Timer which uses the new Delegate.
Based also on the Timer updates from JustACatRu.
Please comment


anakod:
I'm should think about optimisation details on this pull-request. Will be beter to have special case for standard functions calls (for high frequiency timers).

hreintke:
OK, I will update the PR. Also do some timing tests so we know what are the differences

PR #54, opened 2015-05-03
added README.md to BMP180 folder with instructions.

caquino:
added README.md to BMP180 folder with instructions.

This follows the same idea as the Liquidcrystal PR, it aims to provide more information to the beginners.

Thanks


anakod:
> Use `make` and `make flash` to build and flash the firmware to the ESP8266 board.
I think it isn't right way for Windows users

PR #57, opened 2015-05-11
Add block read to HardwareSerial

AlmightyFrog:
Feature to read multiple bytes from Serial buffer more efficient.

### DIFF
--- a/Sming/SmingCore/HardwareSerial.cpp
+++ b/Sming/SmingCore/HardwareSerial.cpp
@@ -100,6 +100,27 @@ int HardwareSerial::read()
 	return res;
 }
 
+int HardwareSerial::readBlock(uint8_t* buf, int max_len) {
+	RcvMsgBuff &rxBuf = UartDev.rcv_buff;
+	int num = 0;
+
+	noInterrupts();
+	if (rxBuf.pWritePos != rxBuf.pReadPos) {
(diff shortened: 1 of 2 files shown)


anakod:
Can you discover details please. Read direct to memory, yes?

AlmightyFrog:
It's pretty much a combination of how available() and read() uses rxbuf but optimized to use as little overhead as possible for big data.
Looking at it again, there might be slight optimisation, please comment:
- HardwareSerial.cpp:108 is futile as in following line checked again as first condition
- HardwareSerial.cpp:117 is missing brackets as it was c&p from read()
- HardwareSerial.cpp:110, 112, 115 could be comined if requested but should be without effect

I did use this block read trying to make a real transparent UART bridge but failed there getting UDP packets in time - every few seconds lwip just didn't spit out packets anymore (though they were transmitted in time).

anakod:
OK I will look all this points. Next, sorry, I'm not understand, with new method transparent bridge work correctly? Of course String make life easer, but cost additional resources.

AlmightyFrog:
Forget about the point with transparent bridge - i just did want to mention for what i did use this function.

Using String would be a completely different usecase. Idea here is to get as low level as possible the raw data. String would make sense to me only when writing a function like read a complete line or so - but then you have to think about buffer strategy as you can't tell how long a line could be and on embedded systems  resourcs are limited of course.

PR #62, opened 2015-05-14
Replace ${ESP_HOME}/sdk/include with ${ESP_HOME}/ESP8266_SDK/include

ozayturay:
Replace ${ESP_HOME}/sdk/include with ${ESP_HOME}/ESP8266_SDK/include in .cproject files to prevent warnings in Eclipse


anakod:
As I know, it's a problem, because in Open SDK (linux, mac) this pacth are different.
May be we can add both paths without conflicts?

ozayturay:
Isn't .cproject file Eclipse specific?

anakod:
No, it is the same for all platforms.

ozayturay:
Does Linux and Mac guys use Eclipse too, or do they prefer command line?

Then the solution is renaming or symlinking ESP8266_SDK to sdk in C:\Espressif right?

anakod:
Yes, them use.
May be we can just add both paths? Will it provide errors?

anakod:
Also please update, new project was added :)

ozayturay:
I'll try to add both anc check if Eclipse gives warnings, but i guess it will. I'll let you know. :)

ozayturay:
Putting both paths in .cproject file gives warnings. The reasonable solution is renaming ESP8266_SDK folder to sdk in C:\Espressif folder. It can break makefiles in Espressif/examples though.

anakod:
Which warnings it provide? Can we make some workaround?

kireevco:
When you install sdk via chocolatey it creates a symlink automatically. Same happens when you update it.

Issue #67, opened 2015-05-15
Timer issue, large interval


danielnilsson9:
Hi

Think i found an issue with the timers. Setting a long interval causes issues, for example 10 minutes, 10 * 60 * 1000 = 600000ms. Its not related to overflow, but that would happen if setting a longer interval than 71 minutes in the way the Timer class is implemented, intervalUs = intervalMs * 1000 should cause overflow in the uint32 if intervalMs > approx 71 minutes.

Anyway thats not the issue here:
If setting timer interval repeating 10 minutes the callback will be called twice in a row every approx 340 seconds, which is strange. Even if doing it manually with the ets_timer functions the same issue persist in Sming. Also the callback function gets called right away when using large interval, that doesn't happen otherwise.

BUT: if not using Sming framework (only using ESP SDK) and doing it with ets_timer functions there is no issue at all, callback called every 600 seconds as expected. (Same SDK version used)

Have been looking but cannot find the cause, using the exact same ets_timer code without using Sming works, but when using Sming it doesn't.

Any ideas? Seems like a tricky one... :(

anakod:
Is it on last version (yersterday\today changes)?

danielnilsson9:
Updated aprox 12 hours ago, haven't updated since then. It was when you merged my pull request about bug fixes in NtpClient. I will try.

anakod:
I have commited new changes now. It will be uint64_t to cover all cases.

anakod:
Miss one moment, fixed.

danielnilsson9:
Yes thats a good idea, will fix that 71 minute problem. But I don't think it will solve the other one which happens even if not using the Timer class.

If you run the SystemClock_NTP example and set update interval to 10 minutes you should set that it will request update every ~5 minutes with two callback each time   If its not something wrong with my board..

anakod:
I think I know difference - Sming using system_timer_reinit() at the start for PWM.

j0hncc:
I did some pretty thorough testing of different values.
I can't seem to find my notes!  
But my conclusion was very clear, there is a modulo going on something 450000-500000 msec. (Can't remember exact number though I did determine it).  
Eg  460000 would delay correct 7.6 min but the 465000 would delay only 5 seconds!
Eg 925000 would also delay 5 seconds.

danielnilsson9:
[...]

PR #71, opened 2015-05-16
Update NtpClient

hreintke:
Use UDPClient instead of UDPServer
Add fuctionality to :
- use timezone
- If no callback, set systemtime in ntpclient
- wait until wifi connected
- use both NTP_VERSION 4 and 3 answers

### DIFF
--- a/Sming/Services/DateTime/DateTime.h
+++ b/Sming/Services/DateTime/DateTime.h
@@ -52,6 +52,8 @@ public:
 	DateTime();
 	DateTime(time_t time);
 
+	operator time_t() { return toUnixTime(); }
+
 	void setTime(time_t time);
 	void setTime(int8_t sec, int8_t min, int8_t hour, int8_t day, int8_t month, int16_t year);
 	bool parseHttpDate(String httpDate);
--- a/Sming/SmingCore/Network/NtpClient.cpp
+++ b/Sming/SmingCore/Network/NtpClient.cpp
@@ -14,12 +14,15 @@ NtpClient::NtpClient()
 	requestTime();
 }
 
-NtpClient::NtpClient(String reqServer, int reqIntervalSeconds, int reqTimezone, NtpTimeResultCallback onTimeReceivedCb /* = NULL*/)
+NtpClient::NtpClient(String reqServer, int reqIntervalSeconds, NtpTimeResultCallback onTimeReceivedCb /* = NULL*/)
 {
 	autoUpdateTimer.initializeMs(NTP_DEFAULT_AUTO_UPDATE_INTERVAL, Delegate<void()>(&NtpClient::requestTime, this));
 	this->server = reqServer;
 	this->onCompleted = onTimeReceivedCb;
-	setTimezone(reqTimezone);
+	if (!onTimeReceivedCb)
+	{
+		autoUpdateSystemClock = true;
+	}
 	if (reqIntervalSeconds == 0)
 	{
 		setAutoQuery(false);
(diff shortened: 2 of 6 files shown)


anakod:
I think it's not very obviously what we will set system clock only if callback isn't set. Maybe add description or some other way to make it simpler?

hreintke:
I understand your remark, by description you mean documentation ?
The ratio behind only when callback is not set is 
-  if a user sets callback he/she want control on what is happening so only callback
- if no callback, "normal behaviour" of ntp is to keep time in sync -> so set systemtime

danielnilsson9:
Maybe it would be more clear if having a method on the SystemClock instead, perhaps "setAutoSyncWithNtp(bool enable)" or something which would set up auto syncing using the NTP class? Then the user wouldn't even have to create the NtpClient to have autosyncing time. And maybe the timezone method could be moved there aswell then. Just a thought.

hreintke:
@danielnilsson9 : My idea was indeed to make the usage as simple as possible. 
(now implemented) Just instantiate ntpclient with your preferences on server, freq and TZ and your done keeping time in sync. Moving (part of ) that to SystemClock is definitely an option

anakod:
1. Move timezone information in clock - great idea. Like in .NET we have two properties: Now and UtcNow. Also that mehods should return DateTime, not uint.
2. I think that the methods should be assigned to who carries out the action, so may be add to ntpclient  method setAutoUpdateSystemClock(bool) ? Ans set it by default depending on the callbac parameters and etc.

danielnilsson9:
@anakod Sound good to me!

@hreintke While you are at it working with these classes:
- The NtpClient should probably also be updated with those nice Delegate callbacks. 
- A workaround for issue #67 would be easy to implement, with a e.g. a callback every 60 seconds and then a counter to se if it is time to update.

hreintke:
OK, Will work on it and come with an updated proposal

hreintke:
@anakod : Are you sure you want now() and the new utcnow() to return DateTime ?
As far as I understand, all arduino/esp use the time_t = uint as return value.
Of course we can add methods like nowdt and utcnowdt (or any name you prefer) to provide DateTime results but then we keep sming compatible to arduino/esp

tprochazka:
I think that now() should always return time as time_t in the UTC, as as seconds since Jan 1 1970
But possibility to get local time as DateTime should be very ussefull.
Arduino has also great library which can handle automatically timezone change for summer and winter time: https://github.com/JChristensen/Timezone which contain Timezone classwith method toLocal() which is able to convert utc now() to local time.

hreintke:
@anakod : made updates in the way you suggested (I think).
Needs some updates and cleaning of code but shows the direction/options.
Is this the way to go ?
@tprochazka : I had seen the TimeZone lib. First want to make a simple version before (possible) go for a fullblown version.

anakod:
> I think that now() should always return time as time_t in the UTC
1. Why not DateTime object? (it's more flexible)
2. Why in UTC?

I'like .NET architecture, but of course, it has many details related to that functionality.

hreintke:
@anakod : 
DateTime vs Time_t : In arduino everyone is using the fucntion now() which there is returning time_t.
If you want compatibility with that you need to return time_t and make additional other options.
In my latest update I user uint32_t as you did.
UTC : I was wrong in that. in the latest update now() returns localtime and I added the option to call now(bool reqUtc). so both options are available

anakod:
Converting DateTime to time_t can be automatic (implicit convesrsion).

hreintke:
OK, have done some initial reading on implicit conversion.
Am I correct that adding something like "operator int() { return toUnixTime(); }"' to the public part of DateTime class is the way you are thinking ?
If so I check/test further with that

anakod:
Yes, it is "operator time_t()"

hreintke:
OK, will work in that direction.

anakod:
Also, as example of implicit conversion, you can look IPAddress class too.

hreintke:
Next version of ntpclient, systemclock and datetime.
Is this the direction you envisioned ?
Can you have a look at the systemclock setNtpSync ?
I used two options in application.cpp (identical in my view) 
	SystemClock.setNtpSync("fritz.box",30);
//	nt = new NtpClient("fritz.box",30);
both use the same ntpclient way of working but I get system crashes using the systemclock implementation

anakod:
I think you just fogot initialize "ntpClient" clock class property, and it isn't null by default (and it trying to delete wrong pointer).

Also, two version with delegates and methods callbacks isn't needed here (we provide two callback ways in Timer class for high frequency timers and perfomance), but it isn't problem at all, just case.

Thank you for great work! You will fix setNtpSync method problem or I should?

hreintke:
@anakod : I think ntpClient is initialized by : 	NtpClient* ntpClient = nullptr; in SystemClock.h 
If you can have a look at setNtpSync would be great, I am stuck and out of ideas at the moment.
Otherwise I leave out the systemclock sync option and only implement the ntpclient version, but am a little reluctant to that because we might have a undiscovered issue in that.

For the delegated and callback both implemented : I agree on just having delegates.
I just wanted to create a constructor like : 
NtpClient(String reqServer, int reqIntervalSeconds, NtpTimeResultCallback onTimeReceivedCb)
 : NtpClient(reqServer, reqIntervalSeconds, Delegate<void(time_t ntptime)> (onTimeReceivedCb))
and only use delegates in NtpClient but didn't find the right syntax yet (lack of c++ knowledge from my side), same is for a nice typedef for delegates.
So decided for this intermediate version to keep both

tprochazka:
Only example is now completely confusing.

PR #81, opened 2015-05-22
NTP example cleanup

tprochazka:
I tried to do some cleanup in NTP example. If I understand it correctly.

But I found weird issue.
After 60s when SystemClock should repeat NTP sync request it crash:

```
NtpClient request Time
UDP connect to 81.0.235.220:123
NtpClient request sent
Fatal exception (28):
epc1=0x40002d4e, epc2=0x00000000, epc3=0x00000000, excvaddr=0x13433cb8, depc=0x00000000
rm match
del if0
Fatal exception (28):
epc1=0x40002d4e, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00008000, depc=0x00000000
bcn 0
del if1
usl
sul 0 0
Fatal exception (28):
epc1=0x40002d4e, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00008000, depc=0x00000000

 ets Jan  8 2013,rst cause:4, boot mode:(1,7)

wdt reset
```

### DIFF
--- a/SystemClock_NTP/app/application.cpp
+++ b/SystemClock_NTP/app/application.cpp
@@ -4,15 +4,9 @@
 
 // Put you SSID and Password here
 #define WIFI_SSID "ssid"
-#define WIFI_PWD "pw"
-
-
-void onNtpReceive(NtpClient& client, time_t timestamp);
+#define WIFI_PWD "password"
 
 Timer printTimer;
-//NtpClient ntpClient(onNtpReceive);
-//NtpClient ntpc("fritz.box",30, onNtpReceive);
-
 
 void onPrintSystemTime() {
 	Serial.print("Time    : ");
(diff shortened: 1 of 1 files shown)


hreintke:
@tprochazka : I also experienced this crash. Only with my config I get fatal exception 9 instead of 28.
Did not yet find a solution for it, that's why I asked anakod to have a look at it.
If you know a direction for the solution, let me know and i'll give it a try.
Also created PR which only uses current working options of NtpClient.
@anakod : Maybe we should have a dev branch where intermediate code like this can be placed on worked on.

tprochazka:
I'm relatively beginner in this topic (I'm primary Android Developer).
I found only this http://www.esp8266.com/viewtopic.php?f=13&t=2379&sid=cfec3da676d74d453d762ba6440aecfe&start=10 yet.
Like small clue.

dev branch is good idea (or feature branch).

tprochazka:
I think that this is obsolete already

PR #82, opened 2015-05-23
Update SystemClock_NTP example

hreintke:
Only uses currently working NtpClient functionality

### DIFF
--- a/Sming/SmingCore/Network/NtpClient.cpp
+++ b/Sming/SmingCore/Network/NtpClient.cpp
@@ -14,8 +14,8 @@ NtpClient::NtpClient()
 	requestTime();
 }
 
-NtpClient::NtpClient(String reqServer, int reqIntervalSeconds, NtpTimeResultCallback onTimeReceivedCb /* = NULL*/)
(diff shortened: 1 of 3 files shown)


hreintke:
@ anakod : I try to get the ntptime delegate callback working.Used HTTP_Server (and others) as example. See the latest commit for the results.
However, when compiling I get the following error : 
```
app/application.cpp: In constructor 'ntpClientDemo::ntpClientDemo()':
app/application.cpp:14:71: error: no matching function for call to 'NtpClient::NtpClient(const char [13], int, NtpClientResultCallback (ntpClientDemo::*)(NtpClient&, time_t))'
     ntpcp = new NtpClient("my_ntpserver",30, &ntpClientDemo::ntpResult);
```
I also called with input a string but that gave the same results.
Can you point me in the right direction to solve this ?

anakod:
You should declare it in full form as Delegate<XXX>(YYY) when you trying to provide class method as callback.

hreintke:
I declared it as a typedef : 
```
typedef Delegate<bool(NtpClient& client, time_t ntptime)> NtpClientResultCallback;
```
Is that correct or should it be : 
```
typedef Delegate<bool(time_t ntptime)> NtpClientResultCallback;
```

PR #83, opened 2015-05-24
Delegates in Interrupts

sfam:
Added the ability to use delegate functions in interrupts.


anakod:
I have merged it manually now (git conflict), thank you.

PR #85, opened 2015-05-25
Add POST support

dmarkey:
Pretty simple.

### DIFF
--- a/Sming/SmingCore/Network/HttpClient.cpp
+++ b/Sming/SmingCore/Network/HttpClient.cpp
@@ -12,6 +12,8 @@
 HttpClient::HttpClient(bool autoDestruct /* = false */) : TcpClient(autoDestruct)
 {
 	waitParse = false;
+	body = "";
+	content_type = "";
 	reset();
 }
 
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -64,11 +66,43 @@ bool HttpClient::startDownload(URL uri, HttpClientMode mode, HttpClientCompleted
...
+void HttpClient::setContentType(String _content_type){
+    content_type = _content_type;
+}
+
+String HttpClient::getContentType(){
+    return content_type;
+}
+
+void HttpClient::setPostBody(String _body){
+    if (content_type == NULL){
```
anakod: Is it correct? It isn't pointer.
Also I think we can initialize it by default. 

C++11 should allow in-place class fileds initialization.

Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -64,11 +66,43 @@ bool HttpClient::startDownload(URL uri, HttpClientMode mode, HttpClientCompleted
...
+        content_type = "application/x-www-form-urlencoded";
```
anakod: We have special ContentType namespace, what can be used instead of this string: ```ContentType::SOMETHING```

Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -64,11 +66,43 @@ bool HttpClient::startDownload(URL uri, HttpClientMode mode, HttpClientCompleted
...
 	connect(uri.Host, uri.Port);
-	sendString("GET " + uri.getPathWithQuery() + " HTTP/1.0\r\nHost: " + uri.Host + "\r\n\r\n");
+	if (body.length() == 0){ //Not a POST
+        sendString("GET " + uri.getPathWithQuery() + " HTTP/1.0\r\nHost: " + uri.Host + "\r\n\r\n");
+	}
+	else{
+        sendString("POST " + uri.getPathWithQuery() + " HTTP/1.0\r\nHost: " + uri.Host + "\r\n");
+        sendString("Content-Type: " + content_type + "\r\n");
+        String content_length = "";
+        content_length += body.length();
```
anakod: Why "+="?

anakod:
Thank you very much. It very useful, but need small corrections.

dmarkey:
I think I covered everything. Thanks for the review, i'm a Python developer at heart.

Issue #86, opened 2015-05-25
Compilation issues with SDK 1.1


slaff:
During compilation with the latest version of the esp-open-sdk (https://github.com/pfalcon/esp-open-sdk/) using Espressif SDK 1.1 I got the following error:

```bash
C+ SmingCore/Digital.cpp
SmingCore/Digital.cpp: In function 'void pullup(uint16_t)':
SmingCore/Digital.cpp:78:43: error: 'PIN_PULLDWN_DIS' was not declared in this scope
  PIN_PULLDWN_DIS((EspDigitalPins[pin].mux));
                                           ^
SmingCore/Digital.cpp: In function 'void pulldown(uint16_t)':
SmingCore/Digital.cpp:86:42: error: 'PIN_PULLDWN_EN' was not declared in this scope
  PIN_PULLDWN_EN((EspDigitalPins[pin].mux));
                                          ^
SmingCore/Digital.cpp: In function 'void noPullup(uint16_t)':
SmingCore/Digital.cpp:92:43: error: 'PIN_PULLDWN_DIS' was not declared in this scope
  PIN_PULLDWN_DIS((EspDigitalPins[pin].mux));
                                           ^
make: *** [out/build/SmingCore/Digital.o] Error 1
```

slaff:
That link here can point out the issue: http://bbs.espressif.com/viewtopic.php?f=7&t=494

anakod:
Please check it.

slaff:
Works like a charm :)

PR #87, opened 2015-05-26
Added initial support for SDK 1.1.

slaff:
See issue: #86.

### DIFF
--- a/Sming/SmingCore/Platform/System.cpp
+++ b/Sming/SmingCore/Platform/System.cpp
@@ -51,7 +51,7 @@ void SystemClass::setCpuFrequency(CpuFrequency freq)
 	else
 		REG_CLR_BIT(0x3ff00014, BIT(0));
 
-	os_update_cpu_frequency(freq);
+	ets_update_cpu_frequency(freq);
 }
 
 CpuFrequency SystemClass::getCpuFrequency()
--- a/Sming/compiler/ld/eagle.app.v6.cpp.ld
+++ b/Sming/compiler/ld/eagle.app.v6.cpp.ld
@@ -20,6 +20,11 @@ PHDRS
 
 /*  Default entry point:  */
 ENTRY(call_user_start)
+EXTERN(_DebugExceptionVector)
+EXTERN(_DoubleExceptionVector)
+EXTERN(_KernelExceptionVector)
+EXTERN(_NMIExceptionVector)
+EXTERN(_UserExceptionVector)
 PROVIDE(_memmap_vecbase_reset = 0x40000000);
 /* Various memory-map dependent cache attribute settings: */
 _memmap_cacheattr_wb_base = 0x00000110;
@@ -163,6 +168,30 @@ SECTIONS
   {
     _stext = .;
     _text_start = ABSOLUTE(.);
+     *(.UserEnter.text)
+    . = ALIGN(16);
+    *(.DebugExceptionVector.text)
+    . = ALIGN(16);
+    *(.NMIExceptionVector.text)
+    . = ALIGN(16);
+    *(.KernelExceptionVector.text)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    . = ALIGN(16);
+    *(.UserExceptionVector.text)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    . = ALIGN(16);
+    *(.DoubleExceptionVector.text)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    LONG(0)
+    . = ALIGN (16);
     *(.entry.text)
     *(.init.literal)
     *(.init)
--- a/Sming/Libraries/WS2812/WS2812.cpp
+++ b/Sming/Libraries/WS2812/WS2812.cpp
@@ -43,7 +43,7 @@ int ICACHE_FLASH_ATTR ws2812_writergb(uint8_t gpio, char *buffer, size_t length)
     os_delay_us(1);
 
     // Send the buffer:
-    os_intr_lock();
+    ets_intr_lock();
     const char * const end = buffer + length;
     while (buffer != end) {
         uint8_t mask = 0x80;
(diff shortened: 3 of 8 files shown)


anakod:
> PERIPHS_IO_MUX_PULLUP2
We just remove the PULLDOWN definition because our pins do not support inside-pulldown function now.
Functionally, It's the same as PERIPHS_IO_MUX_PULLUP.

I think we should remove all references to pull-down in Sming because previous API from Espressif was wrong, and pulldown doesn't suported

kzyapkov:
For me, building any of the examples with SDK 1.1 results in a stream of:

    Fatal exception (0): 
    epc1=0x40105ad3, epc2=0x00000000, epc3=0x00000000, excvaddr=0x00000000, depc=0x00000000

Everything works fine with 1.0.1.

Thanks for the great framework!

slaff:
@anakod Unfortunately I get the same result as @kzyapkov :(. Can it be that the ld script needs updating Similar to https://github.com/esp8266/Arduino/commit/92147999b997e7e167ab0365cc98d1f2e99824c7 ? This is just a blind guess.

dmarkey:
Seems probably, anyone know how to generate a new one?

dmarkey:
I applied it by hand, seems to work for me:

https://github.com/dmarkey/Sming/commit/8e74effdd397d9c3677cab240306ea4eeac063f2

slaff:
@dmarkey I will check tonight on my Olimex ESP. It would be great if it works :)

dmarkey:
@kzyapkov if you could have a go that would be helpful thanks!

kzyapkov:
@dmarkey , yes, updating the ld script fixed it, tested on ESP-12 (nodemcu devkit) and an Olimex' module.

dmarkey:
Hmm is there some issue with this request?

Sming/SmingCore/Platform/System.cpp
```diff
@@ -51,7 +51,7 @@ void SystemClass::setCpuFrequency(CpuFrequency freq)
 	else
 		REG_CLR_BIT(0x3ff00014, BIT(0));
 
-	os_update_cpu_frequency(freq);
+	ets_update_cpu_frequency(freq);
```
anakod: Can you discover, why this change needed?

Sming/compiler/ld/eagle.app.v6.cpp.ld
```diff
@@ -20,6 +20,11 @@ PHDRS
 
 /*  Default entry point:  */
 ENTRY(call_user_start)
+EXTERN(_DebugExceptionVector)
+EXTERN(_DoubleExceptionVector)
+EXTERN(_KernelExceptionVector)
+EXTERN(_NMIExceptionVector)
+EXTERN(_UserExceptionVector)
```
anakod: Already merged, yes?

Sming/SmingCore/SdkCompat.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef os_intr_lock
+#define os_intr_lock ets_intr_lock
+#endif
+
+#ifndef os_intr_unlock
+#define os_intr_unlock ets_intr_unlock
+#endif
+
+#endif /* SDKCOMPAT_H_ */
```
anakod: I think this file should be removed, because pull-down isn't accessible in hardware (is it right, yes?). So we just should remove all relations to pull-down mode in code.
.

Sming/SmingCore/Platform/System.cpp
slaff: It seems that this function, along with others, was removed from osapi.h. An abbreviated version of the diff is given below. Lines starting with "-" mean that the code was removed and lines starting with "+" mean that this code was added. I guess we have to ask Espressif for the reasons behind these changes.

```diff
diff -u /opt/Espressif/esp_iot_sdk_v1.0.1/include/osapi.h /opt/Espressif/esp_iot_sdk_v1.1.0/include/osapi.h
--- /opt/Espressif/esp_iot_sdk_v1.0.1/include/osapi.h	2015-04-21 14:45:08.000000000 +0000
+++ /opt/Espressif/esp_iot_sdk_v1.1.0/include/osapi.h	2015-05-22 19:45:30.000000000 +0000
-#define os_install_putc2 ets_install_putc2
-#define os_intr_lock ets_intr_lock
-#define os_intr_unlock ets_intr_unlock
-#define os_isr_attach ets_isr_attach
-#define os_isr_mask ets_isr_mask
-#define os_isr_unmask ets_isr_unmask

-#define os_putc ets_putc
-#define os_str2macaddr ets_str2macaddr
-#define os_timer_done ets_timer_done
-#define os_timer_handler_isr ets_timer_handler_isr
-#define os_timer_init ets_timer_init
-#define os_update_cpu_frequency ets_update_cpu_frequency
+unsigned long os_random(void);
+int os_get_random(unsigned char *buf, size_t len);
```

Sming/compiler/ld/eagle.app.v6.cpp.ld
slaff: Yes, that's correct.

Sming/SmingCore/SdkCompat.h
slaff: > I think this file should be removed, because pull-down isn't accessible in hardware (is it right, yes?). 

According to Espressif the hardware does not support this. I would say that we should listen to them and do as you suggested - remove all occurences of PULL_DOWN. 

As for the os_intr_unclock and os_intra_lock if you want I can replace them directly in the framework. 

And finally I can remove them SdkCompat.h because it won't be needed any longer.

What do you think? Comments from your side?

anakod: Solution:
1. Replace all what needed
2. Remove all what doesn't needed :)

slaff:
@anakod I rebased the changes. Please, take a look at them and check if they are ready for merging.

anakod:
Very big thanks for all! It working on previous SDK version without any problems too.

Issue #88, opened 2015-05-26
Be able to define extra libraries per application/project.


slaff:
We can have an application that is using the smarconfig library, for example. In order to use the existing Makefiles it would be great if we can define something like EXTRA_LIBS per application/project in the project/Makefile-user.mk.

anakod:
Example needed to close this issue.

PR #89, opened 2015-05-26
Added EXTRA_LIBS variable that can be defined per project to add addi…

slaff:
…tional libraries.

Related issue: #88.


anakod:
Thank you, very quickly problem solving :)
I think will be good add commented-out example of value for this variable in all *Makefile-user.mk* files
Example: https://github.com/anakod/Sming/blob/master/Basic_Blink/Makefile-user.mk

PR #90, opened 2015-05-26
Example for using EXTRA_LIBS.

slaff:
Added initial skeleton version of Basic_SmartConfig project that uses the additional "smartconfig" library.
See: https://github.com/slaff/Sming/blob/extra-libs/Basic_SmartConfig/Makefile-user.mk


slaff:
Related to issue #88.

anakod:
Thank you! Can you add some readme file or comments how to use it?

PR #91, opened 2015-05-26
Added working example of Delegate Callback

hreintke:
Updates SystemClock to use enums

### DIFF
--- a/Sming/SmingCore/Network/NtpClient.cpp
+++ b/Sming/SmingCore/Network/NtpClient.cpp
@@ -88,7 +88,6 @@ int NtpClient::resolveServer()
 
 void NtpClient::requestTime()
 {
-	debugf("NtpClient request Time");
 	if (!WifiStation.isConnected())
 	{
 		connectionTimer.initializeMs(1000, Delegate<void()>(&NtpClient::requestTime, this)).startOnce();
@@ -118,7 +117,6 @@ void NtpClient::requestTime()
 
 //	Send to server, serverAddress & port is set in connect
 	NtpClient::send((char*) packet, NTP_PACKET_SIZE);
-	debugf("NtpClient request sent");
 }
 
 void NtpClient::setNtpServer(String server)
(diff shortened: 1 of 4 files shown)


hreintke:
@anakod : 
Now with working example of delegate callback
Is there a specific reason you made the callback function returning bool ? If not, maybe we should change to void.
When using the ntpClientdemo class in the example I found that a declaration of 
ntpClientDemo dm1 = ntpClientDemo(); is working correctly but 
ntpClientDemo dm1();  is not working.
I expected both being the same : Do you have an explanation for that ?
Still working on getting the ntpClient internally ony working with delegate, no luck in that at the moment. 
Can you point into a direction I should update 
```NtpClient::NtpClient(String reqServer, int reqIntervalSeconds, NtpTimeResultCallback onTimeReceivedCb)
 : NtpClient(reqServer, reqIntervalSeconds, NtpClientResultCallback (onTimeReceivedCb))
```
in order to move static callback to delegate callback ?

anakod:
> Is there a specific reason you made the callback function returning bool ? If not, maybe we should change to void.

It isn't my code, but I think yes, it can be changed.

> ntpClientDemo dm1(); is not working.
It depended on constructor, but should work, I think.

hreintke:
'''
ntpClientDemo dm1 = ntpClientDemo(); is working correctly but
ntpClientDemo dm1(); is not working
'''
I expected in both options above the same constructor would be used. Is that correct ?

anakod:
```C++
ntpClientDemo &dm1 = ntpClientDemo();
ntpClientDemo dm1();
```
- this lines should be equals, I think. In your example - no.

hreintke:
If I use 
```
ntpClientDemo &dm1 = ntpClientDemo();
```
I get the compile time error 
```
app/application.cpp:56:36: error: invalid initialization of non-const reference of type 'ntpClientDemo&' from an rvalue of type 'ntpClientDemo'
 ntpClientDemo &dm1 = ntpClientDemo();
```
When using 
```
ntpClientDemo dm1();
```
It looks like no instantiation at all.
If you can think of something please let me know. Next steps from me on thursday

PR #92, opened 2015-05-26
Added more logic to the smartconfigDone callback.

slaff:


anakod:
Great. Can you provide information how to use it? :)

slaff:
@anakod Is that readme (https://github.com/slaff/Sming/commit/c16d4328f99f4d51503e1b5c519bd45c4e4e85e3) ok?

anakod:
Yes it is. One question: can we use this solution without Espressif mobile application?

anakod:
Thank you, @slaff

slaff:
> Yes it is. One question: can we use this solution without Espressif mobile application?

@anakod The answer is yes. We can but someone had to implement that code. The code can also be run on normal PC with network adapter that can act as Access Point. Here comes the question about the legal side. As far as I've heard the SmartConfig code contains parts that are used under strict proprietary licence agreement (I might be wrong here).  Plus it is not quite secure. 

Which brings me to the idea of having free/open source implementation of a smart config plus the possibility to use shared secret (just a long number or string) that only the device and the mobile phone know. If there is an interest I can describe that idea as a separate issue and will need the help of skilled C/C++ programmers to help me implement it :)

kireevco:
What's the protocol? Is it open?

On Thursday, May 28, 2015, slaff <notifications@github.com> wrote:

> Yes it is. One question: can we use this solution without Espressif mobile
> application?
>
> @anakod <https://github.com/anakod> The answer is yes. We can but someone
> had to implement that code. The code can also be run on normal PC with
> network adapter that can act as Access Point. Here comes the question about
> the legal side. As far as I've heard the SmartConfig code contains parts
> that are used under strict proprietary licence agreement (I might be wrong
> here). Plus it is not quite secure.
>
> Which brings me to the idea of having free/open source implementation of a
> smart config plus the possibility to use shared secret (just a long number
> or string) that only the device and the mobile phone know. If there is an
> interest I can describe that idea as a separate issue and will need the
> help of skilled C/C++ programmers to help me implement it :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/anakod/Sming/pull/92#issuecomment-106221240>.
>


-- 
--
Sincerely,
 Dmitry Kireev
 cel.: +1 (818) 850-5577
 email, gtalk: dmitry@kireev.co
<dmitry@kireev.co> skype: kireev.co
 aim: 336719228

PR #93, opened 2015-05-27
NtpClient internally only working with delegates/Cleanup

hreintke:
Cleanup NtpClient code
Remove setNtpSync from Systemclock (not working correctly)

### DIFF
--- a/Sming/SmingCore/Network/NtpClient.cpp
+++ b/Sming/SmingCore/Network/NtpClient.cpp
@@ -1,39 +1,18 @@
 #include "NtpClient.h"
 
 NtpClient::NtpClient(NtpTimeResultCallback onTimeReceivedCb)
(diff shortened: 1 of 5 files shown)


hreintke:
@anakod : Finally found the issue why I couldn't use delegate code for static callback.
(delegate was defined as bool, static as void). Now both are void.
SystemClock ntpSync still gives crash -> removed from code until solution found.
Cleanup of NtpClient code

anakod:
Great, thank you.

hreintke:
Now also found why 
```
ntpClientDemo dm1();
```
Is not working. It should be 
```
ntpClientDemo dm1;
```
The first one is wrong and getting unexpected behaviour error because compiler thinks of
ntpClientDemo dm1();
as of function declaration with name 'dm1' and the return type 'ntpClientDemo'.
Updated the application to use the correct variant

PR #96, opened 2015-05-28
Use consistent naming for Callback/Delegate and provide macro to defi…

hreintke:
…ne them

Implemented in delegate.h If you want another location please let me know
Used in Timer, NtpClient and SystemClock_NTP example application
f.e. : 
DELEGATE_CALLBACK(Test,void,int req) defines
TestCallback and TestDelegate as void functions accepting int as parameter

I created this because I spent a lot of time searching for issue, finding that I had a small discrepancy in the definition of callback and delegate.

### DIFF
--- a/Sming/SmingCore/Delegate.h
+++ b/Sming/SmingCore/Delegate.h
@@ -10,6 +10,10 @@
 
 #include <user_config.h>
 
+#define DELEGATE_CALLBACK(PARTIAL_NAME, REQTYPE, ...) \
+	using PARTIAL_NAME ## Callback = REQTYPE (*)(__VA_ARGS__); \
+	using PARTIAL_NAME ## Delegate = Delegate<REQTYPE(__VA_ARGS__)>;
+
 
 template<class ReturnType, typename... ParamsList>
 class IDelegateCaller
--- a/Sming/SmingCore/Network/NtpClient.cpp
+++ b/Sming/SmingCore/Network/NtpClient.cpp
@@ -1,7 +1,7 @@
 #include "NtpClient.h"
 
 NtpClient::NtpClient(NtpTimeResultCallback onTimeReceivedCb)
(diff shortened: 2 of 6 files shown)


anakod:
I don't like macros magic, but I will look.

hreintke:
In general I agree with limited use of "maco magic" 
But the positive on this is that we would get a very consistent way of usage and naming of callback and delegates.This is also the case for the "users" of the functionality.
And I see the use of callback/delegate increase more and more now we get a good idea of the possibilities

anakod:
I think what both versions Delegates and Callback wasn't needed. We have added it to Timer class only for performance reasons (double insurance, I think). In most cases just Delegate is enough.

anakod:
I have completed refactoring.

hreintke:
I implemented both because : 
- Most people are used to "normal callbacks" that is then still available -> lesser/no learning curve
- Internally all works with callback -> only addition to use both is an additional constructor -> not more maintenance needed

anakod:
"normal callbacks" will be available without additional code. Delegate class functionality already cover this, you can look.

anakod:
I have update Serial example also. Can you look to SystemClock_NTP?

hreintke:
Why did you remove the = NULL from last parameter 
	NtpClient(String reqServer, int reqIntervalSeconds, NtpTimeResultDelegate onTimeReceivedCb);
and the declaration 
NtpTimeResultDelegate delegateCompleted;

Now you cannot instantiate anymore with the simple NtpClient("myntpserver",myinterval)

anakod:
Yes, you are right, it's my mistake. But better will be use "nullptr" for same style.

hreintke:
OK, Do you want me to include it again or will you do it.
Have no option to do it today. Will work on it tomorrow,

anakod:
I have fixed this issue now.

anakod:
Example simplification still will be actual.

PR #103, opened 2015-05-30
Update SystemClock_NTP example

hreintke:
Updated SystemClock_NTP example


hreintke:
Anakod,
I want to add another callback to NtpClient : 
NtpClientErrorDelegate(ntpClient &Client, int errorcode)
to be called when timeout or other error occurs when requesting time
And also add status info to NtpClient (initial, inprogress,compete,waiting for next,...) 
Agree ?

anakod:
I think we can add it, yes.

PR #104, opened 2015-05-30
Add Full Control option to HardwareSerial Callback

hreintke:
By adding "Character Received"and option not to use HWSerial buffer a user can take full control over HWS input

### DIFF
--- a/Basic_Serial/app/application.cpp
+++ b/Basic_Serial/app/application.cpp
@@ -13,7 +13,7 @@ void sayHello()
 	Serial.println();
 }
 
-void hwsDelegate(Stream& stream, unsigned short charCount)
+void onDataCallback(Stream& stream, char recvChar, unsigned short charCount)
 {
 
 }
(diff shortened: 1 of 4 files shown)


anakod:
Is here any reason to disable internal buffer? I think what working with buffer is simpler than with one char, and also this buffer never overflow.

hreintke:
I came to this because : 
Returning rcvChar : Currently you only can "see" first char in buf by using peek, If you want to leave chars in buf until special char (f.e. \n) you need to know when that happens. This now can be done easily.-> you can make more use of the RxBuff from HardwareSerial.
Option "do not use RxBuff" :  If you want to use larger buffers, or do activities where you have to store the received chars in your own buffer anyway. In the current implementation you need to use available/read on every received char. By using useRxBuff=false you have the shortcut. You know the received char, it is not stored in HardwareSerial so no need to call additional functions -> optimization in an interrupt routine.
If you want to use the RxBuff from HardwareSerial as it is now -> no problem just keep useRxBuff true (= default) and no change to current functionality.
BTW :  The internal buffer overflows after 255 chars.

PR #109, opened 2015-06-04
Added PWM example.

ystrem:


anakod:
Will be good to refactor:
1. We can use analogWrite direct, without class
2. Why two vars countUp\countDown? I think it can be simple combined.

anakod:
Thank you for participation!

ystrem:
You are welcome, I'll look at the comment's.

-----Original Message-----

PR #113, opened 2015-06-08
Add Printf to print.cpp and update basic serial example

hreintke:
Added printf function to print.cpp.
In this way available for all "Serial" -> also for HardwareSerial.

### DIFF
--- a/Basic_Serial/app/application.cpp
+++ b/Basic_Serial/app/application.cpp
@@ -4,6 +4,7 @@
 
 Timer procTimer;
 SerialReadingDelegateDemo delegateDemoClass;
+int helloCounter = 0;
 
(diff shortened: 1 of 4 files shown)


anakod:
"vsnprintf" from esp_microc wasn't work?

hreintke:
I always got link errors when I tried to use the vsnprint from microc, that's why I finally went to use directly the ets variant. Could be because of "missing" include files but I spent days to find a solution without any luck. I am not sure how you include(d) the microc header files in order to be able to use them in the sming framework. BTW, the link errors occur not when building sming framework but only when building an application which uses the functionality.

anakod:
Ok, agree.

PR #114, opened 2015-06-09
Added Delegate support for the Station::waitConnection method.

slaff:


anakod:
We can remove old c-style callbacks methods, because Delegates already cover this functionality and provide back compatibility.

slaff:
> Delegates already cover this functionality and provide back compatibility.

@anakod Cool! Is there an example how to turn normal C-style function into delegate? For example I have the following:

```CPP
typedef Delegate<void(int,int)> XDelegate;

void ClassName::doSomeWork(XDelegate xDelegate)
{
     xDelegate(1,2);
}
```

How can I  use: 

```CPP
void xCallback(int a, int b)
{
  //...
}

ClassName obj;

obj.doSomeWork(???? xCallback ???)
```

anakod:
obj.doSomeWork(xCallback);

slaff:
Nice! That means that there is a need for general refactoring of all methods/functions that accept callbacks :)

anakod:
Yes, it is. Part of them already refactored.

slaff:
@anakod Can you check if this refactoring will work?

PR #116, opened 2015-06-11
sdk 1.1.1 patch

alonewolfx2:
sdk 1.1.1 patch


slaff:
IMHO that should not be part of the framework but a patch for the SDK that is used. See for example the patch for the esp-open-sdk (https://github.com/pfalcon/esp-open-sdk/pull/55/files).

alonewolfx2:
ok. right title is "patch for sdk1.1.1" right?

tprochazka:
Interesting that user_rf_pre_init() is already in the user_main.cpp and it is still not enough.

avr39-ripe:
current version of user_mail.c's user_rf-pre_init() lacks 'extern "C" ' declaration. adding it solves linking issues.

anakod:
@alonewolfx2 better push different features in different P-R to accept it by items. About `extern "C" `, yes it's my bug, sorry :(

alonewolfx2:
@anakod i know. but i am new in github i didnt make different pr for each features :( :)

PR #117, opened 2015-06-12
Switch from C-Style callbacks to more powerful Delegates.

slaff:
SmingCore/System callbacks are switched to delegates.

### DIFF
--- a/Sming/SmingCore/Platform/System.cpp
+++ b/Sming/SmingCore/Platform/System.cpp
@@ -34,7 +34,7 @@ bool SystemClass::isReady()
 	return state == eSS_Ready;
 }
 
-void SystemClass::onReady(SystemReadyCallback readyHandler)
+void SystemClass::onReady(SystemReadyDelegate readyHandler)
 {
 	readyHandlers.add(readyHandler);
 }
--- a/Sming/SmingCore/Platform/System.h
+++ b/Sming/SmingCore/Platform/System.h
@@ -11,10 +11,11 @@
 #include <user_config.h>
 #include "../../Wiring/WString.h"
 #include "../../Wiring/WVector.h"
+#include "../SmingCore/Delegate.h"
 
 class BssInfo;
 
-typedef void (*SystemReadyCallback)(void);
+typedef Delegate<void()> SystemReadyDelegate;
 
 class ISystemReadyHandler
 {
@@ -57,7 +58,7 @@ public:
 	CpuFrequency getCpuFrequency();
 	bool deepSleep(uint32 timeMilliseconds, DeepSleepOptions options = eDSO_RF_CAL_BY_INIT_DATA);
 
-	void onReady(SystemReadyCallback readyHandler);
+	void onReady(SystemReadyDelegate readyHandler);
 	void onReady(ISystemReadyHandler* readyHandler);
 
 	void applyFirmwareUpdate(uint32_t readFlashOffset, uint32_t targetFlashOffset, int firmwareSize);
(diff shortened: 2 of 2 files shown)


avr39-ripe:
Do please give me some link where I can figure out What is delegate and why it is better than good old callbacks. Thanks in advance!

kennpie:
As I work with Sming and C++, I am wondering the following: do any of the more advanced C++ functionality, like creating and using classes, or like delegates, have any penalty on the ESP8266, like bigger binary or slower operation? After all, this is still a microcontroller, and usually just 512 kb of flash. Thanks.

anakod:
Delegates add just around 500 bytes for complete implementation and 4 or 8 bytes additional RAM memory per created instance. It designed to have very low overhead.

slaff:
@avr39-ripe 
> Do please give me some link where I can figure out What is delegate and why it is better than good old callbacks.

If you are programming without using classes then you do not need delegates.
But if you are using classes and want to integrate them better in your code you may need to call a method from a class instance. C/C++ does not allow you to do this out of the box BUT there are some solutions to achieve this and delegates is one of the most elegant ones.

Let's see some examples. You may have the following code:

```CPP
void onConnect()
{
   // ... do something after successful connection
}


class Device
{

public:
  void bootstrap()
  {
     // .. do something smart ..
 

     WifiStation.waitConnection(onConnect);
  }

}
```
Above we have a class called Device. And a method in this class called bootstrap. At the end of this method we set up a callback that will be executed once the device is successfully connected. The callback is a normal C-Style function that is not part of the class. You may want to encapsulate the onConnect logic in the class itself and call back that method. After some refactoring the code will start to look like this:

```CPP
class Device
{

public:
  void onConnect()
  {
     // ... do something after successful connection
  }

  void bootstrap()
  {
     // .. do something smart ..
 

     WifiStation.waitConnection(???); // <!--- what shall we write here???
  }

}
```

If we replace the ??? from the example above with onConnect
```CPP
WifiStation.waitConnection(onConnect)
```

that won't work as you might expect it. Neither the example below will work:
```CPP
WifiStation.waitConnection(&Device::onConnect)
```

What will work is using delegates. The code below WILL work as expected:
```CPP
typedef Delegate<void()> ConnectionDelegate;

class Device
{

public:
  void onConnect()
  {
     // ... do something after successful connection
  }

  void bootstrap()
  {
     // .. do something smart ..
 

     WifiStation.waitConnection(ConnectionDelegate(&Device::onConnect, this)); // This WILL work.
  }

}
```
There are two important things related to Delegates in the code above. The first is 
```CPP
typedef Delegate<void()> ConnectionDelegate;

```
This says,  roughly, create a new type called ConnectionDelegate that derives from Delegate and expects C-style functions or class methods that have no parameters and have void return type. If we want to declare a Delegate that expects to have a function or method that has two parameters: int and float and float as a return type then 
the declaration will be:
```CPP
typedef Delegate<float(int, float)> MyDelegate;

```

The second important thing is the usage of Delegates
```CPP
..onnection(ConnectionDelegate(&Device::onConnect, this));

``` 
We specify that we want to have as a callback the method onConnect from the class Device for the instance ```this``` and the method onConnect should return void value and have no parameters, as defined in the ConnectionDelegate type. If I had to use MyDelegate in the code above then the usage can look like
```CPP
float Device::onConnect(int a, float b)
{
   // do something ...
   return b;
}

//...

..onnection(MyDelegate(&Device::onConnect, this, 5, 3.14));
``` 

Of course I have tried to answer your question without going into details, but if you are interested you can start reading more about C++ operator overriding, functional classes, class method callbacks, passing ```this``` "silently".

AlmightyFrog:
Your examples are as far as i can se correct, but still i'm wondering like kennpie if it is a good idea to use delegates everywhere just for sake of cause we can.
In the end it's for me comparing delegates to function pointers is mostly just getting "nicer" syntax in trade for ram/rom-space and runtime.
If there is only one function intended as callback - use callback as it is simple stupid superfast.

If more than one function needs to be called using delegate is the way to go. And as soon as you need that the 0.5k once is ok. Otherwise why waste 0.5k if you can get it for free; of course at this target we have mass of flash...

So in the end as i think at least on lib which you want to use the only drawback which i could see critical is RAM and time - last point is the more critical the more frequently a delegate is called.

PR #118, opened 2015-06-13
Remove quotes for path - breaks on linux

kireevco:


anakod:
It's tested on Windows, yes?

kireevco:
Yes, I tested it on windows 7.

PR #121, opened 2015-06-16
PROGMEM implementation

programmerby:
Also includes "Basic_ProgMem" example application.


Sming/Wiring/FakePgmSpace.h
```diff
@@ -17,6 +16,37 @@ typedef uint16_t prog_uint16_t;
 typedef int32_t prog_int32_t;
 typedef uint32_t prog_uint32_t;
 
+#ifdef ICACHE_FLASH
+
+#define PROGMEM STORE_ATTR ICACHE_RODATA_ATTR
+
+#define memcpy_P(dest, src, num) ets_memcpy((dest), (src), (num))
```
anakod: I think it's duplication of line 50 (internally implemented in esp_microc), but some parts was lost  in this branch (strcmp_P for example).

Can we combine it (move outside ifdef region)?

programmerby: Ok. I have made better memcpy_P and implemented missing functions.

PR #123, opened 2015-06-18
[makefile] Add flexibility for own project folders + debug info

ADiea:
This alloows me to have other module/source folders/subfolders besides the "app". These folders can be configured by defining MODULES var in own project Makefile before including the SmingCore Makefile.
Also memory size information added at the end of the build process.

### DIFF
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -108,7 +108,7 @@ TARGET		= app
 
 # which modules (subdirectories) of the project to include in compiling
 # define your custom directories in the project's own Makefile before including this one
-MODULES		+= $(SMING_HOME)/appinit app
+MODULES		+= $(SMING_HOME)/appinit
 EXTRA_INCDIR    ?= include $(SMING_HOME)/include $(SMING_HOME)/ $(SMING_HOME)/system/include $(SMING_HOME)/Wiring $(SMING_HOME)/Libraries $(SMING_HOME)/SmingCore $(SDK_BASE)/../include
 
 # libraries used in this project, mainly provided by the SDK
@@ -318,6 +318,7 @@ spiff_update: spiff_clean $(SPIFF_BIN_OUT)
 $(TARGET_OUT): $(APP_AR)
 	$(vecho) "LD $@"	
 	$(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) $(LD_SCRIPT) $(LDFLAGS) -Wl,--start-group $(LIBS) $(APP_AR) -Wl,--end-group -o $@
+ifeq ($(UNAME),Windows)	
 	$(vecho) "------------------------------------------------------------------------------"
 	$(vecho) "Section info:"
 	$(Q) $(OBJDUMP) -h -j .data -j .rodata -j .bss -j .text -j .irom0.text $@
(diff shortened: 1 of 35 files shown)


Sming/Makefile-project.mk
```diff
@@ -107,7 +107,8 @@ SPIFF_START_OFFSET = $(shell printf '0x%X\n' $$(( ($$($(GET_FILESIZE) $(FW_BASE)
 TARGET		= app
 
 # which modules (subdirectories) of the project to include in compiling
-MODULES		?= $(SMING_HOME)/appinit app
+# define your custom directories in the project's own Makefile before including this one
+MODULES		+= $(SMING_HOME)/appinit app
```
ADiea: Tested with regular project that does not have MODULES defined in own Makefile => PASS

Sming/Makefile-project.mk
```diff
@@ -317,6 +318,13 @@ spiff_update: spiff_clean $(SPIFF_BIN_OUT)
 $(TARGET_OUT): $(APP_AR)
 	$(vecho) "LD $@"	
 	$(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) $(LD_SCRIPT) $(LDFLAGS) -Wl,--start-group $(LIBS) $(APP_AR) -Wl,--end-group -o $@
+	$(vecho) "------------------------------------------------------------------------------"
+	$(vecho) "Section info:"
+	$(Q) $(OBJDUMP) -h -j .data -j .rodata -j .bss -j .text -j .irom0.text $@
+	$(vecho) "------------------------------------------------------------------------------"
+	$(vecho) "Section info:"
+	$(Q) $(SDK_TOOLS)/memanalyzer.exe $(OBJDUMP).exe $@
```
anakod: I think it will not work on Linux and MacOS.

Sming/Makefile-project.mk
anakod: Yes, it's a good idea.

Sming/Makefile-project.mk
ADiea: Ah, indeed that's correct!

Sming/Makefile-project.mk
ADiea: Maybe define MODULES += $(SMING_HOME)/appinit in this file and  MODULES = app as a starting point in all projects. I'll do this in 30 mins

anakod: It can break compatibility with user makefiles, so I think it's not very good.

Sming/Makefile-project.mk
programmerby: It could be fairly simple to make Python (or Bash..) version of the MemAnalyzer - https://github.com/Sermus/ESP8266_memory_analyzer/blob/master/MemAnalyzer/Program.cs

kireevco:
@programmerby I would suggest you submitting a PR to https://github.com/themadinventor/esptool, otherwise we will need to worry about one more dependency

ADiea:
I defined MODULES in all Makefile-user.mk files to be the app folder. Then the user can extend with his own folders

Sming/Makefile-project.mk
anakod: We can accept Pull-Request without memanalyzer, and add it later.

ADiea: Removed memanalyzer

anakod:
Sorry, it is my misunderstanding, we can use it under Windows, because it already included in UDK package

anakod:
Updated.

ADiea:
Thanks for merging have a nice day!

kireevco:
@ADiea Thank you for your contribution. We really appreciate it!

PR #132, opened 2015-06-23
One TcpClient constructor missed the , asyncCloseAfterSent(false) init

hreintke:
Found this missing initialization during my telnet testing


hreintke:
Another, in my view bette,r solution would be to initialize the asyncCloseAfterSent in the declaration (TcpClient.h : bool asyncCloseAfterSent;= false) and remove the  asyncCloseAfterSent(false) from the constructors. If you agree I'll change my PR to that

anakod:
Yes, I agree move initialization to H file.

anakod:
Merged, thank you!

Issue #133, opened 2015-06-23
HTTPServer_WebSockets example not building


smitjj:
Cannot build Websockets example, getting following error.

LD out/build/app.out
out/build/app_app.a(application.o):(.text._Z14startWebServerv+0x38): undefined reference to HttpServer::setWebSocketBinaryHandler(Delegate<void (WebSocket&, unsigned char*, unsigned int)>)' out/build/app_app.a(application.o): In functionDelegate':
/opt/Sming/Sming/SmingCore/Delegate.h:96: undefined reference to `HttpServer::setWebSocketBinaryHandler(Delegate<void (WebSocket&, unsigned char*, unsigned int)>)'
collect2: error: ld returned 1 exit status
make: * [out/build/app.out] Error 1

Any ideas?

robotiko:
Same here with latest code from today .
Tested yesterday without issues.


2015-06-23 22:13 GMT+02:00 smitjj <notifications@github.com>:

anakod:
Pease rebuild SmingFramework project.

robotiko:
@anakod, 
I did but didn't solve the issue.

anakod:
Try full SmingProject rebuild again please.

robotiko:
Fixed!
Thanks a lot

PR #135, opened 2015-06-24
Missing method for binary frame support

MrAPierce:


robotiko:
Hi @MrAPierce, 
I tried this patch to see if it would solve  #133, but keeps failing. 
Any hint?
Thanks

MrAPierce:
Hi @robotiko, I have build the SmingFramework project and the HttpServer_WebSockets project successfully with these changes, and tested that the binary frame is working. What keeps failing? building the HttpServer_WebSockets example project or the SmingFramework project?

@anakod could you please test building with these changes also?

robotiko:
Hi @MrAPierce, 
SmingFramework project build without problems, it is just the HttpServer_WebSockets project that fails (the trace is in #133).

robotiko:
Seems that is fixed with latest merge. 

Thanks

PR #139, opened 2015-06-25
Fix for SystemClock - RTC class added

Tavalin:
Fix for SystemClock wdt crash:
1) RTC class implemented so system time should be more accurate.
2) Implemented saving of the time to the RTC memory which enables time to be maintained across
restarts.


hreintke:
@Tavalin : First remark without detailed look at all code.
Wouldn't the os_delay_us(300) in your updateRtcTime interfere with high speed timers and interrupt processing ?

Tavalin:
@hreintke 
* Have you got any examples in mind? I'll try the PWM example later and see.
* Having looked at the example it seemed unintuitive that the additional constructors started the automatic sync but the default constructor doesn't. Maybe we could add a flag to the default constructor to specify if you want auto sync enabled at start?

Sming/SmingCore/Rtc.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+
+void RTC::updateRtcTime(RtcData &data) {
+
+	uint32 rtc_cycles;
+	uint32 cal, cal1, cal2;
+	cal1 = system_rtc_clock_cali_proc();
+	os_delay_us(300);
+	cal2 = system_rtc_clock_cali_proc();
+	cal = (cal1 + cal2) / 2; // get average cal in case one is out
```
anakod: Why it is needed, it's Espressif recomendation? (just use current CPU speed isn't right way, yes?)

Sming/SmingCore/Rtc.cpp
```diff
@@ -0,0 +1,66 @@
+#include "../SmingCore/Rtc.h"
+
+RTC::RTC() {
```
anakod: I think what Rtc.cpp/h files should be moved to Sming\SmingCore\Platform folder to keep same style (platform related stuff).

Sming/SmingCore/Rtc.cpp
anakod: Sorry, I have misunderstanding system_rtc_clock_cali_proc method functionality, but still want to know why addiotional delay is needed here? To be more accurate? What if updateRtcTime will be called to often (from getRtcSeconds at first)?

anakod: On other side 300 microseconds is very very small delay, of course :)

Sming/SmingCore/Rtc.h
```diff
@@ -0,0 +1,31 @@
...
+class RTC
+{
+public:
+	RTC();
+	uint32_t getRtcSeconds();
+	bool setRtcSeconds(uint32_t seconds);
+
+private:
+	bool hardwareReset;
+	bool saveToRtcMemory(RtcData &data);
```
anakod: Will be better to rename it, because RTC memory can be used in different real tasks, and should be accessible for user, I think. May be just saveTime/loadTime?

Sming/SmingCore/Rtc.cpp
Tavalin: I think sometimes system_rtc_clock_cali_proc doesn't always give a reliable value but I am doing some more testing to see if the delay can be taken out.

Sming/SmingCore/Rtc.h
Tavalin: Rename which bits?

anakod: About saveToRtcMemory\read
Because RTC memoty can be used for diferent tasks, not only last clock value storing

Tavalin:
Files moved to Platform folder and methods renamed to loadTime and saveTime.

Still testing removing the delay.

Tavalin:
@anakod @hreintke I've tested removing the delay and timekeeping seemed OK.  Are you happy to merge now?

PR #141, opened 2015-06-26
Telnet implementation

hreintke:
Implementation of TelnetServer
Includes the TcpServer/TcpClient Handling


hreintke:
Anakod, I created the rudimentary command implementation as a placeholder to show the possibilities.
Will work on a complete object/delegate/tabledriven solution.
Next steps are also integration with the previous Debug PR

kireevco:
I really want this. I want to be able to call methods from the command line for testing .

Great job!!!

robotiko:
Thanks a lot. 
Great functionality. 

For the command processor, you might find useful this project
https://github.com/joshmarinacci/CmdArduino

kireevco:
I feel like cmd and FreeRTOS SDK would be a perfect match...

AlmightyFrog:
Console sounds also quite goot do me, also that linked example looks good straight simple implemented.

I would suggest to use that implementation or make another generic one to use it both flexible for telnet aswell as for uart.

0xPIT:
There's a very nice implementation for a Commanline processor in 
http://www.microbuilder.eu/Projects/LPC2148ReferenceDesign.aspx

I have it running with the ESP RTOS API, currently trying to port to Sming+FreeRTOS.

robotiko:
@0xPIT  can you share a working example on Sming that can be used for serial?

hreintke:
Don't know what you expect on Serial from telnet.
The callbacks from telnetserver/clients are shown in the http_ajax example, updated in the PR
HardwareSerial example is basic_serial which also shows callback/delegate

robotiko:
@hreintke, you are totally right. 
I was contextually thinking of the command line/ console generalization to be reusable.so it could handle serial or telnet commands.  
Sorry, out of scope.

hreintke:
@robotiko : No problem, and yes, I am indeed trying to make the commandprocessrr usable for both serial and telnetclients. In fact for all "Streams" in Sming.

robotiko:
Thanks @hreintke 
That sounds great.  Being a generic stream.. we can reuse it for any "source" serial, telnet, websockets...
PLEASE!  :)

0xPIT:
I've found some time to fork the current RTOS-SDK and merge my sample port of the LPC Monitor: https://github.com/0xPIT/esp_iot_rtos_sdk/tree/master/app/user

hreintke:
Currently I am checking an implementation based on http://www.realtime.bc.ca/articles/cmdproc.html.
That, whit the inclusion of the Delegates available in sming, would give a maintenance free solution in which the functionality can be expanded by any class implementing a cmd callback.
That means also creators of libraries and people who "just" building applications would be able to extend. The sming core would then provide some basic commands (status, restart, debug,...)

anakod:
**cmdproc** very powerful solution, but it require different stdlib C++ classes (istream, ostream, and etc). Problem is what it would cause really large increasing of firmware compiled size, and it willn't be compatible with standard Arduino streams (Stream class).

**CmdArduino** - it isn't so powerfull solution, but looks good. Only thing, in my opinion, will be better wrap arguments to String's classes arrays, to avoid double pointers usage (`char **args`).

Will be glad to heare you opinions!

Sming/SmingCore/Network/TelnetServer.cpp
```diff
@@ -0,0 +1,139 @@
...
+			}
+		}
+	}
+	curClient->send(prompt,1);
+	return true;
+}
+
+void TelnetServer::processCommandLine( char *cmdBuf, int size)
+{
+	debugf("Telnet Command %s", cmdBuf);
```
anakod: It, of course, should be splitted to separated methods in future.

hreintke:
I will look at CmdArduino and see how we can adapt to sming.
On cmdproc, I am/was not planning to use the istream/ostream etc but replace that with the Stream class we already have in sming. We already have HardwareSerial as Stream. For tcpserver, I see two options. Make TCPclient also derived from Stream or add a "TcpStream" class derived from both TCPClient and Stream.

Sming/SmingCore/Network/TelnetServer.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+bool TelnetServer::onClientReceive (TcpClient& client, char *data, int size)
+{
+	debugf("TelnetServer onClientReceive : %s, %d bytes \r\n",client.getRemoteIp().toString().c_str(),size );
+	debugf("Data : %s", data);
+
+	for (int i=0;i<size;i++)
+	{
+		if (data[i] >= 240) // part of Telnet command
```
anakod: Can you discover, why 240? Also, I think, will be good move it to some define with clear naming.

Sming/SmingCore/Network/TelnetServer.cpp
hreintke: Yes, sure. I just included some very rudimentary commands to show some functionality.
definitely not as the final implementation. My focus in this PR was on getting TelnetSever working on top of/together with TcpServer and TcpClient

anakod:
*I'm really very like remote console and controll idea!* @hreintke, can you please, move Telnet server usage sample to another example project, to make it more simple and doesn't complicate HttpServer_AJAX project?

anakod:
Something like **TcpServer_Telnet**, for example, will be fine, I think

Sming/SmingCore/Network/TelnetServer.cpp
hreintke: Yes, should be define but also here : it's a quick way to show possibilities.
The 240 comes from the telnet protocol. That uses chars 240-255 for protocol negotiation (f.e echo yes/no and other options)

hreintke:
OK, Will create new example project for telnetserver

hreintke:
@All : This is the initial Command Processor implementation
Known issues : 
- Command input source is already generic but output is only telnet.
Next version will also get generic output
- API will change in next version due to above
- No line editing -> not possible to "stop/cancel" current input

Testapplication added,it is my first push of application code. Works in my environment but needs confirmation in others

Please comment on implementation, usage, testenvironment and all other.

robotiko:
@hreintke  thank a lot . 
Really nice addition.

hreintke:
@All : Added the following functionality : 
- Make CommandOutput generic, now usable for TcpClients and Stream (f.e. Serial)
- Added CommandProcessing functionality to Serial 
-- use Serial.commandProcessing(true/false); for enable/disable Commans from Serial, default = off
-- Be aware : set local echo on on terminal as current CommandProcessor has no echo func.
- implemented two additional commands : 
debugon -> set debug to Serial on
debugoff -> set debug to Serial off

As earlier : Please comment.

anakod:
Great, ready to merge this functionality with Debug, but few small final correction needed:
1. No reason to complicate HttpServer_AJAX example, because we already have separated Telnet usage example now. But if you think it's really needed, you can move code to additional file, like SerialReadingDelegateDemo in Basic_Serial example - to keep main code cleaner. But I think what telnet here isn't needed, because it's HTTP example :)

2. Can you remove unused files please: eagle.app.v6.data.bin, eagle.app.v6.rodata.bin...., Telnet_TCPServer_TCPClient/web/build/index.html
Or it's needed and I miss something may be?

Sming/Services/CommandProcessing/CommandExecutor.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+int CommandExecutor::executorReceive(char recvChar)
+{
+	if ((recvChar >= 240) || (recvChar == '\n'))
+	{
+//		debugf("CommandExecutor : Character ignored value = %u\r\n", recvChar);
+	}
+	else
+	{
+		if (recvChar == '\r')
```
anakod: Better ignore \r and wait \n. It's more standard way in all examples and in Arduino too.

hreintke:
@Anakod : Is OK, I will clean up the examples HTTP_AJAX & Telnet_Example.
Will change 
```
if ((recvChar >= 240) || (recvChar == '\n'))
```
to : 
```
if (!isprint(recvChar) || (recvChar == '\r'))
```
And add some minor changed I've done already to cleanup and make little more robust.
Regarding Debug. That needs now some rework to get it clean integrated this CommadProcessing environment. As it then will be based on this, would be easiest if you already commit the telnetso I can work from that.
Agree ?

hreintke:
LS,
Updates are : 
- End-of-Line charter updated to '\r'
- option to set prompt and EOL thru memberfunction of Commandexecutor
- ESC character deletes current line
- Only one client simultanious allowed on tcpserver (if needed can be updated in a later version if required)

kireevco:
I am wondering what will happen if there is an http mqtt server running?

hreintke:
@kireevco : Not sure what you are referring to. I've been running the telnetserver in combination with httpserver without issues. I have little experience with mqtt and see no example project running mqtt server. If you provide the information/examples I can test.

anakod:
> what will happen if there is an http mqtt server running

It should wor together.

hreintke:
@kireevco @anakod : Rereading my comment I noticed I made a mistake.  The sentence : Only one client simultanious allowed on tcpserver.
Should be : Only one client simultanious allowed on telnerserver
Is that where you were referring to ?
I will update the comment.

anakod:
Merged!

hreintke:
Thanks.

ADiea:
@hreintke @anakod Hi guys,

For me the latest telnet commit caused a massive increase on RAM over 1KB - which is serious business considering i only has 20KB RAM free and my project need additional code

Can you please point me to the main reasons for this? I do not use the Telnet server in my current project...just recompiled libsming and recompiled my unmodified prj and I have this increase in RAM.

My suspicion is this could be due to modification of tcpclient/tcp server but 1K seems too much. Could there be anything else?

I love this addition but for me it costs too much RAM and I would like to guard the RAM-eating parts with #defines ... can you help me identifying the areas?

Also, for ROM I see we have the --gc-sections in linked flags and so it should not link in unused code. It's not too much 4K as it is, but if I could isolate this more I'd love to.

Thanks

#Memory / Section info:

|  Section|                   Description| Start (hex)|   End (hex)|Used space |
|:--------:|:--------------------------------:|:------------------------------:|
|    data|        Initialized Data (RAM)|    3FFE8000|    3FFE8582|    1410 |
|    rodata|           ReadOnly Data (RAM)|    3FFE8590|    3FFEB4A4|   12052 (+1104) |
|       bss|      Uninitialized Data (RAM)|    3FFEB4A8|    3FFF5670|   41416 (+64) |
|      text|            Cached Code (IRAM)|    40100000|    40105D99|   23961 (+28) |
| irom0_text|           Uncached Code (SPI)|    40209000|    40243FDF|  241631 (+4192) |

Total Used RAM : 54878
Free RAM : 27042
Free IRam : 8825

anakod:
@hreintke can you look it, please.

ADiea:
@hreintke @anakod pleae take a look at #169 for me it reduces RAM/ROM feedback is welcome

PR #145, opened 2015-06-28
mDNS Advertise demo using ESP SDK

hrsavla:
/*** mDNS Demo
 * The multicast Domain Name System (mDNS) resolves host names to IP addresses
 * within small networks that do not include a local name server.
 * More info on mDNS can be found at https://en.wikipedia.org/wiki/Multicast_DNS
 * mDNS has two parts 1. Advertise 2. Listen
 * Bellow code (using ESP SDK) just does Advertise
 * Listen implementation is still on going work and will be basically porting
 * http://gkaindl.com/software/arduino-ethernet/bonjour to "Sming"
 *
 * In short this code will advertise other machines about its ipaddress.
 * But you can not convert other mDNS advertiser's host name to ipaddress. (this is work of Listening)
 *
 * How to use mDNS
 * 1. ADD your WIFI_SSID / Password
 * 2. Flash the Complied code to ESP8266
 * 3. According to OS of your PC / phone
 * 		A. Mac OS (iphone/ipad/ mac)
 * 			in Safari browser type "http://test.local/" to open the Ajax sample webpage.
 * 		B. Windows (You need Bonjour Service running. If you do not have it Install it from
 * 		   http://download.cnet.com/Bonjour-for-Windows/3000-18507_4-93550.html
 * 		   After installing in IE or Chrome or other browser type
 * 		   "http://test.local/" to open the Ajax sample webpage.
 * 		C. Linux  (You need to install Avahi mDNS/DNS-SD daemon)
 * 			In your browser type "http://test.local/" to open the Ajax sample webpage.
 * 		D. Android
 * 			You need to install ZeroConf Browser or Bonjour Browser
 * 			In those app you would be able to see IP address of your ESP module
 * 			In android Chrome "http://test.local/" does not translate to IPaddres
 * 			So android Chrome is not supporting mDNS.
 * 			But you can make your own app using Network Service Discovery. Look at bellow link for details
 * 			http://developer.android.com/training/connect-devices-wirelessly/nsd.html
 *
 */


kireevco:
I like this

anakod:
Thank you for participation. Some points to discuss:
1) I have plan to implement mDNS and DNS natively in future, with object oriented style and fully open source, of course (because we wasn't very like espconn implementation), but for now it is great and very useful example, I will be glad to merge it
2) We need rename it because too many Basic examples now :). I think "UdpServer_mDNS" name will be good (because mDNS based on UDP protocol)
3) I think no reason to make full HttpAjax example copy here, we can just make minimal single page, to keep mDNS example transparent and minimalistic.

hrsavla:
Implementation for mDNS with Name resolution (SD) will be a great addition to Sming. 
I have also started work of porting http://gkaindl.com/software/arduino-ethernet/bonjour. But it is not on my priority list right now as mDNS is not supported directly on Android Browser Chrome, etc. 

In above PR, I have changed the name to UdpServer_mDNS and replace Ajax with simple one line Http page. 

**Also I would like to know if websocket "client mode" is implemented in Sming?

anakod:
Websocket supported only in server mode. If yo want client, please create Issue with feature request and describe details why this functionality needed.

PR #163, opened 2015-07-06
Improve pullup handling for Arduino compat

ADiea:
Currently pullup is enabled by default.
To better emulate Arduino there are 2 cases
1. for versions pre 1.0.1 - pupllup is enabled by setting pinMode(INPUT) and then digitalWrite(HIGH)
2. for version post 1.0.1  - A new mode exists INPUT_PULLUP - pullup is enabled by setting the pinMode(INPUT_PULLUP)

See https://www.arduino.cc/en/Tutorial/DigitalPins

### DIFF
--- a/Sming/SmingCore/Digital.cpp
+++ b/Sming/SmingCore/Digital.cpp
@@ -8,20 +8,12 @@
 #include "../SmingCore/Digital.h"
 #include "../Wiring/WiringFrameworkIncludes.h"
 
-//Store pin modes to be able
-// to detect pin writes to INPUT pins
-// and activate the pullup accordingly.
-static bool gIsPinInput[TOTAL_PINS+1] = {0};
-
 void pinMode(uint16_t pin, uint8_t mode)
 {
 	if (pin < 16)
 	{
 		// Set as GPIO
 		PIN_FUNC_SELECT((EspDigitalPins[pin].mux), (EspDigitalPins[pin].gpioFunc));
-
-		//Default Pull-up
-		//pullup(pin);  Handled below and in digitalWrite() for older Android libs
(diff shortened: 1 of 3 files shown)


anakod:
It is great idea, but some improvements is needed for merge

Sming/SmingCore/Digital.cpp
```diff
@@ -8,6 +8,11 @@
 #include "../SmingCore/Digital.h"
 #include "../Wiring/WiringFrameworkIncludes.h"
 
+//Store pin modes to be able
+// to detect pin writes to INPUT pins
+// and activate the pullup accordingly.
+static bool gIsPinInput[TOTAL_PINS+1] = {0};
```
anakod: We should detect pin state dynamically, because it can be changed with different way.
This should be useful: http://esp8266.ru/esp8266-gpio-register/

ADiea:
@anakod I have modified to decide dinamically if pin is input. Now we have 2 cases
1. In pinMode if user explicitly sets mode INPUT_PULLUP (arduino >1.0.1) then set pullup
2. for older arduino pinMode will be called with INPUT and then when performing digitalWrite if pin is input activate pullup depending on value

Tested with setup containing a DHT22 sensor and a WS28xx RGB led + sdcard using software SPI
(picture of hardware setup is here https://github.com/ADiea/libDHT/blob/master/docs/ESP8266_Hardware.jpg)

anakod:
Can you also make pull request for sdcard, please? It's very intresting.

ADiea:
@anakod will integrate it into sming/libraries soon. It's a FAT32 "lib" from elChan

PR #168, opened 2015-07-11
:beetle:Fix isInputPin reversed logic

ADiea:
+ fix stupid inverted logic issue on isInputPin oops 
+ make digitalWrite() function with single return point

### DIFF
--- a/Sming/SmingCore/Digital.cpp
+++ b/Sming/SmingCore/Digital.cpp
@@ -72,7 +72,7 @@ bool isInputPin(uint16_t pin)
 	{
 		result = (READ_PERI_REG(RTC_GPIO_ENABLE) & 1);
 	}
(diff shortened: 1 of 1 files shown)


ADiea:
@anakod this is *urgent* pull request dunno how it escaped me, I still had pullup() calls in my code and this got unobserved

anakod:
Did you complete hardware tests?

ADiea:
@anakod yes hw test passed I have data from DHT22 device. 
If i remove pullup from DHT code, without this modification won't work
Also in my code I did not return in digitalWrite so for outputs pullup was activated == nop and then the write works but if you activate pullup and then return output will no longer work.
Rootcause isInputPin has reversed logic...

ADiea:
thanks!

PR #173, opened 2015-07-12
fix imports on case sensitive platforms

zendes:
after these two commits, my code compiles again on linux (ubuntu 14.04.2)


anakod:
Merged, thank you! Please, make P-R without libsming library in next time to avoid conflicts.

PR #182, opened 2015-07-17
Added __FlashStringHelper support for PROGMEM

brunnels:
Added in methods and macros to assist with PROGMEM usage

### DIFF
--- a/Sming/Services/ArduinoJson/src/JsonBuffer.cpp
+++ b/Sming/Services/ArduinoJson/src/JsonBuffer.cpp
@@ -14,7 +14,7 @@
 using namespace ArduinoJson;
 using namespace ArduinoJson::Internals;
 
-JsonStringStorage JsonStringStorage::_invalid(NULL);
+JsonStringStorage JsonStringStorage::_invalid("");
 
 JsonArray &JsonBuffer::createArray() {
   JsonArray *ptr = new (this) JsonArray(this);
--- a/Sming/Makefile-windows.mk
+++ b/Sming/Makefile-windows.mk
@@ -2,7 +2,7 @@
 ESP_HOME ?= c:/Espressif
 
 # Default COM port
-COM_PORT	 ?= COM3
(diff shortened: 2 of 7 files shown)


anakod:
ICACHE_FLASH_ATTR isn't needed, because Sming building system locate all code to IROM by default. Instead you should mark with iram attrinbute to methods what should go to IRAM memory.

__FlashStringHelper will be very fine freature, I think.

brunnels:
Okay, thanks for the info. I didn't know that. I'll modify this evening.

brunnels:
Okay, should be good to go now.

Sming/Services/ArduinoJson/src/JsonBuffer.cpp
```diff
@@ -14,7 +14,7 @@
 using namespace ArduinoJson;
 using namespace ArduinoJson::Internals;
 
-JsonStringStorage JsonStringStorage::_invalid(NULL);
+JsonStringStorage JsonStringStorage::_invalid("");
```
anakod: Why not NULL or nullptr for invalid values?

brunnels: NULL is macro'd to 0 which is int so throws a build error. You can try
nullptr and see if it builds but I think I tried it.

kireevco:
@brunnels thank you so much for your PR. 
For example _feature/progmem-flashstring-helper_?
Also, please squash your commits to one - too many of them.

hreintke:
@brunnels : 
We are restructuring and cleaning Sming repository and PR's.
Can you please squash your commits as some are reverting others and complete PR is unclear.
I have seen updates to ArduinoJson source. Can you explicitly mention why you want that update as it would diverge form the used library. 
PS There are plans to upgrade to a newer version of ArduinoJson.

hreintke:
This issue has been inactive for a long period and is being closed. If the issue is still valid please open a new with complete description

PR #188, opened 2015-07-24
Float support in printf, also for Wiring class

ADiea:
Implement printf equivalent of os_printf  that supports %f with precision.

To minimise code
* ftoa/itoa functions in system/stringconversion.cpp were used
* implementation is kept at minimum without printf flufs

Still wip, opened for discussion


Sming/Makefile-project.mk
```diff
@@ -143,7 +143,7 @@ ifneq ($(WIFI_PWD), "")
 endif
 
 # linker flags used to generate the main object file
-LDFLAGS		= -nostdlib -u call_user_start -Wl,-static -Wl,--gc-sections
+LDFLAGS		= -nostdlib -u call_user_start -Wl,-static -Wl,--gc-sections -Wl,-Map=$(FW_BASE)/firmware.map
```
ADiea: this is very useful to generate mapfile where one can see what functions got linked and their sizes

Sming/SmingCore/SmingCore.h
```diff
@@ -24,6 +24,7 @@
 #include "PWM.h"
 #include "Timer.h"
 #include "Wire.h"
+#include "SPISoft.h"
```
ADiea: this file should not be here yet, will revert

Sming/system/m_printf.cpp
```diff
@@ -0,0 +1,179 @@
...
+			*str++ = *fmt;
+			continue;
+		}
+
+		flags = 0;
+		fmt++; // This skips first '%'
+
+		do
+		{
+			//skip width and flags data - not supported
```
anakod: What about additional formating for int types? Is it will miss them?

ADiea: I believe they are all pretty print fluffs...if you feel we should support something let me know.

for ints: 
* no width support -> needed probably when displaying table data, i think rarely
* no left adjustment support -> rarely used, maybe for tables
* hex numbers always display 0x prefix to avoid confusion (printf has the # flag)
* the star flag saying that width and precision info are taken from parameters and not from format string -> rarely used
* no zero padding in front of int -> maybe usefull for table and pretty printing?

for floats
* presision is supprted as it is useful. 
* width not supported, only useful probably for pretty printing
* zero padding / space padding related to width above not supported

I could add width specifing for floats, unfortunately the ltoa function used here for ints does not take width parameter and it is already used like this in the project

anakod:
I think, if we want improve printf functionality we should make it in [esp_microc](https://github.com/anakod/esp_microc/) and reload standard printf, sprintf and etc. functions implementation, not user side debugf or Serial.printf.

But at first I'm not sure what this implementation will cover all cases what covered by current ets_printf. Can you provide more different test and details, please?

Sming/system/m_printf.cpp
ADiea: Theese are my test cases, I will probably include some of them in the basic console example 

```
	m_printf( "pi=%f\n", 3.14f);
	m_printf( "1.002=%f\n", 1.002);
	m_printf( "1.050250=%f\n", 1.050250);
	m_printf( "pi=%d %f %c\n", 3, 3.14, '3');
	m_printf( "etc %d %f %s %f %f\n", -99, -3.01040, "abc", 3.0/4, 8.0/9);

	m_printf( "etc %3.0d %.3f %s %.5f %f\n", -99, -3.01040, "abc", 3.0/4, 8.0/9);

	m_printf( "etc %3.0d %% %.3f %%%s %-+.5f %#*f %%%%\n", -99, -3.01040, "abc", 3.0/4, 8.0/9);

	m_printf( "etc %3.0d % %.3f %j %s %-+.5f %#*f %z%k\n", -99, -3.01040, "abc", 3.0/4, 8.0/9);

	int min_i =
	m_printf("Int limits %d .. %d Uint limits 0 .. %u Long limits %ld .. %ld  \n",
			INT_MIN, INT_MAX, UINT_MAX, LONG_MIN, LONG_MAX);

	 m_printf("Color %s, number1 %d, number2 %05d, hex %#x, float %5.2f, unsigned value %u, octal %o.\n",
	        "red", 123456, 89, 0x00C0FFEE, 3.14159, 250, 06675);
```

Sming/system/include/esp_systemapi.h
```diff
@@ -25,8 +25,11 @@
 #define STORE_TYPEDEF_ATTR __attribute__((aligned(4),packed))
 #define STORE_ATTR __attribute__((aligned(4)))
 
+extern int m_vsnprintf(char *buf, uint32_t maxLen, const char *fmt, va_list args);
+extern int m_printf(const char *fmt, ...);
+
 #undef assert
-#define debugf(fmt, ...) os_printf(fmt"\r\n", ##__VA_ARGS__)
+#define debugf(fmt, ...) /*m_printf(fmt"\r\n", ##__VA_ARGS__)*/
```
anakod: Why it is disabled here?

ADiea: work in progress, testing debugf impact on code size

anakod: Please provide more info in related thread: https://github.com/anakod/Sming/pull/169
But t shouldn't be here, of course.

ADiea:
@anakod can we get rid of os_printf i believe it is in user_interface.o/libmain.a 

I think the other esp libs are already linked against os_printf symbol and what we will do with etc_microc will be to only add more code. 

My idea was to provide super lightweight implementation with main focus on basic capabilities that we can use until hopefully Espressif provides a os_printf with float support. Because our alternate implementations are duplicate code as long as os_printf still gets linked in...

ADiea:
@anakod took a quick look at esp_micro and read that it is possible to rebuild e.g. libmain.a to exclude os_printf and replace with ours maybe then modify the PROVIDE address for linker? Is this correct?

If we can remove os_printf I suggest using ee_printf implementation from here https://github.com/jpbonn/coremark_lm32/blob/master/ee_printf.c

It looks almost complete I would even remove some things from it. Originally I started from that but removed almost all because I'm worried about code space :)

anakod:
I think yes, we can replace address for linker. On other side Sming standard is using usual methods (printf and not os_printf, strcpy and not os_strcpy), so redefining printf, sprintf, ..., etc can be enough.

ADiea:
@anakod yes sming uses printf but the espressif libs use os_printf => duplicate code in final image

anakod:
I understand you. Now printf is just proxy to os_printf

anakod:
But it wouldn't make more duplication, because os_printf always exist in chip static ROM code.

ADiea:
@anakod implemented width modifiers so now %3d and %11.3f are supported. Implementation is still kept minimal, please see Basic_Serial/app/application.cpp for the implemented tests and   Basic_Serial/printf_test_output.txt for output.

Also let me know if you have other suggestions for tests.

robotiko:
+1

ADiea:
@anakod please let me know if you feel something else is missing here or not ok.

anakod:
I'm not completely sure about this feature (and related breaking changes), but if peoples want %f printf I think we should support it. Big thank you for participation anyway!

anakod:
I think we miss one important feature: "%X" and "%x" support. It is also used in Sming code.

ADiea:
@anakod In the example output we have: 

```
Show a decimal 123456, a hex 0xc0ffee, an unsigned 250, an octal 6675.
```

In m_printf.cpp

```
case 'x':
+			*str++ = '0';
+			*str++ = 'x';
+			base = 16;
+			break;
```

The implementation does not support upper case(%X) it will always print lowercase(%x) letters to save program space but if that is wanted i can easily add support for %X.

ADiea:
@anakod I will submit a PR to correct #235 and will also include uppercase X to do the same thing as x for now

```
case 'x':
case 'X':
...
```

Is this ok with you?

PR #195, opened 2015-08-02
Debug implementation

hreintke:
Debug implementation which includes Debug to Serial & telnet, both optional & configurable.
Does not takeover as that would interfere with tcp debug output.
Please comment


anakod:
I am very sorry for delay with merging debugging pull request. I'm still have few ideas and features related to debugging and\or command functionality, what I think should be implemented here, but it isn't formalized and completed yet. I'm should go deeper into that, and than, I think, I will make additional commits to that module.  Sorry, again for delay.

hreintke:
@anakod : Please share your initial thoughts on the new features you are thinking about. 
We can combine the effort and come to a good and functional implementation.

PR #199, opened 2015-08-03
activeClients debug message shows wrong information

hreintke:
Update debugf message in TcpServer

### DIFF
--- a/Sming/SmingCore/Network/TcpServer.cpp
+++ b/Sming/SmingCore/Network/TcpServer.cpp
@@ -128,7 +128,7 @@ err_t TcpServer::onAccept(tcp_pcb *clientTcp, err_t err)
 void TcpServer::onClient(TcpClient *client)
 {
 	activeClients++;
-	debugf("TcpServer onClient  %s, activeClients = %d\r\n ",client->getRemoteIp().toString().c_str(),activeClients);
+	debugf("TcpServer onClient  %s\r\n ",client->getRemoteIp().toString().c_str());
 	if (clientConnectDelegate)
 	{
 		clientConnectDelegate(client);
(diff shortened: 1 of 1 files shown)


anakod:
Why it calculates wrong? I think it should be substracted somewhere.

hreintke:
Yes I agree but...
The decrement of activeclients is in tcpserver class. The httpserver/httpconnection does not callback/make use of that fucntionality -> it keeps incrementing. Possibly the solution is in the destructor of httpconnection but I am not sure yet and want to verify/test.
I like to make it work as it also would give the possibility to add functionality to limit the number of active clients to limit memory usages
The current display of the (wrong) increasing number of activeclients creates confusion and discussions of heap usage. I wanted to get that out of the equation and temporary remove the display from the debug statement

kireevco:
@hreintke thank you so much for your PR. 
We are trying to implement git-flow (#230). Can you please re-submit your PR to a hotfix branch?
For example _hotfix/active-clients-debug_?

hreintke:
@kireevco : This is not a feature but a bug-fix.
In my opinion these should not be pushed to a feature branch but either to develop or a generic bug-fix branch

kireevco:
@hreintke sorry,that's what I meant - _hotfix/active-clients-debug_

Sming/SmingCore/Network/TcpServer.cpp
```diff
@@ -128,7 +128,7 @@ err_t TcpServer::onAccept(tcp_pcb *clientTcp, err_t err)
 void TcpServer::onClient(TcpClient *client)
 {
 	activeClients++;
-	debugf("TcpServer onClient  %s, activeClients = %d\r\n ",client->getRemoteIp().toString().c_str(),activeClients);
+	debugf("TcpServer onClient  %s\r\n ",client->getRemoteIp().toString().c_str());
```
raburton: bit picky, but do you want to remove the blank space on the end of the message?

hreintke: Sharp.. Should I update the PR or submit a new one ?

PS will be offline from now. Further tomorrow

raburton: *if* you want to change that the ideal thing to do would be squash the second commit and the first together, but it doesn't look like the original repository still exists so i'd probably just leave it

hreintke:
@raburton : 
Just noticed that this patch did not go into the 1.4.0 release.
Can you merge with develop so I goes in next ?

Issue #174, opened 2015-07-13
DNS record not found 


flexiti:
After 2 and more hours of work DHT example reading loop ( every 2 minutes) I have : 

8:17:52.338, connect to: api.thingspeak.com
8:17:52.338, realloc 121 -> 252
8:18:05.993, DNS record _not_ found: api.thingspeak.com
8:18:05.993, -TCP connection
8:18:05.993, Failed

Using direct IP work all the time

kireevco:
@flexiti Please provide us your code and your device type that you are using.

Thank you

flexiti:
Now I have another symptoms (more code?), in any case, as you get an error like this below- from this moment each call is completed with error until  a soft reset.
it happens sometimes after an hour and sometimes several.
In my opinion it is associated with errors in the network (I have many devices on my WiFi)
I'll try to prepare in my free time some sources simplified to suit the tests. I use 3 pieces:ESP01 + DHT22 ,  ESP01 + BS1820 and ESP12 +BS1820+ BMP180 . In all the symptom it is similar, generally hard for me to get continuous work more than 24 hours.
Also I will try to get a more extensive log.

Debug with error:
==============================================
Download: http://api.thingspeak.com:80/update?key=YOXULSNC3XXXXXXX&field6=23.9&field2=52.7&field3=13&status=dev_ip:192.168.0.188
+TCP connection
connect to: api.thingspeak.com
realloc 131 -> 262
MB_TCPConnection_DNS record found: api.thingspeak.com = 184.106.153.149
TcpConnection::connect result:, 0
MB_HttpClient_onFinished Write error code: 0, Mode: 0, Finish state: 4, Body: 

onDataSent  --> ResponseCode: 0
onDataSent  --> ResponseString: 
onDataSent  --> PostBody: 
onDataSent  --> ResponseHeader: none
onDataSent  --> ResponseState: 4
onDataSent  --> Successful: Error

MB TcpConnection error: -8
MB_TCPConnection_staticOnError err

Normal debug:
============================
Download: http://api.thingspeak.com:80/update?key=YOXULSNC3RXXXXXX&field6=23.9&field2=52.9&field3=11&status=dev_ip:192.168.0.188
+TCP connection
connect to: api.thingspeak.com
realloc 131 -> 262
MB_TCPConnection_DNS record found: api.thingspeak.com = 184.106.153.149
TcpConnection::connect result:, 0
OnConnected
TCP connected
onReadyToSendData: 0
TcpClient request completed
[...]

Issue #198, opened 2015-08-03
TCP Client Stops with "TCP connection error: -8" 


hrsavla:
"TCP connection error: -8" mean ?? error number is either -8 or -9 . I have been receiving this error in most TCP client application like Mqtt, downloadString() (httpclient), ftp (Big files). The occurrence of this is random. But occurs onces in atleast 5 attempts. My usual fix for it has been restarting the TCP client which works, but now analyzing and have auto restart in Tcpclient will be helpful to all.

On further looking on error codes in espconn.h (which is in  this folder C:\tools\sming\Sming\system\include\lwip\app)

#define ESPCONN_ABRT       -8    /* Connection aborted.      */
#define ESPCONN_RST        -9    /* Connection reset.        */
#define ESPCONN_CLSD       -10   /* Connection closed.       */
#define ESPCONN_CONN       -11   /* Not connected.           */

So it seems that esp8266 is aborting /resetting TCP connection. 
** "Is this aborting/resetting of TCP connection due to timeout ? "  
** Also suggest about way to implement "Auto-reconnect".

flexiti:
Please read issue:
 https://github.com/anakod/Sming/issues/174
and check https://github.com/flexiti/HttpClient_Bug
Is not this another symptom of the same problems?

hrsavla:
Yes @flexiti , all your symptom seems to be pointing to same "Exception" (lets not say call it  bug). We need to handle this "Connection aborted/reset" exception. In my Mqtt client I have a timer periodically checking connection status and reconnecting it. But there should be better way of "Auto Reconnect" in TCPClient.

hreintke:
@hrsavla : Currently I thing the concept of the implementation is OK.
It calls the callback oncompled with status true/false based on correct/not correct completion of http request. It's up to the application to reconnect or take other actions.
Of course the status in the callback should be correct (which it not now) to let this work

flexiti:
I found a solution to  "not existing server" problem. (https://github.com/flexiti/HttpClient_Bug )
 it's probably solves few other problems. 
On network  error, the "asyncTotalLen" remained  not cleared
What is your opinion?

SimngCore-Network-TcpClient.cpp
----------------------------------------------
void TcpClient::onError(err_t err)
{
	      asyncTotalLen = 0;  //this should be added in my opinion
	      asyncTotalSent = 0; // and this ?? 

	state = eTCS_Failed;
	onFinished(state);
	TcpConnection::onError(err);
}

[...]

PR #200, opened 2015-08-04
TCPClient re-initializy async variables after finishing connection

hreintke:
TCPClient does not re-initialize variables after finish connection.
This gives errors in the next use (connect) of the TcpClient.

PS : Solving was combined effort with @flexiti 

Should at least fix issue #174 and possible #198 (needs to be re-tested)


anakod:
My mistake, thank you very much.

PR #202, opened 2015-08-07
Added WebSocket Client support for Sming.

hrsavla:
I have created Websocked Client library and a demo program. Kindly merge it with Sming.


hreintke:
@hrsavla : Did you test this part of your code : 
```
+    if (!(connect(uri.Host, uri.Port))) 
+    {
+        Mode = ws_Disconnected;
+        debugf("Could not connect with Server :(");
+        return false;
+    }else
+    {
```
From my work on tcp error I remember that the returncode from connect is only that connect has been correctly started (f.e. enough memory available for internal tcp strctures).
The actual success of connect will come from onConnect and onError callbacks.

Other : any reason why you did not implement the onConnected onCompleted callbacks ? 
This would make the wsclient implentation more like the tcpclient functionality

anakod:
Thank you for great work, I will review and reply.

Sming/SmingCore/Network/WebsocketClient.cpp
```diff
@@ -0,0 +1,422 @@
...
+    
+}
+/* Function Name: startWS
+ * Description: Start Websocket Client. It connects to Server and initiates 
+ *              handshakes
+ * Parameters: url - Url of the server in format ws://echo.websocket.org
+ *              _callback - Callback function which receives incoming data
+ *              eg. void wsMessageReceived(String message)             
+ */
+bool WebsocketClient::startWS(String url, WebSocketRxCallback _callback /* = NULL*/)
```
anakod: Better name it in standard way. "connect" is most usual for that case in Sming environment. Another is "begin".
You can look MQTT client as naming example also.

hrsavla:
Implemented things suggested by @hreintke . Updated the demo application.

hreintke:
@hrsavla : I had a different implementation in mind. Only did a quick check, so these are first remarks.
In your implementation the application still has no possibility to be "called back" when the wsclient enters specific states and thus needs some kind of polling to get forward. See the
```
void wsMessageSent()
{
    if((wsClient.getWSMode() == ws_Connected) )
```
in your demo. 
And how would an application know when the connection has been closed.
I was thinking of having multiple constructors like in tcpclient f.e.
```
TcpClient::TcpClient(TcpClientCompleteDelegate onCompleted, TcpClientEventDelegate onReadyToSend /* = NULL*/, TcpClientDataDelegate onReceive /* = NULL*/)
	: TcpConnection(false), state(eTCS_Ready)
{
	completed = onCompleted;
	ready = onReadyToSend;
	receive = onReceive;
}
```
Not sure either whether there should be a distinction in onconnect between "tcp level connection" and "wsclient level connection"

anakod:
> And how would an application know when the connection has been closed.
I was thinking of having multiple constructors like in tcpclient f.e.

In my opinion we have to many constructors in current implementation. I'm thinking about move this functionality to **setSomeCallback** methods, like in WebSocketServer.

hreintke:
We also could then get back to two contructors. 
One default xxClass() and one xxClass(callback1,callback2,callback3,..)
And of course also the setSomeCallback routines.
Then both possibilities are present. If you want to construct with limited number of callbacks you can use xxClass() with some setCallback calls or xxClass(callback1, nullptr, callback2)
If we define that as the way to go we can implement step by step across all classes but should be aware of the API changes we are introducing

anakod:
I think we can have constructors together with setXxxCallback methods, but in my opinion constructor with large amount of arguments is hard to use. So better move only most important delegates to constructors, I think. Now it is possible with nice Sming delegates.

hrsavla:
@hreintke @anakod I have added call back functions in Websocket client . 

Also I need your inputs about adding SDK PWM (which uses hardware timer) to sming. 
What should be  name of the Class. 
2 files in above PR are adding SDK's PWM.h function in esp_systemapi.h and Makefile-project.mk. Or we need to include ESP_SDK/include/pwm.h in esp_systemapi.h.

robotiko:
+10

hrsavla:
I have added ESP8266 SDK pwm api  to Sming. Esp8266 SDK PWM  uses hardware timer (Flicker free compare to Smings PWM which uses software timer). Flickers were generated in Sming's PWM as software timer were not of higher priority.  HwPWM output is flicker which is useful for applications like LEDs lighting.

hreintke:
@hrsavla : was out for a short holiday, so delayed response.
This is a remar from someone who is not a user of pwm so just on style and generic functionality.
- Shouldn't  analogWrite, setDuty, and setPeriod be boolean instead of void to return error in case of wrong input ?

hreintke:
@hrsavla : Sorry for two comments which could have been one.
Maybe classname should be something like HWPwm or HW_pwm, and although it would break backwards compatibility change current PWM to SW_pwm. 
Possible for intermediate time have both PWM and SW_pwm and phase out PWM

kireevco:
@hrsavla thank you so much for your PR. 
For example _feature/websocket-client_?

kireevco:
Closed in favor of #258

PR #208, opened 2015-08-11
Reduce firmware size by 16kB for apps not needing spiffs

ADiea:
Removing spiff_mount() from user_main.cpp and calling it when needed by the relevant apps, reduces the firmware size by 16kB

This is the output of build process after rebuilding Basic_Serial example

```
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8554|    1364
    rodata|           ReadOnly Data (RAM)|    3FFE8560|    3FFE8F70|    2576 (-472)
       bss|      Uninitialized Data (RAM)|    3FFE8F70|    3FFF22B8|   37704 (-1992)
      text|            Cached Code (IRAM)|    40100000|    40105B49|   23369
irom0_text|           Uncached Code (SPI)|    40209000|    40232D56|  171350 (-16724)
```


anakod:
Will be better make it as Makefile configuration option, to keep it simpler by default.
Also, if needed, user always can redefine initialization function `user_init` and change standard behavior.

ADiea:
Second option with the redefining would be best but not sure if possible. I have tried right now to redefine it in application.cpp and it says: error already defined, because it is linked already in libsming.a
Also, searching on the web I could not find anything except for some macro trickery. Do you know of a way to redefine the C function in application.cpp?

anakod:
About "C" - no problem, `extern "C"`
About redefining - we can add attribute, I think I fogot about that at first.

hreintke:
I was sure that I had rad something on this before. Searching again helped.
I have no experience in this and don't know if all platforms support it.

It is about the weak attribute : __attribute__((weak))
with the definition : 
```
weak The weak attribute causes the declaration to be emitted as a weak symbol rather than a global. 
This is primarily useful in defining library functions that can be overridden in user code,
though it can also be used with non-function declarations. 
Weak symbols are supported for ELF targets, and also for a.out targets when using the GNU assembler and linker
```

ADiea:
I also found out about using --wrap in linker params (-Wl,--wrap=user_init) but a quick test was not successful. Maybe it would be best to define a CFLAGS = -DNO_SPIFFS in Makefile-user.mk and call spiff_mount in user_init() if flag not defined.

ADiea:
The weak attribute works like a charm thanks @hreintke 
I have left user_init() weak to provide flexibility in the future, and also added NO_SPIFFS makefile argument in makefile and code.

To use it you must use "make all NO_SPIFFS=y" you need to define a new target in Eclipse for your project. For the Sming project, I already created the new target configuration: "rebuild NO_SPIFFS=y"

Also modified makefile to fix #218.

avr39-ripe:
Looks nice! We really need this! +1

anakod:
I think **NO_SPIFFS=1** is more usual convention

anakod:
Oк may be **DISABLE_SPIFFS=1**

ADiea:
sure why not

ADiea:
@anakod let me know if you feel something else is missing or not ok, from my side code is ready to integrate

anakod:
It is fine now, thank you.

PR #225, opened 2015-08-21
Allow spiff customization

piperpilot:
This change set allows you to customize the size of the SPIFFs and also specify the location.  I also updated 3 sample projects so that they will automatically build the SPIFFs instead of having to copy to files or change the config.

This is in reference to issue https://github.com/anakod/Sming/issues/224

### DIFF
--- a/HttpServer_AJAX/Makefile-user.mk
+++ b/HttpServer_AJAX/Makefile-user.mk
@@ -28,3 +28,6 @@ MODULES = app
 
 # Com port speed
 # COM_SPEED	= 115200
+
+# SPIFFs Location
+SPIFF_FILES = web/build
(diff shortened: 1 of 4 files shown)


anakod:
Great, thank you! Why you select 512 Kb size for 4M flash? I think it can be larger. Of course, best way will be calculate storage size from total files size, but I'm not sure about right calculation spifffs internal requirements.

piperpilot:
I had problems before when I tried to make the SPIFFS extend past the 1MB boundary.  I'll have to try it again and make sure it works and can then make the adjustment.  I figured 512Kb was a "SAFE" number for the 4M flash.

robotiko:
@piperpilot   
You will find this interesting about the 1M limitation.
http://richard.burtons.org/2015/06/11/memory-map-limitation-workaround/

PR #237, opened 2015-08-27
Update TFT_ILI9163C.cpp

flexiti:
change for correct horizontal display


anakod:
I'm not sure what it's right for all cases. How you tet it?

flexiti:
if "setRotation" exist in library TFT_ILI9163C , it should work well ;)
in the case of "3" was inconsistency 
(rotation 1 -> +90degrees, 2 -> next 90 and so on)
 It was tested, see picture on #238

Also used in sming example:  ScreenTFT_ILI9163C 
void init()
{
	Serial.systemDebugOutput(true); // Allow debug output to serial

	Serial.println("Display start");
	tft.begin();
	tft.setRotation(2); // try yourself   <-------------------------------------------------------------- 
	tft.fillScreen();

anakod:
Ok, agree.

PR #241, opened 2015-08-28
fix systemDebugOutput()

ADiea:
* fix systemDebugOutput()
* remove os_printf from framework to only use m_printf alternative
* support %X format specifier, but still prints lowercase

Testcases for m_printf can be found in Sming/Basic_Serial/app/application.cpp

Thanks

### DIFF
--- a/Sming/Services/SpifFS/spiffs_config.h
+++ b/Sming/Services/SpifFS/spiffs_config.h
@@ -23,7 +23,7 @@
 #define SPIFFS_CHACHE 0
 
 #define c_memcpy os_memcpy
-#define c_printf os_printf
+#define c_printf m_printf
 #define c_memset os_memset
 
 /*typedef int32_t s32_t;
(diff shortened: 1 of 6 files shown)


hreintke:
@ADiea @anakod : Running Basic Serial
I get reset of esp when I include any of these examples, 
```
//	Serial.printf( "Display -99:%d, \t\t-3.0104:%f, \t'abc'=%s, \t\t0.75:%f, \t\t0.888888889:%f\n",
//			-99, -3.01040, "abc", 3.0/4, 8.0/9);

//	Serial.printf( "Display '   -99':%6d, \t-3.010:%.3f, \t\t'abc'=%5s, \t\t0.75000:%-+.5f, \t0.888888889:%#*f\n",
//			-99, -3.01040, "abc", 3.0/4, 8.0/9);

//	Serial.printf( "Display -99:%3d, \t\t-3.010:%.3f, \t\t'abc'=%#s, \t\t' 0.75':%5f, \t\t0.889:%.3f\n",
//			-99, -3.01040, "abc", 3.0/4, 8.0/9);
```
And maybe better to : 
- Run testPrintf() in SayHello(), now it interferes with other initial output.
- Should we include \r\n instead of \n in the printf examples ?

ADiea:
@hreintke 
I will investigate and move it to say hello. I can use \r\n no problem, my terminal is set up to output \r for every \n etc

PR #244, opened 2015-08-29
spiffs mounting improvements

raburton:
Allows the location and length of the spiffs to be specified by the app, the app
should be compiled with DISABLE_SPIFFS to prevent auto mounting and then call
spiffs_mount_manual with the desired parameters from the init function.

This was requested by rBoot users and allows mounting of multiple different
spiffs images along side multiple OTAable rom images.

### DIFF
--- a/Sming/Services/SpifFS/spiffs.c
+++ b/Sming/Services/SpifFS/spiffs.c
@@ -59,19 +59,18 @@ spiffs_config spiffs_get_storage_config()
 	return cfg;
 }
 
-bool spiffs_format_internal()
+bool spiffs_format_internal(spiffs_config *cfg)
 {
-  spiffs_config cfg = spiffs_get_storage_config();
-  if (cfg.phys_addr == 0)
+  if (cfg->phys_addr == 0)
   {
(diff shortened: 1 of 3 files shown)


kireevco:
Should it be synced with the upstream? https://github.com/pellepl/spiffs

raburton:
> Should it be synced with the upstream? https://github.com/pellepl/spiffs

I don't think so, none of the code altered exists upstream. spiffs.c isn't there at all and the small change to the declarations in spiffs.h aren't in the upstream version.

kireevco:
Ah, You're right. spiffs.c is sming-specific.

piperpilot:
Richard,

I don't see  a change to support bypassing the spiffs_mount() in user_main.cpp.  Above you mention "DISABLE_SPIFFS"...did you forget to commit the change for that?

Curtis

raburton:
That's right, it's already there from a recent commit. It saved me having to add something similar myself.

anakod:
I think that's right direction for lager flash chips, thank you.

Issue #230, opened 2015-08-23
Community proposal: Git flow for Sming development


kireevco:
I propose switching Sming framework to git flow. This will highly improve stability for the release, will allow us implement backwards-incompatible changes (like upgrading SDK) easily. Currently we can't do that at all without breaking Sming and causing issues for users.
This will also allow us to run automated tests.

__master__: Branch that contains latest production (stable) release
__develop__: Main development branch: contains all current new features 
__feature/websocket-client__. Gets deleted when merged to __develop__: Branch for a new feature while in development process.
__release/1.2.5__: Special release branch that exists only when develop gets merged to master. This branch is used when a feature-set in __develop__ is frozen, and the branch is used to send latest fixes, before it gets merged into master and published to [sming/releases](https://github.com/anakod/Sming/releases).

This will mean that __all Contributors__ will have to submit a PR to a feature branch, it will be tested and then merged to __develop__ for automated integration testing (via TravisCI, Jenkins or SemaphoreCI), as well as manual testing on a real device.  All Commiters will need to submit a PR to __develop__ branch, which then will be either merged by a Contributor or moved to a separate feature branch.

#### Committer
- Fork _Sming_ repo
- Create a local branch off the __develop__ : __feature/websocket-client__
- Build, test your code
- Commit changes
- Push your changes to your fork on github
- Submit PR to the main Sming repo, to __develop__ branch.

#### Contributor
- Fork _Sming_ repo
- Create a local branch off the __develop__ : __feature/websocket-client__
- Build, test your code
- Commit changes
- Push your changes to anakod/sming repo and optionally your fork on github
- Submit PR to the main Sming repo, to __Sming/develop__ branch from __Sming/feature/websocket-client__.

### Reviewing PRs & Manual Testing
Pull requests should be reviewed by peer Contributors, or @anakod. We should have a team of core-contributors that we trust. This should speed up development process and scale Sming Framework horizontally.

Real device testing preferred and encouraged, currently real device testing should be optional for features and mandatory for releases. 

### Continuous Integration & Continuous Testing
__Every feature__ and every branch should have a continious build triggered on every commit. I will configure Travis similar to this demo project: https://travis-ci.org/sming-framework/Sming.
This will prevent broken builds before merging and allow us to do other automated things like style checks and doxygen comments validation as a part of a build, as well as unit tests eventually (__TBD__) 

### Releases
[...]

PR #252, opened 2015-09-04
WebSockets CommandProcessing

hreintke:
Implements CommandProcessing for Websockets
Adding example for CommandProcessing (all options)


kireevco:
@hreintke thank you so much for your PR. 
For example _feature/websocket-command-processing_?

hreintke:
@kireevco : As also mentioned in other PR. 
As far as I know I can only push to an existing branch in the anakod/sming github.
If that is created, I will submit this PR to that.
However, when using the feature/... workflow there should be the possibility to create the branch by the original committer.

kireevco:
Closed in favor of #255

Issue #249, opened 2015-09-03
timer callback restrictions and errors


flexiti:
@anakod @hreintke 

I think I found the reason for the appearance of # 198, I also think that the reason many other problems.  #216 ? #222 ?

As soon as I corrected program in accordance with the instructions as follows problems have been resolved.
For example. I measured the DS1820 in a timer callback,

also you have to be careful because:

eg 
	Serial.println("I'm sending ....");  //                          IT'S 2ms
        Serial.println(float);  //                 IT'S  4.2ms
        os_printf("Debug: %x  | %x..%x (%d bytes)\n\r", 1,2,3,4);  //            IT'S 3 ms

also  in many of the examples in the timer callback function using delay()   > 10 ms

If the execution time of the timer call back is more than 500 ms, start TCP errors, and when it reaches 1500-2000 ms, the system hangs

What is your opinion ?
What about threads in Sming ?

-----------------------------------------------------
from SDK pdf:
Notice：
• Using non-OS SDK which is single-threaded, the CPU should not take long to execute tasks:
‣ If a task occupies the CPU too long, ESP8266 can't feed the dog, it will cause a watchdog
reset；
‣ If interrupt is disabled, CPU can only be occupied in us range and the time should not be
more than 10 us; if interrupt is not disabled, it is suggested that CPU should not be
occupied more than 500 ms.
• We suggest using a timer to check periodically, if users need to call os_delay_us or function
while, or function for in timer callback, please do not occupy CPU more than 10 ms.
• Using non-OS SDK, please do not call any function defined with ICACHE_FLASH_ATTR in the
interrupt handler.
• We suggest using RTOS SDK, RTOS to schedule different tasks.

----------------------------------
from SDK 1.3 version info:
4.Optimize software timer to solve the  ----->>>>>  a connecting problem.
Please do not call "os_delay_us" or "while" or "for" to occupy CPU more than 10 ms in timer callback.
http://bbs.espressif.com/viewtopic.php?f=46&t=919

Please read my question in the forum bss http://bbs.espressif.com/viewtopic.php?f=66&t=1051

flexiti:
It seems that other functions can't also be too long  
(eg. connectOk (), connectFail ()).
Therefore, all applications should be event driven  (event not very long < 500 ms ??, Delay function < 10 ms)
or .... threads

hreintke:
@flexiti : I also noticed those remarks in release notes 1.3 and was wondering what the impact of that was for sming and the application it uses.
[...]

PR #256, opened 2015-09-05
Change  "Temperature_DS1820" example

flexiti:
to meet the ESP8266 requirements for use of the Delay() function
 See #249

### DIFF
--- a/Sming/Libraries/DS18S20/ds18s20.cpp
+++ b/Sming/Libraries/DS18S20/ds18s20.cpp
@@ -1,0 +1,170 @@
+/*
+ * ds18s20.cpp
+ *
+ *  Created on: 01-09-2015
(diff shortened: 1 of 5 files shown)


kireevco:
@flexiti Thank you for your contribution, can you please resubmit your PR to  feature/Temperature_DS1820-delay ?

flexiti:
ok, but in my opinion is rather a bug fix. If someone embarks on an adventure with ESP1820, it is often a temperature measurement using DS1820.Present example use Delay (1000)- it makes the system works but is very unstable. It took me several days before I figured out why the system works uncertainly. Espressif warns against the use Delay() in timer callback function

Second ,I don't know if #230 is a well thought out, I have may question, see my first on #230

PR #260, opened 2015-09-06
Resubmitting HW_PWM . 

hrsavla:
It based on ESP8266 SDK PWM api. It uses hardware timer and does not suffer from flickers. Output on LED will be much better.

pwm.h is added in Sming/system/include/espinc/
Also  Sming/system/include/esp_systemapi.h is modified (few lines added)
Above two changes is needed to access ESP8266 SDK PWM api  in Sming.


kireevco:
@hrsavla I'm confused. Is this agains master branch?

kireevco:
closing in favor of #264

PR #263, opened 2015-09-07
Add --dtr 0 --rts 0 options to miniterm.py

avr39-ripe:
On good devboards with proper reset circuit it will reset esp8266 on every miniterm.py reconnect, no harm on other boards.


kireevco:
@avr39-ripe Thanks for redoing the PR, but this is exactly the same one. You need to PR against anakod/sming/feature/miniterm-dtd-rts

PR #268, opened 2015-09-07
Feature/rboot

kireevco:
Creating a master merge PR for @raburton and https://github.com/anakod/Sming/pull/253


robotiko:
Example build Fails in Windows: 
**** Build of configuration Sming for project Basic_rBoot ****
make all 
OC out/build/libmain2.a
make -C /c/tools/Sming/Sming/rboot
make[1]: Entering directory `/c/tools/Sming/Sming/rboot'
CC rboot-stage2a.c
LD /c/tools/sming/Basic_rBoot/out/build/rboot-stage2a.elf
FW /c/tools/sming/Basic_rBoot/out/build/rboot-hex2a.h
CC rboot.c
LD /c/tools/sming/Basic_rBoot/out/build/rboot.elf
FW /c/tools/sming/Basic_rBoot/out/firmware/rboot.bin
make[1]: Leaving directory `/c/tools/Sming/Sming/rboot'
C+ app/application.cpp
C+ app/rBootHttpUpdate.cpp
C+ /c/tools/Sming/Sming/appinit/user_main.cpp
CC /c/tools/Sming/Sming/rboot/appcode/rboot-api.c
CC /c/tools/Sming/Sming/rboot/appcode/rboot-bigflash.c
AR out/build/app.a
LD out/build/app_0.out
c:/espressif/xtensa-lx106-elf/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: cannot open linker script file eagle.rom.addr.v6.ld: No such file or directory
collect2.exe: error: ld returned 1 exit status
make: *** [out/build/app_0.out] Error 1

gernoteger:
after a clean checkout , `make all` from ./Sming fails with:
```
In file included from SmingCore/Network/../../SmingCore/SmingCore.h:46:0,
                 from system/m_printf.cpp:10:
SmingCore/Network/../../SmingCore/Network/rBootHttpUpdate.h:14:23: fatal error: rboot-api.h: No such file or directory
```

raburton:
Doesn't here:
```
raburton@debian:~/dev$ git clone https://github.com/anakod/Sming.git
Cloning into 'Sming'...
remote: Counting objects: 4807, done.
remote: Compressing objects: 100% (57/57), done.
remote: Total 4807 (delta 22), reused 0 (delta 0), pack-reused 4748
Receiving objects: 100% (4807/4807), 28.27 MiB | 3.33 MiB/s, done.
Resolving deltas: 100% (2922/2922), done.
Checking connectivity... done.
raburton@debian:~/dev$ cd Sming
raburton@debian:~/dev/Sming$ git checkout feature/rboot
Branch feature/rboot set up to track remote branch feature/rboot from origin.
Switched to a new branch 'feature/rboot'
raburton@debian:~/dev/Sming$ cd Sming/
raburton@debian:~/dev/Sming/Sming$ make all
AS system/irq_check.s
CC system/flashmem.c
CC Services/FATFS/ff.c
CC Services/SpifFS/spiffs_cache.c
CC Services/SpifFS/spiffs_gc.c
CC Services/SpifFS/spiffs_nucleus.c
CC Services/SpifFS/spiffs_check.c
CC Services/SpifFS/spiffs_hydrogen.c
CC Services/SpifFS/spiffs.c
CC Services/libemqtt/libemqtt.c
CC Libraries/Adafruit_GFX/glcdfont.c
CC Libraries/Adafruit_ILI9341/hspi.c
C+ system/esp_cplusplus.cpp
C+ system/stringconversion.cpp
C+ system/m_printf.cpp
C+ Wiring/WString.cpp
C+ Wiring/WMath.cpp
C+ Wiring/Print.cpp
C+ Wiring/I2cMaster.cpp
C+ Wiring/Stream.cpp
C+ Wiring/IPAddress.cpp
C+ Wiring/SplitString.cpp
C+ Wiring/WShift.cpp
C+ Wiring/FakePgmSpace.cpp
C+ SmingCore/SPISoft.cpp
C+ SmingCore/Clock.cpp
C+ SmingCore/FileSystem.cpp
C+ SmingCore/ESP8266EX.cpp
C+ SmingCore/Interrupts.cpp
C+ SmingCore/PWM.cpp
C+ SmingCore/Digital.cpp
C+ SmingCore/Wire.cpp
C+ SmingCore/Timer.cpp
C+ SmingCore/DataSourceStream.cpp
C+ SmingCore/SystemClock.cpp
C+ SmingCore/Debug.cpp
C+ SmingCore/SPI.cpp
C+ SmingCore/HardwareSerial.cpp
C+ appinit/user_main.cpp
C+ SmingCore/Platform/RTC.cpp
C+ SmingCore/Platform/WDT.cpp
C+ SmingCore/Platform/Station.cpp
C+ SmingCore/Platform/AccessPoint.cpp
C+ SmingCore/Platform/System.cpp
C+ SmingCore/Network/UdpConnection.cpp
C+ SmingCore/Network/TcpServer.cpp
C+ SmingCore/Network/NetUtils.cpp
C+ SmingCore/Network/HttpClient.cpp
C+ SmingCore/Network/WebSocket.cpp
C+ SmingCore/Network/HttpServer.cpp
C+ SmingCore/Network/FTPServerConnection.cpp
C+ SmingCore/Network/NtpClient.cpp
C+ SmingCore/Network/URL.cpp
C+ SmingCore/Network/HttpResponse.cpp
C+ SmingCore/Network/TcpConnection.cpp
C+ SmingCore/Network/rBootHttpUpdate.cpp
C+ SmingCore/Network/HttpRequest.cpp
C+ SmingCore/Network/MqttClient.cpp
C+ SmingCore/Network/TcpClient.cpp
C+ SmingCore/Network/FTPServer.cpp
C+ SmingCore/Network/TelnetServer.cpp
C+ SmingCore/Network/HttpServerConnection.cpp
C+ SmingCore/Network/HttpFirmwareUpdate.cpp
C+ Services/WebHelpers/base64.cpp
C+ Services/WebHelpers/escape.cpp
C+ Services/DateTime/DateTime.cpp
C+ Services/ArduinoJson/ArduinoJson.cpp
C+ Services/CommandProcessing/CommandHandler.cpp
C+ Services/CommandProcessing/CommandOutput.cpp
C+ Services/CommandProcessing/CommandDelegate.cpp
C+ Services/CommandProcessing/CommandExecutor.cpp
C+ Services/cWebsocket/websocket.cpp
C+ Libraries/LiquidCrystal/FastIO.cpp
C+ Libraries/LiquidCrystal/I2CIO.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_I2C.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR2W.cpp
C+ Libraries/LiquidCrystal/LCD.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR3W.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR.cpp
C+ Libraries/IR/IRremote.cpp
C+ Libraries/si4432/si4432.cpp
C+ Libraries/RF24/RF24.cpp
C+ Libraries/WS2812/WS2812.cpp
C+ Libraries/MMA_7455/MMA_7455.cpp
C+ Libraries/SDCard/SDCard.cpp
C+ Libraries/BMP180/BMP180.cpp
C+ Libraries/DHT/DHT.cpp
C+ Libraries/Adafruit_PCD8544/Adafruit_PCD8544.cpp
C+ Libraries/BH1750FVI/BH1750FVI.cpp
C+ Libraries/HMC5883L/HMC5883L.cpp
C+ Libraries/Adafruit_SSD1306/Adafruit_SSD1306.cpp
C+ Libraries/Ultrasonic/Ultrasonic.cpp
C+ Libraries/TFT_ILI9163C/TFT_ILI9163C.cpp
C+ Libraries/I2Cdev/I2Cdev.cpp
C+ Libraries/Adafruit_GFX/Adafruit_GFX.cpp
C+ Libraries/Bounce/Bounce.cpp
C+ Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
C+ Libraries/Mirf/Mirf.cpp
C+ Libraries/OneWire/OneWire.cpp
C+ Libraries/CapacitiveSensor/CapacitiveSensor.cpp
AR out/build/app_app.a
Installing libsming
Done
raburton@debian:~/dev/Sming/Sming$
```

raburton:
Check the expected files are there. How did you switch to the branch? Best guess is it looks like you might have the second PR but not the first.

hreintke:
LS,
I have been out the last days -> missed the discussion, so if duplicate or already decided just mention.
When integrating rboot fully into sming and "easy to use" I would : 
- Make OtaUpdate_CallBack a delegate so also object functions can be callbacks
- Define a rBoot object (or add to System object) which creates a "easy interface" to functionality
- Move the Ota_Update to a rBoot obect  
- Move the Ota_Update callback to the rBoot object (or add an internal callback so Ota_Update can fully take care of update
- Move ota_updater and switch to rboot object
- Move all the spiff mounting code to a rBoot/System.MountSpiff function
- Change SerialCallback code to a CommandCallback, then all the "commands" are usable by Serial or telnet (and shortly websockets) interface. By using Telnet (and websocket) would make it a "full internet" option

I can imagine that the current version (if it complete, tested and ready to merge) be released first but would consider a quick update afterwards (if agreed that my thoughts are positive).

gernoteger:
looks as I don't correctly resolve EXTRA_INCDIR on my lokal insatllation:
I followed your exact procedure, with the same result as before, still getting the error; however, I'm on WIndoes (mingw), and I checked my make database - it shows:
```
EXTRA_INCDIR := -Iinclude -Iinclude -Isystem/include -IWiring -ILibraries -ISmingCore -I/c/Espressif/ESP8266_SDK/../include -I/c/dev/ESP8266/sming/sming_orig/Sming/Sming/rboot -I/c/dev/ESP8266/sming/sming_orig/Sming/Sming/rboot/appcode
```
So probably have to check in my local environmnet - sorry for the inconvenience

raburton:
>I have been out the last days -> missed the discussion, so if duplicate or already decided just mention.

Yes, a little late to the party :-)

>When integrating rboot fully into sming and "easy to use" I would :
    Make OtaUpdate_CallBack a delegate so also object functions can be callbacks
    Define a rBoot object (or add to System object) which creates a "easy interface" to functionality
    Move the Ota_Update to a rBoot obect
    Move the Ota_Update callback to the rBoot object (or add an internal callback so Ota_Update can fully take care of update
    Move ota_updater and switch to rboot object

I think that sounds reasonable, although personally I'm not sure if it's really necessary. If people want to enhance the integration I have no objection. If you want to wrap the API in a more C++/Sming fashion I suggest a separate discussion (perhaps in the forum) to get something that does what everyone wants and in a way people agree with. Then PR against this code after it has been merged, as an enhancement rather than holding up the main integration.

>    Move all the spiff mounting code to a rBoot/System.MountSpiff function

Disagree with that one. Spiffs mounting is really not rBoot specific, it's just that rBoot has necessitated some improvements (made in #244) to spiffs mounting. There are other issues open on github about the flexibility of spiffs mounting (now largely fixed by #244), so if any improvements need to be made to make it easier to use they aren't going to be rBoot specific.

>    Change SerialCallback code to a CommandCallback, then all the "commands" are usable by Serial or telnet (and shortly websockets) interface. By using Telnet (and websocket) would make it a "full internet" option

That's fine, but it's only a sample intended to demonstrate rBoot integration and functionality. There are plenty of other samples to demonstrate fancy Sming things. I'm not saying the sample shouldn't be improved, I'd be happy if it was, perhaps to give it a simple web gui equivalent to the serial interaction, to be a bit more user friendly for those who aren't much into serial comms, but the purpose of the sample is to demonstrate rBoot. In theory any other sample (with fancier features) can be built for rBoot so we don't have to demonstrate everything Sming can do in this sample.

> I can imagine that the current version (if it complete, tested and ready to merge) be released first but would consider a quick update afterwards (if agreed that my thoughts are positive). 

It is ready. I agree - we get this in and out there to a wider audience. Then people can start to enhance it if they want, after they've had a play with it and worked out what they really want to do with it.

Richard.

robotiko:
I do agree with Richard, 
This thing evolved quite a lot, is functional, working, a great improvement over existing OTA  and allows to take full advantage of large flash devices. 
Once it is merged, there is room for improvements but I would not keep it on hold any longer  considering current waiting times for merge. 
Once the big change is in, I bet and hope that the other small improvements will come in much easier because they wont be that risky .

kireevco:
@anakod Come to the discussion?:)

hreintke:
@robotiko @raburton 
Don't misunderstand and maybe I should have put my last line first : 

>I can imagine that the current version (if it complete, tested and ready to merge) 
> be released first but would consider a quick update afterwards 
> (if agreed that my thoughts are positive)

So yes PLEASE go ahead with merging into master asap. It was difficult enough to get to this state.
And discuss my remarks afterwards

gernoteger:
My compile issue is clearly associated with the way SMING_HOME is defined and resolved in Windows. It didn't show up  in the master branch. It will happen when you compile Sming standalone, unmodified in this branch when not beeing on the standard location. I reported this in Issue #278, and I'll create a seperate PR for this issue.

raburton:
@hreintke I did read that line, don't worry. Further development of rBoot in Sming is welcome, not sure where best to discuss that. Possibly open an enhancement issue?
@gernoteger I've got your issue sussed now, thanks for bringing that to my attention, fixed.

anakod:
I think rboot functionality - one of the best way for OTA. About next improvements - we can open new issue (feature request) to talk about that.

Sming/rboot/eagle.rom.addr.v6.ld
```diff
@@ -0,0 +1,347 @@
+PROVIDE ( Cache_Read_Disable = 0x400047f0 );
```
anakod: Why this file needed? SDK already contains ROM defines.

raburton: The copy of rBoot included in the Sming merge is unmodified from the upstream version (ideally it will stay this way, I've made some tweaks for Sming in the upstream rather than make a it a fork). So why is it included upstream then? A good question, it could probably be taken out with a few tweaks to the Makefile, I'll have a look...

raburton: I have a small change that works nicely upstream and in Sming that eliminates the need for both eagle.rom.addr.v6.ld and eagle.app.v6.ld in rBoot. There were some odd windows specific problems with the linker script include previously, so before I push it I'll ask @robotiko if he'd mine testing the patch out for me
```
diff --git a/Basic_rBoot/rom0.ld b/Basic_rBoot/rom0.ld
index 8fdb75d..7b2a9f7 100644
--- a/Basic_rBoot/rom0.ld
+++ b/Basic_rBoot/rom0.ld
@@ -216,4 +216,4 @@ SECTIONS
 }
 
 /* get ROM code address */
-INCLUDE "../Sming/rboot/eagle.rom.addr.v6.ld"
+INCLUDE "../ld/eagle.rom.addr.v6.ld"
diff --git a/Sming/rboot/Makefile b/Sming/rboot/Makefile
index b106d0d..30be645 100644
--- a/Sming/rboot/Makefile
+++ b/Sming/rboot/Makefile
@@ -4,6 +4,7 @@
 #
 
 ESPTOOL2 ?= ../esptool2/esptool2
+SDK_BASE ?= /opt/esp-open-sdk/sdk
 
 RBOOT_BUILD_BASE ?= build
 RBOOT_FW_BASE    ?= firmware
@@ -17,8 +18,8 @@ LD := $(addprefix $(XTENSA_BINDIR)/,xtensa-lx106-elf-gcc)
 endif
 
 CFLAGS    = -Os -O3 -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals  -D__ets__ -DICACHE_FLASH
-LDFLAGS   = -nostdlib -Wl,--no-check-sections -u call_user_start -Wl,-static
-LD_SCRIPT = eagle.app.v6.ld
+LDFLAGS   = -nostdlib -Wl,--no-check-sections -u call_user_start -Wl,-static -L $(SDK_BASE)/lib/
+LD_SCRIPT = $(SDK_BASE)/ld/eagle.app.v6.ld
 
 E2_OPTS = -quiet -bin -boot0
 
diff --git a/Sming/rboot/rboot-stage2a.ld b/Sming/rboot/rboot-stage2a.ld
index b263c71..d494212 100644
--- a/Sming/rboot/rboot-stage2a.ld
+++ b/Sming/rboot/rboot-stage2a.ld
@@ -208,4 +208,4 @@ SECTIONS
 }
 
 /* get ROM code address */
-INCLUDE "eagle.rom.addr.v6.ld"
+INCLUDE "../ld/eagle.rom.addr.v6.ld"
```
And delete:
`Sming/Sming/rboot/eagle.app.v6.ld`
`Sming/Sming/rboot/eagle.rom.addr.v6.ld`

robotiko: @raburton 
I take note of my homework :)
Let me check the patch and I will get back with the results.

robotiko: @raburton  
I cannot apply the patch with the main repo /feature/rboot  (fa66616610)
I guess it is to be applied in your repo but commits hash doesn't seem to match.
Please let me know the reference repo and commit hash. 

Thanks

raburton:
I was going to say don't use fit to apply the patch just use the patch command, but then you're on windows so you may not have it. The diff is against the feature/rBoot branch in my fork of sming, but if you can't get it going you could just manually apply the changes. There are only about 6 lines to change and two files to delete.

Sming/rboot/eagle.rom.addr.v6.ld
raburton: I was going to say don't use git to apply the patch just use the patch command, but then you're on windows so you may not have it. The diff is against the feature/rBoot branch in my fork of sming, but if you can't get it going you could just manually apply the changes. There are only about 6 lines to change and two files to delete.

robotiko: I was using eclipse patch,  but the main point was to know the reference code  for a useful test. 
I let you know.

raburton: Just the latest commit on the feature/rboot branch in raburton/Sming:
```
raburton@debian:~/dev/Sming$ git remote -v
origin	https://raburton@github.com/raburton/Sming.git (fetch)
origin	https://raburton@github.com/raburton/Sming.git (push)
raburton@debian:~/dev/Sming$ git branch -v
* feature/rboot c53fbcf makefile fix closes #278 plus other tweaks
  master        c000e9c actually use SPIFF_BIN_OUT and allow override
```
I notice if you copy and paste the above diff it loses the single space char at the start of each blank line in code, so that might be part of the problem. Could try patch file [here] (https://dl.dropboxusercontent.com/u/5500141/robotiko.diff) instead.

robotiko: Ok, tested. 
-SmingFramework builds. 
-Basic_rBoot builds. Tested on physical device, it boots OTA works, all working. 

Any other Sming example failing. 
As example: 
19:00:03 **** Build of configuration Sming for project Basic_Serial ****
make all 
C+ app/application.cpp
In file included from c:/tools/Sming/Sming/SmingCore/SmingCore.h:46:0,
                 from app/application.cpp:2:
c:/tools/Sming/Sming/SmingCore/Network/rBootHttpUpdate.h:14:23: fatal error: rboot-api.h: No such file or directory
 #include <rboot-api.h>
                       ^
compilation terminated.
make: *** [out/build/app/application.o] Error 1

19:00:04 Build Finished (took 557ms)

I recall this was expected as side effect of the new Makefile.

robotiko: @raburton, 
copied the Basic_rBoot Makefile and the Makefile-user.mk  setting RBOOT_ENABLED ?= 0

and keep getting : 
c:/tools/Sming/Sming/SmingCore/Network/rBootHttpUpdate.h:14:23: fatal error: rboot-api.h: No such file or directory
 #include <rboot-api.h>
                       ^

Something broke somewhere :D 
Any hint?

raburton: Yeah, fixed. Add $(SMING_HOME)/rboot $(SMING_HOME)/rboot/appcode to the end of EXTRA_INCDIR in the Makefile-project.mk file. As you've tested the change to rBoot ld files on Widnows (which was my only real concern) I'll push that change now (and this little one).

raburton:
@anakod I have removed the offending file (eagle.rom.addr.v6.ld) from rBoot and as a bonus also eagle.app.v6.ld. Great thanks to @robotiko for Windows testing. The updated rBoot has been added to the existing PR #279, waiting to go into feature/rboot. Once that's in hopefully we are somewhere near a merge into master!

robotiko:
ok , so it seem that the whole rboot is ready to merge?
@anakod, with latest changes is all ok?

The sooner this is merged,the better. 
Current repo with several PR some merged in /feature/rboot, some environment related in  head, some bugfixes waiting to be merged in the feature branch...  
@kireevco  can you merge the PR bugfixes in the PR?  right now we  need a "treasure map" to have a test version. 

Thxs

raburton:
What do you mean about environment stuff in head?

robotiko:
@raburton, 

I was referring to the fixes previous to the rboot integration that where merged in head some time ago. 
Reviewing them, were pure Sming bugfixes related to spiffs and file mounting, not just  rBoot requirements (#244,#273..) .
The whole feature branch would be self contained once the pending PR in the rboot PR get merged.  

updated the comment to avoid "false warning trigger". 

Sorry for raising the alert

raburton:
I see. Yes, this is where things stand / will stand:
1) PR #279 needs to be merged into feature/rboot
2) feature/rboot will then contain the complete rboot, integration and sample
3) perhaps a short final round of testing is needed once everything is in the same branch again
4) then merge feature/rboot to master.

Thoughts on the workflow:
It's a bit of a pain when you're working on something non-trivial, because you need help testing. People don't seem to have wanted to switch to a fork repository, they want it in the feature branch in the main repository. A contributor doesn't have write access to that, so they have to create a fork and make multiple pull requests between the feature branch in the fork and the main repo, which can take days to be merged. The original document on this workflow appeared to assume contributors had write access to the repo and weren't working with forks. If a contributor could get write access to the feature branch that has been created for them, so they could work directly with it, I think this would work a lot better. Either do the development in a fork and PR to master in the main repo or develop in a feature branch in the main repo, doing both just complicates and slows things down.

piperpilot:
I hope this is the right PR to comment against.   I just pulled all of the latest from Sming feature/rboot and tried to build one of the standard examples "HttpServer_ConfigNetwork" and get the following:

`
16:00:06 **** Build of configuration Sming for project HttpServer_ConfigNetwork ****
make all 
C+ app/application.cpp
In file included from /Users/cpope/git/Sming/Sming/SmingCore/SmingCore.h:46:0,
                 from app/application.cpp:2:
/Users/cpope/git/Sming/Sming/SmingCore/Network/rBootHttpUpdate.h:14:23: fatal error: rboot-api.h: No such file or directory
 #include <rboot-api.h>
                       ^
compilation terminated.
make: *** [out/build/app/application.o] Error 1

16:00:06 Build Finished (took 115ms)
`

rboot-api.h exists outside of that directory...seems a true relative path needs to be provided.

I tried unwinding it and providing paths to get there, but kept running into issues.

robotiko:
Fixed in PR #279

raburton:
Yep, we are waiting for PR #279 to be accepted into the feature/rBoot branch, until then there is no complete branch you can directly checkout.

raburton:
#279 has been merged. All fixes and requested changes are now in feature/rboot. I believe this is now complete and if people wouldn't mind giving it a quick spin and percussing feedback we can hopefully get it merged to master soon.

piperpilot:
OK...did some more testing...ran into an issue...here's what I did....

1.  Take the HttpServer_ConfigNetwork example
2.  Add the following to the bottom of Makefile-user.mk

```
#### overridable rBoot options ####
## use rboot build mode
RBOOT_ENABLED ?= 1
## enable big flash support (for multiple roms, each in separate 1mb block of flash)
RBOOT_BIG_FLASH ?= 1
## two rom mode (where two roms sit in the same 1mb block of flash)
#RBOOT_TWO_ROMS  ?= 1
## size of the flash chip
SPI_SIZE        ?= 4M
## output file for first rom (.bin will be appended)
#RBOOT_ROM_0     ?= rom0
## input linker file for first rom
#RBOOT_LD_0      ?= rom0.ld
## these next options only needed when using two rom mode
#RBOOT_ROM_1     ?= rom1
#RBOOT_LD_1      ?= rom1.ld
## size of the spiffs to create
#SPIFF_SIZE      ?= 65536
## option to completely disable spiffs
#DISABLE_SPIFFS  = 1
## flash offsets for spiffs, set if using two rom mode or not on a 4mb flash
## (spiffs location defaults to the mb after the rom slot on 4mb flash)
#RBOOT_SPIFFS_0  ?= 0x100000
#RBOOT_SPIFFS_1  ?= 0x300000
## esptool2 path
#ESPTOOL2        ?= esptool2
## path to spiffy
#SPIFFY          ?= spiffy
```
3.  Change Makefile to use the rboot makefile:
```# Include main Sming Makefile
include $(SMING_HOME)/Makefile-rboot.mk```

4.  Clean and build and get the following output.

```
20:47:34 **** Build of configuration Sming for project HttpServer_ConfigNetwork ****
make all 
OC out/build/libmain2.a
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C /Users/cpope/git/Sming/Sming/rboot
CC rboot-stage2a.c
LD /Users/cpope/git/Sming/HttpServer_ConfigNetwork/out/build/rboot-stage2a.elf
FW /Users/cpope/git/Sming/HttpServer_ConfigNetwork/out/build/rboot-hex2a.h
CC rboot.c
LD /Users/cpope/git/Sming/HttpServer_ConfigNetwork/out/build/rboot.elf
FW /Users/cpope/git/Sming/HttpServer_ConfigNetwork/out/firmware/rboot.bin
C+ app/application.cpp
C+ /Users/cpope/git/Sming/Sming/appinit/user_main.cpp
CC /Users/cpope/git/Sming/Sming/rboot/appcode/rboot-api.c
CC /Users/cpope/git/Sming/Sming/rboot/appcode/rboot-bigflash.c
AR out/build/app_app.a
LD out/build/app_0.out
/opt/esp-open-sdk/xtensa-lx106-elf/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld: cannot open linker script file rom0.ld: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [out/build/app_0.out] Error 1

20:47:35 Build Finished (took 1s.77ms)
```

I know that has been addressed before...not sure if it got stomped on with one of the merges.  I am on the head of feature/rboot.

I'm trying to build this example because I'm running into an issue with the WifiStation in my own project not scanning properly for networks.  Wanted to see if something got broken with the latest rboot changes which I just pulled in around the time my app broke.

**Comment by piperpilot** [comment]
There seems to be some kind of major bug in the latest version.  I noticed it in my own project, but just confirmed it by integrating rboot with HttpServer_ConfigNetwork.  Seems that maybe the SDK functions aren't being called correctly or something.  The initial place I noticed it is in the ScanNetworks routine.  HttpServer_ConfigNetwork calls this:

```
	WifiStation.startScan(networkScanCompleted);
```

Problem is, it never seems to either get called or completed.  With the HttpServer_ConfigNetwork example, the serial output is supposed to look like this:

```
Set AP configuration in background
mode : sta(18:fe:34:fe:50:cf) + softAP(1a:fe:34:fe:50:cf)
add if0
add if1
dhcp server start:(ip:192.168.4.1,mask:255.255.255.0,gw:192.168.4.1)
bcn 100
f 0, addUser: me 123
'/' registered
'/ipconfig' registered
'/ajax/get-networks' registered
'/ajax/connect' registered
AP configuration was updated on system ready event
scandone
f 0, scandone
scan completed: 5 found
```

With rboot integrated, it looks like this:

```
Set AP configuration in background
mode : sta(18:fe:34:fe:50:cf) + softAP(1a:fe:34:fe:50:cf)
add if0
add if1
dhcp server start:(ip:192.168.4.1,mask:255.255.255.0,gw:192.168.4.1)
bcn 100
f 0, addUser: me 123
'/' registered
'/ipconfig' registered
'/ajax/get-networks' registered
'/ajax/connect' registered
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 2
cnt 
```

You notice that after scandone, the statement "scan completed: 5 found" is missing and the callback is never triggered.

I have confirmed that disabling rboot and using the standard makefile fixes the issue...so what is missing from the rboot makefile that could be causing this?

**Comment by raburton** [comment]
I can't imagine how that could be related to rBoot at all. I'll have a look when I get home, but you could compare the map films and see if there is any obvious difference between the two. Are you saying building it for rBoot before the last commit worked but with the last PR it stopped working?

**Comment by piperpilot** [comment]
Yeah, I'm not sure exactly what point this broke...but prior to the change where rBootHttpUpdate was moved into Sming Core, everything worked fine.  I suspect there is something in the makefile that is missing or changed.  When you say "compare the map films"...not exactly sure what you are asking there...if you can provide a little bit of detail, I can surely do a comparison.  I see an app_0.map file...I will do a compare there and see if that is what you are looking for.

**Comment by piperpilot** [comment]
I have no idea what I am looking at, but here are the two map files.  firmware.map is the working version WITHOUT rboot, app_0.map is the rboot version with the problem.

Doesn't look like I can attach the files, so I put them on dropbox here:

https://www.dropbox.com/s/15rs1x28n1zpdxm/map%20files.zip?dl=0

**Comment by raburton** [comment]
First thing I notice is that you're linking against /opt/esp-open-sdk/sdk/lib/libmain2.a
That file shouldn't exist and appears to be left over from previous playing with rBoot. The Sming integration creates libmain2 in the build directory. It's possible the old version you are linking against is out of sync with the rest of your sdk.

**Comment by piperpilot** [comment]
OK, I cleared that out and can now see that it is linking against out/build/libmain2.a but that didn't seem to solve the problem.

**Comment by raburton** [comment]
Found another interesting difference. In the non-rboot map your user_init function function is coming from libsming.a. In your rBoot version it is coming from app_app.a. The way this is supposed to work, although I don't know why it is done like this, is that Sming/Sming/appinit is compiled into both libsming and your app. This is existing behaviour and not related to rBoot changes. Presumably the one compiled into your app is supposed to take precedence over the one compiled into libsming, however it does not. Both are marked as weak so which ever is encountered first by the linker will be used. In the non-rBoot build this is the libSming version, in the rBoot build it is the app version. This appears to be a bug in base Sming. I don't know if it has any bearing on your problem, it's hard to imagine it does, but you can easily test it out. In Makefile-rboot.mk remove "-u user_init" from LDFLAGS. When you build check the map file for the string "(user_init)" to see where this function if being found.

**Comment by raburton** [comment]
The rboot build changes the link order to include the app before the libraries. There could be other similar bugs lurking in Sming that are very sensitive to this order. Assuming the above doesn't help you (I suspect it won't unfortunately), try this. Edit Makefile-rboot.mk and swap the order on the linker line from "$(APP_AR) $(LIBS)" to "$(LIBS) $(APP_AR)" this is the same order as the non-rBoot build. Give that a spin, but you'll only be able to work with rom0 because big flash support won't work with this link order. In the mean time I'm going to try to replicate your problem...

**Comment by piperpilot** [comment]
I don't think its link order because I had it that way before.  I had all of this working previously, but due to the multiple PRs, etc I'm not exactly sure when it broke.  I'll look at the user_init LDFLAGS and see what thats all about.  I thought the way its supposed to work is user_init from the library is supposed to execute first and then the user_init from the app.  In the library it sets up the timers, initializes the system, etc.  Maybe thats the problem.  Here's the pertinent user_init code....

```
	system_timer_reinit();
	uart_div_modify(UART_ID_0, UART_CLK_FREQ / 115200);
	cpp_core_initialize();
	System.initialize();
#if !defined(DISABLE_SPIFFS) && !defined(DISABLE_SPIFFS_AUTO)
	spiffs_mount();
#endif
	init(); // User code init
```

**Comment by raburton** [comment]
Oh yes, user_init() runs first before calling the init() function in your app. But the user_init() function is also compiled into your app (the exact same one as the one in libsming). So it's in libsming and the app itself. Only one will go through to the final binary, so presumably it's supposed to be the copy that gets compiled into the app. The only reason I can see for doing that is to allow users to modify the behaviour of user_init with #defines (I added DISABLE_SPIFFS_AUTO myself to take advantage of this behaviour) otherwise you'd be stuck with the version compiled when libsming was created. However that doesn't actually work with base Sming, to get it to work you have to put the app before the libs on the linker command and undefine (the -u bit) it so it does a proper job of looking for it in your app.

**Comment by piperpilot** [comment]
OK...update.  Removing the -u user_init from the LDFLAGS seems to have fixed it.  I'll do more testing, but its probably something we want to change before this is committed to master.  What was the purpose of adding that in?  Is that so that it picks up the defines?

**Comment by raburton** [comment]
Confirmed, that fixes it.
Without it the copy of user_init from libsming will be used in your app, which will contain the spiffs_mount call (the bad one that looks for the spiffs is a calculated location and may create and format a new one there). With that flag the copy of user_init will be the newly compiled one, that respects the defines (so without the spiffs_mount call). I've tested removing the spiffs_mount call from Sming/Sming/appinit/user_main.cpp and compiling it into libsming to check that the problem isn't caused by not having called the function.

**Comment by raburton** [comment]
Ok, problem appears to exists in base Sming. If you use "-u user_init" and put $(APP_AR) before $(LIBS) on the non-rBoot build you get the same problem. Obviously that's not how the current non-rBoot build does things, but the current build override for user_init doesn't actually work if you don't make that change. So the addition of the Sming user_init.cpp to the modules list in Makefile-project.mk so that it gets built as part of the users app (as currently happens in Sming) is completely pointless - it never actually makes it into the binary! So either way we've found a bug in base Sming and need to find a way to work around it for rBoot.

**Comment by raburton** [comment]
Opened #294 for the issue, it'll be interesting to see what feedback that gets. What's still odd though, is that if we force it to use the local version of the function it causes some network functionality to stop working. I really don't understand that, unless there are more places where overrides are in use, e.g. Sming overriding SDK that perhaps is also subject to some gcc flakyness.

**Comment by raburton** [comment]
Ok, I'm a really stuck on this one. If you force the copy of user_init that has been built with the app to be used (by having `-u user_init` in the LDFLAGS) the it pulls in library functions in a different order to if you allow it use the one in libsming. This appears to cause the AP ready callback to fail to get called. This does not make the slightest bit of sense! If you don't force this version of user_init then 99% of the time it will use the one from Sming which contains the bad mount call. 1% of the time it will use the right version and by some magic when this happens everything works fine. If I compile a clean (totally untouched by rBoot) copy of Sming and the HttpServer_ConfigNetwork sample it will mostly pull in user_init from libsming, but occasionally from the app like it should. I've tried setting the app compiled version as global and it still fails to override the weak version in libsming. I think this must be a gcc bug, but we have to use a group for ours libraries to resolve circular dependencies, so maybe it isn't supposed to work in this scenario.

So I have no idea how to fix #294. That isn't an rBoot bug but it has consequences for rBoot - we can't prevent the libsming copy of user_init trying to mount a spiffs at the wrong location. the only fix I can suggest is to recompile libsming without the mount call in it when using rBoot (perhaps with a nice build option, not a manual hack) or the auto mounting needs to be taken out altogether and made a user call for the apps init function. But that would be a change to Sming that would affect existing users. Any thoughts???

**Comment by hreintke** [comment]
@ADiea : 
You have been working on the user_init and weak attribute.
Do you have any thoughts on this ?

**Comment by ADiea** [comment]
Hi

will take a look

**Comment by ADiea** [comment]
The relevant PR is #208 where I implemented the possibility to bypass SPIFFS because I use SDcard FAT32 fs on my project.

The first time I proposed a solution to remove the spiffmount altogether from user_init and include it in init() of sample projects that needed spiffs but on @anakod 's suggestion I left the mount there so that new users don't have to worry about why spiff is not working on their new project. 

So instead made the DISABLE_SPIFF makefile switch + made the user_init weak to maybe use it in the future

In my project I don't override user_init so I cannot provide feedback if the gcc is doing its job or not but from my tests back then it looked that the proper user_init is called when defined in the application.

**Comment by raburton** [comment]
@ADiea thanks for looking at this. Really there are two issues:

1) The fact that the right user_init doesn't get pulled in most of the time (for which I have opened #294). When you say you don't override user_init, do you mean you don't override it again in your own code or that you aren't making use of the override the comes with the samples (the compilation of user_init.cpp into the local project, probably after defining DISABLE_SPIFF)? I'm not overriding it with my own code, just trying to use the standard override with DISABLE_SPIFF. But that is failing - a look at the map file shows it isn't using the locally compiled version but linking in the copy that's compiled into libsming.

2) It is possible to force it use the correct user_init by rearranging the order of libs on the link command and using `-u user_init` then all seems well until you run the rom then the AP ready callback never gets called. This seems like a pretty random problem I know!

**Comment by ADiea** [comment]
@raburton 

1)
I mean that i don't override it again in my code, I do not provide an implementation of the user_init in my code then it will use the weak symbol from libsming
I don't understand what override you are referring to that comes with the samples - do you mean the Sming samples? When making the final project I do not make use of the samples, only libsming

I also compile sming lib with DISABLE_SMING then the sming user_init() will not mount siffs - so in short I'm not making use of the weak attribute but compile everything with spiff_disable

2)
Could it be that AP ready callback location is differing between the 2 cases when comparing the map files? Maybe it will not fit in it's section or maybe esp cannot execute code from that address? Have you tried with standard 512KB flash? FOr my tests I use a esp12 ebay module with 512KB flash kind of "stock" i guess:)

**Comment by raburton** [comment]
1) I see. That makes sense and is probably what we're going to have to do for rBoot. But what I was referring to when I mentioned the existing override is that the Makefile-project.mk (as used by all the samples) adds Sming/appinit to the modules list. This causes user_init.cpp to be compiled again with the app. When an app is then linked there is one weak user_init in libsming and another (compiled from the same source file) in the app.a file (also weak) for the linker to choose from. Why is this? I assumed it was so that you could override the libsming version with a version compiled with per app defines (such as DISABLE_SPIFF), rather than having to re-compile libsming with the defines set. However that doesn't happen, it almost always uses the copy from libsming. Even making the app version global instead of weak still doesn't cause it to override the weak on in libsming, so the linker really isn't doing it's job properly.

2) I agree it must be some odd quirk of the linking. But probably not the position on flash - my first rom is all sat within the first 256k. Also, if it the code doesn't fit in a section the linker should throw an error (and it shouldn't affect the size as it's the same code). I can only assume that being forceful with the linker is having some other effect, perhaps other weak symbols are not getting correctly overridden by global versions between the libraries.

**Comment by hreintke** [comment]
@raburton : @ADiea : @anakod : 
First : I am not experienced in this so if the text below doesn't make sense just let me know.
1/ Raburton, "Even making the app version global instead of weak" does that mean you remove the weak attribute from applications user_init ? If not, is making sure that app version is taken not handled by that one "not having the weak attribute ?
2/ If EVERY sming application includes the user_init function. Why then not "just remove" from sming core. I removed appinit from MODULES statement in Sming Makefile and could compile both sming and application. Did not run and test !

**Comment by piperpilot** [comment]
So the only reason we are having to do this is to disable the default SPIFFs_Mount.  My vote is to remove spiffs_mount from the init and let the user call it if they need SPIFFs.  There are a LOT of applications that don't/won't use SPIFFs, so it is a wrong assumption that it should be mounted by default every time.

The other things in there...timer, core, system...those are needed by every app.

Just my two cents...

**Comment by raburton** [comment]
@hreintke 1) Yes, I removed the weak attribute from the function before the compile of the app, so the one that ends up in app.a is not weak. It still doesn't get chosen, instead the weak one from libsming is used.
2) Yes, it could either be removed from libsming or from the list of modules to compile for the app.

@piperpilot correct and I agree with you, especially as spiff_mount can be destructive. If the mount command finds 0xffffffff where it expects the spiffs to be it will format a new spiffs at that location, possibly wiping out half your chip. However this change will require users who are already used to auto mounting of spiffs to add an extra line to their apps, which will need to be communicated somehow.

**Comment by raburton** [comment]
@hreintke I've tested removing `$(SMING_HOME)/appinit` from the modules list and it literally makes no difference - identical map files produced and tested it on hardware to be sure.

@piperpilot Having done the above in Makefile-rboot.mk and then removed the `spiffs_mount` call in app_init.cpp (and then recompiled sming, which I initially forgot to do!), everything works fine. I then just added the appropriate `spiffs_mount_manual` call to my apps init function and I'm serving pages.

I'd really like to remove spiffs_mount from user_init, it's inappropriate to be there and potentially destructive if it doesn't find a spiffs where it expects to. But I think that change will need higher approval from @kireevco or @anakod

**Comment by hreintke** [comment]
@raburton @piperpilot 
I we then take a two step approach.
1/ Remove user_init from sming keep it in application.
Keep the spiffs compile options and default spiffs_mount in user_init.
That produces a working environment, compatible with current (no break of current) and possibility to remove by setting env variables
Add the spiff_mount_manual to prepare for future updates.
2/ Move user_init from application to sming, removing the spiffs_mount.

If we do 1/ now and 2/ in a following release (and communicate correctly) it would give a transition process

**Comment by raburton** [comment]
Yes, this could work, but it's a two step process when it could be done in one. When would you do step two? You say in a following release, but we are working on the development branch so anything that goes in now doesn't officially land until the next release anyway. I know many people probably use Sming direct from github, but if you're using the development version you should expect some occasional changes.

I can do some more testing to check this way around works too, no guarantee!

**Comment by hreintke** [comment]
Sming is moving to more specific releases. But the idea was (as the last functionality to go this way) to merge rboot branch to  master. That what I see then includes step 1.
After that, new functionality would go from PR -> Develop -> Release. In the first of that process I see step 2.

**Comment by raburton** [comment]
Ok, I'll do some testing with removing it from libsming and leaving it in the app compilation and see what we get...

**Comment by anakod** [comment]
I'm sure we shouldn't mix different problems together.
If this P-R ready to merge we will merge it and then will think about user_init issues. How you think?

**Comment by raburton** [comment]
@anakod I think that's a fair point, the user_init issues can be dealt with under bug #294. In the mean time I have implementation a workaround for the gcc oddity associated with the override and a build flag to disable the auto-mount at build time for libsming. PR #296 against feature/rboot contains the changes, if that could be merged in to the feature branch asap we can get @piperpilot to confirm it fixes his issue (I've tested it here already of course) and then hopefully get a merge of the whole thing to master soon.

**Comment by raburton** [comment]
@piperpilot fix for the GCC linking issue now merged. Can you respin your tests. Thanks.

**Comment by piperpilot** [comment]
@raburton Yes, I will give this a try later tonight (US time) or tomorrow and report back.

**Comment by anakod** [comment]
@piperpilot, did you have a news about linking issue check? I think we can merge this branch now, but will be very fine to have community feedback before.

**Comment by raburton** [comment]
@piperpilot @robotiko @hreintke 
Guys? Where did everyone go? Are we good to merge to master?

**Comment by piperpilot** [comment]
Hey guys...sorry for being offline...I had to go out of town on Sunday unexpectedly  I forgot to bring my ESP module with me.  I'll be back tomorrow and can finish testing, but I did a code review and I think its fine.

**Comment by raburton** [comment]
Thanks.

**Comment by piperpilot** [comment]
I can confirm that this seems to be a good fix!  Ship it!

**Comment by flexiti** [comment]
"Set ESPTOOL2 (env var or in the Makefile) to point to the esptool2 binary."

Is it not included in the project ? why ?
pp1

**Comment by raburton** [comment]
> "Set ESPTOOL2 (env var or in the Makefile) to point to the esptool2 binary."
Is it not included in the project ? why ?

It is a separate external dependency, just like spiffy, the compiler and the sdk.

**Comment by raburton** [comment]
FYI: tested with sdk 1.4, sming, rboot, spiffs, bigflash all working fine.

**Comment by flexiti** [comment]
spiffs is included in sming or I am wrong ?

**Comment by flexiti** [comment]
I had a free half hour and I could not even run rboot, is a little complicated especially since all other demo programs to run sming require a few minutes of work - but I'm not discouraged  ;)

**Comment by raburton** [comment]
>spiffs is included in sming or I am wrong ?

spiffy, the tool for making spiffs, is not included in Sming

>I had a free half hour and I could not even run rboot, is a little complicated especially since all other demo programs to run sming require a few minutes of work - but I'm not discouraged ;)

Assuming you have a 4MB esp12 you just set a couple of environment variables (path type stuff) as explained in the readme file, then build and flash like any other sample. If you are using a different size flash there are a couple of other variable to tweak and these are documented in the readme. It really is no more complicated than the other samples. I can't imagine how you have not managed to do it in half an hour, but if you need talking through it in more detail this is probably not the place for it, send me a message. It takes me less than two minutes to build sming, build the sample and flash it, and that's not because I'm more familiar with it - I'm not doing anything clever I'm just running the make commands!
```
raburton@debian:~/dev/Sming$ export WIFI_SSID=MySSID
raburton@debian:~/dev/Sming$ export WIFI_PWD=MyPasswd
raburton@debian:~/dev/Sming$ export SMING_HOME=/home/raburton/dev/Sming/Sming/
raburton@debian:~/dev/Sming$ export ESP_HOME=/opt/esp-open-sdk/
raburton@debian:~/dev/Sming$ export SPIFFY=/home/raburton/dev/spiffy/build/spiffy
raburton@debian:~/dev/Sming$ export ESPTOOL2=/home/raburton/dev/esp8266/esptool2/esptool2
raburton@debian:~/dev/Sming$ cd Sming/
raburton@debian:~/dev/Sming/Sming$ make
AS system/irq_check.s
CC system/flashmem.c
CC Services/FATFS/ff.c
CC Services/SpifFS/spiffs_cache.c
CC Services/SpifFS/spiffs_gc.c
CC Services/SpifFS/spiffs_nucleus.c
CC Services/SpifFS/spiffs_check.c
CC Services/SpifFS/spiffs_hydrogen.c
CC Services/SpifFS/spiffs.c
CC Services/libemqtt/libemqtt.c
CC Libraries/Adafruit_GFX/glcdfont.c
CC Libraries/Adafruit_ILI9341/hspi.c
C+ system/esp_cplusplus.cpp
C+ system/stringconversion.cpp
C+ system/m_printf.cpp
C+ Wiring/WString.cpp
C+ Wiring/WMath.cpp
C+ Wiring/Print.cpp
C+ Wiring/I2cMaster.cpp
C+ Wiring/Stream.cpp
C+ Wiring/IPAddress.cpp
C+ Wiring/SplitString.cpp
C+ Wiring/WShift.cpp
C+ Wiring/FakePgmSpace.cpp
C+ SmingCore/SPISoft.cpp
C+ SmingCore/Clock.cpp
C+ SmingCore/FileSystem.cpp
C+ SmingCore/ESP8266EX.cpp
C+ SmingCore/Interrupts.cpp
C+ SmingCore/PWM.cpp
C+ SmingCore/Digital.cpp
C+ SmingCore/Wire.cpp
C+ SmingCore/Timer.cpp
C+ SmingCore/DataSourceStream.cpp
C+ SmingCore/SystemClock.cpp
C+ SmingCore/Debug.cpp
C+ SmingCore/SPI.cpp
C+ SmingCore/HardwareSerial.cpp
C+ appinit/user_main.cpp
C+ SmingCore/Platform/RTC.cpp
C+ SmingCore/Platform/WDT.cpp
C+ SmingCore/Platform/Station.cpp
C+ SmingCore/Platform/AccessPoint.cpp
C+ SmingCore/Platform/System.cpp
C+ SmingCore/Network/UdpConnection.cpp
C+ SmingCore/Network/TcpServer.cpp
C+ SmingCore/Network/NetUtils.cpp
C+ SmingCore/Network/HttpClient.cpp
C+ SmingCore/Network/WebSocket.cpp
C+ SmingCore/Network/HttpServer.cpp
C+ SmingCore/Network/FTPServerConnection.cpp
C+ SmingCore/Network/NtpClient.cpp
C+ SmingCore/Network/URL.cpp
C+ SmingCore/Network/HttpResponse.cpp
C+ SmingCore/Network/TcpConnection.cpp
C+ SmingCore/Network/rBootHttpUpdate.cpp
C+ SmingCore/Network/HttpRequest.cpp
C+ SmingCore/Network/MqttClient.cpp
C+ SmingCore/Network/TcpClient.cpp
C+ SmingCore/Network/FTPServer.cpp
C+ SmingCore/Network/TelnetServer.cpp
C+ SmingCore/Network/HttpServerConnection.cpp
C+ SmingCore/Network/HttpFirmwareUpdate.cpp
C+ Services/WebHelpers/base64.cpp
C+ Services/WebHelpers/escape.cpp
C+ Services/DateTime/DateTime.cpp
C+ Services/ArduinoJson/ArduinoJson.cpp
C+ Services/CommandProcessing/CommandHandler.cpp
C+ Services/CommandProcessing/CommandOutput.cpp
C+ Services/CommandProcessing/CommandDelegate.cpp
C+ Services/CommandProcessing/CommandExecutor.cpp
C+ Services/cWebsocket/websocket.cpp
C+ Libraries/LiquidCrystal/FastIO.cpp
C+ Libraries/LiquidCrystal/I2CIO.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_I2C.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR2W.cpp
C+ Libraries/LiquidCrystal/LCD.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR3W.cpp
C+ Libraries/LiquidCrystal/LiquidCrystal_SR.cpp
C+ Libraries/IR/IRremote.cpp
C+ Libraries/si4432/si4432.cpp
C+ Libraries/RF24/RF24.cpp
C+ Libraries/WS2812/WS2812.cpp
C+ Libraries/MMA_7455/MMA_7455.cpp
C+ Libraries/SDCard/SDCard.cpp
C+ Libraries/BMP180/BMP180.cpp
C+ Libraries/DHT/DHT.cpp
C+ Libraries/Adafruit_PCD8544/Adafruit_PCD8544.cpp
C+ Libraries/BH1750FVI/BH1750FVI.cpp
C+ Libraries/HMC5883L/HMC5883L.cpp
C+ Libraries/Adafruit_SSD1306/Adafruit_SSD1306.cpp
C+ Libraries/Ultrasonic/Ultrasonic.cpp
C+ Libraries/TFT_ILI9163C/TFT_ILI9163C.cpp
C+ Libraries/I2Cdev/I2Cdev.cpp
C+ Libraries/Adafruit_GFX/Adafruit_GFX.cpp
C+ Libraries/Bounce/Bounce.cpp
C+ Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
C+ Libraries/Mirf/Mirf.cpp
C+ Libraries/OneWire/OneWire.cpp
C+ Libraries/CapacitiveSensor/CapacitiveSensor.cpp
AR out/build/app_app.a
Installing libsming
Done
raburton@debian:~/dev/Sming/Sming$ cd ../Basic_rBoot/
raburton@debian:~/dev/Sming/Basic_rBoot$ make
OC out/build/libmain2.a
make -C /home/raburton/dev/Sming/Sming//rboot
make[1]: Entering directory '/home/raburton/dev/Sming/Sming/rboot'
CC rboot-stage2a.c
LD /home/raburton/dev/Sming/Basic_rBoot/out/build/rboot-stage2a.elf
FW /home/raburton/dev/Sming/Basic_rBoot/out/build/rboot-hex2a.h
CC rboot.c
LD /home/raburton/dev/Sming/Basic_rBoot/out/build/rboot.elf
FW /home/raburton/dev/Sming/Basic_rBoot/out/firmware/rboot.bin
make[1]: Leaving directory '/home/raburton/dev/Sming/Sming/rboot'
C+ app/application.cpp
C+ /home/raburton/dev/Sming/Sming//appinit/user_main.cpp
CC /home/raburton/dev/Sming/Sming//rboot/appcode/rboot-bigflash.c
CC /home/raburton/dev/Sming/Sming//rboot/appcode/rboot-api.c
AR out/build/app_app.a
LD out/build/app_0.out
E2 out/firmware/rom0.bin
Checking for spiffs files
files directory exists. Creating out/firmware/spiff_rom.bin
Creating rom spiff_rom.bin of size 65536 bytes
Adding files in directory files
testfile.txt added to spiffs (46 bytes)
raburton@debian:~/dev/Sming/Basic_rBoot$ make flash
Killing Terminal to free /dev/ttyUSB0
/home/raburton/dev/Sming/Sming//Makefile-rboot.mk:376: recipe for target 'flash' failed
make: [flash] Killed (ignored)
/opt/esp-open-sdk//esptool/esptool.py -p /dev/ttyUSB0 -b 115200 write_flash -ff 40m -fm qio -fs 32m 0x00000 out/firmware/rboot.bin 0x02000 out/firmware/rom0.bin 0x100000 out/firmware/spiff_rom.bin
::snip::
```

**Comment by flexiti** [comment]
I will test it to the end today and I will let You know  :)

**Comment by flexiti** [comment]
btw. I am not familiar with spiffy(s)  (I promise I will read out :) ), but to make file system in Sming I did not need anything beyond ..sming

You've written: spiffy, the tool for making spiffs, is not included in Sming
Can You bring me a problem in two sentences ? thanks

**Comment by hreintke** [comment]
@raburton :
I am also starting to test/use the rboot functionality.
Initial remarks on that : 
1/ The basic_rboot example doesn't include the eclipse project files -> cannot import to eclipse like other examples
2/ I have trouble compiling esptool2 (on windows) 
- I cloned your raburton/esp8266 repo
- in command, cd to esptool2 directory
- running make gives errors
```
C:\Userdata\esptool\esptool2>make
CC esptool2.c
gcc -O2 -Wall -c esptool2.c -o esptool2.o
In file included from esptool2.c:26:
esptool2.h:33:20: stdint.h: No such file or directory
In file included from esptool2_elf.h:27,
                 from esptool2.c:27:
elf.h:25: error: parse error before "Elf32_Addr"
elf.h:25: warning: type defaults to `int' in declaration of `Elf32_Addr'
elf.h:25: warning: data definition has no type or storage class
elf.h:26: error: parse error before "Elf32_Off"
elf.h:26: warning: type defaults to `int' in declaration of `Elf32_Off'
elf.h:26: warning: data definition has no type or storage class
elf.h:27: error: parse error before "Elf32_Half"
elf.h:27: warning: type defaults to `int' in declaration of `Elf32_Half'
elf.h:27: warning: data definition has no type or storage class
....
```
Do I need to set any specific environment variables or configure anything else ?
Or maybe I am using the wrong procedure creating esptool2

flexiti:
still I do not understand why we have to play in the compilation esptool2

raburton:
Guys, this is the master PR for the feature, not a discussion/support thread. If you have problems with a third part tool (even if it is one I wrote) please open an issue against it (I suspect you are missing a proper stdint.h). If you need help using rBoot please take the discussion to an appropriate place, such as the forum or open an issue against my fork. This was all ready to merge, but now it looks like a mess because of the new discussion of people who are only now trying it and finding problems caused by their environment, not reading the readme and various other issues unrelated to the PR. I am happy to try and help you all but I do expect some some effort from the other end, and as long as it isn't making a mess of the PR or the attempt to merge the feature which those who have been following from the start have done a pretty good job of testing now. This has dragged on for a long time unnecessarily now, and I've spent a lot of time on this and helping people try to fix their issues, not least because it has exposed a few real issues along the way (all of which have been fixed). I spent half of my evening last night trying to help someone who clearly hadn't even bothered to look at the readme, let alone follow the set by step instructions on how to build the sample, and who's first response to every error message was to open an issue, before making the slightest attempt to figure it out for himself, often posting his next problem while I was in the middle of writing a reply to him for the first one, having magically managed to fix himself within a few minutes of opening the issue. The wonderful collaborative potential for the internet is often subverted by people with no desire to make any effort or learn anything for themselves and to just get someone else to figure everything out for them. This is not the attitude I expect of engineers.

Apologies for the rant.

flexiti:
I just wanted to indicate that attempt to use rboot causing trouble - for me is not an option reading a long, long description - success is then like the description is short and by the way I do not have to compile ..windows or linux ;). I'm sorry that I offended.
already I'm not here :)

raburton:
There are 7 steps to compile in the readme, not a lot and several of those are common to building other sming apps. I'm not sure many people would be sympathetic to helping you once you've admitted you don't want to read the readme, but if you open an issue against the fork I'll talk you thought it anyway because I'm nice like that.

robotiko:
hi @raburton, 
sorry I was away for a couple of days. 
My last tests where ok and I integrated it into my own projects. 
I will test again but, for me, it is ready to be merged.  
I just pushed into the branch (PR) with the eclipse project files usually included in the sming examples that some people asked me for. 

Thanks

raburton:
@robotiko thanks for the eclipse for and for all your testing and assistance.

@anakod or @kireevco I think we are ready to merge this PR (plus #298 containing the eclipse files from @robotiko).

hreintke:
@raburton : First of all : I did my reading and have put my remarks in here as, in my opinion, it was directly related to the PR involved.
1/ Non eclipse -> user experience for anyone who wants to get started quickly (now indeed solve by an additional PR)
2/ Esptool2 -> Maybe I overlooked but also my second reading did not evolve to a working environment.
I did not see a link/documentation on where to find esptool2. Google searching led to your esp8266 repo and from there the compiling issue occurred. -> same as above, user experience who wants to get started with sming/rboot functionality based on the current feature/rboot branch
If I missed documentation from the rboot PR, just point me to it. If it is not there  probably it is wise to get that link in the sming readme file

PS : Again : don't misunderstand, I agree that rboot should be merged asap, and as mentioned before in this this thread, I hold my horses until that has happened before suggesting rboot/sming integration improvements.

raburton:
I haven't used eclipse in 10 years but last time I did it had the option to import a project from a Makefile, anyway @robotiko has provided this. You're correct there is no link to esptool2 in the readme and that is a sensible suggestion. I've added the url but unfortunately feature/rBoot is using the original git flow proposed in #230 (before it was refined) so this one line addition to the readme has it's own PR #300. If you still haven't got esptool2 compiling in your environment please open an issue against it in my repo and I'll be happy to try and get you a solution.

robotiko:
@raburton 
Just completed BASIC testing under Win. All good. 
Tested with basic_rboot example against a remote shared webserver with the OTA files to update and all went smooth. OTA, switch, etc. 

Then tried with a non OTA project (Basic_Serial) and also good.  Compiles, flashes and runs. 

There are a couple of simple improvements suggested by @hreintke, but as he said, lest  merge ASAP, forget about this flow PR nightmare and then the minor changes. 

To me, it is ready. I'm using it as it is. 

Again, for the nth time, thank you for your great contribution, time and patience.

PR #274, opened 2015-09-10
new multi DS8120 reading exmaple

flexiti:
DS18S20 example
You can connect multiple sensors to a single port
 (At the moment 4 pcs - it depends on the definition in the library)
Measuring time: 1.2 seconds * number of sensors
The main difference with the previous version of the demo:
  - Do not use the Delay function () which discourages manufacturer ESP8266
  - We can read several sensors
 Usage:
  Call Init to setup pin eg. ReadTemp.Init(2);   //pin 2 selected
  Call StartMeasure();
   if MeasureStatus() false read sensors
  You can recognize sensors by the ID or index.

### DIFF
--- a/Sming/Libraries/DS18S20/ds18s20.cpp
+++ b/Sming/Libraries/DS18S20/ds18s20.cpp
@@ -2,12 +2,23 @@
  * ds18s20.cpp
  *
  *  Created on: 01-09-2015
- *      Author: mariuszb
+ *      Author: flexiti and Anakod
  */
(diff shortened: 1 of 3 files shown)


flexiti:
@kireevco  ,why is still "hanging" somewhere?

kireevco:
@flexiti this should be in the branch already. I'm not sure why it is still a PR here.

flexiti:
ok now

PR #286, opened 2015-09-14
Fix I2C bus driver: in SoftI2cMaster::start set SDA & SCL HIGH at the begining

avr39-ripe:
In SoftI2cMaster::start there was wrong assumption that SDA & SCL are both HIGH at the time this method is calling. Fixup this enables proper I2C start condition. Tested on DS3231 - without this fix every second try to read from I2C slave fails, with - do not.

### DIFF
--- a/Sming/Wiring/I2cMaster.cpp
+++ b/Sming/Wiring/I2cMaster.cpp
@@ -94,6 +94,9 @@ bool SoftI2cMaster::restart(uint8_t addressRW) {
  * \return The value true, 1, for success or false, 0, for failure.
  */
 bool SoftI2cMaster::start(uint8_t addressRW) {
+  digitalWrite(sdaPin_, HIGH);
+  digitalWrite(sclPin_, HIGH);
+  delayMicroseconds(I2C_DELAY_USEC);
   digitalWrite(sdaPin_, LOW);
   delayMicroseconds(I2C_DELAY_USEC);
   digitalWrite(sclPin_, LOW);


Hukuma23:
@avr39-ripe will this fix i2c connection between master esp8266 and slave arduino?

avr39-ripe:
@Hukuma23 It can, cause problem was in issuing proper i2c start condition which is crucial for i2c communication. I do not have any arduino i2c to test. May be you test it yourself or point me to some uno/mini i2c slave sketch to flash and test.

Hukuma23:
@avr39-ripe I tried to use code based on this (https://www.arduino.cc/en/Tutorial/MasterReader) example with minor changes. When esp master write to arduino slave work well, but when esp master read from arduino slave didn't work. I checked this on arduino in different power mode (5v and 3.3v) and got the same result.

avr39-ripe:
@Hukuma23  So I tested esp-master read arduino-slave write. It works mostly. Mostly means that sometimes there is some mess in esp output, but I think it because of no voltage negotiation from 5V arduino to 3.3 esp, no 4.7K pull-up resistors, breadboard wires and so on.  Below is terminal capture from esp (something ate Letter "A" at the beginning :)

rduino as I2C slave test
rduino as I2C slave test
rduino as I2C slave test
9�����������������������
9�����������������������
rdui7�������������������
rdu4��������������������
rdu4��������������������
rduino as I2C slave test
rduino as I2C slave test

Hukuma23:
@avr39-ripe excellent news! Thanks! I will try this later.

hreintke:
@Hukuma23 : @avr39-ripe : 
We are cleaning up Sming repo and PR's.
This PR is currently merged into develop but seeing the discussion above I am not sure on the status of this.
Can you provide status and necessary steps to complete this PR ?

avr39-ripe:
This fix fully works for me. I can read data from ds3132 rtc module without any error. Without it there is errors. It isnt ds3231 specific, it fix i2c bus driver low level. So I think it can be merged to master safely.

PR #289, opened 2015-09-14
Add overloading to provide compatibility to old version codes

gustavobreunig:
Added overload with default pin 0. 

This PR is linked to https://github.com/anakod/Sming/pull/277.

### DIFF
--- a/Sming/Libraries/IR/IRremote.cpp
+++ b/Sming/Libraries/IR/IRremote.cpp
@@ -27,6 +27,12 @@ IRsend::IRsend(int sendpin)
   irparams.sendpin = sendpin;
 }
 
+//provide backwards compatibility
+IRsend::IRsend()
+{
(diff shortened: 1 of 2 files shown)


Sming/Libraries/IR/IRremote.cpp
```diff
@@ -27,6 +27,12 @@ IRsend::IRsend(int sendpin)
   irparams.sendpin = sendpin;
 }
 
+//provide backwards compatibility
+IRsend::IRsend()
+{
+  irparams.sendpin = 0;
```
anakod: Can you discover why "0", please?

gustavobreunig: In the previous version there is a define:
```
#define IR_OUT_PIN 13 // GPIO13
```
It can be any pin, I will change to 13.

PR #298, opened 2015-09-25
Added eclipse project files for direct import of basic_rboot example.

robotiko:


robotiko:
Looks like this is gone after merging feature/rboot.

PR #301, opened 2015-09-26
Wait for header parsing to writeRawData

dpwhittaker:
Some web servers (python 2.7 SimpleHTTPServer, at least), send the HTTP/1.1 200 OK line as one packet, the headers in the next packet, and the body starting in the third packet.  Without this change, writeRawData will write the headers on these servers, but start with the body on servers that send the status code and headers together in one packet.  By waiting until the headerEnd is confirmed, this makes the HTTP client act consistently across all servers.

### DIFF
--- a/Sming/SmingCore/Network/HttpClient.cpp
+++ b/Sming/SmingCore/Network/HttpClient.cpp
@@ -254,7 +254,7 @@ err_t HttpClient::onReceive(pbuf *buf)
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


hreintke:
@dpwhittaker : 
We are structuring and cleaning Sming Repo.
This PR has been merged into develop branch. Can you please test the branch by verifying correct working under several server and report back the results ?

PR #326, opened 2015-10-08
rBoot Delegate implementation

hreintke:
Change otaCallback to use Delegates
Is standard implementation within Sming
Fully backward compatible with current implementation


alonewolfx2:
@hreintke thank you. can you provide example usage of new delegate usage?

raburton:
Looks like you'd use it in exactly the same way, I'm don't really understand the change but it must be following some kind of c++ pattern.

hreintke:
Yes, you use it exactly as a "normal callback". Difference is that you also can use an object function as callback.
`MyOtaUpdater.setDelegate(otaUpdateDelegate(&MyCleverObject::objectUpdate,this));`
instead of only a static callback function.

All callbacks in sming are defined in such a way

PR #340, opened 2015-10-14
include spiffy as standard

raburton:
Improved and bug fixed spiffy. Compiles as part of the libsming build, no need to obtain and build separately. Drives the same copy version (and copy) of spiffs that is used by libsming to ensure you always have the correct version.

### DIFF
--- a/Basic_rBoot/Makefile-user.mk
+++ b/Basic_rBoot/Makefile-user.mk
@@ -55,5 +55,3 @@ SPIFF_SIZE      ?= 65536
 #RBOOT_SPIFFS_1  ?= 0x300000
 ## esptool2 path
 #ESPTOOL2        ?= esptool2
-## path to spiffy
-#SPIFFY          ?= spiffy
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -188,7 +188,12 @@ endef
 
 .PHONY: all checkdirs clean
 
-all: checkdirs $(APP_AR) $(FW_FILE_1) $(FW_FILE_2)
+all: checkdirs $(APP_AR) spiffy/spiffy
+
+spiffy/spiffy:
+	$(vecho) "Making spiffy utility"
+	$(Q) $(MAKE) --no-print-directory -C spiffy V=$(V)
+	$(vecho) "Done"
 
 $(APP_AR): $(OBJ)
 	$(vecho) "AR $@"
@@ -214,5 +219,6 @@ clean:
 	$(Q) rm -rf $(BUILD_DIR)
 	$(Q) rm -rf $(BUILD_BASE)
 	$(Q) rm -rf $(FW_BASE)
+	$(Q) $(MAKE) --no-print-directory -C spiffy clean V=$(V)
 
 $(foreach bdir,$(BUILD_DIR),$(eval $(call compile-objects,$(bdir))))
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -26,8 +26,8 @@ SPI_MODE ?= qio
 # SPI_SIZE: 512K, 256K, 1M, 2M, 4M
 SPI_SIZE ?= 512K
 
-# Full path spiffy command, if not in path.
-SPIFFY ?= spiffy
+# Path to spiffy
+SPIFFY ?= $(SMING_HOME)/spiffy/spiffy
 
 ## ESP_HOME sets the path where ESP tools and SDK are located.
 ## Windows:
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -319,8 +319,7 @@ else
 	$(vecho) "Checking for spiffs files"
 	$(Q) if [ -d "$(SPIFF_FILES)" ]; then \
     	echo "$(SPIFF_FILES) directory exists. Creating $(SPIFF_BIN_OUT)"; \
-    	$(SPIFFY) $(SPIFF_SIZE) $(SPIFF_FILES); \
-    	mv spiff_rom.bin $(SPIFF_BIN_OUT); \
+    	$(SPIFFY) $(SPIFF_SIZE) $(SPIFF_FILES) $(SPIFF_BIN_OUT); \
 	else \
     	echo "No files found in ./$(SPIFF_FILES)."; \
     	echo "Creating empty $(SPIFF_BIN_OUT) ($$($(GET_FILESIZE) $(SMING_HOME)/compiler/data/blankfs.bin) bytes)"; \
--- a/Sming/Makefile-rboot.mk
+++ b/Sming/Makefile-rboot.mk
@@ -361,8 +361,7 @@ else
 	$(vecho) "Checking for spiffs files"
 	$(Q) if [ -d "$(SPIFF_FILES)" ]; then \
 		echo "$(SPIFF_FILES) directory exists. Creating $(SPIFF_BIN_OUT)"; \
-		$(SPIFFY) $(SPIFF_SIZE) $(SPIFF_FILES); \
-		mv spiff_rom.bin $(SPIFF_BIN_OUT); \
+		$(SPIFFY) $(SPIFF_SIZE) $(SPIFF_FILES) $(SPIFF_BIN_OUT); \
 	else \
 		echo "No files found in ./$(SPIFF_FILES)."; \
 		echo "Creating empty $(SPIFF_BIN_OUT) ($$($(GET_FILESIZE) $(SMING_HOME)/compiler/data/blankfs.bin) bytes)"; \
--- a/Sming/Makefile-rboot.mk
+++ b/Sming/Makefile-rboot.mk
@@ -20,7 +20,7 @@ RBOOT_LD_1 ?= rom1.ld
 # esptool2 path
 ESPTOOL2 ?= esptool2
 # path to spiffy
-SPIFFY ?= spiffy
+SPIFFY ?= $(SMING_HOME)/spiffy/spiffy
 # filenames and options for generating rBoot rom images with esptool2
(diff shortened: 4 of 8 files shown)


flexiti:
Several days ago I mentioned about using the same sources to spiffy and spiffs .. 
You're doing a great  job. :))
Can You also do Basic_spiffs?
with eg.
#define ESP01 or  (512K)
and
#define ESP12 or  (4M)
and for those  ESP maximum file size and maximum loaded (filled) content by spiffy
This theme often returns
(it can be check by serial or ftp)

kireevco:
I would prefer decoupling components as much as possible, spiffy imho should go to its own repo otherwise we'll be stuck  in monolithic code.

raburton:
This is no longer an external component, it is a single C file which is custom to Sming. There are a number of advantages to having it in Sming and none, that I can see, to having it separate.

kireevco:
@raburton first of all I would love to thank you for such great work. Sming spiffs + spiffy tool based on same code. Great. 
As usually, I am thinking about distribution, and I'm trying to wrap my head around how a user will receive this? Where spiffy binary will be placed? Sounds like we need to re-visit "Include binaries topic".
I just created https://github.com/SmingHub/Sming/issues/341, feel free to comment.

I just want to minimize amount of work needed for a new user to start (which is download a ready-to-go binary)

flexiti:
@kireevco 
spiffy will compiled together with sming - I suppose :)   will be ok :)
I am right @raburton  ?

kireevco:
@flexiti You are right. And this is not good for those who doesn't want a development version. As a new user I want things pre-built as much as possible.
Building spiffy with sming-framework is ok, we just need to separate development users and normal users.

raburton:
@flexiti yes, it would be built for with you build libsming form the same sources, although it will probably be made optional so we don't need to depend on host gcc for users that don't intent to use spiffs. So instead of just calling `make` in the Sming dir as you do now you'd call `make && make spiffy` if you wanted spiffy building as well.

@kireevco I don't think I understand what you mean by development users and normal users.

Sming is a development tool, so in my book that makes everyone a development user. To slightly go off on a tangent I think the idea that we have non-development end users is a problem. While devices like the esp8266 have opened up development to a wider audience, I think it's gone too wide. At the end of the day if you don't want to buy an off the shelf IoT device you need to do some electronics and programming. I get no end of really stupid emails from people who haven't got a clue about either of these concepts and have copied and pasted bits of code from all over the place without a clue what any of it means, and wonder why it doesn't do what they expect. Then email me expecting me to give them zero-to-expert-programmer lesson in an email / do it for them. These people are all using Windows btw. I don't think we should abandon Windows, it's a perfectly viable developer platform, but everything there is more difficult and a huge amount of effort is being expended fixing issues for windows that could be better spent on Sming itself, if users could just accept installing a linux vm on windows would make everyone's life simpler and take about half an hour of their time.

As for shipping binaries this doesn't really change anything (because we haven't really worked that out yet anyway). You just build spiffy where it is and include it in the binary release, as you would libsming.a. That's a whole separate issue, so #341 will be useful to discuss that. I have some thoughts I will share there when I get chance...

kireevco:
@rabutron I'm totally with you on the zero-to-expert matter. On the other hand arduino had great success with such (or a little bit better) end users that were able to implement a simple copy pasted project. I really want sming to be as popular as arduino, in order to do that we need to make it simple to use for anyone. That's how great products become great. Take a look at other platforms like particle.io or arduino for instance. I know it is a tough task but it's doable, we just need to keep that in mind.

In terms of Windows as a platform I'm with you too. That's why I created sming-docker. That's the closest to vm and best to maintain way, have you ever checked it out?
On the other hand it's impossible to force people to use foreign technology. As mentioned by @flexiti and @hreintke most people don't want to learn thing A in order to make thing X work... Windows is OK If you can control it well on the distribution phase (your app is self sufficient and doesn't rely on users environment which could be hell).

hreintke:
@raburton @kireevco : 
I did a first test on windows with the newly included spiffs/spiffy, testing with rboot example.
Using Mingw -> gcc version 4.8.1
Compiles OK, Runs OK 
Using Mingw64 -> i686-5.2.0-posix-dwarf-rt_v4-rev0 (win32) and x86_64-5.2.0-posix-seh-rt_v4-rev0 (win64) 
gcc version 5.2.0.
I get compilation errors : 
```
Making spiffy utility
CC spiffy.c
spiffy.c: In function 'main':
spiffy.c:259:13: error: 'struct dirent' has no member named 'd_type'
      if (ent->d_type == DT_REG) {
             ^
spiffy.c:259:25: error: 'DT_REG' undeclared (first use in this function)
      if (ent->d_type == DT_REG) {
                         ^
spiffy.c:259:25: note: each undeclared identifier is reported only once for each function it appears in
make[1]: *** [spiffy.o] Error 1
make: *** [spiffy/spiffy] Error 2
```
In my halfway from zero-to-expert knowledge I checked the dirent.h of the different versions and those were : 
Mingw : 
```
/* wide char versions */

struct _wdirent
{
  long            d_ino;                /* Always zero. */
  unsigned short  d_reclen;             /* Always size of struct _wdirent. */
  unsigned short  d_namlen;             /* Length of name in d_name. */
  unsigned        d_type;               /* File attributes */
  wchar_t         d_name[FILENAME_MAX]; /* File name. */
};
```
Mingw-W64
```
/* wide char versions */

struct _wdirent
{
        long            d_ino;          /* Always zero. */
        unsigned short  d_reclen;       /* Always zero. */
        unsigned short  d_namlen;       /* Length of name in d_name. */
        wchar_t         d_name[260]; /* [FILENAME_MAX] */ /* File name. */
};
```
Errors come both from eclipse as cmd make compiles
What I not (yet) understand is how the gcc compiler(s) (and eclipse) come to the include path.
Are they using the "windows system path" to find includes ?

raburton:
I am slowly coming to the conclusion that mingw64 is rubbish. This is the second bug in a week I've found that just affects that version. First it's not posix complaint in fopen, now this. I'll have to work around it...

kireevco:
From my knowledge include paths can be checked running gcc with specific commands. Usually they are assumed by gcc and also used from parameters. Sorry, I'm on the road , can't find what is the parameter to ask gcc to show current include paths, but it is not Windows path by its own.

raburton:
Oh, and thanks for testing it!

kireevco:
Another way you can see what your gcc sees as include paths-run eclipse and Check what are the include paths there.

flexiti:
UDK 2.0.0 windows , SmingFramework compilation 

>.......
CC Services/SpifFS/spiffs_hydrogen.c
In file included from Services/SpifFS/spiffs.h:14:0,
                 from Services/SpifFS/spiffs_hydrogen.c:8:
Services/SpifFS/spiffs_hydrogen.c: In function 'SPIFFS_vis':
Services/SpifFS/spiffs_config.h:203:43: error: implicit declaration of function 'c_printf' [-Werror=implicit-function-declaration]
 #define spiffs_printf(...)                c_printf(__VA_ARGS__)
                                           ^
Services/SpifFS/spiffs_hydrogen.c:997:11: note: in expansion of macro 'spiffs_printf'
           spiffs_printf("%4i ", bix);
           ^
cc1.exe: all warnings being treated as errors
make: *** [out/build/Services/SpifFS//spiffs_hydrogen.o] Error 1

flexiti:
sorry, UDK 2.0.8 :)

raburton:
Ok, both these are fixed. I can't believe how much trouble mingw64 is causing with it's lack of features and not being posix compliant (despite the name).

hreintke:
@raburton : 
Tested both the 32 as 64 bit versions of mingw-w64.
Spiffy compiles and runs OK.

Remarks on usage, possibly by knowledge from me. 
- make spiffy is not visible in eclipse. 

Don't know whether this is a eclipse or makefile setting

- I could not "force new build" of spiffy.

So when running make spiffy after compiler update I got 
Nothing to build
After removing spiffy.exe -> Incompatible spiffy.o
After removing *.o -> compile OK.

Maybe we should always rebuild completely. Only takes seconds

raburton:
I have no idea what that means, but then I don't use eclipse. Perhaps the eclipse project file for Sming needs to have the spiffy target added, if so I'm afraid someone else will have to do that for me. As for rebuilding, a `make clean && make` is the traditional way to rebuild. If you need another make target adding to the spiffy Makefile that does a clean then a build that's easy enough.

raburton:
Thinking about it, if you're going to add a new target to eclipse for spiffy, might want to add a target for spiffyclean (or something similar) too. At the moment spiffy isn;t built with the library, you have to call the separate target, but it is cleaned with it, which is a bit inconsistent. Mind you you'd need to add a rebuildspiffy target too, because the rebuild target on the library shouldn't rebuild spiffy, if neither the all or clean targets do anything with spiffy.

hreintke:
I will check further or the eclipse.

For make spiffy : 
I now understand where I went wrong (?)
I was in directory SMING_HOME\sming where I called `make spiffy` and that worked.
But to use the "clean" option I had to go to spiffy subdirectory and call make clean there.
So if we update the "sming makefile" whichs calls the "spiffy makefile" to get it clean & make.
Then : 
make spiffy from sming directory -> complete update
make spiffy from spiffy directory -> all options available

PS : hope I use the right wording. If not please let me know

hreintke:
PS : On rboot :  
Should we move rom0.ld to the sming/compiler/ld directory ?
It's not that logical that you have to copy a rom0.ld from an example program to make your own program rboot capable

raburton:
>Should we move rom0.ld to the sming/compiler/ld directory ?
It's not that logical that you have to copy a rom0.ld from an example program to make your own program rboot capable

Except that this file needs to be customised based on user choice of flash layout.

PR #343, opened 2015-10-14
Feature/newspiffs

raburton:


flexiti:
@raburton 
Tests so far well. (udk 2.0.8) You are fast :)
By the way, the resulting spiffy.exe is 800KB, debug options are set? (I had a 170 kb)

What are the current limits for Sming spiffs now? 
How to check how big spiffs can be for a given model of ESP and if is dependent on the size of the program? (rboot version)

raburton:
On windows mine is ~100k, on linux it's ~60k so it must be whatever is default in your environment. Maybe it's statically linked.

Don't know size limits for spiffs. If it doesn't have any inherent limits (or at least within our realistic range) you will be limited only by the size and layout of your flash.

Btw, you can now specify the spiffs size in hex on the spiffy command line (by starting the number with 0x), which is much nicer.

hreintke:
Mine on windows, compiled with the mingw, using the newspiffs branch from sminghub is 
`14-10-2015  19:15           124.204 spiffy.exe`

flexiti:
Spiffy.exe size is not yet significant, I it check later.
Yes,Hex is nicer.

How if You set  SPIFF_SIZE  > 1MB ?
If ESP addresses  1 MB -  spiffs and the program must be less than 1 MB? 
What size set to not  exceed the ESP  limits and what happens if you cross?

raburton:
spiffs does not use memory mapped flash, so it is not limited to being within the same 1mb as the rom. See the rBoot sample code and readme for example use of spiffs_mount_manual. You don't need to be using rBoot for it, but it's demonstrated and discussed there.

flexiti:
ok, now it is clear to me.
Tomorrow I'll try to test different sizes.

PR #348, opened 2015-10-14
Feature/mqtt

slaff:
Added code to set Last Will And Testament.
 The keepAlive can be changed now.
 The MqttClient can connect to an IP and Port in addition to Host and Port.


slaff:
Cleaner PR against origin develop is here #349.

PR #349, opened 2015-10-14
Added code to set Last Will And Testament.

slaff:
The keepAlive can be changed now.
The MqttClient can connect to an IP and Port in addition to Host and Port.

### DIFF
--- a/MqttClient_Hello/app/application.cpp
+++ b/MqttClient_Hello/app/application.cpp
@@ -38,6 +38,7 @@ void onMessageReceived(String topic, String message)
 // Run MQTT client
 void startMqttClient()
 {
+	mqtt.setWill("last/will","The connection from this device is lost:(", 1, true);
 	mqtt.connect("esp8266");
 	mqtt.subscribe("main/status/#");
 }
--- a/Sming/Services/libemqtt/libemqtt.h
+++ b/Sming/Services/libemqtt/libemqtt.h
@@ -300,6 +300,13 @@ int mqtt_unsubscribe(mqtt_broker_handle_t* broker, const char* topic, uint16_t* 
  */
 int mqtt_ping(mqtt_broker_handle_t* broker);
 
+/**
+ * Frees dynamically allocated resources
+ *
+ * @param broker Data structure that contains the connection information with the broker.
+ */
+void mqtt_free(mqtt_broker_handle_t* broker);
+
 #ifdef __cplusplus
 }
 #endif
--- a/Sming/Services/libemqtt/libemqtt.h
+++ b/Sming/Services/libemqtt/libemqtt.h
@@ -168,6 +168,8 @@ typedef struct {
 	// Will topic
 	uint8_t will_retain;
 	uint8_t will_qos;
+	char* will_topic;
+	char* will_message;
 	uint8_t clean_session;
 	// Management fields
 	uint16_t seq;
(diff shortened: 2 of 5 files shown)


MqttClient_Hello/app/application.cpp
```diff
@@ -38,6 +38,7 @@ void onMessageReceived(String topic, String message)
 // Run MQTT client
 void startMqttClient()
 {
+	mqtt.setWill("last/will","The connection from this device is lost:(", 1, true);
```
Tavalin: This doesn't look correct.

slaff: @Tavalin Can you please explain what exactly does not look ok in the line above? The method mqtt.setWill is called with 4 parameters, 2 strings, one int with value 1 and one boolean with value true.

Tavalin: You are correct, I thought I saw an unclosed open bracket but it was a sad face instead!

Sming/Services/libemqtt/libemqtt.h
```diff
@@ -168,6 +168,8 @@ typedef struct {
 	// Will topic
 	uint8_t will_retain;
 	uint8_t will_qos;
+	char* will_topic;
+	char* will_message;
```
raburton: These two variables are defined as char* but you treat them like the other strings defined as fixed size char[] (i.e. you copy to them and memset them) but I don't see you ever malloc any memory to these pointers. Am I missing something?

Sming/SmingCore/Network/MqttClient.h
```diff
@@ -54,6 +60,7 @@ class MqttClient: protected TcpClient
 	uint8_t *current;
 	int posHeader;
 	MqttStringSubscriptionCallback callback;
+	int keepAlive = 50;
```
raburton: this is two and a half times the previous value, does it really need to be increased this much? 20 second is quite a while to wait for a modern network, 50s seems excessive

slaff: You are right. I will set it back to 20 and then the developer can decide how often the ping should happen.

slaff:
@raburton Thanks for the review. When you have time can you take a look at the changes that I committed https://github.com/SmingHub/Sming/commit/73b8760ceb69750b937966c3858041394e80a2fa and suggest better changes if needed.

raburton:
You can't use `sizeof` on a pointer to a string (char*) to get the string length, it will just return the pointer size. You'll need to use `strlen + 1` for the `malloc`, then you can just use `strcpy` instead of `strncpy` when you copy the string to your version. I'm very surprised your testing hasn't found this bug or previous ones!

Also, you also don't ever `free` the memory you `malloc` so this code is going to leak. You need to find a way to clean this up when it's done with.

I only just noticed that some oif this code isn't ours it from somewhere else, have you checked up stream to see if they have already implemented this? If not it's worth trying to get it in upstream, the authors will understand it better (I don't know anything about mqtt, I'm just reviewing the code) and we can avoid forking their component so we can easily accept new upstream versions in the future.

hreintke:
The remark on upstream is exactly what I wanted to say. Agree.

On keepalive : The definition of keepalive in mqtt is number of seconds a client waits after the last publish before sending a ping to show the broker it is still alive. 
That means in reliable networks it can be higher because of low probability of loosing connection.
Keepalive to low means flooding the connection with keepalives (when a lot of devices is connected).
I don't know why anakod has set the keepalive by default to 20, but I would certainly propose a larger value (60 ?, 120 ?). The implementer always can update.
If I am wrong about the definition : please let me know.

slaff:
@raburton On one side the last time  I programmed in C was probably 15 years ago and my knowledge is quite rusty. So it would be great if you can show me the right changes for the will_topic and will_message memory allocation.On the other side my MQTT knowledge is not rusty and that is what the protocol says if you want to send Last Will And Testament. I checked also the original project https://github.com/fcvarela/liblwmqtt/blob/master/src/libemqtt.c and there that functionality is missing. With your help we can make the Mqtt code more useful and rock-solid :)

raburton:
merged as pr #390

PR #354, opened 2015-10-16
Added building of projects with changed files.

slaff:


hreintke:
I am not familiar with travis so correct me if I am wromg.

In the proposed PR you compile the examples that are changed.
I think that after an update to sming, there should be verified whether ALL projects still compile OK.

slaff:
> I think that after an update to sming, there should be verified whether ALL projects still compile OK.
Sounds like a good idea to me. Shall I add this code or you want to give it a try?

kireevco:
It look a little bit hacky to me. Ideally all example projects should be moved to their own repo/repos. There was an ticket created. Let's use that pr for now imho until projects are moved out.

slaff:
@hreintke @kireevco With those initial changes Travis CI should be able to catch issues with the standard compilation of the code. Later on we can add also getting env variables from the git commit message and use it in the compilation. But for now this should be enough.

kireevco:
Sounds good. FYI, Travis is a temporary solution until I get Jenkins with a
good build matrix going. But that should not be an issue.

On Friday, October 16, 2015, slaff <notifications@github.com> wrote:

> @hreintke <https://github.com/hreintke> @kireevco
> <https://github.com/kireevco> With those initial changes Travis CI should
> be able to catch issues with the standard compilation of the code. Later on
> we can add also getting env variables from the git commit message and use
> it in the compilation. But for now this should be enough.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SmingHub/Sming/pull/354#issuecomment-148765285>.
>


-- 
Sincerely,
  Dmitry Kireev
  Cell: +1 (818) 850-55-77
  Skype: kireev.co

hreintke:
Sure that is an issue.
If we put energy in something we not going to use we are wasting time.

hreintke:
Will be solved when restructuring smingcore/application repository

PR #366, opened 2015-10-23
added hostname getter/setter to StationClass

ChrisScheffler:
simple wiring for wifi_station_get_hostname and wifi_station_set_hostname

### DIFF
--- a/Sming/SmingCore/Platform/Station.cpp
+++ b/Sming/SmingCore/Platform/Station.cpp
@@ -115,6 +115,12 @@ void StationClass::enableDHCP(bool enable)
 		wifi_station_dhcpc_stop();
 }
 
+String StationClass::getHostname()
+{
+	char* hname = wifi_station_get_hostname();
+	return String((char*)hname);
+}
+
 IPAddress StationClass::getIP()
 {
 	struct ip_info info = {0};
@@ -149,6 +155,15 @@ IPAddress StationClass::getNetworkGateway()
 	return info.gw;
 }
 
+bool StationClass::setHostname(String hostname)
+{
+	char* hname = new char[MAX_HOSTNAME_LEN + 1];
+	memset(hname, 0, sizeof(hname));
+	strcpy(hname, hostname.substring(0, hostname.length()).c_str());
+	bool res = wifi_station_set_hostname(hname);
+	delete[] hname;
+	return res;
+}
 
 bool StationClass::setIP(IPAddress address)
 {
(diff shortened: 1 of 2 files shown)


hreintke:
@ChrisScheffler 
I think that when you call StationClass::getHostname with a string parameter longer than the MAX_HOSTNAME_LEN the strcopy will go wrong

raburton:
Yes, there is a buffer overflow there, and all kinds of oddness on that code. Why do you can substring to get the entire string again? Needs work.

ChrisScheffler:
@hreintke
You are absolutely right, it should read
```
strcpy(hname, hostname.substring(0, min(MAX_HOSTNAME_LEN, hostname.length())).c_str());
```
The whole substring behaviour was meant to prevent the buffer overflow in the first place.
I also had to fix the memset as i realized the pointer size issue.
```
char* hname = new char[MAX_HOSTNAME_LEN + 1]();
memset(hname, 0, MAX_HOSTNAME_LEN + 1);
```
@raburto
Besides the substring and the memset what other kind of oddness I should fix?
I really appreciate the feedback.

raburton:
Oh my.

Why are you trying to copy the string in the first place? And have you ever come across strncpy?

Your entire function could be written as one line:
`return wifi_station_set_hostname(hostname.c_str());`

ChrisScheffler:
@raburton 
I was not aware of that possibility. My initial intention was to limit the value to 32 bytes. I didn't realize that it is not relevant for the underlying api call. I thought if the value is to large it might cause issues.
> Oh my.
As you *might* have noticed from my code I'm not very familiar with C. But you're right, I should not have created a pull request for this kind of code quality.
I'm sorry about that. I'll close the request.

raburton:
Don't be disheartened, everyone learns new languages at some point and we appreciate your effort to help make Sming better! Please create a new (simpler) PR if you wish, or one if us could implement the requested feature for you (probably have to be after the weekend if it's me).

hreintke:
Richard,
I first expected that your solution was available and much simpler,  but when entering this code in my application 
```
void sethost (String reqHost)
{
	wifi_station_set_hostname(reqHost.c_str());
}
```
I get a compilation error : 
```
app/application.cpp: In function 'void sethost(String)':
app/application.cpp:26:41: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
  wifi_station_set_hostname(reqHost.c_str());
```
What am I missing in this ?

raburton:
You'll need to recast it as const, the headers obviously don't state the function takes a const. Annoying when they don't do that properly but I don't expect a lot from the espressif headers! Even with a cast some versions of GCC might complain, but you can ignore them. When I get back I'll raise a bug with espressif to see if they can correctly mark the headers.

PR #371, opened 2015-10-26
Changed WS2812 example: removed delay and WDT refresh

flexiti:
WS8212 example   it must be tested because I do not have WS2812
 Also added 2 new metods getRssi() and getChannel() in StationClass
(can read actual WiFi signal level and the channel number)

### DIFF
--- a/Basic_WiFi/app/application.cpp
+++ b/Basic_WiFi/app/application.cpp
@@ -32,6 +32,14 @@ void connectOk()
 {
 	debugf("I'm CONNECTED");
 	Serial.println(WifiStation.getIP().toString());
+	
+	Serial.print("Connected to: ");
+	Serial.print(WifiStation.getSSID());
+	Serial.print(" Channel: ");
+	Serial.print(WifiStation.getChannel());
+	Serial.print(" Strength: ");
+	Serial.print(WifiStation.getRssi());
+	Serial.println(" dBm");
 }
 
 // Will be called when WiFi station timeout was reached
--- a/LED_WS2812/app/application.cpp
+++ b/LED_WS2812/app/application.cpp
@@ -1,21 +1,29 @@
 #include <user_config.h>
-
+#include <SmingCore/SmingCore.h>
 #include <WS2812/WS2812.h>
 
 #define LED_PIN 2 // GPIO2
 
-void init()
+Timer procTimer;
+int step;
+
+void readWS2812()
 {
-    while (true) {
+    if ((step++) & 1) {
         char buffer1[] = "\x40\x00\x00\x00\x40\x00\x00\x00\x40";
         ws2812_writergb(LED_PIN, buffer1, sizeof(buffer1));
-        os_delay_us(500000);
-        
-        //We need to feed WDT.
-        WDT.alive();
-        
+    }
+    else
+    {
         char buffer2[] = "\x00\x40\x40\x40\x00\x40\x40\x40\x00";
         ws2812_writergb(LED_PIN, buffer2, sizeof(buffer2));
-        os_delay_us(500000);
-    }
+    }        
+}
+
+void init()
+{
+	Serial.begin(SERIAL_BAUD_RATE); // 115200 by default
+	Serial.systemDebugOutput(true); // Allow debug output to serial
+
+	procTimer.initializeMs(500, readWS2812).start();   // every 0.5 seconds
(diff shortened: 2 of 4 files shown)


flexiti:
Renaming function

flexiti:
Can someone merge it? I think it's safe to change

kireevco:
@flexiti can you please update PR description? What is it doing and what is the impact?
(We need this to be included in the changelog).

flexiti:
@kireevco 
its ok  now ?  
There is also 2 new metods  getRssi() and getChannel() in StationClas but I do not know where to write info - in the same line as info about new example  versiona of WS2812 ?

kireevco:
@flexiti I change the wording a little bit. Does that make sense?

you can write here:
>It can be done in this way, it must be tested because I do not have WS2812

flexiti:
@kireevco 
its ok

tprochazka:
Why this is still not merged?

tprochazka:
I tried use it. Example works correctly, but when I use the library in real project where WifiStation is connected it doesn't work, it produce completely different color. I know that it is not problem of this PR, but I would ask if you tried it on the more complex application than example?

alonewolfx2:
@tprochazka what is your not working code part? i am using in one project and its connected to wifi and its working fine

anakod:
> I tried use it. Example works correctly, but when I use the library in real project where WifiStation is connected it doesn't work, it produce completely different color.

I think what may be we should try to add interrupts lock during color data transmission?

alonewolfx2:
> I think what may be we should try to add interrupts lock during color data transmission?

right. i am using it. maybe i can push PR

flexiti:
@kireevco 
Can You merge it to develop ?
Previous example was taken from the Arduino, without the necessary changes specific to sming and ESP, this one is good.
@tprochazka 
after the publication can check and then let me know?
I think that the example will run well, certainly there are any errors in another library - I can check it out but first we must have same version

hreintke:
@flexiti : 
I see comments from @alonewolfx2 that additional PR would be needed and @tprochazka @anakod .
That there are issues when wifi is connected .

flexiti:
@hreintke 
I think that You can publish it, because there is no error in this demo.
The fault is in the library es2812, function ws2812_writergb () (previous version of demo also have this fault)
I can improve this library in my free time. (I see that ws2812_writergb () not pay attention to the time of transmission !! - so important for ws2812)


To sum up, certainly this demo will not be worse than it is now, but better  ;)

alonewolfx2:
i think we can delete this example and library. because adafruit library very powerfull.

flexiti:
@alonewolfx2 
Your demo, you can delete the demo :) 
 Although copying everything from Arduino is not necessarily effective, maybe perhaps it our library is already so big.
@hreintke 
Please keep however, getRssi and getChannel.
(maybe the easiest way to publish it, and then remove the demo WS2812)

hreintke:
@flexiti : @alonewolfx2 : 
I am lost in this discussion. No idea anymore which library/example is OK/not OK.
Please clarify.

@flexiti : Pulling something in, just to remove it afterwards is not a way to keep history clean.
Maybe should have said it before (as I did in other case lately) but now we the discussion above : 
Please split the PR in two. 1/ SmingCore update (Rssi) and 2/ Library and example.
In that way we keep the repo clean and updates understandable/traceable.

flexiti:
@hreintke , please publish it, next @alonewolfx2 remove his example...It is the simplest and clean  solution and it will take a few seconds, otherwise I'll have to play the amendments and publish them again - a waste of time

alonewolfx2:
@flexiti which example ? ws2812 example isnt mine. i ported just adafruit library and ported adafruit example.
@hreintke both examples working fine but just 2 library for ws2812 unnecessary. Adafruit_neopixel library ad example can stay and we can remove old ws2812 library for keep clean.

flexiti:
colleagues, the only time we lose here :) Please @hreintke  publish it, and then @hreintke or @alonewolfx2 delete this example.
I have just been asked to improve this example, neopixel example  and OLED display example- What I did.
 I would also like to organize itself, my  develop... so please, no more comments please...just let do it :)

flexiti:
@alonewolfx2 @hreintke 
Colleagues, can you? I'm already tired of asking to merge it

PR #374, opened 2015-10-27
rework of hw pwm

raburton:


raburton:
oops, against wrong branch

PR #375, opened 2015-10-27
rework of hw pwm

raburton:


Vity01:
Great. Thanks for your hard work.

hreintke:
Richard,
Included is a nbproject directory. Suppose is is netbeans.
Not sure whether we should include that in this application only.

Did not compare it yet but is the implementation of pwm and hw_pwm identical/comparable with regards to function, parameter naming and usage ?
Maybe we could even combine the two and select hw/sofware as an option.

raburton:
> Great. Thanks for your hard work.

No problem. If I can get some feedback I'll get it merged.

>Included is a nbproject directory.

Not in the reworked version there isn't, stripped all the junk out (well, except the .settings folder as I didn't know for sure if that was important or not when I created the pr).

>Did not compare it yet but is the implementation of pwm and hw_pwm identical/comparable with regards to function, parameter naming and usage ?

It's quite different, not compatible.

raburton:
I've pulled out the .settings folder now.

raburton:
I've changed setduty to use pin numbers now, instead of channel numbers, so analogwrite is just the same (just there to make it easier for arduino users), and the api is now more consistent as it only uses pin numbers.

tprochazka:
Is there any reason then HWPWM is completely separate from current software implementation with Arduino compatible API? I mean analogWrite(). I think that it should replace old one. It is not usable.

And I had problem with compilation of this.
PWM.h of sming framework is included instead if espressif pwm.h
I renamed sming PWM.h to different name and update references to it and it works now.
Why you don't have such problem?

tprochazka:
I also tried remove 
delay(25); // delay is needed otherwise PWM is not set
And it still works correctly.
It only slow down other things...

raburton:
It probably can replace the software version, but at the moment it needs more feedback before it can be merged.

>PWM.h of sming framework is included instead if espressif pwm.h
I renamed sming PWM.h to different name and update references to it and it works now.
Why you don't have such problem?

Probably because I have case-sensitive filesystem, I assume you're on windows? I'll shall fix this...

>I also tried remove
delay(25); // delay is needed otherwise PWM is not set

I didn't write the original code, and I can't test it (HW pwm doesn't actually work on any of my devices!!!), so thanks for the info, I'll remove that.

hreintke:
@raburton : 
Did my testing as promised.
1/ I am on windows and also had the issue with pwm.h mentioned by tprochazka 
2/ I tested using the basic_hwtimer example.
- If not changed all works on several pins. Did not try all so if you have trouble with one specific one let me know and I'll repeat my test
- How many simultaneous channels do you use ? 
I changed the dopwm function to : 
```
void doPWM() {
	if (countUp == true) {
		i += inc;
		if (i >= 22222) {
			i = 22222;
			countUp = false;
		}
	} else {
		i -= inc;
		if (i <= 0) {
			i = 0;
			countUp = true;
		}
	}
	debugf("i = %d",i);
	for (int x=0;x<4;x++)
		HW_pwm.analogWrite(pins[x], i);
}
```
When updating multiple channels in each call I noticed : 
0-3 channels -> works OK
4+ channels -> reset in the 29th doPWM (always exactly, no variation)
Also mixed the pinnumbers -> doesn't seem to be related to pinnumbers only # of channels updating.

tprochazka:
Mazbe maximum value shouldn't be hardcoded as 22222.
Actually maximum is defined in the documentation as ```getPeriod() * 1000 / 45```
22222 it is only for default period and there are setPeriod method which change it.
Maybe there should be some method to get maximum value for duty.

It works much better than current SW implementation, but it is weird that sometime if CPU do something more complicated duty cycle is sometimes changes and LED will light on maximum level for some time. Do you somebody know if it is really HW PWM or it is still only emulated but on lower level of SDK than current SW implementation?

And I also found one weird think.
If I'm using the LED active by HIGH on the pin, I'am able to control light from zero (duty 0) to maximum intensity. But If I have LED active by LOW (like RGB led with common anode), I'm not able to get the same behaviour. Also with 22222 duty LED intensity is still big, but it should be completely off.
I don't understand why. It works correctly on Arduino board with analogWrite with 255 value.

raburton:
It's still done in software (code in libpwm), just uses a hardware timer to trigger it. You are right about the hard coded limit. I can add a function to calculate the maximum and use it when verifying the user supplied value.

tprochazka:
I found the reason for "the weird think"
http://bbs.espressif.com/viewtopic.php?f=7&t=631
It is very sad :-(

raburton:
@hreintke thanks for testing, i think the problems you've found relate to libpwm rather than our integration, so shouldn't be a problem for this pr
i'll fix the pwm.h issue by renaming the software version, we shouldn't have files with the same names as ones in the sdk, likely to cause problems eventually

raburton:
please test latest changes, outstanding issues fixed
note that the changed software pwm driver (which I doubt anyone uses anyway since it's even worse than the hardware version) no longer has a default instance so you need to declare it if you want it (and analogWrite needs to be called against that instance, as there is no longer a default one).

jmccoy555:
So after finally managing to get this out of git (why is there no 'command line' as per the help page??) it's all working rather nicely on a RGBW MQTT dimmer with OTA update. It's very usable now, so thanks!

raburton:
What command line are you referring to? Is that a custom dimmer or is it
the H801 rgbww controller? Thanks for the feedback!

jmccoy555:
Hi, step 3 - https://help.github.com/articles/checking-out-pull-requests-locally/ - anyway I gave up with that and switched to Linux to make a zip. I'm driving 4 MOSFETs off of pins 4, 5, 13, and 14 and now also a DHT22 off of pin 2. All stable so far. Will keep an eye out for any more updates to test.

raburton:
@jmccoy555 The command line bit appears to show up next to the merge button, and you won't have permission to merge so I guess you don't see the command line either.

You can do this:
```
git fetch origin pull/375/head:hwpwm
git checkout hwpwm
```

raburton:
If there aren't any more problem reported I'll probably merge this tomorrow evening, things that crop up after that can always be dealt with later.

raburton:
Not had any more feedback (good or bad) so going to merge. Please open new issues if you have problems after this. Also please remember that pwm support on esp is a bit rubbish at the best of times and we are depending on libpwm from the sdk, so it's probably buggy too. Also, hw pwm doesn't work at all for me on any of my devices, so if you do have problems please have a go at fixing them yourself first.

tprochazka:
Great work! Thanks.

Issue #385, opened 2015-10-31
Proposal, sdk v1.4


raburton:
I propose for the next release of Sming we move to (and require) SDK v1.4. A number of changes have been made in SDK v1.4 which we need to update our code to use, but these are not all easily backwards compatible. All common build platforms (esp-open-sdk for linux, unofficial esp8266 dev kit for windows) all use v1.4 by default now.

kireevco:
@raburton in my opinion it is always good to move forward. And as for me nothing stops us from creating a separate branch (like _feature/sdk1.4_ or so), and make work there. 

What people think?

profra:
Hi folks, I installed Sming yesterday evening, everything was functioning at the first time but I tried only Basic_Blink before I went to bed... also the  whole library was build with SDK 1.4 without any erros... as far as I can see, I am total newbie... If I can recommend please switch to new SDK 1.4 .... before the version 1.5 is released :-) .... But what I really miss is better documentation, especially the story about Sming to get the right backroud idea... WHY Sming, WHAT are the parts of Sming, HOW these parts communicate.... So, thank you, Sming seems to be excellent tool for projects with ESP8266!!!

raburton:
I agree about documentation, we need some, at the moment if there is no api
documentation you have to look through the many samples as a substitute.
Problem is nobody I'd interested in doing that, I'm certainly not, so
unless someone volunteers I don't expect we'll ever have any.

hreintke:
As far as I know the latest UDK Releade 2.0.8 is still on Espressif SDK 1.3.0
However the UDK can be updated manually to use SDK 1.4.0 but that requires an additional step for windows users of Sming.

If we still decide on moving to SDK 1.4.0 into Sming 1.5.0 we should opt for a quick integration into the develop branch. This to have as much of testing/usage with the new SDK.
Whether we do the quick integration with use of PR (like hw_pwm) or feature branch (like rboot) is not a big difference.

For the backward compatibility issues which come with the upgrade we do need them documented in the PR/Branch from the start otherwise we will get a number of questions/remarks which can be prevented.

profra:
Yes, you are right, I had to go through some examples to understand the matter... The good documentation is really hard task, I would like to help but I don't have a head  good enough for it and also time... IMO it is task for creators, otherwise Sming cannot be widely spread and used....

raburton:
> IMO it is task for creators, otherwise Sming cannot be widely spread and used...

[...]

PR #389, opened 2015-10-31
sdk v1.4 support

raburton:
malloc related issues fixed
please report more v1.4 issues in #385 (not here)


raburton:
@kireevco any chance you can create a new zip of the build environment for sdk v1.4, and pop it in the appropriate location https://github.com/raburton/Sming/blob/7cfd0b28601111c6da2b227dd9359dcf53e12aaf/.travis.yml#L5

no rush, thanks

tprochazka:
Integration smartConfig to the WifiStation is very good idea, but there is missing (optional) callback with operation result, sometimes is neccesary to react to configuration in application... or do you think that ```WifiStation.waitConnection(connectOk, 10, connectFail)``` is enought?

raburton:
> or do you think that WifiStation.waitConnection(connectOk, 10, connectFail) is enough?

I don't know, I don't use smart config. I'll take your feedback on whether it's really necessary to have a call back. Do you just want to know the process is complete?

tprochazka:
I just trying playi with the SmartConfig. It is very interesting thing which allow to configure any IoT device automatically from another device that is already configured. In my case from the mobile phone.

I think that best would be possibility to get original sc_status status and void *pdata in the callback.


And I'm a little bit confused about changes in malloc. I'm using this code
```
struct mdns_info *info = (struct mdns_info *)os_zalloc(sizeof(struct mdns_info));
info->host_name = (char *) "socket1";
...
espconn_mdns_init(info);
```
os_zalloc doesn't work now with just one argument.

And in the mem.h is still this definition, which is invalid now
```
#ifndef os_zalloc
#define os_zalloc(s) mem_zalloc((s))
#endif
```
I'm not so good in the C++, what I should use instead of os_zalloc in my case?

raburton:
i have no idea what that extra lwip related mem.h is there for, that needs further investigation, but i've fixed your issue, thanks for reporting it

raburton:
> I think that best would be possibility to get original sc_status status and void *pdata in the callback.

Then you are playing with the underlying c code/data instead of the nice c++ object. Tell me what you are trying to achieve and I'll see what needs to be done to do it properly.

tprochazka:
Just display status of smartConfig operation on display.
And run the application itself if it is done or display error message if timeout expire.

Btw. Maybe also ```esptouch_set_timeout``` should be possible to define as optional parameter of smartConfigStart()

hreintke:
If we are creating callbacks. Let's use the Sming standard in callback using delegates.

Basic_SmartConfig/app/application.cpp
```diff
@@ -36,5 +8,5 @@ void init()
 
 	WifiAccessPoint.enable(false);
 	WifiStation.enable(true);
-	smartconfig_start(SC_TYPE_ESPTOUCH, smartconfigDone);
+	WifiStation.smartConfigStart(SCT_EspTouch);
```
slaff: WifiStation.smartConfigStart must have a callback as parameter. A Sming-style smartConfigDelegate should be fine.

Since smartconfig works only when AP is off and STA is on then may be the following two commands can be inside of the body of WifiStation.smartConfigStart

```C
  WifiStation.enable(true);	
```

tprochazka:
I have stupid question.
Why was ```include <smartconfig.h>``` added to the user_while? Why not to Station.h where is used?
What is the key to add something to user_config?

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -351,6 +351,43 @@ const char* StationClass::getConnectionStatusName()
...
+			debugf("SC_STATUS_WAIT\n");
+			break;
+		case SC_STATUS_FIND_CHANNEL:
+			debugf("SC_STATUS_FIND_CHANNEL\n");
+			break;
+		case SC_STATUS_GETTING_SSID_PSWD:
+			debugf("SC_STATUS_GETTING_SSID_PSWD\n");
+			break;
+		case SC_STATUS_LINK:
+			{
```
zendes: i don't think the curly braces here are necessary

Sming/SmingCore/Platform/Station.h
```diff
@@ -25,6 +25,13 @@ enum EStationConnectionStatus
 	eSCS_GotIP
 };
 
+enum SmartConfigType
+{
+    SCT_EspTouch = SC_TYPE_ESPTOUCH,
+	SCT_AirKiss = SC_TYPE_AIRKISS,
```
zendes: your indentation is off here

Sming/SmingCore/Platform/Station.cpp
raburton: yes they are

Sming/SmingCore/Platform/Station.h
raburton: thanks, i'll fix that (and the one above, but at least they match!)

raburton:
I shall add the callback once i get chance then, i was hoping the user wouldn't need it, but if you lot (as people who actually use this (i don't)) think it's needed I'll add it.

raburton:
> have stupid question.
Why was include <smartconfig.h> added to the user_while? Why not to Station.h where is used?
What is the key to add something to user_config?

Good question. Stuff in there seems to be more sdk related, and are all c headers instead of cpp, but I've moved it back into station.h

raburton:
I've added a callback option, as well as a default that will do what it was previously doing (receiving the statues updates and setting the wifi credentials once received), so there is the easy option if you want it.

hreintke:
@raburton : I propose 
Make pdata the right type 
`typedef Delegate<void(sc_status status, station_config* stationConfig)> SmartConfigDelegate`

and smartConfigDefaultCallback a member function
`void StationClass::smartConfigDefaultCallback(sc_status status, station_config* stationConfig)`

You then need to update 
`smartconfig_start(smartConfigDefaultCallback);`
to 
`smartconfig_start(SmartConfigDelegate(StationClass::smartConfigDefaultCallback,this));`

raburton:
pdata isn't always station_config, the type depends on the event.

hreintke:
OK, Did not catch that from the code.
Proposal to class member stays, of course then with pdata argument.

BTW : I fetched #389 and at compiling Sming Framework I get : 
`c:\userdata\sming_hr1022local_project\sming\smingcore\platform\station.h:36:25: error: 'SC_TYPE_ESPTOUCH_AIRKISS' was not declared in this scope
  SCT_EspTouch_AirKiss = SC_TYPE_ESPTOUCH_AIRKISS`

raburton:
That was added at sdk v1.4, are your sure your compiling with 1.4?

My understanding is you can't pass a member function as a call back to c,
which is why I've had to do it the odd way I have, but I'll look into your
code suggestion when I get chance.

tprochazka:
Is a little bit weird, that every application now need to setup
EXTRA_LIBS=smartconfig
also when it doesn't use smart config

tprochazka:
@hreintke What is benefit of use SmartConfigDelegate? It looks more complicated.
@raburton It is possible also add possibility to configure timeout for smart config?

And it is a little bit weird, that every application now need to setup
EXTRA_LIBS=smartconfig
also when it doesn't use smart config

raburton:
>@raburton It is possible also add possibility to configure timeout for smart config?

There is no timeout provided by the api. If you want to add your own you can easily do that by setting a timer and calling smartconfig stop function.

>And it is a little bit weird, that every application now need to setup
EXTRA_LIBS=smartconfig
also when it doesn't use smart config

It doesn't. None of my applications have that added to them.

hreintke:
@raburton : 
OK, I'll do another check on callback, let you know.
Shouldn't we add smartconfig lib to LIBS ?
When not used doesn't add anything to application. 
And any application (with or without smartconfig) will work

tprochazka:
Interesting. You have right. I don't understand where I so 45s default timeout in the SDK :-(

hreintke:
@raburton 
You were right about my mistake on member functions and c callbacks.

I updated the StationClass to have included :  
- All smartconfig functionality internal in the class (same approach as other within Sming)
- Renaming scanCompletedCallback to Delegate
- Update smartConfigStop to include smartConfigDelegate = NULL

I don't know (yet) how to initiate a PR to a PR, that's why I added the patch file for it in this comment.
Had to rename to .txt in order to be allowed to attch.
If you want it in another way, just let me know.

[0001-SmartConfig-functions-into-StationClass.txt](https://github.com/SmingHub/Sming/files/26030/0001-SmartConfig-functions-into-StationClass.txt)

tprochazka:
I don't understand a reason for the patch file. Which you simply don't commit all changes directly to this pull request? It would by easier to comment specific changes.

But. I'm OK with all the changes. It looks good.
Only I'm not sure if is good idea to skip internal callback when user callback is set.
I'd like more variant before which allows to choose by return true or false from the custom callback-
I think that it is more flexible.

hreintke:
@tprochazka : 
The reason for patch : I don't know how submit a PR to a PR or committing changes to the PR. 
It's just a matter of experience, will get better over time.

Regarding skip internal callback processing. 
I think I did not change functionality. Also in the "original PR" the processing is completely left to the application when callback is set.
What is your idea on custom callback & true/false return ?

raburton:
Not had any more feedback about other issues with sdk v1.4 support, so are we ready to merge this and make the development branch support and require sdk v1.4?

hreintke:
@raburton : 
Against moving now to SDK 1.4.0 is that the latest UDK 2.0.8 is still on SDK 1.3.0.
As @kireevco does not have esp-open-sdk available for windows users yet, most of them will start from UDK.
What is the functionality we cannot deliver/support when staying on SDK 1.3.0 ?

raburton:
The main issue is we can't support the current espressif sdk, we have to require an old version.

hreintke:
Yes, I know. But all the Sming Windows install instructions are based on UDK -> SDK 1.3.0

raburton:
And on Linux the install instructions are based on esp-open-sdk, which is sdk v1.4. So which do we support and which do we not?

hreintke:
That's why I suggested to @kireevco to create esp-open-sdk as standalone.
Windows users can then get the toolchain from udk, sdk from sming based on esp-open-sdk
The we have the same SDK for all users, which is the preferred option (personal opinion).

hreintke:
PS : Have to go now, next answers/remarks tomorrow.

tprochazka:
> @tprochazka : 
> The reason for patch : I don't know how submit a PR to a PR or committing changes to the PR. 
It's just a matter of experience, will get better over time.

Just push another commit to the branch which is part of PR, it will be automatically added to the curent PR.


And I personally think that update espressiff SDK in the UDK is very simple, just replace folder.
At least we can have 1.4 SDK support in the development branch for now.

hreintke:
@tprochazka : 
I know with a PR initiated by me, committing to the "PR branch in my fork" will update the PR in Sminghub/sming. But This PR was initiated by @raburton, so I do not have the PR branch in my fork.

tprochazka:
Sorry. I miss that :-( So, you have right it is not possible in this situation.

tprochazka:
But would be great if @raburton will push your (@hreintke) PR to his own if he will agree with the changes, rather that it will be forgotten accidently during the merge :-(

hreintke:
@raburton : 
Saw your comment on "not upgrading to 1.4.0 due to downvoting from windows"
As we are getting several times the issue reported on using SDK 1.4.0 probably we should include this PR into develop. 
Before releasing we either will need to update the windows install instruction to include Espressif SDK or esp--opn-sdk when @kireevco finishes his  windows support.

 When merging, van you include my patch for station class update ?

raburton:
@hreintke I've added your changes to the callback/static stuff, but I can't bring myself to rename the variables 'delegate', variable names are supposed to tell you what a variable is for and calling it a callback does but calling it a delegate is just repeating its type.

hreintke:
Calling it callback is also "just repeating the type"
Everywhere in Sming the naming is delegate -> keep it identical over Sming code

raburton:
>Calling it callback is also "just repeating the type"

No it isn't, the type is Delegate, not callback. As for keeping things the same as the rest of Sming, there is another Delegate type variable with the name of callback in this very class.

hreintke:
Maybe I did not use the exact correct wording.
A delegate is a callback with added functionality.

If I missed to update one in Station class it was a mistake. 
Take a look at other locations in sming and you will see the usage.
And yes, also at some/limited other places there is still callback in use. Same as above, should not be there. But not updated by a PR to just make it sming compliant but done when a PR touches code anyway.

Issue #397, opened 2015-11-03
GPIO9 and 10 is overwritten during  Adafruit_SSD1306:display() method


tprochazka:
If you will use GPIO9 a 10 as output pins together with Adafruit_SSD1306 library, state of these GPIOS will be overwritten every time that OLED is updated by display() method.

I'm not sure just now with the exact reason now.
But I'm very close. I tried to replace 
```
*csport |= cspinmask;
*dcport |= dcpinmask;
*csport &= ~cspinmask;
*csport |= cspinmask;
```
by
```
digitalWrite(cs, HIGH);
digitalWrite(dc, HIGH);
digitalWrite(cs, LOW);
digitalWrite(cs, HIGH);
```
In Adafruit_SSD1306.cpp.
And problem is gone!

It looks that bit mask is incorrect and change more pins!
I checked that uint8_t is used for PortReg and PortMask it is correct for ESP like 32 bit CPU? 
Or it is problem in digitalPinToBitMask() or portOutputRegister() ?

tprochazka:
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/Arduino.h#L227

anakod:
Ok, got it. I think we should make minimal test code for repeating this problem and when we can start working on register implementation part.

tprochazka:
I already localize and found the problem, 8 bit variable was used in Sming instead of 32 bit for GPIO register.

anakod:
Please look my comment to related P-R. I think it isn't right and willn't work for high pin numbers.

tprochazka:
@anakod  ESP8266 Arduino project use also 32 bit register, why do you think that it will not work?

hreintke:
Issue should be moved to SmingRTOS.

tprochazka:
I don't understand why this was fixed in RTOS version and not here, id this is still under development, or not?

PR #398, opened 2015-11-03
Fix for wrong GPIO reg type (issue #397)

tprochazka:
I think that is should be like this.  (issue #397)

### DIFF
--- a/Sming/SmingCore/pins_arduino.h
+++ b/Sming/SmingCore/pins_arduino.h
@@ -20,7 +20,7 @@ extern const unsigned int A0; // Single ESP8266EX analog input pin (TOUT) 10 bit
 #define PB 2
 #define PC 3
 
-#define GPIO_REG_TYPE uint8_t
+#define GPIO_REG_TYPE uint32_t
 
 // We use maximum compatibility to standard Arduino logic.
 
@@ -34,9 +34,9 @@ extern const unsigned int A0; // Single ESP8266EX analog input pin (TOUT) 10 bit
 #define STD_GPIO_IN (PERIPHS_GPIO_BASEADDR + GPIO_IN_ADDRESS)
 #define STD_GPIO_ENABLE (PERIPHS_GPIO_BASEADDR + GPIO_ENABLE_ADDRESS)
 
-#define portOutputRegister(P) ( ((volatile uint8_t*)(P != PC ? STD_GPIO_OUT : RTC_GPIO_OUT)) + ( ( ((int)P) == PB ) ? 1 : 0) )
-#define portInputRegister(P)  ( ((volatile uint8_t*)(P != PC ? STD_GPIO_IN : RTC_GPIO_IN_DATA)) + ( ( ((int)P) == PB ) ? 1 : 0) )
-#define portModeRegister(P)	  ( ((volatile uint8_t*)(P != PC ? STD_GPIO_ENABLE : RTC_GPIO_ENABLE)) + ( ( ((int)P) == PB ) ? 1 : 0) ) // Stored bits: 0=In, 1=Out
+#define portOutputRegister(P) ( ((GPIO_REG_TYPE*)(P != PC ? STD_GPIO_OUT : RTC_GPIO_OUT)) + ( ( ((int)P) == PB ) ? 1 : 0) )
+#define portInputRegister(P)  ( ((GPIO_REG_TYPE*)(P != PC ? STD_GPIO_IN : RTC_GPIO_IN_DATA)) + ( ( ((int)P) == PB ) ? 1 : 0) )
+#define portModeRegister(P)	  ( ((GPIO_REG_TYPE*)(P != PC ? STD_GPIO_ENABLE : RTC_GPIO_ENABLE)) + ( ( ((int)P) == PB ) ? 1 : 0) ) // Stored bits: 0=In, 1=Out
 
 
 #endif /* WIRING_PINS_ARDUINO_H_ */
--- a/Sming/Libraries/Adafruit_PCD8544/Adafruit_PCD8544.h
+++ b/Sming/Libraries/Adafruit_PCD8544/Adafruit_PCD8544.h
@@ -32,8 +32,8 @@ All text above, and the splash screen must be included in any redistribution
  typedef volatile RwReg PortReg;
  typedef uint32_t PortMask;
 #else
-  typedef volatile uint8_t PortReg;
-  typedef uint8_t PortMask;
+  typedef volatile GPIO_REG_TYPE PortReg;
+  typedef GPIO_REG_TYPE PortMask;
 #endif
 
 #define BLACK 1
--- a/Sming/Libraries/Adafruit_SSD1306/Adafruit_SSD1306.cpp
+++ b/Sming/Libraries/Adafruit_SSD1306/Adafruit_SSD1306.cpp
@@ -517,14 +517,18 @@ void Adafruit_SSD1306::display(void) {
   if (sid != -1)
   {
     // SPI
+	//digitalWrite(cs, HIGH);
     *csport |= cspinmask;
+	//digitalWrite(dc, HIGH);
     *dcport |= dcpinmask;
+	//digitalWrite(cs, LOW);
     *csport &= ~cspinmask;
 
     for (uint16_t i=0; i<(SSD1306_LCDWIDTH*SSD1306_LCDHEIGHT/8); i++) {
       fastSPIwrite(buffer[i]);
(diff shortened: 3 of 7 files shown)


tprochazka:
I would appreciate some feedback ;-)

Sming/SmingCore/pins_arduino.h
```diff
@@ -32,9 +32,9 @@
 #define STD_GPIO_IN (PERIPHS_GPIO_BASEADDR + GPIO_IN_ADDRESS)
 #define STD_GPIO_ENABLE (PERIPHS_GPIO_BASEADDR + GPIO_ENABLE_ADDRESS)
 
-#define portOutputRegister(P) ( ((volatile uint8_t*)(P != PC ? STD_GPIO_OUT : RTC_GPIO_OUT)) + ( ( ((int)P) == PB ) ? 1 : 0) )
-#define portInputRegister(P)  ( ((volatile uint8_t*)(P != PC ? STD_GPIO_IN : RTC_GPIO_IN_DATA)) + ( ( ((int)P) == PB ) ? 1 : 0) )
-#define portModeRegister(P)	  ( ((volatile uint8_t*)(P != PC ? STD_GPIO_ENABLE : RTC_GPIO_ENABLE)) + ( ( ((int)P) == PB ) ? 1 : 0) ) // Stored bits: 0=In, 1=Out
+#define portOutputRegister(P) ( ((GPIO_REG_TYPE*)(P != PC ? STD_GPIO_OUT : RTC_GPIO_OUT)) + ( ( ((int)P) == PB ) ? 1 : 0) )
```
anakod: I'm not sure what it's right correction. (uint32_t* + 1) will go out of register range.

anakod:
Originaly in Sming I have used "pseudo" sub-registers with type uint8_t*  (wich cover part of original full 32-bit ESP register) to be compatible with large amount of old Arduino libraries. On other side now we have many ported libraries, and we can switch to incompatible with old Arduino 32bit register, but it need more detailed checking and refactoring. Did we have reason for that? What will be better way for end users?

tprochazka:
The reason for that is described here
https://github.com/SmingHub/Sming/issues/397
Masking operation used during SPI communication change also GPIO9 and GPIO10,  probably because these pins are outside your original 8 bit register.

Best way how reproduce is use SSD1331 example and add for example blink code for GPIO9 and 10.

robotiko:
@tprochazka 
do you think that this also will fix the  direct crash in esp-12e (nodemcu devkit 1.0) when  trying to set the pin?
Simply  
pinMode(9, OUTPUT);
digitalWrite(9, 0);)

makes a nice reset

hreintke:
@tprochazka : @robotiko : 
Can you update with status ?
If there needs to be a discussion on " "pseudo" sub-registers" before I suggest to close this PR and open an issue for further communication.

robotiko:
@hreintke 
I'm waiting for @tprochazka reply to check in related issues that are affected by this. 
So far, no news.

Sming/SmingCore/pins_arduino.h
tprochazka: I'm also not sure. I just changed uint8_t to uint32_t according Arduino/ESP8266 project to solve my problem #397 which works.

tprochazka:
@robotiko Sorry for delay. I don't think that it is related. If you check digitalWrite() implementation it already using uint32 as register if I understand it correctly.

alonewolfx2:
gpio9 issue flash chip related. if you cut gpio9 to flash connection you can use gpio9. tested,confirmed.

tprochazka:
I have no problem with using GPIOP9 on Nodemcu 1.0 without any modification.

alon24:
is your nodemcu amica brand,  do u flash dio,  and if so how,  can u attach
rooms here,  do u use linux,  Mac,  or what.  you r the only one who says
this works

robotiko:
@tprochazka 
I' m using a couple of amica nodemcu devkit 1.0  without success. 
Sming Dev code with SDK 1.4 on windows, setting the DIO mode in rboot and non rboot code examples. 
GPIO10 working without issues.. GPIO9 .. not working at all .. always reset when setting output. 
We recently found that DIO setting was having no effect while flashing ( this bug recently just got fixed)  

I read in many forums.. (many people having the same issue), that Amica modules should work due to the flash module they use, but other ESP12e.. depend on flash chip. 
So there is no reason why we are having this issue (amica users). 
Can you provide a code example or binary to test? so we can know if it is  HW related or SW related issue?

As @alon24 pointed, What is you environment? SDK, SO, Sming version? 

Thanks a lot.

zhivko:
I use nodekit 1.0 devkit12E from aliexpress:
http://www.aliexpress.com/item/10-PCS-New-Wireless-module-4M-4FLASH-NodeMcu-Lua-WIFI-Networking-development-board-Based-ESP8266/32454268280.html
And tried GPIO write with develop branch SMING, and sdk 1.5.
Result is as soon as gpio is marked for output:
pinMode(9, OUTPUT);
module freezes and then wdt reset happens.

zhivko:
According to this: http://bbs.espressif.com/viewtopic.php?t=654, the gpio9 and gpio10 must not be physically connected to flash.
Since we can do qio on nodemcu esp12e dev 1.0 it means GPIO9 and GPIO10 are connected to flash - and there fore we could not use those two pins??

alon24:
can someone run that code on devkit 1.0?
And see if this is the truth (need to flash in dio mode ofcourse)

alonewolfx2:
> the gpio9 and gpio10 must not be physically connected to flash.

as i said before.

zhivko:
I did a test on nodemcu and flash in dio mode and was NOT able to use gpio9.

alonewolfx2:
> I did a test on nodemcu and flash in dio mode and was NOT able to use gpio9.

without cutting gpio9 to flash chip connection right ?

zhivko:
Exactly.

robotiko:
Hi, since the GIPIO9 issue doesn't seem to be related to this issue, please continue side conversation in the #474

hreintke:
@All : 
Can you clarify the status of the PR. ?

hreintke:
@tprochazka @robotiko @alonewolfx2 

I will probably release a bugfix V2.1 release shortly and would like to have this included.
Can you finalize is PR and confirm correctness ?

tprochazka:
Is any action needed from my side?

robotiko:
@hreintke  my GPIOP9 issues  looks like it is not directly related to this issue and moved to #474. 
I had no additional involvement with this issue.  Didn't test.

tprochazka:
Is there still something preventing this from merge?
Or just fix this problem in better way, that I did and close this PR?

hreintke:
@zardam : @tprochazka : 
I don't how this does or does not interfere with https://github.com/SmingHub/SmingRTOS/issues/70

I prefer to have one PR which can be included in both NONOS and RTOS

hreintke:
Outdated

tprochazka:
I don't understand why this was fixed in RTOS version and not here, id this is still under development, or not?

slaff:
@tprochazka Can you check if this PR is still working with the latest develop version?

slaff:
> Fix itself is still working, but there are now some more examples which expecting bad variable type.

Can you rebase your PR based on the latest develop and then fix all examples that are affected? 

> but the Wire implementation is very outdated and there are too differences against Arduino

The Wire implementation was updated recently (0c2d053a1d22d019d117158ccb6a20cdd205345b) and  PR #1193,  that will be merged after 3.3.0 is released,  takes care to decrease the differences with Arduino

tprochazka:
I updated my pull request, there was just one more place where was necessary to use macro  GPIO_REG_TYPE. But I did not test it on the real device.

Sming/SmingCore/pins_arduino.h
```diff
@@ -34,9 +34,9 @@ extern const unsigned int A0; // Single ESP8266EX analog input pin (TOUT) 10 bit
...
+#define portInputRegister(P)  ( ((GPIO_REG_TYPE*)(P != PC ? STD_GPIO_IN : RTC_GPIO_IN_DATA)) + ( ( ((int)P) == PB ) ? 1 : 0) )
```
slaff: `volatile uint8_t*` was changed to `GPIO_REG_TYPE*` which is wrong. Without the `volatile` keyword the compiler may optimize-out the code. Please, bring back the `volatile` keyword.

slaff:
@tprochazka Can you get the latest `develop` version and tell me if the issue that you had is fixed for you?

PR #316, opened 2015-10-05
Websockets Command Processing

hreintke:
Squased PR to Develop

PR #400, opened 2015-11-04
Websockets CommandHandler implementation

hreintke:
Replaces #316, now based on latest develop


robotiko:
Tested. 
Very powerful demo. 
Working as expected. 
As suggestion to the example, add an additional command  to see how to extend the commands.

hreintke:
@robotiko : 
Thanks for the feedback. As you suggested I extended the example and the documentation of that.

@raburton : 
Can you merge this PR into Develop ? Thanks

PR #401, opened 2015-11-04
Add esp_mqtt lib as MqttClient2

sguilly:
Alternative of the MqttClient of Sming.
esp_mqtt is more robust
Thanks to tuanpmt for this lib.


hreintke:
@sguilly : 
First remark without having evaluated the full PR.
The PR doesn't compile in my environment. Do I need to set specific options or updates ? 
```
Services/espmqtt/mqtt.c: In function 'MQTT_Connect':
Services/espmqtt/mqtt.c:625:5: error: implicit declaration of function 'UTILS_StrToIP' [-Werror=implicit-function-declaration]
  if(UTILS_StrToIP(mqttClient->host, &mqttClient->pCon->proto.tcp->remote_ip))
```
Besides that : 
-The MQTT Client uses the espconn layer from espressif. Sming is using the LWIP layer underneath.
In this way we would use two network layers in Sming.
-The MQTT client is a "C implementation", Sming is C++ based, would need a "Class layer" around the library 

Can you elaborate what the issues are of the current MQTT Client library ?

sguilly:
The MQTT client is a "C implementation". 

The wrapper C++ file is https://github.com/sguilly/Sming/blob/master/Sming/SmingCore/Network/MqttClient2.cpp

First, you need to recompile the sming lib : 

In my ubuntu :
cd Sming/Sming
make

Then, you could find an example project to use it : 
https://github.com/sguilly/Sming/tree/master/MqttClient2_Hello

I know that the lib use espconn layer from espressif but i try to minimize modifications (just modify some headers) to be able to apply bug fix from the original project.

/BR

hreintke:
I know I have to compile Sming to get libsming but...
The compile error is in compiling the Sming Framework -> That is the first step I have to take and is failing.

espconn usage is a big drawback, then we get into supporting two network layers.

Indeed missed the c++ wrapper, will need rework on callback -> delegate usage, not sure how that effects mqtt layer.

And main question : What are you trying to solve with bringing a new component into Sming.

PS : Yes, if we can use quality software which is maintained, it's worth a try. For now I am just finding out the consequences.

sguilly:
This lib has a queue, a better network management when the TCP link is broken and a large community around this lib (more than 300 github starts).

I agree with the drawback to use two network layers. May be wrap the 'espconn' call to LWIP ?

I don't understand about your request to rework on callback -> delegate usage.

Could you describe more or give a name of a sming module with this implementation for example ?

I don't have compilation issue on my environnement maybe we don't have the same esp_open_sdk or os.

Travis succes to compile it. -> https://travis-ci.org/SmingHub/Sming/builds/89230250

Could you describe the default environnement use by this project ?  

I discover this project recently. I'll help you if I can.

hreintke:
- Network

Espconn is a wrap around LWIP. 
We need to check whether/how we can base this mqtt on top of the Sming network layer.

- Queue

Is there a description of the queue mechanism in use ?

- Callback/Delegate

You are right, there is no decent documentation of this important Sming component.
Looked into but there is no "easy location" where you can see the implementation details.
I will create a wlki page about callback/delegates in the coming week. Hopefully a good starting point for you. Feedback on usability appreciated.

- Environment : 

I am on Windows, Espressif SDK 1.3.0, Sming 1.4.0 (latest release).

anakod:
I'm not sure what duplication of internal component is needed (if it willn't introduce new features for end user). But I think it can be external library for user who want use esp_mqtt lib for some reason.

May be create separated repository for this component?

cucabr:
I had some problems when sending messages with  the function **MqttClient2::publish**.
I have to change in line 61 of MqttClient2.cpp
```c
(int) sizeof(message)
```
to
```c
strlen(message.c_str())
```

ashubhatt:
@cucabr  Does this library work?

cucabr:
@ashubhatt yes. i use it perfectly in my project.

ashubhatt:
@cucabr That's great. I still get error in make process.
```
Services/espmqtt/mqtt.c: In function 'MQTT_Connect':
Services/espmqtt/mqtt.c:625:5: error: implicit declaration of function 'UTILS_StrToIP' [-Werror=implicit-function-declaration]
  if(UTILS_StrToIP(mqttClient->host, &mqttClient->pCon->proto.tcp->remote_ip))
```

Any solution?

hreintke:
PR is obsolete.
Please open new issue if discussion is still necessary on esp_mqqt

Issue #59, opened 2015-05-12
Help: callback after response sent


wejedge:
Hi,
is there a way to use a callback after the response has been sent?
i.e. so that the list can be deleted?

Thanks.

Vector<String> list;

void onSendList(HttpRequest &request, HttpResponse &response)
{
     debugf("onSendList\r\n");

     JsonObjectStream* stream = new JsonObjectStream();
     JsonObject& json = stream->getRoot();

     json["count"] = list.count() / 2;

     for (int i = 0; i < list.count(); i += 2)
     {
        json[list[i + 0].c_str()] = list[i + 1].c_str();
     }
 
     response.sendJsonObject(stream);  
}

anakod:
1. You should save list in global variable
2. JSON requrie using static string variables (json lib doesn't copy pointers, just reference to them)
3. Very soon I will provide WiFi Network selection example

wejedge:
Hi,
thanks for ultra fast response.
Yes I thought about using a global, but I’m unsure how big the list is it also change's over time and don’t want to waste memory.
Maybe any other ideas?

Thanks

anakod:
I don't know you task, so I can't provide solution. Also I'm sure what without saving (to memory/to file/etc..) yo can't use data anyway.

wejedge:
I’m collecting data from a sensor (adc).
It’s not a regular period, sometimes more samples than others. The samples are recorded in ram and frequently sent to the browser.

anakod:
Main problem is what JSON doesn't save copy of your data. If you store integers/doubles/bools/etc.. (as keys or as a values, no difference) it will be saved in JSON library memory, but if you save strings, it just store pointer to you external object. So if you data can be represented as number - just save numbers. If not - additional work will be requried.

wejedge:
Ok understood thanks for the advice.

kireevco:
Closing for now

PR #402, opened 2015-11-04
ArduinoJson Upgrade from R4.2 to R5.05

hreintke:
This is the PR for upgrading the ArduinoJson package within Sming.
Current version is Release 4.2
New version is Release 5.0

You can view the changes on the ArduinoJson project changelog : 
https://github.com/bblanchon/ArduinoJson/blob/master/CHANGELOG.md

Besides the changes mentioned in there is a specific Sming update : 
**The implementation is completely ArduinoJson project -> No Sming adaptions**
This was possible due to the native String implementation within R5.
There might however be some application updates necessary in order to use R5.

I am trying to open an easy way of getting this update information gathered and accessible. 

I updated the examples which uses Json to use the new version.
HttpServer_AJAX, HttpServer_ConfigNetwork, HttpClient_ThingSpeak


tprochazka:
Great work!

HttpClient_Instapush/app/application.cpp
```diff
@@ -45,15 +45,17 @@ class InstapushApplication : protected HttpClient
 
 		DynamicJsonBuffer jsonBuffer;
 		JsonObject& root = jsonBuffer.createObject();
-		root["event"] = event.c_str();
```
anakod: Sming support storing String classes in JSON memory. Without this feature very hard to control living time of objects. It's why addCopy implemented in our ArduinoJson fork.

Related issue: https://github.com/SmingHub/Sming/issues/59
(solved in last Sming versions)

anakod:
I think it's wrong way remove all Sming specialize features in ArduinoJson library.
Arduino have very limited memory, but ESP - no, so implementation isn't the same. I'm sure update should be merged as patch to current implementation.

hreintke:
@anakod : 
The String implementation within the Sming specialized code has a serious drawback, 
When using a String, it copies the .c_st() into an object to be sure that it is still available when the string variable goes out of scope, That object is never deleted -> With every use of String available memory is going down.
The native ArduionJson String does from functionality point of view, exactly what the Sming specialization was supposed to do : Saving the .c_str() info. But now in such a way that is is deleted when jsonBuffer object is deleted or goes out of scope -> no more memory loss.

We do not limit/lower the functionality of Sming but just "fixing a serious bug" in the implementation within Sming by upgrading to "standard version" 

With the added benefit that we easily can follow the Releases from ArduinoJson project, and as I have seen in action already : Very quick feedback from project i case of issues/bugs.

I still propose as above mentioned to take the "Standard ArduinoJson implementation"

anakod:
No, ArduinoJson alvays allocate buffer for integers and internal variables (and not for strings, but for string pointers - "const char*"). addCopy just use this buffer to store full string copy. It's much better in some cases when we can't be sure how long original variable will be needed.

Example where it's important - WiFiConfig. We generate JSON with dynamic string variables and send it out as JSON. Original variablies will be freed immediately, but network stack can access to it only after some time (depends on TCP work). On other side after addCopy we can be sure what Tcp connection outgoin buffer contains full copy of strings in their JsonBuffer. It will be freed on connection closing (release Tcp connection -> release JSON obkject -> release allocated buffers).

It costs some memory but without that this didn't work and user should garant what data will be available all time while connection isn't completed.

Another issue in my point of view - Int's variables saving to JSON buffer but String isn't. It hard to understand and debug. This limitation added because Arduino have **very** limited memory, what isn't right for ESP.

> I still propose as above mentioned to take the "Standard ArduinoJson implementation"
I agree use standard lib implementation if it will contain full functionality. May right way make path for original library?

hreintke:
@anakod : 
As I said above : 
>The native ArduionJson String does from functionality point of view, exactly what the Sming specialization was supposed to do : Saving the .c_str() info. 

In order to be sure I asked the question to the ArduinoJson project (before I created the PR)
```
In the construction :

{
    StaticJsonBuffer<200> jsonBuffer;

    JsonObject& json = jsonBuffer.createObject();
    int i = 4;
    while (i--) {
        String k = "testing" + String(i) +String("_key");
        String v = "value"+ String(i+5) +String("_val");
        json[k] = v;
    }
    json.printTo(Serial);
}

Which uses local instead of global variables for key and value.

Is this a valid application ?
In other words : Will both Key and Value be copied into the jsonBuffer 
and be kept after Stings k and v are out of scope ?
```
I received the following answer : 
```
Yes, that's exactly what will happen.

The key is duplicated in JsonObject.ipp, line 194.
The value is duplicated in JsonObject.ipp, line 204.

As a consequence, you probably need to increase the size of the StaticJsonBuffer.
```
Question : 
- What additional features are you looking for ?
- Can you point me to the WIFIConfig code you are referring to so I can check ?

hreintke:
@anakod : 
Any further questions/remarks/opinions ? 

I like to merge to Develop asap to get as much as testing possible.

robotiko:
Great contribution. 
I do also believe that having a jsonArduino compatible version and all the great support that it has, will allow avoid serious  lib issues as we are having now due to not being updated for a long time. 
Tested for several days with json reading and writing with large objects with no issues.

anakod:
Yes, you are right, author of ArduinoJson library have added this functionality this summer:
https://github.com/bblanchon/ArduinoJson/commit/c161f698fca9808879e30ee3a543af8e8b4487ad

So we can switch to compatible library version without trubles. Will be fine check WiFi config example deeper before merging branch to master.

hreintke:
@anakod : @robotiko : 
Thanks for the feedback.

@raburton : 
Can you merge this PR into Develop ? Thanks

piperpilot:
I've switched over to using the new implementation in my project.  Running into a problem with JSON parsing.  First off, I get a string coming in that looks like this:

```
{"cmd":"RC", "data":{"c1":100,"c2":100,"c3":150}}
```
In different cases, cmd will change based on what the message is for.  I first check for the existence of cmd and print it out with this code:

```
				if (root.containsKey("cmd")) {
					Serial.println("GotCmd");
					root["cmd"].printTo(Serial);
```
The containsKey is new and great functionality!  What gets printed out in serial is:

```
GotCmd
"RC"
```
I then check to see what is in the cmd value with:
```
if (root["cmd"].asString() == "RC") {
```
but that never passes.

I had a working version of these checks previously...but not now.  Any suggestions?

hreintke:
@piperpilot : 
Do not now the direct answer on your question on "why" but might have to do with typeconversion/returntypes between string and const char* which is new in this version of arduinojson.

I tried using `as<String>()` instead of `asString()` and that works OK.
```
	char json[] = "{\"cmd\":\"RC\", \"data\":{\"c1\":100,\"c2\":100,\"c3\":150}}";

	Serial.printf("Json %s\r\n",json);

	StaticJsonBuffer<200> jsonBuffer;
	JsonObject& root = jsonBuffer.parseObject(json);

    if (root.containsKey("cmd")) {
        Serial.println("GotCmd");
        root["cmd"].printTo(Serial);
        Serial.printf("\r\ncmd = %s\r\n",root["cmd"]);

        if (root["cmd"].asString() == "RC"){
        	Serial.printf("asString %s\r\n",root["cmd"].asString());
        }
        if (root["cmd"].as<String>() == "RC"){
        	Serial.printf("as<String> %s\r\n",root["cmd"].as<String>().c_str());
        }
        String tmpCmd = root["cmd"].asString();
        if (tmpCmd == "RC") {
        	Serial.printf("tempString %s\r\n",tmpCmd.c_str());
        }
    };
```
We are using "standard arduinojson" without sming adaptions but of course the issue can be something to do with String implementation in Sming vs Arduinojson.

I do not have experience in handling json/using arduinojson package.
Can you check documentation on asString() within arduinojson release notes/examples (change between 4.2 & 5?).
Depending on the outcome you/me/others could submit an issue on their repo to get answers.

domonetic:
@hreintke can you helpme to install sming? i dont know how.

hreintke:
On the Readme (first page) of the repository are the links to getting started on several platforms.

domonetic:
I did everything, what it says to windows, but nothing appears on eclipse.

hice todo lo que dice para windows, pero no aparece nada en eclipse.

hreintke:
@domonetic : 
Did you also import the projects into your eclipse workspace ?
If that doesn't solve your issue., get all your logfiles at hand and go to https://gitter.im/SmingHub/Sming 
That is the location for these questions. This issue is on arduinojson upgrade and I'd like to keep this clean of other discussions.

PR #404, opened 2015-11-05
Add library for Microchip MCP23S17 SPI variant port expander

avr39-ripe:
Add arduino-based library https://github.com/n0mjs710/MCP23S17 for MCP23S17 port expanders. Needs more polishing and may be additional features, but fully working in current state.


anakod:
Is it tested on hardware?

avr39-ripe:
Yes, example that come with lib adopted by me was tested on real hardware of esp-12e + 2xMCP23S17 one chip for 16 ch input and second - for 16 ch output. It works. Lib itself isn't complet but enough to start from. Later I'll try either extend and polish it or adopt more complete lib.

flexiti:
Maybe this librarry should be also for PCF7584 ? 
(only 0,56 usd with board for LCD)  
http://www.aliexpress.com/item/IIC-I2C-Interface-LCD1602-2004-LCD-Adapter-Plate-Free-Shipping-Dropshipping/32500560356.html

avr39-ripe:
@flexiti , no, chips are incompatible. there is already lib for PCF8574 in https://github.com/SmingHub/Sming/blob/master/Sming/Libraries/LiquidCrystal/I2CIO.h so you can use it.
MCP23S17 is SPI dev, and PCF8574 is I2C.. so as always i2c vs spi :) an MCP23S17 has 16 GPIO while PCF - 8. Also there is MCP23017 same as above mcp but with I2C.

anakod:
Yes, I have tested Sming with displays driven by PCF7584 board. It's Ok.

hreintke:
@avr39-ripe : 
What are your plans regarding the : "Later I'll try either extend and polish it or adopt more complete lib"
How complete is the current implementation and what are the risks of API changes whit your coming updates ?

avr39-ripe:
Current implementation is completely working except of interrupt support. But I found out bit nicer lib. Will try to port it this week. If succeed, I close this PR and open new one. Otherwise will optimize this lib a bit and do more testing.

hreintke:
@avr39-ripe : 
Can you update on status of this PR ?

avr39-ripe:
Look around for better lib, but found that current one is good compromise between speed and features, so will not change library to something else. Restyle this lib not to use #define, but namespace to look more C++-ish, and calculate "read/write command" codes in constructor to speed things up. So it can be merged to develop I think.

hreintke:
@avr39-ripe : Can you submit a PR for SminRTOS too ?

PR #413, opened 2015-11-09
Added 0 padded output to printf.

qtonthat:
Also removed the prefix 0x for hex output.

For example,
  printf("0x%08x", 0x123)
would produce previously
  0x0x     123
now becomes
  0x00000123

Signed-off-by: Quy Tonthat <qtonthat@gmail.com>


anakod:
I think you should update printf usage for hex output in examples too.

qtonthat:
Thanks, anakod for pointing out the examples. Just looked at it and found they were a bit misleading and could use some little fixes, too.

hreintke:
@ADiea : 
As you are the "designer" of the m_print functionality, can you comment on this PR ?

ADiea:
@hreintke, @qtonthat  functionality extensions look ok, good job :)

PR #423, opened 2015-11-12
Correction of test error when using sh as default shell for make

zardam:
Fixes "/bin/sh: 1: [[: not found" message when building an app


raburton:
Tested and working on Linux. Can someone test on Windows please?

PR #429, opened 2015-11-15
digital read and write should be interrupt save

johndoe8967:
all major function got the IRAM_ATTR to guarantee save access within an interrupt


dmarkey:
+1

hreintke:
@johndoe8967 : 
This PR consists of updates for multiple issues/proposals.
Please split the PR so we can discuss and merge independently.

PR #433, opened 2015-11-18
Neopixel demo

flexiti:
1. Tested only teoreticly :) , made for @alonewolfx2 
2. Library <Adafruit_NeoPixel/Adafruit_NeoPixel.h> should also be checked / modified
3. More patterns of light would not add because the example will not be simple
4. Now all (web,tcp) should work, no need to use delay and WDT refresh - teoreticly

### DIFF
--- a/Basic_Neopixel/app/application.cpp
+++ b/Basic_Neopixel/app/application.cpp
@@ -3,6 +3,12 @@
 
 #include <Adafruit_NeoPixel/Adafruit_NeoPixel.h>
 
+
+#ifndef WIFI_SSID
+	#define WIFI_SSID "XXX" 			// Put you SSID and Password here
+	#define WIFI_PWD "XXX"
+#endif
+
 // Which pin on the Esp8266 is connected to the NeoPixels?
 #define PIN            15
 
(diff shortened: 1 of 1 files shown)


flexiti:
@alonewolfx2 @robotiko pls check, its teoretical job only

robotiko:
Hi flexiti, 
"Empirically" tested with 16 led ring. 
It just plays up to demo 5 

Now led0  has severe flickering that was not present before. 
 
what do you mean with > 2. Library <Adafruit_NeoPixel/Adafruit_NeoPixel.h> should also be checked / modified?

just .h?

flexiti:
1. TheaterChase not work yes ? I will check it again at the evening
2. explain please "now led0 (first ?) has severe.."
3. Library should also be checked if its ok, I sow some asm line, hardware depending  etc. (I will check it tomorrow)

flexiti:
@robotiko 
it stops on demo 5, still only demo 5 active ? or ESP not working from that moment ? .Can You check rs  log ?

flexiti:
if  (TheaterChaseQ < 3 )
	 {
		 // erase previous

		 if (TheaterChaseQ!=0)
		 {
			 for (i=0; i < strip.numPixels(); i=i+3)
		                  strip.setPixelColor(i+TheaterChaseQ-1, 0);        //turn prev every third pixel off
		 }

         for (i=0; i < strip.numPixels(); i=i+3)
                        strip.setPixelColor(i+TheaterChaseQ, StripColor);    //turn every third pixel on
         strip.show();


    TheaterChaseQ++  // should be added here

flexiti:
add this line as above   ( TheaterChaseQ++)

flexiti:
@alonewolfx2 
should be checked..- its future time :) 
As I wrote:
>Library should also be checked if its ok, I sow some asm line, hardware depending etc. (I will check it tomorrow)

Today I would like finish this demo.
@robotiko 
Do you have added "TheaterChaseQ++"  ? it is working now  (demo 5,6,7) ?

alonewolfx2:
@flexiti yes that changes just for esp8266.i qill test all animations tonight but why just 5 demo working?

flexiti:
@alonewolfx2 

find:
> if  (TheaterChaseQ < 3 )
 {
     // erase previous
     if (TheaterChaseQ!=0)
     {>  for (i=0; i < strip.numPixels(); i=i+3)
                      strip.setPixelColor(i+TheaterChaseQ-1, 0);        //turn prev every third pixel off
     }
     for (i=0; i < strip.numPixels(); i=i+3)
                    strip.setPixelColor(i+TheaterChaseQ, StripColor);    //turn every third pixel on
     strip.show();

After that add this line:
>TheaterChaseQ++;  

I forgot to add

robotiko:
Hi @flexiti 
Tested
Fixed the issue with demo cycles.
Led 0 timing issues are more visible in last examples.

flexiti:
@robotiko 
Please describe this led0 phenomenon better.
You can change the time (50ms) for a longer or shorter .This may allows easily notice what the nature of problem.
eg. Led0 lights shorter/longer than the other, or it will be lit after the previous demo, etc.
Today I will check Library

>Led 0 timing issues are more visible in last examples.

This issues was visable earlier ?

robotiko:
@flexiti 
looks like it is nothing related to demo but with led addressing.
It was not present in first demo code. 
As i mentioned, it is just a flicker on first led that in some animations even turn the led to green when is should be a different color.
@alonewolfx2  had similar issues with original code with his led stripes. Googling it, seems to be a common issue when strip is not driven with correct timing. 

I will upload previous and after videos somewhere and post them here.

flexiti:
@robotiko 
>common issue when strip is not driven with correct timing.

So it looks like maybe problem of library - I will look on  it

flexiti:
@robotiko 
Neopixel800 timing are 200-500 ns for LOW, and 550-850 for HIGH.
Midle is  350 and 700.
We have 400 for LOW and 800 for HIGH. 
Especially the HIGH is close to the upper limit.

Can You check how it work if we change for 350 and 700.
Just for test change this line in Adafruit_NeoPixel.cpp  (remember to recompile sming after)
>#define CYCLES_800_T0H  (F_CPU / 2500000) // 0.4us
 #define CYCLES_800_T1H  (F_CPU / 1250000) // 0.8us

for this line
>#define CYCLES_800_T0H  28
 #define CYCLES_800_T1H  56

and after this lines
>void Adafruit_NeoPixel::show(void) {
  if(!pixels) return;

 add 
 >digitalWrite(pin, LOW);

robotiko:
@flexiti 
tested. 
flicker is minimized in some cases(ColorWipe 1-4).. but in someothers (TheaterChase 5-7) now it clearly shows incorrect led activations in led 0. Maybe a lower index issue ? 
I can send you some videos on gitter if you show up  there.

alonewolfx2:
@flexiti your calculations seems resonable but one weird thing is if i connect second pixel on ledstrip everythings working exactly what i wrote. how can we explain this?

flexiti:
@robotiko @alonewolfx2 
Can You upload video on eg. Youtube ?
Send me also google link which you mentioned.
Test was done with this,yes ? 
>#define CYCLES_800_T0H 28
#define CYCLES_800_T1H 56
...
digitalWrite(pin, LOW);

WS2812B is 5V (min 3.5V), ESP8266 is 3.3V, you plug it right?

The only function that makes transmission is "strip.show ();"  from library.
In the control NeoPixel no indexing.
You just have to reset  (LOW > 50 us), and then send the appropriate number of bits (no pixels * 24 bits).
The first 24 bits are the first Neo, the second is shifted the second Neo ..etc.
The only thing that could explain such behavior, it brings up a start bit, or some random bits between good transmissions.
if You don't have logic converter from 3v3 to 5V is very easy for it.

I think that this is not a problem of the demo, just different demos can more or less accentuate the problem. In my opinion it is strip.show () or hardware (logic level)

kireevco:
Btw, Jing is a tool to upload screencasts!

robotiko:
Added IRAM_ATTR  to show as suggested by @tprochazka
And fixed flickering.

flexiti:
so this could be the reason, strip.show () should be fixed length time, readings program memory can interfere

Issue #434, opened 2015-11-18
Allow CFLAGS in Makefile-user.mk


piperpilot:
A very simple change to allow the user to pass in their own CFLAGS from Makefile-user.mk.  In my example I want to pass in a project version like this:

```
CFLAGS		+= -DSMOBOTVERSION=$(SMOBOTVERSION)
```

Pull request follows.

raburton:
Already possible, see 0a4404a33.

piperpilot:
Closing as already done.

PR #436, opened 2015-11-18
Allow CFLAGS to be passed in from Makefile-user.mk

piperpilot:
This is a fix for https://github.com/SmingHub/Sming/issues/434 to allow users to pass in their own compiler flags.


raburton:
Note the `$(USER_CFLAGS)` on the end of the current `CFLAGS`? You can use this to pass your own `CFLAGS`, while still allowing us to properly set the rest as we need them, not risking picking up a set of host `CFLAGS` from the environment.

piperpilot:
Got it...closing this issue then.  I didn't see that get added.

PR #447, opened 2015-11-25
* Replace esptool.py by esptool2 for image generation with new sdk >= 1.4.x

avr39-ripe:
* with recent update of develop branch there is no way to generate flashable image by esptool.py. There is workaround based on chat in gitter to generate images with esptool2. This method is for rBootLESS projects as rBoot aware projects already generate images by esptool2.

### DIFF
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -29,6 +29,16 @@ SPI_SIZE ?= 512K
 # Path to spiffy
 SPIFFY ?= $(SMING_HOME)/spiffy/spiffy
 
+#ESPTOOL2 config to generate rBootLESS images
+IMAGE_MAIN	?= 0x00000.bin
+IMAGE_SDK	?= 0x09000.bin
+# esptool2 path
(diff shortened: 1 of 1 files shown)


ystrem:
It's working, but I have to change path for esptool2 and compile spiffy.

raburton:
I've opened an [issue](themadinventor/esptool#69) with esptool.py about this, but in the mean time using [esptool2](https://github.com/raburton/esp8266/tree/master/esptool2) to build the rom is a good workaround. However it does add an extra dependency to those who aren't already using the rBoot in their projects.

> It's working, but I have to change path for esptool2 and compile spiffy.

Can't see any reason that you'd need to build spiffy just to use esptool2, they are unrelated tools.

ystrem:
Yes, that's correct.

2015-11-26 11:46 GMT+01:00 Richard Antony Burton <notifications@github.com>:

zardam:
I think this solution is not good. A symbol will be missing in the produced binary, resulting in at least the ArduinoJson library not working.

See #450 for more explainations.

Issue #450, opened 2015-11-26
__errno symbol not defined & esptool.py errors


zardam:
Missing __errno symbol is the root cause of esptool.py problems on branch develop.

I have tracked down the issue to "JsonVariant.cpp" which includes errno.h from the toolchain (newlib) and not from the custom libc of Sming (musl derivate), which declare errno a different way.

Using the right header from https://github.com/anakod/esp_microc/blob/master/include/errno.h solves the problem (errno is associated to the symbol __errno_location).

I'm not an expert in this domain, so I dont know exactly the right way to solve this problem. If anyone has an advice, I can make a pull request.

hreintke:
Fixed

PR #451, opened 2015-11-27
Proposal for #450 correction

zardam:
Just copied errno.h from esp_microc with some adaptations, solves #450


hreintke:
@zardam : 
I checked jsonvariant.cpp and the new version indeed included errno.h and the previous one not.
Although I do not (yet) understand the solution you propose, I still wonder why not all "json users" have the issue.

hreintke:
Solved by using esptool2

PR #453, opened 2015-11-29
Basic servo hwtimer

johndoe8967:
I created 
- a SmingCore function Hardware_Timer
- a Library for Hardware_Servo and individual Hardware_Servo_Channels

If you like I would love to share


robotiko:
@johndoe8967 
I just checked. 
Suggestions: 
-You included .travis.yml that I think is not related to this and not multiplatform. Remove
- Readme.md is an older version. Remove  
-The code in application has all the testing included that is great but makes it really difficult to follow the "basic example".
I would suggest to move test away from example  

-And finally after rebuilding Sming (current develop) I get this error while compiling 
"c:/tools/Sming/Sming/compiler/lib/\libsming.a(HardwareTimer.o): In function `Hardware_Timer::getIntervalUs()':
C:\tools\sming\Sming/SmingCore/HardwareTimer.cpp:43: multiple definition of `hardwareTimer'
c:/tools/Sming/Sming/compiler/lib/\libsming.a(HardwareServo.o):C:\tools\sming\Sming/Libraries/HardwareServo/../../SmingCore/Boards.h:17: first defined here
collect2.exe: error: ld returned 1 exit status"
That looks like hardwareTimer  is defined twice.

hreintke:
@johndoe8967 : 
Can you split this PR in two separate ones : 
1/ HardwareTimer + Digital  (all smingcore updates)
2/ Library + example for servo ?

The can we discuss each implementation in it's own "thread"
BTW I would suggest to rename the library from HardwareServo to Servo/BaseServo/GenericServo
or even HardwareTimerServo ? Seems more logical for new users.

And please squash your commits for the PR so there is only on commit per PR.

Thanks.

johndoe8967:
@hreintke 
I'm quite new to github and I'm not sure if I will do it right the next time, but I will try.
maybe it will need some time, XMAS is coming already :-)
splitting it will only make sense if you think of just accepting the HardwareTimer and not the servo.
Just tell me then I can safe the effort.

@robotiko 
Seems strange because the CI test is successful, but I will check it as soon I have time.

robotiko:
@johndoe8967 
to make it compile and work (works with no issues),  I had to comment the line 58 that defines Hardware_Timer in HardwareTimer.h   (extern Hardware_Timer hardwareTimer;) 
and also line 157 in HardwareTimer.cpp (//Hardware_Timer hardwareTimer;)

hreintke:
No, don't misunderstand, I definitely would like to include both the HWTimer and Library/example included.
I just envision that for both some updates will come so : 
1/ I'd like to keep the talks about both separate
2/ When having two separate PR's we keep the history within the sming repo clean and understandable.

The question on squashing the commits for the PR is for the same (2) reason. 
It is much clearer to have one commit saying "Servo Library and Example" than having the 27 commits from above with all detailed development commits.

johndoe8967:
ok, I now have some time again to change whats needed 
- make it work, for sure I will fix it @robotiko 
- split in 2 PR is possible
- remove travis (was not intentional)
- remove readme (was not intentional)
- separate sample and test (we really need a concept how to unit test everything)

but I'm not sure if I'm able to do it with one commit because I don't know how to reset my git fork, maybe I need some help with that.

robotiko:
@johndoe8967 
Fast review of the code,  looks ok ,  it has everything :) 
HW timer removed to use the Sming one?
I will test in same hardware tomorrow (if not today) to double check. 

Thanks.

PR #458, opened 2015-12-01
TheaterChase modify

alonewolfx2:
TheaterChase modify for first pixel from @flexiti


raburton:
@alonewolfx2 it doesn't look like this change was included in the version of you added to sming rtos yesterday. Should it be?

alonewolfx2:
I will check it

PR #463, opened 2015-12-03
Fix SDCard initialisation

ADiea:
Fix SDCard init function to return correct status - random results were returned

### DIFF
--- a/Sming/Libraries/SDCard/SDCard.cpp
+++ b/Sming/Libraries/SDCard/SDCard.cpp
@@ -388,18 +388,16 @@ DSTATUS disk_initialize (
 
 	if(ty == 0)
 	{
-		Stat = STA_NOINIT;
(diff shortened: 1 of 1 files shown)


Sming/Libraries/SDCard/SDCard.cpp
```diff
@@ -388,19 +388,17 @@ DSTATUS disk_initialize (
...
 		debugf("SDCard OK: TYPE %d\n", ty);
 	}
 
 	deselect();
 
-
-
 	SDCardSPI->setDelay(SCK_NORMAL);
 
 	return s;
```
raburton: could you just return stat and get rid of s altogether?

ADiea: done, thanks for spotting

Issue #467, opened 2015-12-05
UART receive buffer overflow


festlv:
A fatal exception happens when sending large amounts of data to ESP over UART.

Here is a stripped-down test application: timer executing every 10ms tries to read as much of received UART data as possible and prints out how much did it receive:

app/application.cpp
```c++
#include <user_config.h>
#include <SmingCore/SmingCore.h>

void timer_interrupt();
Timer uart_recv_timer;

void timer_interrupt() {
    static uint8_t in_buffer[RX_BUFF_SIZE];
    int read_len = Serial.readMemoryBlock((char*)in_buffer, RX_BUFF_SIZE);

    if (read_len > 0) 
        debugf("Read: %d bytes\n", read_len); 
}


void init()
{
	Serial.begin(921600);
	Serial.systemDebugOutput(true);
    //start timer to call timer_interrupt every 10ms
    uart_recv_timer.initializeMs(10, timer_interrupt).start();
}
```

Test application in python- writes 1024 bytes of data:
write-test.py
```python
#!/usr/bin/env python2

import serial
import time

s = serial.Serial("/dev/ttyUSB0", timeout=0, baudrate=921600)

print s.read(1024)

data = "B" * 1024

s.write(data)

time.sleep(0.5)

print s.read(1024)
```

And here is the output of running this script:

```

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Read: 256 bytes

Fatal exception 3(LoadStoreErrorCause):
epc1=0x402370ac, epc2=0x00000000, epc3=0x00000000, excvaddr=0x40000000, depc=0x00000000
```
After this the ESP resets and curiously enough, I see reset cause 2 which would normally indicate reset by pulse on nRST.

[...]

PR #468, opened 2015-12-05
Fixes out-of-bounds access when reading from UART

festlv:
This bug was triggered by uart0_rx_intr_handler leaving rxBuf.pReadPos
at the last element of rxBuf and subsequent reads from rxBuf incrementing
pReadPos past the bounds of array and then comparing to array size.


festlv:
Related issue: #467

hreintke:
Close & reopen to trigger travis rebuild

PR #469, opened 2015-12-05
Fix Makefile targets

ADiea:
When setting more patchs for include dirs and/or for module targets the operator ?= will not work anymore.+= is needed, and is also backwards compatible to the case when the custom application makefile does not set any values for MODULES and extra include dir.

### DIFF
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -133,8 +133,8 @@ TARGET		= app
 
 # which modules (subdirectories) of the project to include in compiling
 # define your custom directories in the project's own Makefile before including this one
-MODULES 	?= app  # if not initialized by user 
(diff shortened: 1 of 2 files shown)


hreintke:
@raburton : @kireevco : 
Can you comment on this PR and if OK, merge into develop ?

kireevco:
older .travis.yml here, doesn't pull right sdk. @ADiea can you please get the latest so the test won't fail?

ADiea:
Sure sorry old PR, didn't realise

raburton:
Looks fine to me.

raburton:
I think I'm going to have to reverse my opinion here. Why do we want to append to the modules list? If the user wants to set it that's fine, if they want to keep 'app' in the list they can add it to their own, that's the default in the Makefile-user.mk files shipped with the samples so they just need to add their own modules to the list. If they don't want 'app' at all, they define their own variable and we shouldnt; add app to it. Happy to keep the EXTRA_INCDIR, because we add a lot of stuff that we should, but I think better to separate the two so they can override the default project include directory (as they can for app), but always add the Sming stuff on too. I'll submit a new pr...

ADiea:
You are right for the modules, should be reversed. Will you make a PR out of your latest commit or should I do it?
Thank you for noticing!

raburton:
Done, see #507. Standardised all the sample makefile-user.mk files too (so you'll have to search the changes page to see the most important could of files).

PR #471, opened 2015-12-06
Hardware Timer: Interface similar to Software Timer

johndoe8967:
Digital IO has to be interrupt save to be used in Hardware Timer applications

### DIFF
--- a/Sming/SmingCore/Digital.h
+++ b/Sming/SmingCore/Digital.h
@@ -16,13 +16,13 @@
 
 void pinMode(uint16_t pin, uint8_t mode);
 
-void digitalWrite(uint16_t pin, uint8_t val);
-uint8_t digitalRead(uint16_t pin);
+void IRAM_ATTR digitalWrite(uint16_t pin, uint8_t val);
+uint8_t IRAM_ATTR digitalRead(uint16_t pin);
 
-void pullup(uint16_t pin);
-void noPullup(uint16_t pin);
+void IRAM_ATTR pullup(uint16_t pin);
+void IRAM_ATTR noPullup(uint16_t pin);
 
-bool isInputPin(uint16_t pin);
+bool IRAM_ATTR isInputPin(uint16_t pin);
 
 unsigned long pulseIn(uint16_t pin, uint8_t state, unsigned long timeout =
 		1000000L);
--- a/Sming/SmingCore/HardwareTimer.cpp
+++ b/Sming/SmingCore/HardwareTimer.cpp
@@ -41,6 +41,7 @@ typedef enum {
 
 static void IRAM_ATTR hw_timer_isr_cb(void *arg)
 {
+	if (arg == null) return;
 	Hardware_Timer *ptimer = (Hardware_Timer*)arg;
 	ptimer->call();
 }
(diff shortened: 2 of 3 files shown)


hreintke:
@johndoe8967 : 
Thanks for spiltting the PR into core and lib/example.

My remarks on this HW Timer : 
- Now the interface is (nearly) identical to Software Timer. What are pro's con's in combining the two and get something like Timer.setHardwareTimer(bool) to get to a HW vs Software.

- The defines you are including are they for both hw timer and hw pwm ?

- in consructor you "ETS_FRC_TIMER1_INTR_ATTACH((void*)hw_timer_isr_cb, (void *)this);"
but in deconstructor you don't "unattach' is that correct ?

- You only can use HWTimer once. Is there a possibility to check whether the HW timer callback is already in use and if it is return error ? 

- Same for HW PWM, only one of the two can be active simultaneous. 

- Should we include initializeMS ? hwtimer is for very precise timing, maybe only for microseconds ?

johndoe8967:
My remarks on this HW Timer :
-> and my answers

* Now the interface is (nearly) identical to Software Timer. What are pro's con's in combining the two and get something like Timer.setHardwareTimer(bool) to get to a HW vs Software.
-> I thought about it, but a SW Timer is freely instance able, the HW Timer may only exist once, it would be good to have a check if more than one instance exist.
-> the SW Timer allows function calls which are not in IRAM, the HW Timer uses interrupt so every function has to be in IRAM, this is why it is not possible to use the delegates

* The defines you are including are they for both hw timer and hw ppm ?
-> I never had a look at the hw pwm function, so I don't know

* in consructor you "ETS_FRC_TIMER1_INTR_ATTACH((void*)hw_timer_isr_cb, (void *)this);"
but in deconstructor you don't "unattach' is that correct ?
-> this could be improved :-)

* You only can use HWTimer once. Is there a possibility to check whether the HW timer callback is already in use and if it is return error ?
-> yeah also a good idea, I also thought about static declaration of the HW timer, but then it would always use resources 

* Same for HW PWM, only one of the two can be active simultaneous.
-> I don't know how to check against hw pwm, maybe they need a common base class

* Should we include initializeMS ? hwtimer is for very precise timing, maybe only for microseconds ?
-> up to 8388ms it works great, so I do think it would be easy to use the microseconds.

=> should I commit the changes into this branch or do I have to make a new branch?

johndoe8967:
* in consructor you "ETS_FRC_TIMER1_INTR_ATTACH((void*)hw_timer_isr_cb, (void *)this);"
but in deconstructor you don't "unattach' is that correct ?
-> this could be improved :-)
=> there is no detach function, the only possibility would be to attach a null pointer but I'm not sure if this is better than just doing nothing

hreintke:
On unattch : When you leave it as it is, the hwcallback (which is static) cannot "see" whether the object is still available. If you detach with "ETS_FRC_TIMER1_INTR_ATTACH((void*)hw_timer_isr_cb, NULL);" 
you can check for nullptr in the callback.

When you push updates to the branch in which you have HWTimer, it will automatically update the PR

johndoe8967:
you are right, the PR is containing now the Detach function

hreintke:
close & open to trigger travis recheck

raburton:
@hreintke you can locate the pull request on travis and request a rebuild there

robotiko:
@johndoe8967  
could you please help us to add this functionality to the sming RTOS version? (https://github.com/hreintke/Sming_RTOS_POC/issues/37)

This was finally merged into Sming ;D

thank you!

johndoe8967:
@robotiko 
Yeah, great.
Of course I will help.

PR #472, opened 2015-12-06
Servo library + Basic Servo sample

johndoe8967:
Servo library: supports generic RC servo on GPIO
Basic_Servo: sample for the servo library

### DIFF
--- a/Basic_Servo/.cproject
+++ b/Basic_Servo/.cproject
@@ -1,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?fileVersion 4.0.0?><cproject storage_type_id="org.eclipse.cdt.core.XmlProjectDescriptionStorage">
+	<storageModule moduleId="org.eclipse.cdt.core.settings">
+		<cconfiguration id="cdt.managedbuild.toolchain.gnu.mingw.base.1135534147">
+			<storageModule buildSystemId="org.eclipse.cdt.managedbuilder.core.configurationDataProvider" id="cdt.managedbuild.toolchain.gnu.mingw.base.1135534147" moduleId="org.eclipse.cdt.core.settings" name="Sming">
+				<externalSettings/>
+				<extensions>
+					<extension id="org.eclipse.cdt.core.ELF" point="org.eclipse.cdt.core.BinaryParser"/>
+					<extension id="org.eclipse.cdt.core.GCCErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
+					<extension id="org.eclipse.cdt.core.GASErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
+					<extension id="org.eclipse.cdt.core.GLDErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
+					<extension id="org.eclipse.cdt.core.GmakeErrorParser" point="org.eclipse.cdt.core.ErrorParser"/>
+					<extension id="org.eclipse.cdt.core.CWDLocator" point="org.eclipse.cdt.core.ErrorParser"/>
+				</extensions>
+			</storageModule>
+			<storageModule moduleId="cdtBuildSystem" version="4.0.0">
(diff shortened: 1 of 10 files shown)


robotiko:
@johndoe8967 
just checkout.  Building Sming got this error:
C+ Libraries/Servo/Servo.cpp
Libraries/Servo/Servo.cpp:10:27: fatal error: HardwareTimer.h: No such file or directory
 #include <HardwareTimer.h>

I just saw that you have another PR containing the HW timer that need to be merged before this one.

johndoe8967:
yes, the last PR contained both, but there was the wish to split into the Timer and the ServoLib+Sample
So yes, this PR alone will not work intentionally and we have to wait for merge :-)

But of course someone might use it from my fork in the meantime

robotiko:
This depends on #471

robotiko:
Tested the new lib with the HW Timer PR and working properly in a custom project.

johndoe8967:
yeah, very nice to hear :-)
hopefully it will be pulled into sming

robotiko:
Final delivery is very good. Interfaces, abstraction and easy to use. 
Very similar to the Arduino one, so newcomers will feel at home :) 

My only suggestion would be to make a clearer/ simpler example that doesn't mix with defined the to position modes. I think I can recover my Sweep example once merged. 

By the way, the unit testing was great to see, but was really messy in a  basic example BUT as separated test.. is a great example of to do quality code.   It is gone and Miss it :) 

Thanks a lot for the contribution.

johndoe8967:
If you have a nice sample you may send me the code :-)

The unit test will come again as soon as someone manages to get a unit test framework running.
But you are right, we really need a good tested framework.

raburton:
see #546

hreintke:
@johndoe8967 : 
I tried the Basic_Sevo using 2 pins simultaneous using gpio 2 and 4 and see the folowing :  
Looks like when using the two, only one works.
Can you check this ?

GPIO2 only : 
gpio2

GPIO4 only : 
gpio4

GPIO2 and GPIO4 : 

gpio2_4

johndoe8967:
Oh, thanks for the good measurements 
I will check today evening 

Br
John

Von meinem iPad gesendet

johndoe8967:
@hreintke: could I have a look at your application.
I only have one output on my device, but I will add an additional output for further tests.

hreintke:
@johndoe8967 :
I used your basic_servo example. 
Defined a ServoChannel s2* and just added the same servo statement everywhere it occurs in the example.

hreintke:
@johndoe8967 : 
I have also been looking at the Servo/ServoChannel implementation.
There are some items I think should be changed in order not to get application issues when usiing the libraty : 
- You have a lot of global variables.They should move to class and be class variables.
F.e. I had an int array pins[2] But yu also have pins in the lib -> doesn't work.
- I propose change the naming to : Servo -> ServoControl, ServoChannel -> Servo
- There are checks in parameter input but functions do not return error
- Should class functions be identical to Arduino service class to be compatible

johndoe8967:
Hello Hreintke,

You are right I use global variables, this is just because I didn’t manage to get the Interrupt working with a class method instead of a function (ServoTimerInt) and of course a function can not access the private members directly.
I may have to review the interrupt .

I can change the naming, basically I have no prefer any suggestion, if everybody is ok with the change I can do that

Which checks do you mean? I thought I have a boolean return value on each setter function.

I don’t know arduino but what would be the benefit on being compatible with it?

hreintke:
Hi johndoe8967,
See the HardwareTimerUpdate branch in the SminRTOS repo
https://github.com/SmingHub/SmingRTOS.git

There I updated the servo class to be able to access class variables from static functions.
Only did for hardwareTimer but should be identical for other globals;

I propose change the naming to : Servo -> ServoControl, ServoChannel -> Servo

Arduino compibility : a lot of esp users & libraries come from the arduino world.
Having the same methods eases use/conversion of arduino provided libraries.
That is why I propose to add arduino method names.

hreintke:
Isuue fixed by #599.
Other updates should be in a separate issue/PR

PR #478, opened 2015-12-09
Fix #447, shrink Sming size

### LINKED ISSUE #477: Huge program can't boot
My (probably) huge sming based program can't boot any longer.
I was adding code (and const data) to the program and all worked fine(build download and run)
However after adding some new code today, still compiles and downloads without a problem but do not get beyond the boot message. If I remove this string or comment out some code then everything is fine Details:
sming V1.4.0 on  Windows, installed with Chocolatey.
Boot message(more details follows):
     ets Jan  8 2013,rst cause:2, boot mode:(3,6)
    
    load 0x40100000, len 22364, room 16
    tail 12
    chksum 0x7d
    ho 0 tail 12 room 4
    load 0x3ffe8000, len 880, room 12
    tail 4
    chksum 0x73
    load 0x3ffe8370, len 13652, room 4
    tail 0
    chksum 0xc6
    csum 0xc6
    csum err
    ets_main.c

My link script is the standard sming 'eagle.app.v6.cpp.ld' with a modified line:
`irom0_0_seg :                         org = 0x40209000, len = 0xF0000`

The download command:
`/c/Espressif/utils/esptool.exe -p COM2 -b 256000 write_flash -ff 40m -fm qio -fs 4m 0x00000 out/firmware/0x00000.bin 0x09000 out/firmware/0x09000.bin 0x54000 out/firmware/spiff_rom.bin`

Compiler/linker output:
`#Memory / Section info:
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8370|     880
    rodata|           ReadOnly Data (RAM)|    3FFE8370|    3FFEB8C4|   13652
       bss|      Uninitialized Data (RAM)|    3FFEB8C8|    3FFF32E8|   31264
      text|            Cached Code (IRAM)|    40100000|    4010575A|   22362
irom0_text|           Uncached Code (SPI)|    40209000|    402518C6|  297158
Total Used RAM : 45796
Free RAM : 36124
Free IRam : 10424
\# Generating image...
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8370|     880
    rodata|           ReadOnly Data (RAM)|    3FFE8370|    3FFEB8C4|   13652
       bss|      Uninitialized Data (RAM)|    3FFEB8C8|    3FFF32E8|   31264
      text|          Uncached Code (IRAM)|    40100000|    4010575A|   22362
irom0_text|             Cached Code (SPI)|    40209000|    402518C6|  297158
Entry Point : 40100004 call_user_start()
Total Used RAM : 45796
Free RAM : 36124
Free IRam : 10406 or 26790 if 48k IRam
Generate firmware images successully in folder out/firmware.
Done
# Generating spiffs_bin
Checking for spiffs files
web/build directory exists. Creating out/firmware/spiff_rom.bin
Creating rom 'out/firmware/spiff_rom.bin' of size 0x40000 (262144) bytes.
Adding files in directory 'web/build'.
Added 'bootstrap.css.gz' to spiffs (15615 bytes).
Added 'index.html' to spiffs (12255 bytes).
Added 'ipsettings.html' to spiffs (4208 bytes).
Added 'jquery.js.gz' to spiffs (30153 bytes).
Added 'lockcfg.html' to spiffs (4025 bytes).
Added 'style.css' to spiffs (7329 bytes).
Added 'wifi-sprites.png' to spiffs (1769 bytes).
out/firmware/spiff_rom.bin---------->0x54000`

I don't need  OTA/FOTA. Just want the maximum space available (nodemcu with ESP-12 and a W25Q32BV =>4MBYTE)
Any help will be appreciated. Need more info? let me know

ADiea:
If someone doesn't use the commandExecution geature, he/she can save 3.3KB flash and 1.3KB RAM if making libSming like this
make rebuild DISABLE_FEATURE_COMMAND_EXECUTOR=1 

Might fix #477

   Section |                   Description | Start (hex) |   End (hex) | Used space
------------ | ---------------------------- | -------------- | ---------- | ----
      data |        Initialized Data (RAM) |    3FFE8000 |    3FFE842D |    1069
    rodata |           ReadOnly Data (RAM) |    3FFE8430 |    3FFEA490 |    8288 (-1288)
       bss |      Uninitialized Data (RAM) |    3FFEA490 |    3FFF1018 |   27528 (-120)
      text |            Cached Code (IRAM) |    40100000 |    40105D0B |   23819
irom0_text |           Uncached Code (SPI) |    40209000 |    402430A7 |  237735 (-3372)

### DIFF
--- a/Sming/SmingCore/Network/HttpServer.cpp
+++ b/Sming/SmingCore/Network/HttpServer.cpp
@@ -135,7 +135,9 @@ void HttpServer::processWebSocketFrame(pbuf *buf, HttpServerConnection& connecti
 		msg.setString((char*)data, size);
 		debugf("WS: %s", msg.c_str());
 		if (sock && wsMessage) wsMessage(*sock, msg);
+#ifndef DISABLE_FEATURE_COMMAND_EXECUTOR
 		if (sock && sock->commandExecutor) sock->commandExecutor->executorReceive(msg+"\r");
+#endif
 	}
 	else if (frameType == WS_BINARY_FRAME)
 	{
--- a/Sming/SmingCore/Network/WebSocket.cpp
+++ b/Sming/SmingCore/Network/WebSocket.cpp
@@ -17,10 +17,12 @@ WebSocket::WebSocket(HttpServerConnection* conn)
 
 WebSocket::~WebSocket()
 {
+#ifndef DISABLE_FEATURE_COMMAND_EXECUTOR
 	if (commandExecutor)
 	{
 		delete commandExecutor;
 	}
+#endif
 }
 
 bool WebSocket::initialize(HttpRequest& request, HttpResponse& response)
@@ -66,8 +68,10 @@ void WebSocket::sendBinary(const uint8_t* data, int size)
 
 void WebSocket::enableCommand()
 {
+#ifndef DISABLE_FEATURE_COMMAND_EXECUTOR
 	if (!commandExecutor)
 	{
 		commandExecutor = new CommandExecutor(this);
 	}
+#endif
 }


hreintke:
@ADiea : 
First of all we should first check/find the reason for the increase of memory/rom usage. That would benefit all users, not only the "non command" users. Did a check and maybe the registering of the system commands in the constructor is the one causing the increase. I remember a similar item when first introducing commands. 

When are deciding to implement a "feature choice" that should be a generic solution which facilitates  and makes sure that usage/supporting and debugging stays easy and understandable, both for users and developers.

Currently a lot of options are defined in makefiles, that makes it difficult to maintain and keep the overview, we should move that to config files.

I have been thinking of a "three way level" 
1/ Global Config with all options defined and giving them a default value.
2/ Sming Core config -> Only updates Global Config which are necessary for the required build of Sming Framework
3/ Application config -> Only updates to the Sming Core Config. 

This way the overview/settings will be clear and located in a structural way.

Besides this config files, we also should be implementing the features in an agreed way.
I would prefer that an application, which uses "opted out" features just will not compile. 
Next option -> will not link.

If parts of the code is "silently taken out" we will get confusion and thus support questions.

F.e. in your patch above. All compile, links and runs as expected, even can set the options in code.
But.. will not give the expected output.

ADiea:
**TL;DR:** You are right that this is a workaround and it's not the proper way to do it, but I need your help to make it cleaner -> My suggestion: Why can't we move the calls to command executor in the example application's callback, on the message received user callbacks? That way the Command executor & all will be included in the binary only by the apps that really use it. 

@hreintke  Please don't get me wrong, i believe this is a great feature and already some appreciate and use it - it's just that for others myself included it's not (yet) necessary - also because we are super limited with ram and flash I work very hard to reduce code size so I don't like when I see HardwareSerial / CommandExecutor/Handler/etc in my mapfile eating up 3KB of flash because it is called somewhere in the httpServer by the framework ?!

In my oppinion, to have command executor code inside http server / websocket server / telnet server etc is breaking the framework concept because it is coupling everything together. 

If I write an application and I use telnet and websockets, it seems natural to me that the command execution is handled **in my own user callback**, if i want to define commands etc i don't want to make Sming PRs to extend functionality in CommandExecutor.

ADiea:
diverging opinions here, keeping modifications locally until command executor is decoupled from sming core like websock and httpserver and inserted into user callback Closing

hreintke:
@ADiea 
It is not my opinion that we should not introduce options like the ones you propose.
What I would like is that we introduce this feature in such a way that we also can use it in a similar way for other functionalities.

In the Sming_RTOS version we do introduce the possibility for applications to have "Sming Global" options, set to default value in Sming, changeable  by applications.
Although naming of that option file is not fixed, probably the examples will get an empty user_config.h to in which the application specific values can be set.

This feature can be easily extended for compile time options, which I can do in the current version of Sming_RTOS.
If these compile time options are careful and clear implemented for applications that are using/not using those it is a valid possibility.

>Besides this config files, we also should be implementing the features in an agreed way.
I would prefer that an application, which uses "opted out" features just will not compile or link

>If parts of the code is "silently taken out" we will get confusion and thus support questions.

>F.e. in your patch above. All compile, links and runs as expected, even can set the options in code.
But.. will not give the expected output

I do think that with the move from NONOS to RTOS there are more RTOS SDK/SMING Framework features that can be handled in such a way.
If you have suggestions for those, please let them know and we can discuss the implementation.

PS : On your remark :
>If I write an application and I use telnet and websockets, it seems natural to me that the command execution is handled in my own user callback, if i want to define commands etc i don't want to make Sming PRs to extend functionality in CommandExecutor.

The commandexecutor is designed in the way you mention.
Only a few system commands are implemented in the feature itself, all application commands are within the application callbacks. There is no need to PR for application commands.
You can see examples of that in the Commandprocessing_debug example where `processApplicationCommands` and `ExampleCommand::processExampleCommands` are used to show the possibilities.

PR #486, opened 2015-12-16
DS81S20 temperature readings below zero fix

flexiti:
DS18S20 temperature readings below zero path


hreintke:
Close & reopen to trigger travis rebuild

flexiti:
@hreintke 
should be also improved in version RTOS

hreintke:
@flexiti : 
Yes, This should also be included in Sming_RTOS.
Could make a PR that for that too ?

flexiti:
ok

PR #489, opened 2015-12-17
Serial Callback Update

hreintke:
Update Serial Delegate & CommandProcessing

### DIFF
--- a/Sming/SmingCore/HardwareSerial.cpp
+++ b/Sming/SmingCore/HardwareSerial.cpp
@@ -29,9 +29,8 @@ HardwareSerial::HardwareSerial(const int uartPort)
 {
 	resetCallback();
 	// Start Serial task
-	serialQueue = (os_event_t *)os_malloc(sizeof(os_event_t)*TEST_QUEUE_LEN);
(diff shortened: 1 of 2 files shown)


flexiti:
@hreintke 
eg. void serialCallBack(Stream& stream, char arrivedChar, unsigned short availableCharsCount) 
  from rboot demo  stopped working

hreintke:
@flexiti : 
I should have mentioned the limitations of the patch also in the PR and not only in the issue.

>In there I changed the handling of Serial Delegate & CommandProcessing within the HardwareSerial Class.

>PS : In this patch the availableCharsCount parameter of the delegate does not yet contain the actual count but always is one (1).

The rboot demo uses the availableCharsCount.

hreintke:
I updated the PR to a complete version, supporting all current functionalities of serial delegate & commandprocessing.

Please test and feedback.

raburton:
@hreintke does this need porting to rtos, or is it included / superseded by your rtos serial work?

hreintke:
No, do not merge this one (and any other for HardwareSerial)
HardwareSerial in RTOS has ben rewritten to use freertos specific functions

PR #503, opened 2015-12-23
HttpClient - Rest client support

fcgdam:
The downloadString function for the HttpClient, only supports GET and POST requests. There is no other way to use this function to make, for example a PUT request, or other types of REST operations to a REST server.

So my changes add 5 new functions to HttpClient, that allow to do the GET/POST/PUT/DELETE verbs to be used directly, or any other, more uncommon verbs, like HEAD, PATCH for example to be used, allowing the Sming framework now to be used to build full REST clients.

I added an example for the client and for the server (in Node.Js). The Sming REST Client, cycles through the possible REST requests, while the REST server shows the request type and JSON body.

### DIFF
--- a/Basic_RestClient/README.md
+++ b/Basic_RestClient/README.md
@@ -1,0 +1,15 @@
+#Rest Test Server
+
+Deploy the Rest Test server on a machine that has Node.js installed.
+
+Move to the directory where resttest.js and package.json are copied.
+
+Do a: npm install
+
+to install dependencies.
+
+Run the Server with:  node resttest.js
+
+The REST API will be available at: http://mymachine_address:3003/api/test
+
+The Sming App running on the ESP8266 will call the API and check each REST verb.
--- a/Basic_RestClient/Sming_RestClient/Makefile
+++ b/Basic_RestClient/Sming_RestClient/Makefile
@@ -1,0 +1,24 @@
+#####################################################################
+#### Please don't change this file. Use Makefile-user.mk instead ####
+#####################################################################
+# Including user Makefile.
+# Should be used to set project-specific parameters
+include ./Makefile-user.mk
+
+# Important parameters check.
+# We need to make sure SMING_HOME and ESP_HOME variables are set.
+# You can use Makefile-user.mk in each project or use enviromental variables to set it globally.
+ 
+ifndef SMING_HOME
+$(error SMING_HOME is not set. Please configure it in Makefile-user.mk)
+endif
+ifndef ESP_HOME
+$(error ESP_HOME is not set. Please configure it in Makefile-user.mk)
+endif
+
+# Include main Sming Makefile
+ifeq ($(RBOOT_ENABLED), 1)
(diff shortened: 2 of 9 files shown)


hreintke:
@fcgdam : 
Could be a nice addition to the HTTPClient.
My remarks : 
- There is quite some code duplication in "download" and "request" code.
Can you try to combine the two 
- Your example uses an LCD display. Can you update the example not to require additional hardware ?

fcgdam:
Hi:
  Regarding the code duplication, yes, sure. I did notice the code duplication, but didn't want to change the original code. I'll see how can i merge the two functions. I'm aware the we need to keep the framework foot print as light as possible.
  Regarding the LCD, it also can be removed.

hreintke:
@patrickjahns : 
Can you take a look at this PR and indicate : 
- it is OK to merge
- it is not relevant anymore due to your updates to come

patrickjahns:
@hreintke 
just had a look - the changes are a good addition to sming. My proposed changes will not include this

I will test and see if it is ready for merge

My remarks so far
- I find the naming with `HttpClient::doVerb` a bit odd - I`d use `HttpClient::Verb`
- I would change the explicit String definitions (i.e. `PUT`) to the Webconstant definitions *i.e. `RequestMethod::PUT` )
- Remove unecessary whitespaces

patrickjahns:
@hreintke 

Some further remarks on this:
```
bool doPost( String url, HttpClientCompletedDelegate onCompleted);
bool doPut(  String url, HttpClientCompletedDelegate onCompleted);
```

should take a Post body as argument instead of needing to use:
```
void setPostBody(const String& _method);
```

```
bool doPost(String url, String body, ...)
```

In general I`d suppose it would be good to change the general interface of the HttpClient to be a bit more straight forward.

For example - we require to set RequestHeaders via a seperate function - but then we use the URL in the processing request. It would be more aligned if we have something among having a request object that has the information for a request. A client object processing the request, and a Response object holding the results of the execution of the request.

So seperating the functional part (processing the request) and the object part request/response accordingly. This would align with the talk we had about a client having a connection and being a connection. 

Example code
```
client = HttpClient()
request = HttpRequest()
request.setUrl("url")
request.setHeader("...")
request.setCookie("...")
HttpResponse response = client.POST(request)
body = response.getBody() ...
```

hreintke:
@patrickjahns : 
Good and structural ideas, we should make an overall idea where we want to move to so that we can work on it in successive PR's.

Beware on the async nature of Sming. 
`HttpResponse response = client.POST(request)` would not be possible.

Vity01:
Don't forget to add a support for sending binary content (like s file with > 10kB). Not just String only.

rosiste:
Hi guys, any progrress on that? Realy need this feature (PUT method for strings). Can I help with that to make it live? Thanks.

slaff:
That will be implemented as part of the [http refactoring](https://github.com/slaff/Sming/issues/5).

Issue #491, opened 2015-12-17
LWT feature of MQTT not working


ashubhatt:
I am using Sming 1.4.0 and trying the sample code "MQTTClient_Hello".
Although the LWT function is implemented, the actual LWT message is not received in case of abrupt disconnection.

I have opened issue #337, but it was closed assuming the latest build have added the feature; the bug still remains.

Kindly update the source code.

Informatic:
I can confirm it is working now.

matoushybl:
Is LWT working with the patch? After setting LWT and disconnecting power, it was still not sent.

flexiti:
Are You using last develop ?

matoushybl:
Yes, I do. Is the keep alive interval important for LWT? I mean if it is for example 1 hour, will it send LWT after the one hour?

flexiti:
no, keep alive is not important for LWT if socket is closed eg. no internet connection
(depends on broker)
For me LWT is working after aprox. 30 sec then device is off

> if(!mqtt.setWill("main/device/100","{\"Value\":0,\"No\":100,\"Status\":\"Off Line\",\"Time\":\""+SystemClock.getSystemTimeString()+"\"}", 1, true)) {
> 			debugf("Unable to set the last will and testament. Most probably there is not enough memory on the device.");
> 		}
> 
> 	mqtt.connect("esp8266-100",MQTT_USERNAME,MQTT_PWD);
> 	mqtt.subscribe("main/#");
> 	mqtt.publishWithQoS("main/device/100","{\"Value\":1,\"No\":100,\"Status\":\"On Line\",\"Time\":\""+SystemClock.getSystemTimeString()+"\"}",1,true);
>

matoushybl:
I have a very similar code as yours, I am starting to think that this is caused by broker, at my second attempt the LWT was received after more than two hours. What broker are you using?

flexiti:
Mosquitto on my serwer

matoushybl:
Thanks, me too. That is weird, I updated mosquitto and it still isn't working, I'll dig deeper to find out what the problem is.

matoushybl:
Yeah, by power off, I haven't noticed the question, sorry.

hreintke:
Fixed

PR #505, opened 2015-12-24
MQTT LWT Fix : issue #491

etmmahi:
Fix for #491 
In MqttClient::connect() mqtt_init() cleared the LWT already set. 
Pull request changes order of initialization
In constructor the broker is cleared. Later in connect the clientid is overwritten when supplied.


Informatic:
I can confirm it is working fine now.

hreintke:
@etmmahi : 
- Will this also work when the application calls connect multiple times on an instance ? 
- Better to make mqtt_init(&broker, NULL) -> makes clear it is a pointer instead of a value.

etmmahi:
Hi, thanks for testing an commenting.

I have updated the pull request with the NULL pointer. (Style update)

Furthermore the client can still reconnect with these changes on the same instance.
Even better, there was a memleak in the LWT implementation when a LWT was set for the connection. 
The LWT was allocated again for the new connection instance. (previous LWT was not freed). This is now also fixed.

ashubhatt:
I have tested the code. It works.

etmmahi:
Hi
@ashubhatt 
Where did your fix for "LWT will not work when used with MQTT username and password." go?

ashubhatt:
Hey @etmmahi ,

Kindly find the fix at: https://github.com/ashubhatt/testSming/

gschmottlach:
Looking at the  fix here: https://github.com/ashubhatt/testSming/

It doesn't look like "will retain" and "will QoS" parameters are being encoded into the flags parameter in the mqtt_connect() function.  I think something along the lines of the following should be added:

```
if(willtopiclen) {
	payload_len += willtopiclen + strlen(broker->will_message) + 4;
	flags |= MQTT_WILL_FLAG;

	if ( broker->will_retain ) {
		flags |= MQTT_WILL_RETAIN;
	}

	flags |= (((unsigned)broker->will_qos & 0x03) << 3)
}
```

Without setting the ```flags``` variable you can't set the retained or QoS options for the LWT.

ashubhatt:
@gschmottlach Thanks for noticing that.

I forgot to look at the flags. The library is patched now.

etmmahi:
@ashubhatt  I checked the two patches against MQTT Version 3.1.1 OASIS Standard
both are fine. Now get them merged in develop. So we can get the MQTT lib fully functional with LWT

etmmahi:
@hreintke Can we get this one merged?  

@ashubhatt  can you create pull request to fix the order of the username and message in the connect()

To finalize there is also a PR/pull request needed for adding the correct flags in the connect message (see comment @gschmottlach)

Time to finish up the LWT issues

etmmahi:
Closed due to merge of #565

PR #514, opened 2015-12-30
[Websock Server] Add user data pointer per socket. Add disconnect method

ADiea:
Sometimes you need to store additional information on every websock connection, this can be done by storing a pointer to a user structure. 
Also, the server could decide to close some unused websocks to free up heap, I've added a disconnect method.

There are some debugf traces I will remove them

### DIFF
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -225,6 +225,7 @@ void TcpConnection::close()
 	tcp_poll(tcp, staticOnPoll, 1);
 	tcp_arg(tcp, NULL); // reset pointer to close connection on next callback
 	tcp = NULL;
+	checkSelfFree();
 }
 
 void TcpConnection::initialize(tcp_pcb* pcb)
--- a/Sming/SmingCore/Network/WebSocket.cpp
+++ b/Sming/SmingCore/Network/WebSocket.cpp
@@ -50,7 +50,7 @@ void WebSocket::send(const char* message, int length, wsFrameType type)
 	size_t headSize = sizeof(frameHeader);
 	wsMakeFrame(nullptr, length, frameHeader, &headSize, type);
 	connection->write((char*)frameHeader, headSize, TCP_WRITE_FLAG_COPY | TCP_WRITE_FLAG_MORE);
-	connection->writeString(message, TCP_WRITE_FLAG_COPY);
+	connection->write(message, length, TCP_WRITE_FLAG_COPY);
 	connection->flush();
 }
 
@@ -70,4 +70,9 @@ void WebSocket::enableCommand()
 	{
 		commandExecutor = new CommandExecutor(this);
 	}
+}
+
+void WebSocket::close()
+{
+	connection->close();
 }
--- a/Sming/SmingCore/Network/WebSocket.h
+++ b/Sming/SmingCore/Network/WebSocket.h
@@ -29,6 +29,7 @@ public:
 	void sendString(const String& message);
 	void sendBinary(const uint8_t* data, int size);
 	void enableCommand();
+	void close();
 
 protected:
 	bool initialize(HttpRequest &request, HttpResponse &response);


ADiea:
Example usage (where WebSockUserData  is a user structure)

```
        void wsPruneConnections()
	{
		WebSockUserData *pData;
		WebSocketsList &clients = gHttpServer.getActiveWebSockets();
		for (int i = 0; i < clients.count(); i++)
		{
			pData = (WebSockUserData*) clients[i].getUserData();
			if(pData && !pData->isAlive())
			{
				clients[i].disconnect();
				pData->isInvalid = true;
			}
		}
	}
```

hreintke:
@ADiea : 
I think you can get the same application functionality by using a `HashMap<WebSocket,WebSockUserData*> webSocketHashMap;`

For closing Websockets your are right. 
I would prefer to name the function close() to keep it consistent with other connection types.

ADiea:
@hreintke done, removed the data pointer + associated methods

ADiea:
@hreintke @raburton hi what else needed for merge?

raburton:
Looks alright to me, but any chance you can squash to a single commit? With things being added and removed we've ended up with 4 commits for 8 lines of code. If you can do that for me I'll merge it now.

ADiea:
@raburton done, thanks!

piperpilot:
I think there is a problem with this code..something new is showing up, every time an HTTP request is made, you get:

3fff4cb8 already freed

On the serial line.  It seems to be related to this change:

	tcp_arg(tcp, NULL); // reset pointer to close connection on next callback
	tcp = NULL;
	checkSelfFree();
}

Line 228 in TcpConnection.cpp

I'm not sure exactly what is going on here, but this is called elsewhere in the file with:

con->checkSelfFree();

Perhaps its not actually referencing the connection right?  Anyways, I think this needs to be fixed...not sure the best way to reflect this, in the PR or in a new issue?

ADiea:
@piperpilot sometimes seems to call destructor twice. will look into it

hreintke:
@ADiea : 
First : sorry for misusing this issue for finding help on another.

Then, as you probably know I am porting Sming to the Espressif RTOS SDK.
In there I have problems getting the m_printf to full use. A short description is in here : https://github.com/SmingHub/Sming_RTOS_POC/issues/26.
Can you take a look there and we can discuss reasons/solutions ?

piperpilot:
@ADiea Great...thanks...the "already freed" message is junk that screws with some serial stuff I am doing.  It doesn't look like there is any way to turn off that message either...its part of the core SDK as far as I can tell.  It probably isn't an issue that we are calling the destructor twice...but the error message is an issue.

ADiea:
@piperpilot on a first search i could not find the error message in spource code it is probably like you said in the sdk. This does not happen for every closed connection i have to figure out in what case

hreintke:
@ADiea : 
I see that besides adding the websocket close you also added the checkselfFree into the TcpConnection.
This checkselfFree is also called on the next Poll received from the LWIP.
This "waiting for next poll" is meant to be sure that the other endpoint of the tcpconnection can finalize sending data.

What was the reason for adding that call.

I am not sure but vaguely remember I had similar issues when I was researching why tcp_closure took more time than expected. At that time I limited the time for next poll by the `tcp_poll(tcp, staticOnPoll, 1);` just above your change.

ADiea:
Without that line there are times the tcpConnection object is not destroyed causing mem leaks... will look more into it to see how to avoid this

ADiea:
@piperpilot investigated this a little and found the rootcause
TCPConnection::close() can be called from the outside in which case checkSelfFree needs to be called to actually remove the TCPConnection from heap
But the destructor also calls close()

solution is to modify destructor to be
'''
TcpConnection::~TcpConnection()
{
	autoSelfDestruct = false;
	close();
	debugf("~TCP connection");
}
'''

piperpilot:
Can we fix this?  The "already freed" message is still showing up in the latest develop.  I will open a bug on this and reference this PR.

hreintke:
@piperpilot 
Could not find this reply, although I knew it was there. even created a new issue for it (on which @ADiea  replied). I will PR and merge this.

Issue #495, opened 2015-12-21
RTOS SDK based Sming


hreintke:
@All : 
I have been working on a Proof of Concept of the RTOS SDK based Sming.
Reason is the move from Espressif towards RTOS SDK, also for the upcoming ESP32.
Result is in my https://github.com/hreintke/Sming_RTOS_POC.git repository.

- Based on Sming Develop of 26/11/2015 
- Most of the Sming components included 
- Only limited testing done
- No Framework/Application structure, just application
- I am running on Windows
More details in the "README" 

Have a look, test, feedback and start helping to get to an Alpha version.

Disclaimer : 
- I am NOT planning to create a cloned Sming project -> this should become part of SmingHub repository. 
- I have tried to get into contact with @anakod since december 8th without any success -> I have no access rights to add this to the SmingHub organzation.
- This should be resolved soon, in order not to get a "dead end". Any help appreciated.

alonewolfx2:
@hreintke i will look it. where is task implementation

hreintke:
Task implementation is "just plain" rtos.
In the POC I just start a task without any functionality.

As it is now Sming is running in main task and doesn't use any other.
To change that (if necessary) does also need more/further adaption to Sming to create thread/task safety.

Please focus first on porting current sming functionality before introducing new.

alonewolfx2:
@hreintke second question is can we use app/framework things like main sming with nonos sdk? i think its important for adding new examples

hreintke:
@alonewolfx2 : 
Yes sure we can.
Just realize that this is a Proof of Concept done by me only.
Now opened up, the community should also take further action to test and extend.
So if app/framework is you prime concern, implement and PR.

raburton:
>No Framework/Application structure, just application

Maybe this is stupid question, but what do you mean by this?

hreintke:
In sming you build framework -> libsming.a -> building application using lib -> flash
In here build sming/application as one -> flash

raburton:
On right, that doesn't sound like a big problem. I'm away at the moment, then will be working all over Christmas :-( but I'll have a look when I get chance. Sounds interesting!

hreintke:
No, not a big problem but just one of the things (makefiles etc) where I do not have much experience.
And.. not necessary for my initial working on this

kireevco:
@hreintke why not just create another RTOS branch and then implement selector which sdk to use?

hreintke:
@kireevco : I have been thinking of that but 
[...]

PR #515, opened 2015-12-31
Lwip from sources for SMING

ADiea:
@all succeeded in putting it all together and we can now compile lwip directly from sources. 

Benefits:
-> modify system/include/lwipopts.h to disable enable modules and debug traces. This can help in reducing RAM/firmware size, catching misterious network bugs, keeping up to date with lwip etc
-> somehow decreased firmware size, can be decreased further by stubbing espconn functions (not tested yet)

Big thanks goes to @pvvx all based on (MinEspSDKLib)[https://github.com/pvvx/MinEspSDKLib] 

Also noticed this other (interesting implementation)[https://github.com/kadamski/esp-lwip] -> 
@kadamski could you take a look when you have some time to see if there are fixes/improvements from your implementation we could use here? I will also try to :)

Testing: Tested on a ESP12 / 512KB Flash . When enabling debugging like it is now you must use a bigger starting address for the 0x9000.bin image, I used 0xF000 this must be modified in Makefile-user.mk and compiler/ld/eagle.app.v6.cpp.ld
```
  iram1_0_seg :                         org = 0x40100000, len = 0xF000
  irom0_0_seg :                         org = 0x4020F000, len = 0x45000
```


ADiea:
just noticed this PR also containes #514 :|

alonewolfx2:
i will try that. i am waiting for this :) thank you. are there any issue?

hreintke:
To get all communication in the right place : 
------------------------------------------------------
From @hreintke


@ADiea :
We are moving Swing from NOSOS SDK to RTOS SDK.
In my opinion we should not put effort in the LWIP of Sming NONOS.

Within the conversion process I had some issues with LWIP and tried to use the NONOS LWIP from Espressif in RTOS SDK but failed. Also noticed then that NONOS and RTOS LWIP implementations differ. There is no (yet?) LWIP for RTOS from espressif.
There is a LWIP in the esp-open-rtos project. I was not able to use that either.

I stopped the LWIP compile effort when I solved the Sming RTOS LWIP issue.
That was related to the LWIP IPv6 inclusion of RTOS LWIP

---------------------------------------------------------------------------------

From @ADiea : 


@hreintke
Why not have 2 versions of SMING one NONRTOS (mature let's say) and the new RTOS version

I have succeded in compiling everything up for the NONOS version os sming, will submit a PR
Also my tests show somehow a more stable implementation than just linking with liblwip.a as it's currently done in sming NONOS

Also, how is the new RTOS? Did you do any tests yet?
1. how much heap you are left with,
2. how many tasks/processes you can have
3. lag compared to nonrtos
4. is it stable? based on freertos?
etc

Cheers

----------------------------------------------------------------------

From @hreintke


@ADiea :
Supporting two versions of Sming is not useful, it takes additional effort of scarce resources

For moving to RTOS SDK :

- Espressif is moving support from NOSOS to RTOS -> will keep Sming future proof
- Opens the Sming Support for ESP32
- Based on FreeRTOS which will provide tasking/threads -> Standards based
- Opens better Sming support for "Arduino like" programming paradigms (without getting WDT resets).

And yes/no about testing. See #495 For information.
There is a "Working Proof of Concept" version which include most of the current functionality.
With this version the question you state will be answered.

If you have time, would appreciate your help in testing and maturing the RTOS version.

hreintke:
@ADiea : 
There is also a Espressif supplied LWIP source.
From the text above I understand that your work is not based on that. 

Is that correct ? 
And if so, what is the benefit/of not taking that version ?

ADiea:
@alonewolfx2 no issues until now still testing

@hreintke This is based on Espressif supplied LWIP source. This was a 2 step process first I forked (MinEspSDKLib)[https://github.com/pvvx/MinEspSDKLib] => this project can generate liblwip.a among other binaries. Then I integrated the lwip implementation directly into Sming. MinEspSDKLib is based on (Open source LWIP for ESP_IOT_SDK_V1.4.0.)[http://bbs.espressif.com/viewtopic.php?f=46&t=1221] 

Until RTOS version is stable I thibk we should still support Sming NOOS because probably some people, me included have projects based on it...

alonewolfx2:
@ADiea can you suggest not important modules? i need to reduce firmware size.

ADiea:
This ofcourse depends on your custom app For sensor nodes i would use udp + dhcp
For gateway node a bigger flash esp 1MB tcp udp dhcp etc for connecting to outside world

There is no simple way to shrink the size it is a recursive process
1. Disable the module and see who complains when compiling and linking.  
2. If you can, remove the callers. 
The map file helps you alot in this process.  Good luck

ADiea:
@all anyone had time to also test this? Ty

hreintke:
@ADiea :  I did not test this update and I will stay with my earlier opinion 
>We are moving Swing from NOSOS SDK to RTOS SDK.
In my opinion we should not put effort in the LWIP of Sming NONOS.

The current Sming_RTOS has already most of the functionality, testing is progressing and with good results. I expect it will not be to long when we will slowly move from NONOS to RTOS support.

ADiea:
ok got it

slaff:
@ADiea How about re-submitting this PR as a third-party module? The open LWIP should be optional.

ADiea:
This is the major diff between my "work" branch (https://github.com/adiea/sming/tree/work) and the current sming dev :( My idea is also to make it external module to track the lwip. Because sming nonos is finally alive again i will start converging but it will be a big effort for me to resubmit this - but will do it because it gives me more control to disable some features and make it less vulnerable to attacks

Better to keep this closed because it is no longer relevant and I will open a new PR sometime in the future based on new sming structure(third party module)

PR #518, opened 2016-01-01
MQTT Disconnect Callback and LWT Fix

ashubhatt:
@all Fixed the LWT of MQTT with username and password.
Added a new function setDisconnectCb() which sets the callback function upon MQTT disconnection.

Tested and verified these fixes.

### DIFF
--- a/Sming/Services/libemqtt/libemqtt.c
+++ b/Sming/Services/libemqtt/libemqtt.c
@@ -44,9 +44,9 @@
 
 uint8_t mqtt_num_rem_len_bytes(const uint8_t* buf) {
 	uint8_t num_bytes = 1;
-	
+
 	//printf("mqtt_num_rem_len_bytes\n");
-	
+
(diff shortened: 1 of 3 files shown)


hreintke:
@ashubhatt : 
Fixed LWT : 
There is already a PR for fixing LWT which (as the comments say) need an update for username password. Can you add your fix in that PR so there is only one fix ?
And in there, do not reformat the code. With all the updates in the PR it is very difficult to identify what is really changed to fix the issue.

For MQTT callback.
There is a callback implemented  in TcpClient. MQTT is derived from TcpClient.
I think (needs verification) that if we just add a class function setTcpClientCompleteDelegate (or setCompleteDelegate) to the TcpClient class it will work for both.

Generic : 
Please do not combine PR's for multiple fixes/functionalities. That 1/ eases discussion and 2/ shows the changes separate in the merge history.

PR #522, opened 2016-01-01
Problem with TemplateFileStream

dseliskar:
1)
In case variable is included in template in such a place that TemplateFileStream::readMemoryBlock would start with it, then TemplateFileStream::readMemoryBlock will change state to eTES_Found and return 0. It relies on ::seek to change state to eTES_StartVar, but since it returns 0 the ::seek never gets called so it is stuck in a eTES_Found state returning 0 over and over again. This occurs when variable is at beginning of file or is "split" by FileStream::readMemoryBlock in which case  TemplateFileStream::readMemoryBlock starts with variable at the beginning and same situation appear. The proposed fix changes state to eTES_StartVar immediatelly.

2)
Small typo fix.

### DIFF
--- a/Sming/SmingCore/DataSourceStream.cpp
+++ b/Sming/SmingCore/DataSourceStream.cpp
@@ -223,10 +223,12 @@ uint16_t TemplateFileStream::readMemoryBlock(char* data, int bufSize)
 					memcpy(varname, cur + 1, p - cur - 1); // name without { and }
 					varName = varname;
 					state = eTES_Found;
-					varWaitSize = cur - tpl;
-					debugf("found var: %s, at %d (%d) - %d, send size %d", varName.c_str(), cur - tpl + 1, cur - tpl + getPos(), p - tpl, cur - tpl);
+					int sz = cur - tpl;
+					varWaitSize = sz;
+					debugf("found var: %s, at %d (%d) - %d, send size %d", varName.c_str(), sz + 1, sz + getPos(), p - tpl, sz);
 					skipBlockSize = block;
-					return cur - tpl; // return only plain text from template without our variable
+					if (sz == 0) state = eTES_StartVar;
+					return sz; // return only plain text from template without our variable
 				}
 			}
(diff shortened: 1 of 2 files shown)


alonewolfx2:
tested. working fine. thanks. @dseliskar .
@hreintke it can merge. any thought?

hreintke:
@dseliskar @alonewolfx2 
Thanks for the PR. 
One remark which might require an update to the PR.
You create another variable sz : 
```
int sz = cur - tpl;
varWaitSize = sz;
```
Any specific reason for that ? 
If not, I prefer to keep the code clean and revert to only varWaitSize usage.

dseliskar:
Well, original fix was oneliner and yes, creating a new variable was my concern, but I went 
for it for two reasons, which of course are debatable.  Expression (cur - tpl) is appearing six times in four lines of code:
```C++
          varWaitSize = cur - tpl;
          debugf("found var: %s, at %d (%d) - %d, send size %d", varName.c_str(), cur - tpl + 1, cur - tpl + getPos(), p - tpl, cur - tpl);
          skipBlockSize = block;
          if ((cur - tpl) == 0) state = eTES_StartVar;
          return cur - tpl; // return only plain text from template without our variable
```
Efficiency - I'm not sure if compiler handles this, if not this is very inefficient. If it does - then one reason debunked. Other is readability - with employing another variable the code is much cleaner.
In sz defense - it is ephemeral.  But I agree with your concern - any variable on such device must be
well thought of. So we go back to oneliner ?

dseliskar:
@alonewolfx2, @hreintke 
This was bugging me so I checked (I should done it before - mea culpa!). Sming is compiled with -Os
so -fgcse is included. First reason debunked!. Second is not important enough to add another variable, so shell I make another PR to reverse this? Sorry for wasting your time.

hreintke:
@dseliskar : No problem. Good that you rechecked.
And, I did miss all the cur-tpl occurrences in the original code.

Should we combine the options ? Just use varWaitSize and replace all the cur-tpl ?

dseliskar:
@hreintke 
Sure, even better.

PR #523, opened 2016-01-02
move samples to subdir closes #69

### LINKED ISSUE #69: Sming projects/examples should be separate from libsming
This will be beneficial for referencing other projects from other repositories, as well as will help making every example an actual project.

As a mere of transition to it we can use submodules that will point to github/anakod/Basic_Blink (or better - github.com/sming/Basic_Blink)

raburton:


raburton:
comments please
if no adverse problems i'll merge tomorrow

kireevco:
Why `samples` and not `Samples` or why `Sming` and not `sming`? I prefer consistency. 

Thoughts?

raburton:
Lower case is preferred, much more friendly for the command line. Sming hasn't been changed to keep the change minimal, this is just a change to the samples. Also, changing Sming would potentially inconvenience users who might need to change their scripts or env vars. I'd be happy to do a larger restructure, but that's been discussed previously and never got anywhere, I decided to finally go ahead with this small change to actually make some progress on the most important aspect (the mess of samples in the root dir).

hreintke:
@raburton : Agree. 
Let's merge now, so it gets included in the V2.1 and release.

Issue #501, opened 2015-12-22
I2C support for OLED1306 display without reset pin


harry-boe:
I played wit the new timer based version of the OLED 1306 example project.
After heaving many issues with the older delay Version I cn confirm that it's now working fin for SPI and I2C displays.

There is just one add on. Most of the I2C OLED display do not have a RESET pin.
Proper initialisation of those devices require another constructor:

-> current example:  

    // by default, we'll generate the high voltage from the 3.3v line internally! (neat!)
    display.begin(SSD1306_SWITCHCAPVCC);`

This works for I2C with RESET and SPI 

-> I2C without reset needs to be initialised like that:

    // by default, we'll generate the high voltage from the 3.3v line internally! (neat!)`
    // initialize with the I2C addr 0x3D (for the 128x64)
    // bool:reset set to TRUE or FALSE depending on you display
    display.begin(SSD1306_SWITCHCAPVCC, SSD1306_I2C_ADDRESS, FALSE);`

Maybe we add some comments in the code to explain i2c support

alonewolfx2:
Good catch. Can you make pr ?

sle118:
Wow! This will save me some time! I actually captured the trace of SPI communication from the initialization to understand why the display wasn't firing up, and I was about to start analyzing. so thank you!

harry-boe:
as requested pr made: now set to I2C as default but SPI still there in comments.

hreintke:
@harry-boe : 
I did not read all the details but you mention : "now set to I2C as default"
Suppose that will break code which depends on the old default,

Is so, you should revert to the current default.

harry-boe:
Should not break anything. There are no changes in the libs. 
The change is only in the sample application.cpp

maybe default was the wrong wording here ..

harry-boe:
THX for merging

fvalenta:
Works through I2C without reset. The undocumented 0xe4 command has to be sent.

destroyedlolo:
I'm working on the port of this library to unix SBC (like RaspberryPI, BananaPI, and all other *PIs) : see https://github.com/destroyedlolo/ArduiPi_OLED
I'm experiencing some issues with my Display turn off with 0xAE command but never tune on back with 0xAF one.
Do you think it can be solved with this 0xE4 command ? Where have I to add it ?
Thanks

feer9:
> Works through I2C without reset. The undocumented 0xe4 command has to be sent.

[...]

PR #534, opened 2016-01-05
Issue #501 - I2C support for OLED1306 display without reset pin

harry-boe:
Pull Request for Issue #501 

Source is now set for I2C bus without reset pin. See source comments for SPI and I2C with Reset


hreintke:
@harry-boe : @alonewolfx2 : 
Can you PR this for Sming_RTOS too ?

Issue #538, opened 2016-01-07
FTP server command LIST is not able to show more like 50 files (depends on filename length)


tprochazka:
Steps to reproduce:

Use FtpServer_Files and ad this code to generate more files on FS

```
	for (int i = 0; i<100; i++) {
		WDT.alive();
		fileSetContent("example" + String(i) + ".txt", "hello world!");
	}
```

Then connect to the FTP server and you will see just something like 50 files.

It is because ```class FTPDataFileList``` defined in ```FTPServerConnection``` trying to sent complete list in one iteration, but output buffer for this is limited.

I'm working on fix.

hreintke:
Needs to be documented as a known limitation.

PR #539, opened 2016-01-07
fix FTP LIST command didn't sent complete list if it is bigger than bufer (#538)

tprochazka:
```class FTPDataFileList``` didn't support multiple iteration if list was bigger than send buffer.
I removed also ```fileList()``` because it created vector of Strings which can consume also a lot of memory. I think that code like this should not be used on low memory device.

### DIFF
--- a/Sming/SmingCore/FileSystem.cpp
+++ b/Sming/SmingCore/FileSystem.cpp
@@ -126,7 +126,12 @@ void fileSetContent(const String fileName, const char *content)
 
 uint32_t fileGetSize(const String fileName)
 {
-	file_t file = fileOpen(fileName.c_str(), eFO_ReadOnly);
+	return fileGetSize(fileName.c_str());
+}
+
+uint32_t fileGetSize(const char* fileName)
+{
+	file_t file = fileOpen(fileName, eFO_ReadOnly);
 	// Get size
 	fileSeek(file, 0, eSO_FileEnd);
 	int size = fileTell(file);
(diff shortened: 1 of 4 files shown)


tprochazka:
Just note.
 I don't like that ```writeString()``` and ```write()``` in ```TCPConnection``` return always -1 for all types of errors (defined by ```tcp_write()```). I think that it is not possible to simply repeat the action for every error type like in buffer full situation.

hreintke:
@tprochazka : 
We need to get deeper into this, for now I think you are right that it doen't work  correctly.
On your solution ;  there is a tcpclient class, that takes care of the buffering. 
We should investigate whether the use of tcpclient instead on tcpconnection would solve the issue with available means. 

I see you use direct spiffs functionality to retrieve the directorylist.
If the issue point is that filelist() is using to much memory, I would prefer to add code to filesystem class to improve on retrieving directories. Then it will be available for application too.

tprochazka:
filelist() in the filesystem class class is good for small amount of files. It loads all files names to the memory. It is not good way if you know nothing about files count.

I think that internal framework classes should be as memory save as possible, but filelist() is more easy and clean way how to do it, especially with beginners. Maybe filesystem class can provide better memory save way, maybe is better keep it small and easy and use more low level classes if you want to save memory in application code.

tprochazka:
So maybe this should be merged as quick fix, because I don't think that it is time now for complete rewrite whole FTP support.

tprochazka:
Is here something what I can do for merge this really important fix?

hreintke:
@tprochazka : 
As I said above, I prefer not to use the spiff_.. outside the filesystem class.

I am busy now with new functionality from Spiffs and can handle such an update then on one location.

anakod:
@tprochazka I think you idea can be implemented as new methods files enumeration. In that case we will cover internall spiffs implementation from user and will have working solution.

hreintke:
Outdated, will be documented as a known issue.

PR #541, opened 2016-01-09
Adafruit  ST7735 TFT Support 

harry-boe:
Support for the Adafruit ST7735 based 1.4 and 1.8 inch displays

Port of the Adafruit libraries
Sample app that goes through all Adafruit GFX lib demos (using timers instead of delays)
Additional demo to read a bitmap and display a bitmap from the spins filesystem

All tested in the 1.4" display. But i'm very convinced that it will work on 1.8" displays as well as there is only on line of code changed in the Adafruit_ST7735.h original file


raburton:
Are these tested or just ported? I can't help but notice there are 18 commits! Can you squash these down to a single commit.

BTW - I notice a lot of merge commits from upstream back to your local repo. To avoid these instead of doing `git pull upstream develop` into your branch then having to create the merge commits, just do `git pull --rebase upstream develop`.

PR #542, opened 2016-01-09
Adds (partial) API documentation.

riban-bw:
Also adds scripts for automated documentation build and publishing via Travis-CI. This is not yet enabled.


riban-bw:
This PR adds API documentation comments to a number of files (not all of them). It also provides a Doxygen configuration file to allow building of html API documentation and a script for automated publishing of API documentation to github pages. I have not included the updates to .travis.yml that will enable the automation because this will require a one-off effort from the github repository owner / manager to create encrypted ssh keys. This will be the subject of a separate PR.

There are many more files to document. This is a start.

anakod:
@riban-bw Great work, thank you!

riban-bw:
I used a template header comment which I added to most files which was derived from the comments in one of the earlier files I edited which had a comment header describing the author, copyright, licence, etc. I need some guidance from the community as to whether we should replace all the header comments with a consistent header comment. Should / can we remove existing header (info)? I can take out the javadoc comments about author, copyright and licence or can replace them with a consistent comment that we agree on. Maybe leave the original header comments (though it looks rather untidy in the header file).

raburton:
Looking good. Are we needing to commit or do we need to do anything more on the publish side first?

riban-bw:
I think we should commit this so that we have the API docs started. I can follow up with a separate PR for the automation. (Things are in place but it is neater to do it a step at a time.)

raburton:
looks like it isn't going to merge cleanly, are you able to rebase/merge to current develop or do you need a hand with that? i'm off out now so it'll be tomorrow if you need me

riban-bw:
I merged rather than rebased. I suggest squashing to a single commit before merging back to develop. I can't do that from here. May have a chance tonight but if anyone else can do the squash and merge, please do so.

raburton:
No one else can on your branch, so easier for you to do it later...

PR #548, opened 2016-01-13
Adds automated publishing of API documentation

riban-bw:
Creating pull request to allow testing of Travis-Ci process. Will need a little iteration before it is fully working and can be merged upstream. (Don't merge yet!)


riban-bw:
Can  not test publishing in PR because Travis-CI does not expose encrypted environmental variables to pull requests. How do we progress this? Do we merge it (after due diligence) and check results? Tasks that need to be performed to make this work are:

- Add encrypted env vars to Travis-CI SmingHub/Sming (kireevco says this is done)
- Enable gh-pages on Sminghub/Sming
- Change commit user and email in publish_api.sh (someone tell me what to put in there - do we have a generic user?)
- Merge this PR

raburton:
Do we want to commit the docs to repo or would it be better to just publish them somewhere? It means an extra commit to the repo everytime anything else is committed, often without the docs needing to be updated. And presumably that doc commit will trigger another travis build, causing another commit?

riban-bw:
The commit back to docs only occurs if there are any changes to docs (git handles checking for modifications). In my tests, the commit back to docs does not trigger another Travis build. (Is travis linked to particular branches?) It may be appropriate to publish docs elsewhere (I considered this earlier when looking at the security aspects of this) - any suggestions.

hreintke:
@riban-bw : @raburton :
I leave the talk about documentation to you but when I looked into travis for another reason I found these options for branches. By default every branch which has the travis.yml file in it will trigger run.
```
# blacklist
branches:
  except:
    - legacy
    - experimental

# whitelist
branches:
  only:
    - master
    - stable
``

raburton:
So we could have a branch that is docs only, but that seems a bit odd doesn't it? It's not really a branch if you delete all the normal content and put something else in instead. Maybe just putting the docs in a separate repo would make more sense. That repo could then have master and devlop branches with their own versions of the doc (based on the code branch that generated them), and it doesn't mix source and output in a single repo.

raburton:
It's a pity that readthedocs can't do doxygen properly, then it could simply auto do this and publish on their site without any effort or thought at our end.

riban-bw:
When you enable gh-pages an extra branch is added to the repo called gh-pages. It is the gh-pages branch that is updated by my API docs script and only if the docs have changed. So normal code changes will not cause any extra commits. Only changes to the documentation will cause a commit to the gh-pages branch.

I will have another look at readthedocs. I only gave it a cursory glance before. I used doxygen because it just works. There may be other methods of creating the docs. (I did write a javadoc2markdown application for instance but that is incomplete.)

riban-bw:
Okay, I think this process is inappropriate. I was concerned about the security aspect and there are a lot of hoops to jump through to make it work. Also, it is difficult to test. I will look at Read the Docs and test a process to get the docs in to there. (There is a process but it will take me a few days to test it.) I will close this PR. Will someone (@kireevco ?) remove the encrypted env vars from Travis as we won't need these any more.
Cheers!

raburton:
Hmm, so if it's on its own branch we can only have one version of the docs at a time? Ideally we need the two sets of docs, one tracking the master branch and one the develop branch. This is so much more complicated than it should be.

I fully agree that doxygen is the way to go, which is why I was surprised that readthedocs doesn't properly support it, it's a much better system than just writing your own documentation in markup pages from scratch as they seem to advise (What is the point in their service? You might as well just write up the documentation on the github wiki or github pages directly.) I think there is some king of bridge, but again it's jumping through a few more hoops, but it might still be the better option. Their github hooks, auto publishing and hosting is nice, if we can get it working with doxygen I think we'll have a good solution, I just wish they made it easier.

riban-bw:
I agree we need separate docs for stable and develop. We may also need a third set of docs for PR to allow checking of documents created during PRs. I am working through the process to use doxgen XML output + breath + sphinx + commit hooks. I'll create a PR when I have something working.

riban-bw:
Reopening PR#548. Having looked at readthedocs I see its core functionality is based on documenting python. There are ways to convert doxygen XML output, in to Sphinx compatible syntax but this will still require a lot of manual creation of documentation in Sphinx syntax which is what we are trying to avoid. I have not seen any API documentation on readthedocs that looks good. Very small projects are documented on one page but a framework the size of Sming requires a hyperlink type API doc structure. Doxygen does a good job of creating the API docs (and if it doesn't, we could use something else). The issues I see are:

- Where do we host the API documentation
- How we build the documentation
- How we publish it

Readthedocs has the benefit of being able to build documentation and pull it from the source via commit hooks. By pulling and building we have fewer security issues to deal with. The doc builder host is a client that has read-only permission on GitHub Sming source repository which everyone does.

What I was trying to do with this PR was build and push. Travis would build the documentation in response to a successful build then push the documentation to where it gets published. A push process requires write permission on the site that publishes - hence a security issue.

If we had a site that could pull the source (git client ideally triggered by commit hook), build the documentation (doxygen) and  publish html (a web server) then we have a simple solution. I haven't found such a solution that does not cost.

Thoughts?

hreintke:
My opinion, shoot if you want :smile: 

- We have a sming release every 2 months at the most.
- We have a limited number of PR's which has large effect on develop branch documentation

It is better to have an easy, partly manual process which will do when one of the two above situations occur that have a complicated  automated process that we have to maintain.

riban-bw:
Yes I agree that a manual process may not prove too arduous - I didn't want to add to the burden of producing code and docs, hence the effort to automate. If the project maintainers are happy to manually build docs and commit then fine. I could adjust the script to indicate when a docs update is required but that might not work well if we build docs with a newer version of doxygen than is available to Travis.

Let's start with manual build. Should we create a new repository for docs or use the gh-pages branch?

riban-bw:
SmingHub is a github "Organization" with several repositories. I suggest adding a new repository called SmingHub.github.io. This repository will hold documentation for SmingHub which will be automatically published to http://SmingHub.github.io.

I don't know what the granularity of permissions is for a github organization but I would hope that you can grant write permission to a user to individual repositories, in which case you could add a documentation maintainer to that repository without them having elevated rights on other repositories, facilitating a workflow where a user / group of users maintain the documentation. You can add and remove user permission as appropriate over the life of the project.

By using a separate repository, the source repros are kept clean and we have a single repro for all docs, hopefully reducing the need to work across multiple repros to maintain docs for all of them.

It _might_ be possible to add automation in the future. The security _may_ be more acceptable if we can constrain access to the documentation but I don't think you can currently constrain write access via ssh key to a single repro - but I'm not sure on that. (I haven't worked in a github organization before.)

raburton:
I think we can probably do the automation quite easily if we have a build box for it. @kireevco has a box that builds the SDKs. If he's able to add a script that will pull the latest git source, run doxygen against it and upload the result (either to a normal webserver, or git to a doc repo) in it can be triggered with a github [webhook](https://developer.github.com/webhooks/).

riban-bw:
Okay but let's get the API documentation manually published first. Then we can work on automating it.

Will someone add a repository to SmingHub called, "SmingHub.github.io"?
It would be good if I could be added as an [outside collaborator](https://help.github.com/articles/adding-outside-collaborators-to-repositories-in-your-organization/) on that repository so that I can set up the docs.

I will continue to look at automation once we have the docs published. I have been talking directly with @kireevco about this and will continue that.

raburton:
Done.

kireevco:
@riban-bw  can you look into https://readthedocs.org/ - I think they support accepting pushes via git & markdown, and it is a very nice platform.

raburton:
@kireevco he has (it's discussed in this issue), they don't support doxygen. In fact they really don't do anything very clever at all, just checkout your hand written markdown documentation and publish it. And, as @riban-bw said, the results of that aren't even very good.

kireevco:
@riban-bw wha'ts going on with this PR?
Also, I have a question, are those commits yours or you back-pulled older commits?

raburton:
I think this was originally just a test pr to get travis going, wasn't really an attempt to merge all those random commits. Probably would be best to start a new issue for discussion though, rather than here to avoid confusion.

riban-bw:
Yes, this PR was initially created to add automated build of API docs using Travis-CI. I closed it when it was suggested that we use ReadTheDocs. After investigating that and deciding it looked inappropriate, I reopened this PR as a discussion on how best to handle publishing of API docs. For now I will manually update the docs and will work on an automated publishing solution that _may_ involve Travis-CI or @kireevco hosts or something else. For now I shall close this PR. (I will be submitting a different PR for some more API documentation I have added.)

I am also considering how to track effort for adding API docs. There are may elements yet to document and, although I am plodding through them, it may be advantageous for others to help but I want to avoid duplication. I have a publishing script which I shall use which creates a report, detailing the documentation issues. I might add these as issues to the docs repository to allow individuals to pick ones they want to work on - I'll let you know in gitter.

kireevco:
@riban-bw where is the documentation live right now? Can you update README.md with a link?

Issue #549, opened 2016-01-14
MQTT client does not send PING when receiving messages


Informatic:
MQTT PINGREQ when there is no outgoing activity are handled in `MqttClient::onReadyToSendData`. However, `sleep` variable checked there is reset to 0 in `TcpConnection::staticOnReceive` (as well as `onSent`). In such case `mqtt_ping` is never called when client receives multiple messages with QoS 0 with intervals lower than 10 seconds.

MQTT client should store its own "sleep-like" counter variable, reset only in `onSent` event (or just use millis()?)

fcgdam:
Yes, I've also hit this issue when only receiving data, after 10s the connection is dropped.
My solution was to inform the broker by setting  keepAlive() to a large value.

hreintke:
Solve by #550

PR #550, opened 2016-01-14
Fix MQTTClient not sending PINGREQ when receiving messages

Informatic:
Suggested fix for #549. Tested, seems to work fine.

### DIFF
--- a/Sming/SmingCore/Network/MqttClient.cpp
+++ b/Sming/SmingCore/Network/MqttClient.cpp
@@ -267,7 +267,8 @@ err_t MqttClient::onReceive(pbuf *buf)
 
 void MqttClient::onReadyToSendData(TcpConnectionEvent sourceEvent)
 {
-	if (lastMessage and millis() - lastMessage >= 10000)
+	// Send PINGREQ every keepAlive time, if there is no outgoing traffic
+	if (lastMessage && (millis() - lastMessage >= keepAlive*1000))
 	{
 		mqtt_ping(&broker);
 	}
--- a/Sming/SmingCore/Network/MqttClient.cpp
+++ b/Sming/SmingCore/Network/MqttClient.cpp
@@ -96,6 +96,7 @@ int MqttClient::staticSendPacket(void* userInfo, const void* buf, unsigned int c
 {
 	MqttClient* client = (MqttClient*)userInfo;
 	bool sent = client->send((const char*)buf, count);
+	client->lastMessage = millis();
 	return sent ? count : 0;
 }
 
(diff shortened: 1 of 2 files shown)


hreintke:
@Informatic : 
Is the PING_REQ used for the keepAlive messages ? 
If so, there is a keepAlive variable in the MqqtClient, can we use that value ?
if not, There is a harcoded 10 seconds in the patch. Can you explain the value ?

hreintke:
Mqtt specs : 
>The MQTT specification says the following:

>It is the responsibility of the Client to ensure that the interval between Control Packets being sent does not exceed the Keep Alive value. In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet.

>That means as long as messages are exchanged frequently and the keep alive interval is not exceeded, there is no need to send an extra message to ensure that the connection is still open.

>But if the client doesn’t send any messages during the period of the keep alive it must send a PINGREQ packet to the broker to confirm its availability and also make sure the broker is still available.

>The broker must disconnect a client, which doesn’t send PINGREQ or any other message in one and a half times of the keep alive interval. Likewise should the client close the connection if the response from the broker isn’t received in a reasonable amount of time.

I read this as : If there is a continuous flow of message there is no need to send PING_REC 

There might be a flaw in the handling of keepalive as there is also a harcoded 10 in the current code.
Can you check that too ?

Informatic:
About hardcoded 10 seconds - yes, I think it should be based on keepAlive value, for instance keepAlive/2 (like it was with default 20 second keepAlive value in current `develop` branch) however I wanted to make as little changes as possible to fix that specific issue.

> I read this as : If there is a continuous flow of message there is no need to send PING_REC
Yes, you are right, however the specification talks about packets sent from client to server. When server sends QoS 0 messages client does not send any response for that. (and thus server does not know if client is alive or not and PINGREQ is needed)

hreintke:
@Informatic : 
OK, understand the need for PING.
Can you  extended the patch to handle keepalive without hardcoded values ?

Remarks on that.
In don't understand why we would send PING at a rate of keepAlive/2. Shouldn't that just be keepAlive ?
Can you use parenthesis in the ((lastMessage and (millis() - lastMessage >= 10000)) for readability.
Can you check handling of millis() rollover ?

Informatic:
Changed previous hardcoded 1000ms interval with `keepAlive*1000`, as it seems to be behaviour of python-paho MQTT implementation, which seems to work fine.
Added parenthesis per your suggestion, and changed `and` for `&&` since that looks like a more common version in SmingCore.
millis() rollover is handled properly.

I'll test everything again later this weekend.

hreintke:
Thanks for your effort in this. 
FYI : The appveyor fail is due to the experimental implementation we have there.
What matters is the OK from Travis that a PR doesn't break compilation of other parts.

hreintke:
@Informatic : 
Any results from your testing available ?

hreintke:
@Informatic : 
Can you provide an update on this PR ?

Informatic:
Hi, turns out i don't have enough time to test it thoroughly past two and next two weeks. I only had it running for 30-something minutes and worked well, but more testing might be needed.

anakod:
It's ready for merging, I'm sure.

PR #551, opened 2016-01-14
Fix high latency when sending with empty buffer

### LINKED ISSUE #547: High latency when publishing MQTT messages
I noticed major latency (300-500ms) when publishing messages from Sming with long intervals.
Testing code: http://git.io/vzv5P

Turns out it's probably caused by `send()` called in `MqttClient::staticSendPacket` storing data in a stream and flushing it only when it receives readyToSend event. Simply adding `client->pushAsyncPart();` after sending as mentioned above reduces latency to around 10ms.

As far as my testing went, this can be similarily reproduced with plain `TcpClient`.

In theory `send()` could check if buffer was empty at the time of last readyToSend event and call `pushAsyncPart()` immediately in such case. I tried it with simply adding this in `TcpClient::send()`:

``` c++
// [...]
stream->write((const uint8_t*)data, len);

if(asyncTotalLen == asyncTotalSent)
    pushAsyncPart();

asyncTotalLen += len;
// [...]
```

That seems to work, however I haven't tested it much yet, and am not sure if it's the proper way to do this, since I just dove into Sming codebase. I need some comment on that before submitting a proper PR.

Informatic:
Fixes #547.

### DIFF
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -88,6 +88,12 @@ bool TcpClient::send(const char* data, uint16_t len, bool forceCloseAfterSent /*
 
 	debugf("Storing %d bytes in stream", len);
 
+	// If stream buffer is empty we can already push data instead of waiting for
+	// next onReadyToSend event
+	if(asyncTotalLen == asyncTotalSent) {
+		pushAsyncPart();
+	}
+
 	asyncTotalLen += len;
 	asyncCloseAfterSent = forceCloseAfterSent;
 


Sming/SmingCore/Network/TcpClient.cpp
```diff
@@ -82,6 +82,10 @@ bool TcpClient::send(const char* data, uint8_t len, bool forceCloseAfterSent /*
 		stream = new MemoryDataStream();
 
 	stream->write((const uint8_t*)data, len);
+
+	if(asyncTotalLen == asyncTotalSent)
```
raburton: just from looking at it shouldn't this be after the next line `asyncTotalLen += len;`? i'm not an expert on this bit of the code, so i'm probably wrong, but it just looks odd

hreintke:
@raburton : @Informatic :
I need to check this patch. 
There is also a "tcp buffer combining" code within sming.
Must be sure that that is not broken by this.

Sming/SmingCore/Network/TcpClient.cpp
Informatic: `asyncTotalLen == asyncTotalSent` literally means "there is no data waiting to be confirmed as sent", if I understand it correctly. `asyncTotalSent` is increased in `onSent` callback which is surely executed after returning from `TcpClient::send`.

Informatic:
@hreintke If I understand your definition of "tcp buffer combining" correctly, yes, it might somehow break that, as in - if user were to issue multiple `client.send()` in a row, first call will directly send data (flush the stream), however any consecutive calls will be joined into single packet and then fragmented when needed.

Informatic:
This seems to be working now in current develop branch.

Informatic:
Just wondering... Am I the only one to notice that latency? :)

Maybe good alternative would be to just provide users with some sort of flush method?

slaff:
> Just wondering... Am I the only one to notice that latency? :)

@Informatic Can you rebase your PR? I am planning to merge it if there are no side effects on the other network modules.

Informatic:
Aaaaaand it seems to be fine now. I'll retest it later this week.

Sming/SmingCore/Network/TcpClient.cpp
```diff
@@ -88,6 +88,9 @@ bool TcpClient::send(const char* data, uint16_t len, bool forceCloseAfterSent /*
 
 	debugf("Storing %d bytes in stream", len);
 
+	if(asyncTotalLen == asyncTotalSent)
```
slaff: @Informatic  Please, add curly braces for the if-body. `if() {}`

Informatic:
Added curly braces as well as a short explanation of this code, as it seems not-so-obvious.

slaff:
@Informatic If the CI tests pass I will probably merge it tonight and test how are the HttpClient and HttpServer doing.

slaff:
Yesterday I tested the changes with the `HttpServer_ConfigNetwork` sample and here is the results.

# Advantages:
Slightly faster delivery time. In the range of 10-20 ms per request. Probably that is the time between polls and we are forcing the delivery skip those initial 10-20 ms.

# Disadvantage:
Instability. There are things that we probably do not understand quite well in the ESP network stack. 
If ESP is connected to a router and we access the web server on the ESP via the router, and not directly, then all requests seem to be delivered without issues. I used `apache benchmark (ab)` with 4 concurrent request for 40 requests and there was not a single failed request. 

BUT if we connect to the ESP access point and then access the web server there is an instability. Some of the requests don't get delivered completely. Unknown failures and etc.  Again test with ab showed high ratio of failures. That failure rate is not present or very low if we use the current code in `develop` code. It might be related to
* My network router taking basic care to fix TCP traffic
* Issues with the SDK that I am using to test - 1.5.4
* Some wrong setup from my side ( although I tried it multiple times, just to be sure )

I would be happy if others can test it on their ESPs with their SDKs and publish their results.

PR #554, opened 2016-01-18
Edit SPI class to allow for adjusting the frequency

MisterRager:
The SPI class is currently stuck at a fixed frequency. This adds the ability to set the frequency divider(s) in `SPIClass::begin`.

### DIFF
--- a/Sming/SmingCore/SPI.cpp
+++ b/Sming/SmingCore/SPI.cpp
@@ -17,7 +17,11 @@ SPIClass::SPIClass(uint8_t spiID) : id(spiID)
 	// Only SPI_HSPI tested on hardware for now!
 }
 
-void SPIClass::begin()
+void SPIClass::begin() {
+	SPIClass::begin(2, 4);
+}
+
+void SPIClass::begin(uint16_t predivider, uint8_t divider)
 {
 	//bit9 of PERIPHS_IO_MUX should be cleared when HSPI clock doesn't equal CPU clock
 	//bit8 of PERIPHS_IO_MUX should be cleared when SPI clock doesn't equal CPU clock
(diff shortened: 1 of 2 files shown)


MisterRager:
Is there something I should do to fix the AppVeyor issue?

hreintke:
@MisterRager : 
No, not necessary, the configuration of appveyor is not correct.

Question on the PR.
Would it be better to force always setting of both devider and predivider when not default ?

MisterRager:
@hreintke That's probably a good idea. Do you know if adjusting the frequency using one or the other has any difference? That is, will a divider of `8` with a predivider of `4` behave the exact same using a divider of `2` with a predivider of `16` ?

hreintke:
@MisterRager : 
I don't know that.I have no knowledge of the details on SPI. 

Expect you to know that. 
On second reading of the PR, missing is a check on valid input values.

MisterRager:
What is the protocol for throwing errors for invalid input values?

MisterRager:
@hreintke alternately, how do you like it with stuff that just coerces inputs into safe values?

hreintke:
@MisterRager : 
Just above your code the method is : 	SYSTEM_ERROR("UNSUPPORTED SPI id: %d", id);
You can use the same.

Is there any dependence between the two values ? like if one < x then the other needs to be > y ?
If so, also add those checked.

Did you test the patch on hardware with non-default values ?

MisterRager:
@hreintke After some more reading, the `predivider` value is used to reduce the clock speed before the SPI peripheral does its work. The `divider` value is used to determine how many cycles of the predivided CPU clock each actual SPI clock pulse is equal to. The main thing is to keep them within their bounds for values and to disallow 0's.

MisterRager:
WRT relative values, you can see that the default values are such that the `divider` is larger even though the given range for it is much smaller, so a relationship there doesn't seem likely. I've been using "non-default" values for LED control for a while, so I know that works, at least.

hreintke:
@MisterRager : 
This PR is ready for merging. 
Only update needed is to squash the commits into to get a clean commit history.
Can you do that ?

MisterRager:
@hreintke Squeesh!

PR #559, opened 2016-01-22
Added MCP23017

crosofg:
Added MCP23017


hreintke:
@crosofg : 
I think you mixed a couple of fixes/updates on your side in this PR.
Can you clean and then squash the commits into one ?

PR #563, opened 2016-01-23
Add more API docs.

riban-bw:
Remove timestamp from API docs to avoid committing every file every time.


riban-bw:
This PR adds more API documentation. mostly just adding or amending javadoc comments but does add the name of the parameters to some CommandHandler functions (which makes them consistent with the rest of the framework code).
I suggest that future API documentation pull requests are smaller, e.g. for individual modules which will make it simpler to review. I also think that someone who understands the module should review the content to ensure the commiter has understood and described each element correctly.

riban-bw:
Currently failing appveyor. @kireevco please check what is wrong.

hreintke:
close and open to verify travis/appveyor behavior.

hreintke:
@riban-bw : 
I am the author of commandhandler and will check this documentation.

Question on this : What part is generated automatic and which part is added manual ?

hreintke:
@riban-bw : 
Maybe a little late in the API documentation process but have been working mostly on the RTOS version instead of the API updated NONOS version.

I see that all the documentation is placed in the .h files. 
That makes them very hard to use/overview when developing.
I would prefer to have this information in the .cpp files, that keeps the .h files clean and short. 
For application developers this should not give a difference as I expect them to use the documentation and not the .h files to get API information.

riban-bw:
@hreintke Every project I have previously worked on has placed documentation in the header. This provides the following advantages:
- Documentation can be derived from just headers - no need to provide source code
- Able to document code implemented within the header - often done for getter functions
- Able to document (public) class variables which are usually defined in header
- Keeps source code files manageable and clean / balances size of source / header files

The key advantage of placing the documentation in the source file is that there is an in-place description of what the function _should_ do to which a developer can implement.

If we want to change the location of documentation (from header to source) then much of the work I have already done will need redoing. There has been little appetite from the community to perform this work so far. I have done much of the API documentation and am continuing to do so. (I have much documentation that is done but not yet offered in a PR.)

If the consensus of the community is to move the docs then I can do so but don't want to do so unless necessary. I would prefer avoiding a split economy of some docs in source and some in headers as this may become more difficult to maintain.

riban-bw:
@hreintke I have done a bit more documentation of commandhandler now that I better understand it. (I took a break from documenting and tried using the docs to write an application - which gave me insight of what I had documented poorly!) I will update the PR with these updates soon, hopefully within next few days.

All API documentation is currently created manually by someone (me) running doxygen and committing to github.io. Doxygen parses the header files of SmingCore, Services, Wiring, rboot and sdk. (Actually, not all of those headers yet - I am incrementally adding parts as I add API doc modules.) It was agreed that we would do this manually to start with to get it somewhere and revisit automation once we were over the hump of documenting the API. Docs are available at: http://sminghub.github.io.

hreintke:
@riban-bw : 
First of all : I really appreciate you work on the documentation and as said, can do a part of it when touching parts of the code.

For documentation on header vs source. I did a quick scan on the internet and see there are multiple opinions (as expected).
I came to my earlier remark on documentation within header files when touching the header of Station class. This header file has almost quadrupled in size and because of that you ( I ) lose completely the overview of the class. I updated the station.h file (see attached) into a format which gives both documentation info and class overview. I do not want to commit plagiarism, so here the quote I took from someone else which describes my idea.
>My header is treated as a summary (both for the compiler and myself) - it uses a single-line // comment that briefly describes each method. This gives more information than the (hopefully relatively self-documenting) method/parameter names, but a lot less than the detailed documentation in the cpp. Think of it as a summary or reminder that saves you looking at the actual implementation to get enough details to use the method most of the time

On use of documentation, I expect application users (almost) only use the generated documentation and not the header files. The information therein presented can be taken from source files.

On creating documentation : You are adding a documentation to the sming source by hand. As you probably know we are in the process of creating a Sming_RTOS which is based on a different Espressif SDK. We must take care that the effort you do now doesn't need to be done again when we actually move to Sming_RTOS and cease support for current version.

PS : I cannot attach .h file so changed the extension to .txt. Please revert before using.

[Station.txt](https://github.com/SmingHub/Sming/files/112074/Station.txt)

riban-bw:
I guess this PR is stuck (and probably out of date now!) because of the decision whether to place API documentation in the source or header. If placed in source file there is a problem with documenting functions that are implemented in the header, of which there are several within Sming. I understand the arguments for each (source / header) but don't see significant advantage in changing things at this stage.
Another option is to document in another (not source or header) file. I have done this when documenting the elements of the ESP SDK that are used by Sming. This is suboptimal because it means the docs are remote from the functions and less likely to be done by the coder who writes the code. (It can become second nature to add docs to your code as you write the definitions. I do it by habit now.)
I was very engaged in documenting this project but that enthusiasm is waning with this delay in applying the work I have done. (Documentation is a task few relish!)
We need a decision on location of API docs (source / header) and then activity to process the PR backlog. Anything I can do to move things along just let me know.

hreintke:
@riban-bw : 
First of all : Your activities on documentation are very welcome, appreciated and necessary for gettting Sming (more) friendly for both users as developers.

I did not take further action and react on this PR/issue as the main contact was between you, @raburton and @kireevco . My way of working is then not to get into the detailed discussion but only comment on the points that effects my activities. I am also no in the details of the relationship between the sming/smingrtos and sminghub.guthub.io.

But.. I am more than willing to participate more into this discussion to get a maintainable solution implemented.

Then for me to get clarity on the activities and results we have for the documentation this far. 
Questions : (and maybe they are duplicate from earlier discussion but here to understand) 
- Do we distinguish documentation for "Framework users"  and "Framework developers" ?
- What documentation (content and location) will we provide for those ?
- What will be (currently) manual vs automated ?
- What is the state/level/version of the current documentation ?
- What decisions need to be made to restart the documentation activities ?
- What are the next steps to be taken ?
- Are there further open questions ?

raburton:
I was quite happy with your work, I didn't realise this hadn't been merged already. Your rBoot documentation was already merged upstream and pushed back into Sming (original and rtos). I haven't had time to be involved much recently myself, rotated job and stated working towards my next exam recently, but I wasn't really involved in this PR. If it's just more of the same goodness then lets get it merged, I don't see the controversy!

riban-bw:
@hreintke I shall try to answer your points:

Do we distinguish documentation for "Framework users" and "Framework developers" ?
I have targeted the API documentation at Framework users. The motivation is to provide Framework users with the tools required to use the framework. I tried to use the framework and found the learning curve too steep with insufficient supporting documentation, hence my effort to improve this. Framework developer documentation may best be provided as comments within the source code where needed (in combination with intuitive element names, structured code layout and accompanying guides / wikis / etc.)

What documentation (content and location) will we provide for those ?
Using doxygen to generate HTML documentation from in-source comments, a web based provision of API documentation is hosted at http://sminghub.github.io. Placeholders exist for stable / development versions of Sming / SmingRTOS. I'm not sure all four can be easily maintained - need consideration. I expect development version docs to be current which gets snapshot and copied to release version when a release occurs so probably not much effort.

What will be (currently) manual vs automated ?
Currently the whole operation is manual. This is because there are a few hurdles for automation that, although not insurmountable, would delay getting the docs issued. Pragmatic solution was to just issue them manually (which I can do) and consider automated docs creation later.

What is the state/level/version of the current documentation ?
Current API docs are incomplete and not reviewed. I have been working through each class, adding my take on each. I want to complete this then get peer review. I want users (including me) to try using the API docs and report errors and other issues. I tried using some of the API docs I created and found they were not ideal so updated to make easier to use. The aim is to have all current elements documented and to request all new code be provided with documentation.

What decisions need to be made to restart the documentation activities ?
Location of API documentation (and any high-level style guides). If current style and location are acceptable (i.e. placing API docs in comments within header files, using javadoc style) then I can continue. If this is unacceptable then new docs need to use new rules and old docs need to be moved. I would prefer the current rules to continue and feel it may be easier to copy docs to RTOS because headers are more likely to be similar than source files. (There are the other reasons listed earlier in this discussion also.)

What are the next steps to be taken?
Decision on API documentation location.
Current PR merged. (May need attention which I can do if someone says it will be processed soon afterwards - I don't want to have to do this too many times!)

Are there further open questions?
Who should be reviewing API doc PRs? There is a lot of documentation to add (which I have done more than in this PR) and don't want to lose momentum due to delays in PR processing or having to revisit PRs due to them becoming out of date.

riban-bw:
Also, @anakod wrote a guide to documenting the API in the Wiki (https://github.com/SmingHub/Sming/wiki/Documentation-creation) but it doesn't seem that many (any?) of this was actually applied. I have just written a guide (https://github.com/SmingHub/Sming/wiki/Documenting-the-API) which differs a little. As most of the docs done so far match my latter style (because I did them before I saw anakod's guide), shall I combine these two wiki guides using my style guide and anakod's workflow guidance?

hreintke:
@riban-bw : Was not even aware of the wiki page on documentation.
Thanks for your extensive reply, will give a reaction on that within the next days. Need a decent thought on this to get the way forward.
And sure, I want to prevent that a lot of work has to be redone.

hreintke:
@riban-bw : 
Sorry for the delay in my response, more busy than expected last time.
>Location of API documentation (and any high-level style guides)
Been thinking on this a lot as it have big impact on final source code and activities already done.
I keep to my previous statement that I prefer to have (most of) the documentation in the source files instead of header files. And for documentation within the header the limitation is that is on the same line as the item it documents.  As far as I understand this has no impact for the users of the documentations but keeps the header/source file clean and usable for the framework developers.
We should however find  way to minimize the additional effort for you and keep the documentation progress going. 
I am thinking on : Agree on the above but commit most of the current. New doc will then be based on new, and when touching already documented, it will be moved to new.
Am happy to give you access rights to commit the documentation in the repository..

On reviewing, as always will be difficult to get people to do it in a structural way. Is the a way that we can out the status of documentation (draft, reviewed,..) somewhere in it ? 

There is almost no difference in API and source files between NONOS and RTOS.
We could even decide to get documentation only in the RTOS version as that will be the future version of Sming.

riban-bw:
@hreintke Will you please propose how we document elements of the API that _only_ appear in header files, e.g. MemoryDataStream::getStreamLength()? I wouldn't want to see some of the API javasoc in the source files and some of it in the header files. Cheers

hreintke:
@riban-bw : 
Please correct me if I am wrong.
This is an implementation within Sming that is not like the "general use/implementation" and therefor not heavily used. Probably the best is to move them out of the .h file into the .cpp file.

When we encounter these I suggest to do either : 
- Directly move the code to cpp and document
- Temporary document within h. and list for update
- Temporary leave out documentation and list for to be documented.

When going thru all Sming we will encounter imperfections within Sming implementation.
If that is not a one-off we can decide for each of them what is the easiest way to go.

slaff:
@riban-bw Do you have time and energy to update this PR? Sming is lacking good documentation and that is something that we need. To make things simpler: Can you move documentation comments only in the headers?

riban-bw:
Sorry for the long delay. Work commitments have engulfed me!
I felt (and still feel) that we were at an impasse. The two most vocal contributors to this debate have opposite opinions on its solution. There are others who support one argument but it kind of comes down to the opinion of hreintke (who is a significant code contributor) and riban-bw who is a significant documentation contributor. The former has valid reasons for wanting the documentation tags to exist in the source code files whist the latter (me) has valid reasons for the documentation tags to exist in the header files. Without a steer from the rest of the community, I do not feel sufficiently justified to decide. My experience informed my original approach and the document creation mechanisms are structured for that approach. It will take some effort to change the method which I only want to do if there is overwealming support for the alternative. This PR conflicts with hreintke proposal and hence cannot be processed if we go that way. I spent a lot of time making this work and I'm not sure if I have the appetite to rework it to support the alternative approach, especially when I don't appreciate the benefit being significant. I am the junior partner in this discussion and hence do not feel empowered to contradict hreintke. If we agree to change to hreintke proposal, this PR needs to be rejected.

slaff:
> I spent a lot of time making this work and I'm not sure if I have the appetite to rework it 

@riban-bw I understand and completely agree with you that results from hard work should not be wasted. And I do not want contributors to be disappointed and progress to stall. 

A good project needs a good documentation. And adding a good documentation is an ongoing effort that is taking quite some time. Before we start the documentation initiative we have to decide one simple thing, namely "Where to put the documentation comments (the ones used by library users): header files or C/C++ files".

@hreintke @raburton @anakod @alonewolfx2  @robotiko @piperpilot @patrickjahns @kireevco Please, vote for `documentation` comments in HEADERS or C/C++ code files. We need your vote till 6th of November. Please, vote for the answer that will be make the documentation useful for the users. Give your "vote" as comment to this PR.

hreintke:
@riban-bw @slaff : 
Reading the thread again and re-thinking options meaning not just sticking to earlier writings.

However, I still prefer splitting the documentation between .h and .cpp.
But as always, that is just **my** opinion and if others want to implement otherwise so be it.

Same accounts for NONOS vs RTOS documentation. Don't know how we should get documentation for both in one go.

riban-bw:
Regarding NONOS / RTOS, if the API remains broadly similar between the two, putting API doc's in headers may help. The function definitions will be the same (or similar) but implemention may differ. The same (or similar) headers may be used for both. If the headers differ, it should be by a fairly small amount, allowing fairly easy diff and merge. (Scooter Software's Beyond Compare is excellent for such process but free alternative like Meld also work.)

slaff:
Sounds like we are reaching a solution here. 

I can suggest the following future steps: 
We should add `documentation` comments in the headers files. Those comments should describe **WHAT** are the function/classes/methods/enums, etc doing. At some point we should make these documentation comments required. For example we can add automated checks that run over the changed/added source code in the commit and decide if a commit should fail or succeed.
If the developers feels like it he can also add documentation comments in the C/C++ files describing **HOW** the function is working.  But that should be optional.

If there are no voices against this plan than we can think about scheduling those changes to be part of the upcoming releases ( 2.1, 2.2 ... etc )

slaff:
Merged in PR rebase https://github.com/SmingHub/Sming/pull/794

PR #573, opened 2016-01-28
HotFixes for master branch

anakod:
* Restore analogRead for compatibility reasons
* Fix memory leaks in Vector copy operations


hreintke:
I know about discussion on this but not the details. is some time ago.
It was about the pin parameter. that is not used at all -> it is suggested that analogread is available on multiple pins, which is not.
So : At least the analogread as in the patch is not one we should have in.

anakod:
It's added to have compatibility with Arduino. We should stay comapatiblile while it possible (and didn't have deterioration for end user).
In ESP32 as I know we will have not only single ADC, so this parameter will be used in next version too (and still will be compatible).

hreintke:
I agree with being compatible with arduino but having a function which has the same name as in arduino but acts different is no compatibility.And will confuse Framework users.

There will much more parts of code that will have different values/possibilities with esp8266 vs esp32.
There probably will come a compile time options to distinguish the two.

anakod:
> but acts different is no compatibility
What is difference? It's the same. Just single analog pin.

hreintke:
What is the behavior when you call analogread(pin) on arduino with a pin that does nor exist ?

anakod:
This pin exist, its TOUT pin in ESP. WIll be right add special define A0 for it (like in Arduino).

Main point why it's needed - it's because if user get some Arduino library he will have compilation error without that function. And moreover, he should know low level ESP API method (what can be changed in future) to manually replace all analogRead calls. It isn't good.

anakod:
Ok, @raburton replied, it's removed by mistake: https://github.com/SmingHub/Sming/commit/34fbd5dc7d998beee7da59d3400935584d8a7a5e

I think adding special define for analog pin (A0) will be good idea.

raburton:
Yes, wasn't deliberate. But I agree, accepting a pin number and then always returning the value from one specific pin is confusing. Perhaps validating the pin would be a good idea, or people might be calling it against other pins and wondering why it doesn't work as they expect.

anakod:
@raburton, like this, yes?

hreintke:
@anakod :
Yes, That would make it compatible.

BTW : I think we should not merge into master but in develop branch

anakod:
Hotfix needed for master. But if you plan to merge develop to master in very short time - I agree, it will be beter go over dev.

anakod:
@raburton, what's our the merging plan? How will be better apply fixes to both branches?

raburton:
Ideally we put everything into development and make a new release. For hotfixes we should cherry pick the commit from development to master, to enable future releases to be made easily. But if it's not very urgent waiting till next release would be preferable - this has gone unnoticed for 3 months, so I think it probably fits that category.

anakod:
Main point here what I want to apply hotfix for memory leak (but analogRead also used). It stopping my new project what I'm planing to publish in few days.

Also I think what we should have standartized way to push hotfixes to master when we know about serious bug. Otherwise no reason to support master, we can use only one developement branch :)

alonewolfx2:
@anakod what is new project :)  
We can merge into develop and create new hotfix release I think it's urgent.

hreintke:
@anakod @alonewolfx2 : @raburton 
We have to be careful with the AnalogWrite patch. Especially with the `#define A0 17`
- It will break all applications which have a A0 defined themselves. I can imaging people writing/converting arduino like apps will do so. 
- In arduino you can read and write A0 port both have valid results/pin reaction. That is not the case here. So question is it A0 or A17 ?
- In current Sming we do not have any of the A0.. D0.. defined. If we introduce A0 we better take it all the way and make a patch which defines them and when necessary updates other function.

For Vector it might be different but..
- As Richard mentioned, no complains until now and it has been in since start of Sming
- If it is in develop, applications can still use it.
- In develop we have HardwareTimer and Servo (needs some rework) and other bugfixes on the roll to be merged in there. It will be not to long for actual release. -> Vector fix will not take long to make it to Master

anakod:
It's standard Arduino pin naming notation: 1,2,3....(LastDigitalPinNumber), A0, A1, A2...., A6.

ADiea:
@anakod  thanks for sharing, especially the vector leak!

hreintke:
I was looking for something similar in SmingRTOS and this should/could do the variable ldfile.
Using awk which we use already so no additional dependency.
No real IROM0_SIZE values yet, needs to be set correctly.
I have tested it in the command window but can't get it to run in the makefile-project.mk


on the awk line I get `c:\userdata\rtos_project\sming/Makefile-project.mk:233: *** missing separator.  Stop`
I know it has something to do with tabs/spaces but think I tried all options. 
Can you have look at it ? 
Update/additional base link file with : 
```
MEMORY
{
  dport0_0_seg :                      	org = 0x3FF00000, len = 0x10
  dram0_0_seg :                       	org = 0x3FFE8000, len = 0x18000
  iram1_0_seg :                       	org = 0x40100000, len = 0x9000
  irom0_0_seg :                       	org = 0x4020A000, len = IROM0_SIZE
}
```


Update to makefile-project.mk in RTOS but similar in NONOS
```
ifeq ($(SPI_SIZE), 256K)
	flashimageoptions += -fs 2m
	SPIFF_SIZE ?= 131072  #128K
	IROM0_SIZE ?= 0x52000
else ifeq ($(SPI_SIZE), 1M)
	flashimageoptions += -fs 8m
	SPIFF_SIZE ?= 524288  #512K
	IROM0_SIZE ?= 0x58000
else ifeq ($(SPI_SIZE), 2M)
	flashimageoptions += -fs 16m
	SPIFF_SIZE ?= 524288  #512K
	IROM0_SIZE ?= 0x66000
else ifeq ($(SPI_SIZE), 4M)
	flashimageoptions += -fs 32m
	SPIFF_SIZE ?= 524288  #512K
	IROM0_SIZE ?= 0x78000
	
else
	flashimageoptions += -fs 4m
	SPIFF_SIZE ?= 196608  #192K
	IROM0_SIZE ?= 0x52000
endif

# linker script updated for the above settings
LD_PATH     = $(SMING_HOME)/compiler/ld/

LD_SCRIPT_BASE	= $(LD_PATH)eagle.app.base.ld
LD_SCRIPT		= $(LD_PATH)eagle.app.ld

$(Q) if [ -f "$(LD_SCRIPT_BASE)" ]; then \
	  awk '{gsub("IROM0_SIZE","$(IROM0_SIZE)");print > $(LD_SCRIPT)} ' $(LD_SCRIPT_BASE) 
	fi
```

raburton:
I considered awk or similar when adding rBoot, would be nice to automate the ld files instead of needing user to supply them. Couldn't think how to do it on windows though, and osx probably has a different version of awk, if any.

hreintke:
There is already use of awk in the makefile-project.mk, so everyone using sming must have it on the system.
The use in this is simple and at least it works within windows.
If you give it a try in linux, we just need a volunteer for mac testing.

There a more places in the makefiles where we could use this.
I think it's worth a try.

anakod:
We should build new Sming release because of serious memory leak.
I will close this P-R for now, will be better merge code from develop branch.

PR #575, opened 2016-01-30
TcpClient bug (Mqtt, Http client etc.)

flexiti:
Messages longer than 256 bytes were sent partly. (low value from int len only)
Tested on MQTT
(before work only those messages whose size: ( 6+ topic + payload) were less than 256 bytes)


hreintke:
@flexiti : 
I think you have some other updates in your environment which makes this one conflicting with develop branch

flexiti:
Should I do it again?

hreintke:
@flexiti : 
Yes please. I am not sure what is wrong/missing but you see the line 
`This branch has conflicts that must be resolved` in the PR summary box.

If that is solved I can/will merge.

PR #579, opened 2016-01-31
Tcp Client bug

flexiti:
Messages longer than 256 bytes were sent partly. (low value from int len only)
Tested on MQTT
(before work only those messages whose size: ( 6+ topic + payload) were less than 256 bytes)


flexiti:
@hreintke 
Now is ok ?  (I took a completely new develop from Sming)

hreintke:
@flexiti : Thanks, merged

PR #581, opened 2016-01-31
Add esptool2 as a git submodule

JelmerT:
Adds `esptool2` as a submodule. This gives a better 'works out of the box'-experience. You don't have to install an external tool and then dig in makefiles to set the path to that tool.

By using a submodule, you can just pull in changes if esptool2 gets improved in the future. The source will automatically be checked out by git when you clone the Sming repo.

Also updated the Readme to reflect these changes.

(only tested this under OSX)

note: I have no affiliation with esptool2

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "Sming/esptool2"]
+	path = Sming/esptool2
+	url = git@github.com:raburton/esptool2.git
--- a/Readme.md
+++ b/Readme.md
@@ -32,8 +32,8 @@ Sming - Open Source framework for high efficiency WiFi SoC ESP8266 native develo
 - [MacOS](https://github.com/SmingHub/Sming/wiki/MacOS-Quickstart)
 
 ## Additional needed software 
-- Spiffy  : Source included in Sming repository
-- [ESPtool2] (https://github.com/raburton/esptool2) esptool2 
+- [Spiffy](https://github.com/xlfe/spiffy): Source included in Sming repository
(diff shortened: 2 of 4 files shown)


alonewolfx2:
@raburton what do you think about this?

raburton:
I don't have a problem with including esptool2, but the idea of using sub-modules has been discussed before and never accepted. I don't think it was actively rejected, just never actually accepted. Adding sub-modules makes it more complicated to clone the source and keep your local copy up to date. I can think of at least one other outstanding PR that stalled due to this issue.

kireevco:
This requires full x86 toolchain to build it. Personally I'm against such kinds of integration. It's really doesn't play in favor of portability.

I would rather build a separate binary for esptool2 and ship it with sdk. (Which I already do-check esp-alt-sdk)

raburton:
That's true but building it would be optional and we do the same with Spiffy (also optional). But then having got spiffy down to a single file, which requires spiffs files from within the build tree and not having it's own home anywhere else, there was really no other logical place to put it. Overall I'm more inclined to avoid including third party dependencies that don't need to be included. But if we did go down that route (and start using sub-modules) it would allow us to make modules out of spiffs and rBoot, and possibly others too, which would simplify our code base.

kireevco:
I have an idea. Let's use submodules, building will be optional. On the other hand we can include download binary file procedures.


Honestly platform io will solve most of user delivery issues. I'd rather see work in their direction and sming should state dependencies of external tools .

hreintke:
No agreement on working with submodules.

PR #586, opened 2016-02-01
fixed support for HW SPI for Adafruit_ST7735 TFT's

harry-boe:
fully tested now in SW and HW SPI mode. HW SPI is based on swing-core/SPI


hreintke:
Close & Open to force CI retest

harry-boe:
PR will be obsolete with new SPI implementation as the new SPI support Arduino Style transactions

PR #591, opened 2016-02-05
Hotfixes 20160128

anakod:
https://github.com/SmingHub/Sming/pull/573 for DEV branch too.


anakod:
I didn't think what we should calculate exact accurate size here. If program will be smaller than section size, it will be smaller. I think it's just up limit and should be larger.

PR #592, opened 2016-02-06
-  HttpResponse::sendDataStream(MemoryDataStream* newMemoryDataStream…

harry-boe:
i need a way to send image files (streamed from a vide cam) as HttpResponse 

the Ajax sample app is a test case for the feature. 
it loads a file called favicon and sends it as BinaryData with  from a favicon.ico handler.

```
	MemoryDataStream *stream = new MemoryDataStream();
	file_t fav = fileOpen("favicon", eFO_ReadOnly);
	if (fav) {
		response.setContentType(ContentType::ICO);
		response.sendDataStream(stream);
		while (!fileIsEOF(fav))  {
			len = fileRead(fav, buf, sizeof(but));
			stream->write((unsigned char *) buf, len);
		}
	}
	response.notFound();
```

easy and consistent to use like the other http response types


samples/HttpServer_AJAX/app/application.cpp
```diff
@@ -78,12 +78,40 @@ void onAjaxFrequency(HttpRequest &request, HttpResponse &response)
...
 }
 
+void onByteStream(HttpRequest &request, HttpResponse &response)
+{
+	int len, size = 0;;
+	static void * buf[20];
+
+	MemoryDataStream *stream = new MemoryDataStream();
+
+	file_t fav = fileOpen("favicon", eFO_ReadOnly);
```
anakod: We didn't have this file in example, I think

samples/HttpServer_AJAX/app/application.cpp
```diff
@@ -78,12 +78,40 @@ void onAjaxFrequency(HttpRequest &request, HttpResponse &response)
 	response.sendJsonObject(stream);
 }
 
+void onByteStream(HttpRequest &request, HttpResponse &response)
```
anakod: It's basic exmaple, so I'm not sure what advanced byte streaming sample needed here.

anakod:
Looking good. About example - I'm not sure about it. May be better remove code from here (to don't confuse beginners).

hreintke:
@harry-boe 
What is the difference between your solution and : 

response.setHeader("ContentType",ContentType::ICO)	;	
response.sendFile(file);

harry-boe:
> What is the difference between your solution and :
response.setHeader("ContentType",ContentType::ICO) ;
response.sendFile(file);

The difference is that it is using response.SendDataStream -> Test case for that feature

-> i needed some kind of test case without implementing a new Sample.
-> I can live without that sample code. But i will need the ByteStream for a ArduCAM library and sample
    we are working on.

harry-boe:
Another way to do it is to define a static buffer with the favicon data in it within the application.cpp file

That way it's maybe a bit more clear what it does

hreintke:
What is the behavior/specifics of the "ByteStream" vs the other Streamtypes ?
The response.SendFile/json/memory and as now send all until end of stream.
No need for any other static buffer.

If there is the need for specific response reaction for the ByteStream, it might be better to handle that in the HttpResponseClass instead that it is handled by the application.

harry-boe:
It is not about the response type
It is about binary data
It is handled in the response class

hreintke:
Will see then when you PR this functionality.

I have similar functionality already build in the Sming_RTOS version.
Would like to understand whether our implementations have conflicts.

harry-boe:
This here is actually the PR

hreintke:
I was thinking on making it generic.  
```
bool HttpResponse::sendDataStream( IDataSourceStream * newDataStream , String reqContentType /* = "" */)
{
	if (stream != NULL)
	{
		SYSTEM_ERROR("Stream already created");
		delete stream;
		stream = NULL;
	}
	if (reqContentType != "")
	{
		setContentType(reqContentType);
	}
	stream = newDataStream;

	return true;
}
```

harry-boe:
Can't check source for idatasourcestream right now. But sounds reasonable if idatasourcestream is base class of memorydatastream

anakod:
Yes, IDataSourceStream is base interface. Generic solution is a little better for this case, I think.

harry-boe:
Can u make a PR for that
I'm on vacation

hreintke:
@harry-boe : 
I PR'ed. See #598 

Can you test ?

harry-boe:
It works as expected 

THX

PR #595, opened 2016-02-07
Remove hardcoded timings based on F_CPU in Neopixel lib

### LINKED ISSUE #594: Hardcoded timings in Neopixel driver
Neopixel driver (and many other drivers I suppose) doesn't work when CPU on 160 MHz because of hardcodced F_CPU value and dependent timings. See https://github.com/SmingHub/Sming/blob/develop/Sming/Libraries/Adafruit_NeoPixel/Adafruit_NeoPixel.cpp#L108

How to reproduce: add `system_update_cpu_freq(160);` as first line to the `init()` function in https://github.com/SmingHub/Sming/blob/develop/samples/Basic_Neopixel/app/application.cpp#L178

UncleRus:
Fixes #594. But I'm sure more work is needed to completely remove hardcoded F_CPU in other drivers.

### DIFF
--- a/Sming/Libraries/Adafruit_NeoPixel/Adafruit_NeoPixel.cpp
+++ b/Sming/Libraries/Adafruit_NeoPixel/Adafruit_NeoPixel.cpp
@@ -105,15 +105,8 @@ void IRAM_ATTR Adafruit_NeoPixel::show(void) {
 
   boolean is800KHz=true;
 
- #define CYCLES_800_T0H  (F_CPU / 2500000) // 0.4us
- #define CYCLES_800_T1H  (F_CPU / 1250000) // 0.8us
- #define CYCLES_800      (F_CPU /  800000) // 1.25us per bit
- #define CYCLES_400_T0H  (F_CPU / 2000000) // 0.5uS
- #define CYCLES_400_T1H  (F_CPU /  833333) // 1.2us
- #define CYCLES_400      (F_CPU /  400000) // 2.5us per bit
-
    uint8_t *p, *end, pix, mask;
-   uint32_t t, time0, time1, period, c, startTime, pinMask;
+   uint32_t f_cpu, t, time0, time1, period, c, startTime, pinMask;
 
    pinMask   = _BV(pin);
    p         =  pixels;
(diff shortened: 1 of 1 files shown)


UncleRus:
Is there any reason why it cannot be merged?

hreintke:
@UncleRus : 
First of all :  because we need some time to evaluate PR's which come.
+ : There is an easy workaround -> do not use 160Khz -> Not very high priority
+ : I don not understand the define : NEO_KHZ400 -> How generic is the PR.
+ :  I did not see any remark on : Tested OK on hardware

ADiea:
We only have 2 frequencies 80/160 Mhz could a table with these constants be a better chioce than to compute them over and over every time?

UncleRus:
- "Do not use 160MHz" - really? If it can be used, it will be used.
- It's tested on a hardware, and it's work.
- NEO_KHZ400 was already defined here, don't know why.

And precomputed values is good idea, so I will rewrite the code.

hreintke:
160Mhz : Sure it will be used. What I am saying -> There is a workaround for someone who hits this issue now. So this is not a high priority PR which has to be merged very very fast.

If you don't know what NEO_KHZ400 is : How did you test both options ?
Sure the committer always did a complete test. I'd like to see a confirmation from others too.

On making the PR generic : 
There should be not hardcoded values in the code -> should all move to defines.

Should we have the esp defined values in a "esp8266" ifdef ?
The we only add something to an existing library, it is easier to maintain when new versions arrive.

robotiko:
Hi, 
I do agree with @ADiea   that it is better to have a fixed freq values instead of run heavy computations. 
I do see benefit on supporting the 160MHz since included libs should work with any valid device config  (in fact I do use 160MHz and need Neopixel lib).

About NEO_KHZ400: Neopixel (WS2812 chips) work in 800 KHz bitstream and just legacy "V1" Flora pixels (WS2811 drivers) work in 400KHz mode.  

Most if not all led strips and led rings that you find nowadays online are 800KHz.

hreintke:
No progress on this PR.
If more actions/updates necessary it can be re-opened.

PR #603, opened 2016-02-15
Add DS3232RTC library for supporting DS3232/DS3231 RTC modules

avr39-ripe:
Add DS3232RTC library with small example app for setting up time in DS3232 / DS3231 RTC module to local time (timezone hardcoded to GMT+3 needs change for local TZ) via NTP


hreintke:
@All 
Is this library tested ?

@avr39-ripe : Can you also make a PR for SmingRTOS ?

avr39-ripe:
@hreintke as I promise - I will port all my contribution to nonOS Sming to rtosSming, but I need some time to finish urgent task.. I whant to port WifiEvents, DS3232RTC, getRaw http request body..

I use this lib in one project with is in production.. works so far.. Bit later it worth to change current DateTime library which is a part of Smingcore to Time lib as newer lib with better interface, but even with current DateTime lib it works as a charm.

PR #607, opened 2016-02-17
slight refactor to sample code to highlight the main point of the app…

mralexgray:
…, delegates

### DIFF
--- a/samples/Basic_Delegates/app/application.cpp
+++ b/samples/Basic_Delegates/app/application.cpp
@@ -1,45 +1,43 @@
-#include <user_config.h>
 #include <SmingCore.h>
+#include <user_config.h>
 
-class LedBlinker
-{
+class LedBlinker {
+ public:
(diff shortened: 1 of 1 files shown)


hreintke:
In my opinion he updates you propose would not make the basic_delegates more readable  
```
return reqInterval <= 0 ? false : (ledInterval = reqInterval);
```
Returning the result of an assignment as a bool ?
```
 if (!reqRun) return ledTimer.stop();
```
Returning a value in a void() function ?

```
  ledTimer
          .initializeMs(ledInterval, TimerDelegate(&LedBlinker::ledBlink, this))
          .start();`
```
```
LedBlinker myLed1 = LedBlinker(LEDPIN_1),
					 myLed2 = LedBlinker(LEDPIN_2);
```
Nowhere used in the examples

hreintke:
Original delegates example merged.

PR #608, opened 2016-02-19
Added indiv. library search paths to allow same seach <pattern.h> in …

mralexgray:
Added indiv. library search paths to allow same seach <pattern.h> in core/projects.

This allows "libraries" to depend on each other.  Also, previously.... if you used <Library.h> when compiling for `core`.. everything would be fine.. but then when `#including` a co-/dependent lib in a project... it would fail, expecting a `#include <Library/Lib/Lib.h>` type include.  This fixes that.


hreintke:
@mralexgray : 
The "no inclusion" of the library path is done to make an application developer aware of his use of a library, which is not part of the sming core -> you have to add the "include" for the library.
The same principle is taken with SmingRTOS.

If no specific issues occur with this principle I would like to keep it that way.

Issue #615, opened 2016-02-23
Bug: missing header information (if header split between tcp packets)


patrickjahns:
While working on the HTTP_POST multipart implementation I found a glitch/bug with the current header parsing implementation

I suppose the following code in `HttpRequest::parseHeader` is meant to wait until we received the complete header:
```
int headerEnd = NetUtils::pbufFindStr(buf, "\r\n\r\n");
if (headerEnd == -1) return eHPR_Wait;
```
But what actually what might happen if the header is split between two packets, that the buffer from the first packet will be emptied from lwip.
`HttpServerConnection::onReceive` receives the `eHPR`_Wait flag and calls  `TcpConnection::onReceive(buf);` which empties the buffer from the current packet.

I think we can change this behavior if we either have a local buffer for the header, or we find a way to keep the paket inside the lwip puffer

hreintke:
@patrickjahns : 
Isn't there already a kind of buffer ?. As I understood the code it is getting header information and saving it in the request parameter hashmap. I that is the case there is no need for keeping the tcp packet, only the part of the header being processed (maybe in the same hashmap ?)

patrickjahns:
@hreintke 
Yes the code is getting header information and saving them in a Hashmap. The difficulty now is, that we first search if the header is complete by finding `\r\n\r\n`. If the first packet does not currently have the string we discard the packet and check the next - if the next packet has `\r\n\r\n` we only process that packet. 

I guess what would be also an alternative, is to process the first packet as far as we can. Keep the not processed part in a temporary buffer - then process the temporarybuffer+the next packet as far as we can etc...

I would find it easiest to not need a temporary buffer, but extend the TCP::onRCV with an optional size paramter and only clear lwip buffer until that parameter.

Assuming that calling `tcp_recved(tcp, p->tot_len)` from lwip will only clear the pbuf until the give size;
I am still not too familiar with lwip and the workings around it - maybe you can clarify on this a bit more to pin point me into a direction.

As i`ve sent you in gitter - this idea would also make it easier to implement HTTP Multipart Post (#552)

hreintke:
@patrickjahns : 
LWIP will not clear the buffer. It is done in TcpConnection::StaticOnreceive with the pbuf_free().

I am in favor of the processing of the packet as far as we can. 
The resulting header:Value combinations are now build and kept in strings.
[...]

Issue #552, opened 2016-01-14
Support for POST files and Upload


ram-one:
Is there any support for uploading files with POST requests ?

Bonus question : could we update the device with a manual firmware upload ?

Vity01:
I just wanted to create new thread for this with a same RFE.
@ram-one There is only HttpClient class which you can use for uploading really small files. 
See HttpClient->setPostBody and FileSystem->fileGetContent. See related sample... Note that it's not suitable for bigger files, because it means to load whole file content into memory (+ it's using MemoryDataStream internally).
BUT for big files (eg. >10kB) there is still  no support. There should be some new function aka 
HttpClient->setPostBody(new FileStream("my file")) which would allow to upload bigger file.  

Also you can also use FTP server and download data from external server, if this usecase fits your needs (not for mine, since my AP is not working 24/7).

ram-one:
What I meant is uploading TO the ESP.
Httpclient class is for sending a request.

I used code from PR #499 to be able to read the incoming post request body, but everything goes to the memory. It's not suitable for large files.

Vity01:
So there are 2 use cases:
1) Sming HttpServer - receiving large files
2) Simple "ajax" request (using HttpClient) for sending large files
both needs enhancement

hreintke:
@ram-one @Vity01 : 
I agree that the option for sending a file with HTTP Post would be a viable option.
If you have a proposal/initial design for that I would be happy to discuss that.

With current functionality there is probably a workaround.
For that see the rboothttpupdate class which receives a full file, now sending it to flash but creating a similar class which sends to file should not be to difficult,
If you trigger this by sending a http_post with the "url of the file to be  loaded" you have a "poor mans http post file"

ram-one:
This workaround has a big drawback : It needs a webserver on the other side.

hreintke:
@ram-one : 
I know, but it is available now without major additions to sming.
If this is not an option you either have to create a PR (wiling to discuss the design) or wait until someone else does.

patrickjahns:
I would be interested in this too - but for a different use case scenario: enabling OTA Updates via http post. This way people do not need a webserver of their own and request would never need to leave their own network

patrickjahns:
@hreintke

I had a look at the current http server implementation and this is what I found so far/ideas for changes:

`HttpRequest::parseHeader` needs to be extended to properly detect multipart headers (`multipart/form-data`) - also extract the boundary from there to detect the parts

[...]

PR #622, opened 2016-02-27
Creating a buffer for parseHeader and parsePostData 

patrickjahns:
This addresses #615 and also prepares for working on multipart requests

The previous implementation did not account that header information could be split between several packets. Also post data can be split across several TCP packets.

The changes address this by storing not processed data in a temporary buffer and pre pending the new tcp buffer so it will be parsed together

Has been tested with several requests (GET/POST including Cookies and large headers)

### DIFF
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -186,7 +186,7 @@ HttpParseResult HttpRequest::parsePostData(HttpServer *server, pbuf* buf)
 		tmpbuf = tmpbuf.substring(start, tmpbuf.length());
 	}
 
-	tmpbuf = extractParsingItemsList(tmpbuf, start, tmpbuf.length(), '&', ' ', requestPostParameters);
+	tmpbuf = extractParsingItemsList(tmpbuf, 0, tmpbuf.length(), '&', ' ', requestPostParameters);
 	postDataProcessed += buf->tot_len - start ;
 
 	if (postDataProcessed == getContentLength())
@@ -203,13 +203,13 @@ String HttpRequest::extractParsingItemsList(String& buf, int startPos, int endPo
 	int delimItem, nextItem, startItem = startPos;
 	do
 	{
-		//debugf("item %i  - delim %i - next %i", startItem, delimItem, nextItem);
 		delimItem = buf.indexOf("=", startItem);
 		nextItem = buf.indexOf(delimChar, startItem);
+		//debugf("item %i  - delim %i - next %i", startItem, delimItem, nextItem);
 		if (nextItem == -1) nextItem = buf.indexOf(endChar, delimItem+1);
 		if (nextItem == -1) nextItem = endPos;
-		if (nextItem > endPos || delimItem == -1) nextItem = endPos; break;
-
+		if (nextItem > endPos || delimItem == -1) nextItem = endPos;
+		if (delimItem == -1) break;
 		String ItemName = buf.substring(startItem, delimItem);
 		String ItemValue = buf.substring(delimItem+1, nextItem);
 		char* nam = uri_unescape(NULL, 0, ItemName.c_str(), -1);
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -17,9 +17,10 @@ HttpRequest::HttpRequest()
 	requestGetParameters = NULL;
 	requestPostParameters = NULL;
 	cookies = NULL;
+	headerDataProcessed = 0;
 	postDataProcessed = 0;
-	combinePostFrag = false;
 	bodyBuf = NULL;
+	tmpbuf = "";
 }
 
 HttpRequest::~HttpRequest()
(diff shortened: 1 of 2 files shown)


patrickjahns:
@hreintke 

any comments so far?

Sming/SmingCore/Network/HttpRequest.cpp
```diff
@@ -82,77 +83,94 @@ String HttpRequest::getContentType()
...
-	debugf("path=%s", path.c_str());
 
 	int line, nextLine;
-	line = NetUtils::pbufFindStr(buf, "\r\n", urlEnd)  + 2;
+	line = urlEnd;
 	do
 	{
-		nextLine = NetUtils::pbufFindStr(buf, "\r\n", line);
+
+		nextLine = tmpbuf.indexOf("\r\n", line+2);
```
PTDreamer: Shouldn't the last parenthesis be placed before the "+" operator?

Sming/SmingCore/Network/HttpRequest.cpp
```diff
@@ -82,77 +83,94 @@ String HttpRequest::getContentType()
...
 	} while(nextLine != -1);
 
 	if (headerEnd != -1)
 	{
+		tmpbuf = "";
 		debugf("parsed");
 		return eHPR_Successful;
 	}
+	headerDataProcessed += buf->tot_len;
+	tmpbuf = tmpbuf.substring(line, buf->tot_len);
```
PTDreamer: Shouldn't the size be "buf->tot_len - line" instead of "but->tot_len" ?

Sming/SmingCore/Network/HttpRequest.h
```diff
@@ -47,20 +47,21 @@ class HttpRequest
...
 	String method;
 	String path;
+	String tmpbuf;
 	HashMap<String, String> *requestHeaders;
 	HashMap<String, String> *requestGetParameters;
 	HashMap<String, String> *requestPostParameters;
 	HashMap<String, String> *cookies;
 	int postDataProcessed;
-	bool combinePostFrag;
+	int headerDataProcessed;
```
PTDreamer: headerDataProcessedLenght would be closer to the truth

Sming/SmingCore/Network/HttpRequest.cpp
```diff
@@ -165,61 +183,35 @@ HttpParseResult HttpRequest::parsePostData(HttpServer *server, pbuf* buf)
...
+		//debugf("item %i  - delim %i - next %i", startItem, delimItem, nextItem);
+		delimItem = buf.indexOf("=", startItem);
+		nextItem = buf.indexOf(delimChar, startItem);
+		if (nextItem == -1) nextItem = buf.indexOf(endChar, delimItem+1);
+		if (nextItem == -1) nextItem = endPos;
+		if (nextItem > endPos || delimItem == -1) nextItem = endPos; break;
 
-		String ItemName = NetUtils::pbufStrCopy(buf, startItem, delimItem - startItem);
-		String ItemValue = NetUtils::pbufStrCopy(buf, delimItem + 1, nextItem - delimItem - 1);
+		String ItemName = buf.substring(startItem, delimItem);
```
PTDreamer: Shouldn't it be buf.substring(startItem, delimItem - startitem) ?

Sming/SmingCore/Network/HttpRequest.cpp
```diff
@@ -165,61 +183,35 @@ HttpParseResult HttpRequest::parsePostData(HttpServer *server, pbuf* buf)
...
+		String ItemValue = buf.substring(delimItem+1, nextItem);
```
PTDreamer: Shouldn't it be buf.substring(delimItem + 1, nextItem - delimItem - 1) ?

patrickjahns: Short answer since I am on mobile. Substring uses absolute numbers
substring(Startposition, endposition)  while pbuf used (Startposition,
length). So this should be correct

PTDreamer: Wow this is confusing so this string class is not the same as defined in the c++ std namespace? std::string::substr arguments are to size_t's with the first being start index and the second being length.

PTDreamer: Found the definition here https://github.com/SmingHub/Sming/blob/develop/Sming/Wiring/WString.cpp#L586 and of course you are right. However even if not related to this PR I find this kind of subtleties counter intuitive and should be avoided by not using the same name for functions which exist in "standard c++".

patrickjahns: @PTDreamer 
This is related to SMING being compatible to Arduino framework - see their Substring implementation here: https://www.arduino.cc/en/Reference/StringSubstring

Sming/SmingCore/Network/HttpRequest.cpp
patrickjahns: In regards to  the size see the talk on substring below. But you caught something else - it should
`tmpbuf.substring(line, tmpbuf.length()` - the length of the tmpbuf can be longer than buf->tot_len

totally missed that

Sming/SmingCore/Network/HttpRequest.cpp
patrickjahns: I`ll verify it again

PTDreamer:
No comments regarding coding style since I found no reference to the style adopted by this project. Linux Kernel style is the most common for this kind of projects and this PR doesn't conform in several places.
I don't fully understand the idea behind processing data chunks as they come instead of just increment (append data to) a buffer  waiting for a complete header and process that. It would simplify code and increase readability (IMHO) without any performance impact.
PS: Don't know if my comments are welcomed or not since I haven't contributed with any code to this repo/project, but honestly I had nothing better to do :).

patrickjahns:
@PTDreamer 

See the Summary on the first page - there is the compatibility listed to arduino libraries ...

In regards to processing packets as they arrive (actually you can relate to this as a concept of processing a stream) - it is more memory friendly. While header data might be "small" (compared to post requests) - it requires more ram than just keeping unprocessed packet data in a buffer.

Before this PR there was no processing of headers split accross packets - (see #615 )
This is also meant as basis for processing multipart/post requests - a feature which I am working on in a different PR. While your idea of processing the header at once works for the header - it will not properly work for POST bodies - especially not for multipart requests - latest then I would need to implement the concept - and why not process both the same way? It is more intuitive than to process header and post requests differently

PTDreamer:
> See the Summary on the first page - there is the compatibility listed to arduino libraries ...
> Any further discussion in regard to this should rather be in a different ISSUE - to keep things related to this PR

Not sure if the comment is related to the coding style or discussion about the substring function.


> it is more memory friendly. While header data might be "small" (compared to post requests) - it requires more ram than just keeping unprocessed packet data in a buffer

It is not obvious to me how that is true. You are trading raw data buffer space with hash stored values which looking at the implementation bloats the amount of memory needed to store the data (which is true for all hash implementations).

> Before this PR there was no processing of headers split accross packets - (see #615 )

No argument there. It is obvious there was a bug, the data will need to be appended to a buffer between callback calls or it will get ditched.

> While your idea of processing the header at once works for the header - it will not properly work for POST bodies

Why not? You can just buffer the stream until you find "delimiter "--"" and parse it then.

In the end your code should work fine (although I still think you have a misplaced parenthesis), it is just a matter of personal coding style. Long do/whiles and lots of global variables are not my thing, also I found this PR hard to read not only by the code style but also by the variable naming.
Good job though, I hope I can find something to contribute with so you can take a turn at bashing my code.

patrickjahns:
> Not sure if the comment is related to the coding style or discussion about the substring function.

It was related to the discussion in regards to substring and compatibility with arduino libraries. And as suggested - i think this is best suited to be discussed in it`s own issue since it is not related to this PR

> It is not obvious to me how that is true. You are trading raw data buffer space with hash stored values which looking at the implementation bloats the amount of memory needed to store the data (which is true for all hash implementations).

You need to differentiate between the buffer and the Hashmap - the Hashmap implementation was already in use even before I suggested the buffer. When keeping everything in one buffer before parsing, you would need the space for the buffer and the hashmap at the same time. Parsing as far as it goes, the buffer data needed can be smaller and leave room for the final hashmap to be worked with. 

> Why not? You can just buffer the stream until you find "delimiter "--"" and parse it then.

Since multipart requests are not only variables/fields to be kept in memory, but also can be files - I don`t see your suggestion viable - it would need again 2 different implementations for one task. For suggestions/discussion on multipart and file uploads please join #552 

> In the end your code should work fine (although I still think you have a misplaced parenthesis), it is just a matter of personal coding style. Long do/whiles and lots of global variables are not my thing, also I found this PR hard to read not only by the code style but also by the variable naming.

Please compare the version before and after my PR - I based my solution on the already exiting coding style. In what way would you improve the current implementation ? 

> Good job though, I hope I can find something to contribute with so you can take a turn at bashing my code.

I hope you don`t see this discussion as a need for bashing code - as long as it improves code and usability in the end - it`s good to discuss it. And I am not too well suited for doing c/c++ code - I am more home with python. But since I am working on a project with an esp - I might as well help to improve it right? ;-)

hreintke:
@patrickjahns : 
>In the end your code should work fine (although I still think you have a misplaced parenthesis)
Is there still an update needed from your code.?

On In the PR now there is 
> 	Merge pull request #1 from SmingHub/develop … 
That contains (part of ?) another PR. 
I am not the most experienced in git so am not sure how that effects the merging.
Any change that you can update the PR for that ?

patrickjahns:
@hreintke 

I had trouble squashing - will try again later and update accordingly. Please wait with merging until then. I still want to verify the part with the misplaced parenthesis - didn`t get around to it until now

hreintke:
OK Clear

patrickjahns:
@PTDreamer 
Thanks for the hint with the line numbers - the +2 calculation was unnecessary and was a leftover from previous testing. Removed and now the PR is clean

@hreintke 
squashed and ready to merge - I can prepare another PR for RTOS then

patrickjahns:
@avr39-ripe 
Mentioned that query parameters are not parsed/parsed correctly - will need to have another look

avr39-ripe:
Nither queryParameter nor urlencoded form data DO NOT work ;(( switching to regular sming/develop solve the problem.. and introduce new one with packet hassle

patrickjahns:
@avr39-ripe 

pushed the proposed fix - please test and confirm that query and post parameters are working now as expected (please test only this branch and don`t include rawbody yet)

avr39-ripe:
cherry-pick last two commits on top of develop and it works. tested by using response.getQueryParameter()

patrickjahns:
I`d say this is ready to merge - maybe someone else can also confirm ?

avr39-ripe:
If @dobrishinov try this and confirm.. He expect problems with split/truncated data.. but it needs rawBody fix also..

PR #624, opened 2016-03-01
Added support for "TX only UART".

eadf:
In this mode the esp RX pin can be used as GPIO3.

I have just too many esp-1 modules, and i want to use them for something useful. 
So i simply need more than 2 gpios while still using UART0 for debugging.

I've successfully tested this with SDK 1.5.2. `pinMode(3, INPUT)` and `pinMode(3, OUTPUT)`.

Caution: If you use GPIO3 as output you better make sure that esp RX isn't wired directly to your serial adapter. Instead the esp RX pin should be connected to the serial adapter via 10KOhm or more.

~~Maybe it's better to create a brand new `Serial.begin()` method (e.g. `Serial.beginNoRx()`) instead of just adding a  `disableRx` parameter to it.
That way there will be fewer Arduino incompatibilities.~~ Did that already.

### DIFF
--- a/Sming/SmingCore/HardwareSerial.cpp
+++ b/Sming/SmingCore/HardwareSerial.cpp
@@ -33,8 +33,9 @@ HardwareSerial::HardwareSerial(const int uartPort)
 	system_os_task(delegateTask,USER_TASK_PRIO_0,serialQueue,SERIAL_QUEUE_LEN);
 }
 
-void HardwareSerial::begin(const uint32_t baud/* = 9600*/)
+void HardwareSerial::begin(const uint32_t baud/* = 9600*/, bool disableRx /*=false*/)
 {
+	rxDisabled = disableRx;
 	//TODO: Move to params!
 	UartDev.baut_rate = (UartBautRate)baud;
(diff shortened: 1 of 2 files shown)


alonewolfx2:
Actually you can try serial swap too. that time you can use gpio3 and gpio1 for output i think.

eadf:
From what i can tell, sming serial does not implement `swap()`like [Arduino Esp](https://github.com/esp8266/Arduino) do. It toggles GPIO13/GPIO15 or GPIO3/GPIO(1/2) as RX and TX.

alonewolfx2:
ups. I will add soon. Sorry for late adding

eadf:
No problem..

I had this thought the other day, one of my modifications looks like this:
```
 int HardwareSerial::read()
 {
	if (rxDisabled || available() == 0)
 		return -1; 
```
But is`rxDisabled`really necessary in this case? If the interrupt is disabled then`available()`should always be 0. Maybe we can optimize that a bit. (I have not tested this and I'm rebuilding my dev-board so I can't test it right now)

PR #625, opened 2016-03-02
Fix DS3232RTC lib non-default pins issue and pretify example app

avr39-ripe:
* Remove Wire.begin() from constructor which prevents to setup non-default pins
* Made example app output more pretty


alonewolfx2:
Tested. Working.

PR #631, opened 2016-03-04
Fix rawbody being empty

patrickjahns:
I discovered that the rawbody data might be empty, if header and body are not in the same packet/lwip buffer. 

It is based on the buffer for header (PR #622)

### DIFF
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -186,14 +186,30 @@ HttpParseResult HttpRequest::parsePostData(HttpServer *server, pbuf* buf)
 		tmpbuf = tmpbuf.substring(start, tmpbuf.length());
 	}
 
-	tmpbuf = extractParsingItemsList(tmpbuf, 0, tmpbuf.length(), '&', ' ', requestPostParameters);
+	//parse if it is FormUrlEncoded - otherwise keep in buffer
+	String contType = getContentType();
+	contType.toLowerCase();
+	if (contType.indexOf(ContentType::FormUrlEncoded) != -1)
+	{
+		tmpbuf = extractParsingItemsList(tmpbuf, 0, tmpbuf.length(), '&', ' ', requestPostParameters);
+	}
+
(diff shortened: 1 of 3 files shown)


patrickjahns:
One remark - is it wise to return String with `getBody()` ? Would it save memory when we return a String pointer instead?

hreintke:
This is the internal representation of String : 
```
  protected:
    char *buffer;	        // the actual char array
    uint16_t capacity;  // the array length minus one (for the '\0')
    uint16_t len;       // the String length (not counting the '\0')
```
S it does use pointer already. 
Other maybe to consider is that C/C++ people are used to working with char* when handling buffers. 
But String* doesn't seem necessary

patrickjahns:
So essentially when returning String it is not duplicated? That`s why I wonder if it wouldn`t be better returning a pointer to the String instead? 

(sorry it`s been a while since I worked in c/c++ - usually work in python)

patrickjahns:
I`ve synchronized this with the latest changes in buffering the http header.

avr39-ripe:
Check this patch on top of today develop - WORKS for me! I can still get queryParameters, I can get rawBody so I'm VERY glad. Thank you for your contribution! It will be great if someone who had problem with split packets confirm this to work.

hreintke:
@patrickjahns : 
I am a little lost in the discussions we had and the relationship between them.
Is this PR now ready to merge ?

patrickjahns:
@hreintke 

The discussion was about if the Method `getBody();` should return a pointer to the String object or the object itself (memory wise)

Otherwise it is ready for merge

avr39-ripe:
@hreintke is there something missing/undone to merge this PR? Thanks!

alonewolfx2:
@hreintke any suggestions? I will merge it

hreintke:
@patrickjahns @avr39-ripe @alonewolfx2 : 
Sorry, indeed has taken to long to react (have been busy last time).
And have been doubting about this. 
There is a copy when returning the String. however, it is not on the stack but on heap, and there is almost no usage of String* within Sming.
Let's keep it to String for now.

PR #633, opened 2016-03-04
New Feature: DNS Server - allows to create captive portals

patrickjahns:


alonewolfx2:
Thank you for effort.
Can you make this pr on rtos sming ?

patrickjahns:
Will do - is it okay if I first finish other PRs and then put them over all at once?

alonewolfx2:
no problem. i will test it on 2 different module tonight and let you know. (512kb and 4MB)

PTDreamer:
Code looks good. A nice enhancement would be to allow for the start function to receive an array of structures containing domain/ip pairs so it could handle more that one domain resolution.
Code has many similarities with other GPL'd code so keeping a mention to the original author on the header would be appropriate.

patrickjahns:
@PTDreamer 
Do you have a particular use case in mind in regards to providing more than one domain name or wildcard dns replies?

You are right about the license - actually checked again - thought it was MIT License - will provide credits to the library the code is based on

alonewolfx2:
Code working on smingrtos too. can you make pr for rtos? i will add when you provide credits to the library.

PTDreamer:
> Do you have a particular use case in mind in regards to providing more than one domain name or wildcard dns replies?

Not knowing if it is currently possible but in principle you can have several ip interfaces on the same device and a separate http server running on each. The dns server would then resolve those different domain - ip pairs. Off course there are other ways of achieving the same result.
I mostly suggested it because it seems like a straightforward addition.

patrickjahns:
Added credits - I hope this way is okay, otherwise I change it to a desired format. 

@alonewolfx2 
Will create PR against RTOS 

@PTDreamer 
I don`t require it currently - feel free to extend it

hreintke:
@patrickjahns : 
You should call the UdpConnection::onreceive otherwise the udp delegate is not triggered.

patrickjahns:
Sorry for asking but I don`t understand why I need to call `UdpConnection::onReceive` ?
Since this a specific service it won`t require a delegate - `DNSServer::onReceive` is derive from UdpConnection and thus superseeds it right? 

I don`t see this implemented in other Services (i.e. FTPServer Connection ) 

Will set the onReceive to virtual for derived classes

PTDreamer:
> Sorry for asking but I dont understand why I need to callUdpConnection::onReceive? Since this a specific service it wont require a delegate - DNSServer::onReceive is derive from UdpConnection and thus superseeds it right?

I think @hreintke is saying that you should call the base class function from inside your override (i.e. call UDPConnection::onReceive from your DNSServer::onReceive) passing it the same arguments received. This is to allow the callback (onDataCallback) the be called (triggered).
For this particular use case I don't really see the need for that or even making the function virtual.

patrickjahns:
@PTDreamer  @hreintke 

that was exactly my thought - but if wished I will change it accordingly. How shall I proceed ?

hreintke:
@patrickjahns @patrickjahns : 
I would like to have both updates included.

I try not to assume to much on what users do want to design but I see at least the possibility that an application can monitor/count DNS requests coming to the esp -> if to much you can close.

patrickjahns:
@hreintke 

done - it was not meant as disrespect - more of the idea that I wanted to understand the design decision to learn more about the framework itself and the way things are handled/wanted

hreintke:
@patrickjahns : 
Didn't feel as disrepect. Was a little short on my first answer. Will try to improve :)

PR #637, opened 2016-03-06
use COM_SPEED_SERIAL from Makefile instead of SERIAL_BAUD_RATE

zendes:
the Makefiles define COM_SPEED_SERIAL and use this to start the
terminal, but this value isn't passed to the code. inside user_config.h
there is SERIAL_BAUD_RATE that configures the port speed. this means
it's left up to the user to keep the two in sync, which is less than
ideal. this commit passes COM_SPEED_SERIAL to gcc as a define, and
changes all Serial.begin lines to use this instead of SERIAL_BAUD_RATE.

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -22,15 +22,15 @@
 # MacOS / Linux
 # SMING_HOME = /opt/esp-open-sdk
 
-## COM port parameter is reqruied to flash firmware correctly.
+## Serial port parameter is required to flash firmware correctly.
 ## Windows: 
-# COM_PORT = COM3
+# SERIAL_PORT = COM3
 
 # MacOS / Linux:
-# COM_PORT = /dev/tty.usbserial
+# SERIAL_PORT = /dev/tty.usbserial
 
-# Com port speed
-COM_SPEED = 115200
+# Serial port baudrate
+SERIAL_BAUDRATE = 115200
 
 
 ifeq ($(OS),Windows_NT)
@@ -111,7 +111,7 @@ USER_LIBDIR = compiler/lib
 LIBS		= microc microgcc hal phy pp net80211 lwip wpa main
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
-CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL)
+CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DSERIAL_BAUDRATE_APP=$(SERIAL_BAUDRATE_APP)
 ifeq ($(ENABLE_GDB), 1)
 	CFLAGS += -Og -ggdb -DGDBSTUB_FREERTOS=0 -DENABLE_GDB=1
 	MODULES		 += gdbstub
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -111,7 +111,7 @@ USER_LIBDIR = compiler/lib
 LIBS		= microc microgcc hal phy pp net80211 lwip wpa main
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
-CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106
+CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL)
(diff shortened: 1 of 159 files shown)


eadf:
Wouldn't it have been better to rename COM_SPEED_SERIAL in the Makefile to SERIAL_BAUD_RATE?
COM port is such a windows centric name, you'll never find a com port on a *nix platform.

zendes:
i'm happy to rename them, i just picked the makefile as the one to keep and changed the samples

how about this in Makefile-project.mk:

## Serial port parameters
# Default serial port speed (generic)
SERIAL_SPEED ?= 115200

# Default COM port speed (used for flashing)
SERIAL_SPEED_ESPTOOL ?= $(SERIAL_SPEED)

SERIAL_SPEED_APP  ?= $(SERIAL_SPEED)

and then in all the Makefile-<platform>.mk files
## COM port parameter is reqruied to flash firmware correctly.
## Windows: 
# SERIAL_PORT = COM3

# MacOS / Linux:
# SERIAL_PORT = /dev/tty.usbserial

basically changing any COM_ variable into a SERIAL_ one

hreintke:
@zendes @eadf : 
Still have to read the PR comments careful but already this remark upfront.

The current usage of the serial parameter in the makefile is only used for the speed of the esptool..
That is necessary to be able to flash the esp correctly.

For all "user defines" application developers can add them to the makefile-user.mk and/or in the a user_config.h. That is also the location of the "example define" wiithin the sample appliactions/

Currently I don't think we should add these user items within the generic/framework wide makefiles.

zendes:
Not exactly. The COM_SPEED is also passed to the terminal that gets started after flashing, via COM_SPEED_SERIAL. That value is however not available inside the code, so you have to keep the makefile & the code in sync to keep your terminal working. This flattens the two different values into one (in the makefile), and passes that to the code via a -D on gcc.

eadf:
Ok, to recap: This PR addresses three issues:

- Where should the macro (#define) that sets the default ``Serial.begin()`` speed be defined?
  @zendes suggests it should be defined in the Makefiles instead of ``user_config.h``.

- What should the macro be called? 
   I think the most sensible thing is to keep the old name : SERIAL_BAUD_RATE

- What should the Makefile variable that govern the command-line-terminal serial speed be called?
  ``TERMINAL     ?= python -m serial.tools.miniterm $(COM_PORT) $(COM_SPEED_SERIAL)``
  COM_SPEED_SERIAL is also, by default, defining the esptool flash speed.
  ``COM_SPEED_ESPTOOL ?= $(COM_SPEED)``
  This is a bit unfortunate because they have nothing to do with each other, but it's just the default setting, easily overridden in makefile-user.mk.
 Ideally the Makefile variable should be called the same thing as the macro, and those COM references are an eyesore in a multiplatform project. But it might be a little to late in the game to change that now...

my 2 cents

zendes:
I'm actually against keeping the old SERIAL_BAUD_RATE name, because if someone updates Sming and tried to build a project that they based off of an old example, that example will redefine SERIAL_BAUD_RATE and cause an error for that user (without any real message about what they've done wrong). I think it's best to pick new clean names, and update the examples so that both new & old versions keep working.

hreintke:
@zardam : 
It is an absolute NOGO for updating SERIAL_BAUD_RATE if that breaks current applications.
With this we need to keep full backward compatibility.

zendes:
I've pushed a commit that renames the variables from COM_ to SERIAL_, and updates the descriptions as well. This should make more sense for folks running on *nix platforms.

hreintke:
@zendes : 
This commit breaks backward compatibility.

Let's agree what and how we will update and then create the PR.

zendes:
@hreintke can you explain how this break backwards compatibility? What stops working if you apply this commit right now?

Sming/Makefile
```diff
@@ -111,7 +111,7 @@ USER_LIBDIR = compiler/lib
 LIBS		= microc microgcc hal phy pp net80211 lwip wpa main
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
-CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106
+CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DSERIAL_BAUDRATE_APP=$(SERIAL_BAUDRATE_APP)
```
ADiea: Hi In what file is SERIAL_BAUDRATE_APP defined?

hreintke:
@zendes : 
Did I read the PR incorrect ?
f.e. is the env variable COM_PORT removed so everyone needs to update their environment ?

zendes:
yeah, you're right. let's first get to a point where everyone agress on how & what, and then i'll update the PR

right now I'm in favor of the first PR I sent, and to do the rename (if we agree that's the right thing to do) in a second one. the rename probably broke more things than it improved.

why I made the first PR was because I didn't like having to define the same value twice, and then keep them in sync. example: if you want to see the bootloader messages and you're starting from one of the examples, this is roughly how things currently go:
1. edit your makefile to set COM_SPEED (or COM_SPEED_SERIAL) to 74880. this will be the baud rate of the terminal program that's started and you'll see the esp bootloader messages
2. but then you won't see your app's serial.println messages anymore, cause the baudrate at which you started the serial port didn't change. so then you edit your include file & update the SERIAL_BAUD_RATE there. currently, you have no way of keeping both values in sync so you'll end up making the same change twice. 

This (first) PR makes the COM_SPEED_SERIAL value available inside the code, and updates the examples to use it by default. If you choose to create a new define and use that to initialize your serial port, nothing bad happens & things works as expected. If you based your app off of a previous example (that had SERIAL_BAUD_RATE), nothing breaks and things will still work as they did before this PR.

PR #639, opened 2016-03-06
remove trailing whitespace from all sourcecode files

zendes:
the codebase has lots of trailing whitespace characters
afaik it's generally considered a good idea to configure your editor to
strip these out when saving files, as they can sometimes cause difficult
to spot problems (not all editors visualize them, which is probably why
they appeared in the first place). this cleanup commit does nothing
besides strip those trailing whitespace characters.


raburton:
I agree in principal, that we shouldn't have trailing spaces, however this isn't the way to do it. It'll just cause a lot of pain for anyone trying to merge existing chnages (this PR is already has conflicts), and the spaces aren't actually causing any problems to warrant this method. You've also updated source files that are taken from upstream projects and are currently untouched in our repo, making their situation less clear (and changes will be reversed next time they are updated).

ADiea:
i also vote against, it's just too many files in the change to worth it, don't see the benefit...

PR #640, opened 2016-03-07
add make terminal option

zendes:
sometimes it's useful to be able to start the serial port monitor
without having to remember what com port or baud rate you've configured
for this project. this adds a simple makefile target that just kills &
starts the terminal in the same way make flash does, without flashing or
resetting the mcu.

### DIFF
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -344,6 +344,11 @@ else
 endif
 	$(TERMINAL)
 
+terminal:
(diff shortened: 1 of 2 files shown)


zendes:
Ping. What do I need to do to get some feedback on this? I'm happy to make adjustments if needed, but I haven't heard anything for a while.

zendes:
ok, so deleting a branch on my side auto-closes the pull request. that wasn't intentional...

PR #641, opened 2016-03-07
fixing autocomplete on some libraray for netbeans 

blinkingnoise:
Netbeans has a problem with the short declaration of methods like: 
`
void a(int i),b(int v, int x),c();
`

so i changed it to the more readable declaration:
```
void a(int i);
void b(int v, int x);
void c();
```


hreintke:
@blinkingnoise : 
This updates code which is not "Sming" but libraries taken from other locations.

blinkingnoise:
@hreintke: yes i know.  but the source is in this repo.

hreintke:
@blinkingnoise : 
I know the source is in this repo but for these kind of libraries I want to keep the number of updates to a minimum.

Issue #628, opened 2016-03-03
MQTT sending ping


flexiti:
@Informatic 
Yours last change:
> void MqttClient::onReadyToSendData(TcpConnectionEvent sourceEvent)
>   {
>  -	-if (lastMessage and millis() - lastMessage >= 10000)
>  +	// Send PINGREQ every keepAlive time, if there is no outgoing traffic
>  +	+**if (lastMessage && (millis() - lastMessage >= keepAlive*1000))**
>   	{
>   		mqtt_ping(&broker);
>   	}

keepAlive this is the time  that is sent  to server, 
Based on this time server finds that has lost  connection with the client (client dissconeted) and this this particular time is in seconds.
Setting the same time on the server and client is hazardous.
Therefore, it should be here another time eg. new PingRepeatTime  and should be at least 2 times less down then keepAlive.
I'm not sure but it also should be less than 70 sec ( see code below)

> TcpConnection::TcpConnection(bool autoDestruct) : autoSelfDestruct(autoDestruct), sleep(0), canSend(true), **timeOut(70)**
> {
> 	initialize(tcp_new());
> }
>

Informatic:
As far as i remember MQTT specification is rather unclear on the keep alive behaviour. This is how keep alive is implemented in Paho, the most popular implementation of MQTT in Python.

flexiti:
Whatever it is,setting the same time on the server and client is hazardous.
 -  broker 10s is a litle bit shorter then client - broker sent unnesesery Last Will
 - the lack of a single ping disconnect the connection, the effect as above
You should correct it.

flexiti:
> If the Keep Alive value is non-zero and the Server does not receive a Control Packet from the Client within one and a half times the Keep Alive time period, it MUST disconnect the Network Connection to the Client as if the network had failed [MQTT-3.1.2-24].

Informatic:
...one and a half. Sending PING once every keepAlive period is fine, IMO. Actually, in my first commit I used half a period of keepAlive, but Sming maintainer opposed against that ;)
https://github.com/SmingHub/Sming/pull/550#issuecomment-171638996

flexiti:
Ok, up to You...I will close this issue.

I withdrew this change from my app, since on average every one-two hours I was disconnect and received LWT. (Mosqquito and WiFi clients -easy temporarily lose connection)
Disconnection is not yet a problem - will resume , but the receipt LWT already so :)

I really don't know why risk losing connections when only one ping is not received.
[...]

PR #653, opened 2016-03-14
MQTT ping sending modification (setPingRepeatTime added)

flexiti:
in accordance with the #628  
This solution lets you freely modify the Mqtt client ping frequency regardless of the broker keepAlive.
Default: PingRepeatTime (used on client side) = 20s, KeepAlive (used on broker side) = 60s
It can be changed by:
setKeepAlive(seconds)  and setPingRepeatTime(seconds) metod.
PingRepeatTime should be less than or equal to keepAlive.


hreintke:
@flexiti : 
Can you create the PR for RTOS too ?

PR #642, opened 2016-03-09
New SPI impl. with Arduino style Api - BaseClass, SwSPI, HwSPI  -> Testcase SD Card

harry-boe:
Ok this is a bigger one

New SPI implementation 100% compliant with Arduino HW SPI API.
This eases porting Arduino libs to sming -> in most cases 1:1 

The API is defined by the SPIBase interface, so it can be changed from SW SPI to HW SPI without code mods. See SDCard sample app.

```
//      Software
//	SDCardSPI = new SPISoft(PIN_CARD_DO, PIN_CARD_DI, PIN_CARD_CK, 0);

//      Hardware SPI
	SDCardSPI = new SPIClass();

//      Common
	SDCard_begin(PIN_CARD_SS);

// overwrite default SPI Speed (ignored by SW SPI)
	SDCardSPI->beginTransaction(SPISettings(40000000, MSBFIRST, SPI_MODE0));

```

The API required a new Software SPI implementation. Have teste what  have on Hardware but it requires more test
  
Successful test:  SDCard (The base of the old imply.) ScreentTFT_ST7735, OLED_SSD1306

New HW SPI implementation tested with single byte TX und long buffers (SDCard / ArduCam)
Also tested various frequencies, byte-orders and SPI modes

Includes documentation tags (not sure if they work)

Here are some Benchmarks SW vs HW SPI from the SDCard impl.

```
Software  SPI  SD-Card
4. Write speed benchmark:

Write 1 kBytes in 1 Bytes increment: 17.23 kB/s
Write 1 kBytes in 64 Bytes increment: 17.29 kB/s
Write 1 kBytes in 128 Bytes increment: 17.36 kB/s
Write 1 kBytes in 512 Bytes increment: 17.39 kB/s
Write 1 kBytes in 1024 Bytes increment: 17.90 kB/s
Write 4 kBytes in 1024 Bytes increment: 50.78 kB/s
Write 8 kBytes in 512 Bytes increment: 65.03 kB/s
Write 8 kBytes in 1024 Bytes increment: 73.12 kB/s
Write 8 kBytes in 8192 Bytes increment: 86.05 kB/s

Hw SPI SD-Card

4. Write speed benchmark:

Write 1 kBytes in 1 Bytes increment: 34.26 kB/s
Write 1 kBytes in 64 Bytes increment: 35.45 kB/s
Write 1 kBytes in 128 Bytes increment: 35.33 kB/s
Write 1 kBytes in 512 Bytes increment: 35.46 kB/s
[...]

PR #654, opened 2016-03-14
- initial ArduCAM library and sample replaces #646

harry-boe:
Basic implementation for the ArduCAM, 2MP type OV2640

Sample provides web page to capture JPG images from the camera using the browser. 
The image size can be set using a command handler at runtime by console or telnet

Note: PR replaces #646 to cleanup dependencies after merging #642


hreintke:
@harry-boe : 
Clening issues/PR's
Is this still a valid PR or is it obsolete after you work afterwards ?

alonewolfx2:
@hreintke its still valid :)

harry-boe:
sorry for the late response ..

Yes the PR is still fine. It's based on the new SPI implementation and works fine for GIF image capturing for all supported sizes up to 2MP.
It comes with a nice web ui to capture and display the images.
We used that code already for a training workshop with +20 participants/devices ai it's well tested;)

BTW: The web page uses a css and javascript library that is optimised for size and does not need to load code from a CDN -> might be an option for other build in web sites as well.
Videos are not yet supported -> this would requires a InputStream that creates a multipart response basically adding picture by picture. 
I will add that once i have some time left.

Issues: The BMP mode deliver some strange results wit the picture up-side-down and some weird colours. Maybe a byte order issue - but anyhow JPG's are all fine.

alonewolfx2:
is #642 still necessary?

harry-boe:
#642 got merged already that was the new SPI implementation.

PR #661, opened 2016-03-21
Tests malloc/realloc return values in MemoryDataStream->write

festlv:
malloc/realloc function call return values were not checked and used as buffers which could lead to NULL pointer de-referencing crashes.

Additionally, TcpClient->send now fails when there was no free space in underlying memory stream.

### DIFF
--- a/Sming/SmingCore/DataSourceStream.cpp
+++ b/Sming/SmingCore/DataSourceStream.cpp
@@ -50,10 +50,15 @@ size_t MemoryDataStream::write(const uint8_t* data, size_t len)
 		{
 			capacity = required < 256 ? required + 128 : required + 64;
 			debugf("realloc %d -> %d", size, capacity);
-			buf = (char*)realloc(buf, capacity);
+			char * new_buf;
+			//realloc can fail, store the result in temporary pointer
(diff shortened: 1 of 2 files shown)


hreintke:
@festlv : 
Good catch.
Update to memorydatastream is OK.
On TCPClient. : 
In the current implementation the side effect is that when closing the connection it will be detected as "not completely send data" because of the asyncTotalLen += len; .
After your update this is not updated -> asyncTotalLen == asyncTotalSent -> closed with "all OK'

Maybe should close the connection when data cannot be stored. The principle of TCP connection is a guaranteed transmission of data. Now we leave part of the data out.

festlv:
I actually realized that MemoryDataStream update will leak memory if initial malloc succeeds but realloc fails- buf is NULL and won't be freed. Will update the PR later.

Regarding TcpClient- I see your point. However, my update does not break this guarantee- the data which has been accepted for sending is sent. If TcpClient fails to send data (send returns false), application layer knows it right away and can act accordingly. Otherwise, the current implementation will discard data and application will only learn about this later.

hreintke:
Can you make a PR for SmingRTOS too ?

PR #663, opened 2016-03-22
MQTT Better memory usage.

slaff:
Removed limits for clientid, username and password.

### DIFF
--- a/Sming/Services/libemqtt/libemqtt.c
+++ b/Sming/Services/libemqtt/libemqtt.c
@@ -184,29 +184,54 @@ uint16_t mqtt_parse_pub_msg_ptr(const uint8_t* buf, const uint8_t **msg_ptr) {
 	return len;
 }
 
-void mqtt_init(mqtt_broker_handle_t* broker, const char* clientid) {
+void mqtt_init(mqtt_broker_handle_t* broker) {
 	// Connection options
 	broker->alive = 300; // 300 seconds = 5 minutes
 	broker->seq = 1; // Sequence for message identifiers
-	// Client options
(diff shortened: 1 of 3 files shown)


hreintke:
I see this a a good improvement of the MQTT but do not know the details of the implementation.
Can anyone who uses MQTT comment on correctness and backward compatibility ?

slaff:
The change above allows for longer clientid, username and password so all code that worked up to now with shorter ids should continue to work without a problem. This change also matches better v3.1.1 ( http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718016 ) of MQTT allowing clientid, user name and password to be utf8 encoded string containing 0 to 65535 bytes.

hreintke:
@slaff : 
Can you also PR for RTOS ?

slaff:
@hreitke Yep,  I will add it till the end of the weekend.

marvinroger:
@slaff I am pretty new to embedded programming, and it's my understanding that we should avoid allocating in the heap (malloc, new, strdup...), but prefer stack allocation. So yes, this PR removes length limits, but does it really optimize memory in an embedded point of view?

slaff:
> So yes, this PR removes length limits, but does it really optimize memory in an embedded point of view?

@marvinroger The previous code was allocating 26 bytes for username and password, even though you may not use them at all and also 50 bytes for client id, no matter if the client id is shorted than that. The minimum requirements for the new code are only 6 bytes ( default "emqtt\0" client id. ). Which is 70 bytes less than the previous code.

PR #668, opened 2016-03-23
Refactor WifiStation.config, remove (dis)connect which is harmful here

avr39-ripe:
- Remove wifi_station_disconnect and wifi_station_connect methods which
  cause module to WDT reset if change WIFI Station configuration while
  connected to some AP
  - Conditionally enable STA mode ONLY if it was disabled before call to
    WifiStation.config
  - Remove dhcp client disable, no needed here
  - Conditionally disable dhcp client and/or STA mode itself if they WERE
    disabled before calling to WifiStation.config


slaff:
@avr39-ripe Is this PR still relevant?

avr39-ripe:
yes, I still use it in my projects. May be @anakod can review it? or anybody else review and/or test it?. Again, I still use this patch for my production projects.

slaff:
@anakod Can you review this PR and tell us what you think?

PR #678, opened 2016-04-03
Return content-length with HttpResponse

patrickjahns:
- allow for setting custom response codes / messages 
  The respond messages provided with sming are limited to a small amount, this allows for setting a custom / non-provided response code and makes the webserver more versatile
- return Content-Length header field for files, json and memory messages  
  It is recommended practice for http-server to also return a content-length as response

Unfortunatly I haven`t found a way to create a buffer-length with the template method - any hints or ideas and I will try to also include this

### DIFF
--- a/Sming/SmingCore/Network/HttpResponse.cpp
+++ b/Sming/SmingCore/Network/HttpResponse.cpp
@@ -64,6 +64,17 @@ int HttpResponse::getStatusCode()
 	return status.substring(0, p).toInt();
 }
 
+void HttpResponse::setStatus(String statusline) {
+	status = statusline;
+}
+
+void HttpResponse::setStatus(int code, String msg /* = "" */)
+{
+	status = String(code) + " " + msg;
+}
+
+
+
 bool HttpResponse::hasBody()
 {
 	return stream != NULL || bodySent;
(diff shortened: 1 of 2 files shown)


hreintke:
@patrickjahns : 
With SendString you cannot set the content-length in the call. There can be multiple "sendstrings" which will be combined in the memorydatastream.

With Sendfile we duplicate the sendHeader, should we make that one ? 

Generic : Just a thought, would need some further checking.
Could we add the header after the application callback has finished ?
See http_server::ProcessRequest, around line 100. I think then the stream is completed contains the full response(not sure about the template) and we can "just take the size of the stream.
Probably we need to add a stream::size() method but that is an easy update.

patrickjahns:
@hreintke 

I didn`t know that we can call multiple sendstring instances and they will concatenated - will need to give this some more thoughts.

What do you mean by: 

> With Sendfile we duplicate the sendHeader, should we make that one ?

Did you mean `setHeader`?


I am not to familiar with the stream objects - so I for clarification: is the length of the stream generally known before the first (body) packet is sent? 
I am not sure if `HttpServer::ProcessRequest` is a good place to intercept - I had a look and would suggest then `HttpResponse::sendHeader` for a more general approach? 

The Content-Length needs to be returned in the Header and the latest we can set it, would be just before/while we sent the header? 

Since this turns out to be a bit more work for Content-Length, should I split the PR for custom statuscodes and Content-Length?

slaff:
That will be part of the refactored Http stack. For more information and progress see https://github.com/slaff/Sming/issues/5.

PR #679, opened 2016-04-04
fix ArduinoJson overloading String define

patrickjahns:
Eclipse would show errors, that String is ambiguous. 
This was a result of ArduinoJson providing a second definition for String.

Adding a check for SMING_VERSION and ARDUINO resolves this problem.

I am not sure if defining ARDUINO would provide a better alternative to having to include this modified check with every arduinojson version?

### DIFF
--- a/Sming/Services/ArduinoJson/include/ArduinoJson/Arduino/String.hpp
+++ b/Sming/Services/ArduinoJson/include/ArduinoJson/Arduino/String.hpp
@@ -7,7 +7,7 @@
 
 #pragma once
 
-#ifndef ARDUINO
+#if !defined(ARDUINO) && !defined(SMING_VERSION)
 
 #include <string>
 typedef std::string String;


patrickjahns:
@hreintke 

any thoughts on this? would fix #683

cvgaviao:
this fix have worked in my environment (y)

Bomb3rman:
I am confused. Why was this closed? This pull request has not been merged and it clearly fixes the problem.

hreintke:
@Bomb3rman : 
The issue must me environment specific as only a small number of people are suffering for that.
-> should be solved by that.

the patch is to an external repo's and quite fundamental one.
-> Principle is to keep the number of patches to external to a minimum

patrickjahns:
I agree on the part that this is not a issue that leads to compilation
problems. It's rather a patch to ease the use of Sming.

I would differentiate the part with the modification of the upstream
sources. We utilize and package them together with Sming and for me it is
common to have patched upstream sources specific to projects when they are
shipped together. Any update from upstream may or may not include breaking
changes and thus would also require a patch
On Jun 2, 2016 12:25, "hreintke" <notifications@github.com> wrote:

@Bomb3rman <https://github.com/Bomb3rman> :
The issue must me environment specific as only a small number of people are
suffering for that.
-> should be solved by that.

the patch is to an external repo's and quite fundamental one.
-> Principle is to keep the number of patches to external to a minimum

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/SmingHub/Sming/pull/679#issuecomment-223253615>, or mute
the thread
<https://github.com/notifications/unsubscribe/AGG0q2Sxyp_nGdW61pj6M1Xu5jT7-89Eks5qHq-bgaJpZM4H-ut2>
.

Bomb3rman:
I was mostly confused why this pull request and the corresponding PR were closed without any explanation. Mostly because the problem is not solved. The issue might be environment specific, but that does not mean that it should not be solved in the framework. My setup is not crazy, I basically followed the setup steps on the wiki for OSX and Eclipse and I am hitting that problem so other people must be seeing the same.

> the patch is to an external repo's and quite fundamental one.
The patch is to a codebase of a external repo which was integrated into the sming repo. Imho it is perfectly reasonable to make such changes if necessary. I disagree that it is a fundamental change. The biggest concern should be that changes from upstream can be merged in easily. And I don't think this one-liner creates a problem there.

slaff:
@patrickjahns Merged in https://github.com/SmingHub/Sming/pull/790. Patrick, can you provide a SmingRTOS PR for that issue ?

PR #681, opened 2016-04-04
make header-fields case-insensitive

### LINKED ISSUE #672: Bug: The header field name is case-insensitive but is not processed as such.
There is a bug in the file: "`Sming/Sming/SmingCore/Network/HttpRequest.cpp`".

The header field name is case-insensitive according to:
  https://tools.ietf.org/html/rfc7230
    ...
    _**3.2.  Header Fields**
       Each header field consists of a case-insensitive field name followed
       by a colon (":"), optional leading whitespace, the field value, and
       optional trailing whitespace._
    ...

So in: "`Sming/Sming/SmingCore/Network/HttpRequest.cpp`"
  the function:
    `String HttpRequest::getHeader(String headerName, String defaultValue /* = "" */)`
  must do a case-insensitive compare.

Example: It should accept `Content-Type:` but also `Content-type:` (lowercase t in -type).

patrickjahns:
In order to comply with rfc7230 headerfields are now parsed case-insensitive and retrieved case-insensitive

fixes #672

### DIFF
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -53,6 +53,7 @@ String HttpRequest::getPostParameter(String parameterName, String defaultValue /
 
 String HttpRequest::getHeader(String headerName, String defaultValue /* = "" */)
 {
+	headerName.toLowerCase();
 	if (requestHeaders && requestHeaders->contains(headerName))
 		return (*requestHeaders)[headerName];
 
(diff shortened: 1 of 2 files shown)


hreintke:
@patrickjahns : 
I agree with the update/patch. 
Formally there is  backward compatibility issue. Applications which rely on the (incorrect) case sensitivity will not work anymore. 
What is your opinion on this ?

patrickjahns:
@hreintke 
To be honest I didn't think of backward compatibility issues. 
Generally any existing code would still work as before, no matter the spelling (i.e. `getHeader("Content-Length");` `getHeader("content-length");` `getHeader("cOnTeNt-LeNgTh");` will all return the same. 

While header names are case-insensitive, the values of the header fields are case-sensitive and thus nothing has changed from the previous behavior. 

Is there any planned release notes for the upcoming Sming release?

PR #690, opened 2016-04-10
Feature/http multipart post (file upload to sming webserver)

patrickjahns:
This is the first "fully" working implementaiton for file-uploading/http-multipart-post for http server. (#552)

Currently it provides functionality to:
- upload single files 
- upload multiple files at once
- upload files with additional vars

To enable uploading for a certain path, a delegate has to be provided:
```void addPath(String path, HttpPathDelegate callback, HttpUploadDelegate uploadcallback);```

The uploadcallback function is defined as following:
```
void functionName(HttpRequest&, HttpUpload&)
```
The function has thus access to the current request in the sense of all Header and Query Paramters (as well as path) can be accessed to decide if we accept the upload (or not)

In the pathdelegate information about the uploaded files is accessible by calling:
```
if(request.hasUpload()) {
	Vector<HttpUpload*> uploads = request.getUploads();
}
```

HttpUpload is defined as
```
struct HttpUpload {
	String filename;
	String name;
	String type;
	size_t totalSize;
	size_t curSize; // size of the current data block
	char* bufferdata; //pointer to the data
	file_t file;
	HttpUploadStatus status;
};
```
after a successfull upload `curSize` and `bufferdata` will be NULL and not accessible anymore (to avoid tampering with it
`type` contains the content-type field values
`name` is the variable name defined in the form field

If we receive a http-multipart-request without a filename or if the filename is empty - the data will be treated as normal http-post parameters and is accessible via `getPostParameter` - the limit `NETWORK_MAX_HTTP_PARSING_LEN` applies to all data that is not a file.

For testing see the working examples in `samples/HttpServer_Upload`

Please test it thoroughly and provide feedback. If you encounter any issues, it would be great if you document what has lead to the issue and provide a debug log.

Do you have any suggestion/remarks? 

**Known limitations/issues:**
- when providing a filename but no content - a file with the filename but 0 size will be created  
    (it seems that chrome sends the header again when one presses go-back to the fileupload page)  
- max lenght for filenames is 31chars (current spiffs setting)  
    how should we handle this use case? how to inform the user - automatically or should the "user of the framework" take care of this in the callback?

### DIFF
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -9,6 +9,7 @@
 #include "HttpServer.h"
 #include "NetUtils.h"
 #include <stdlib.h>
+#include "../SmingCore.h"
 #include "../../Services/WebHelpers/escape.h"
 
 HttpRequest::HttpRequest()
@@ -19,7 +20,9 @@ HttpRequest::HttpRequest()
 	cookies = NULL;
 	headerDataProcessed = 0;
 	postDataProcessed = 0;
-	bodyBuf = NULL;
+	postParamSize = 0;
+	multipart = NULL;
+	uploads = NULL;
 	tmpbuf = "";
 }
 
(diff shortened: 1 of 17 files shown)


patrickjahns:
For discussions of the feature please provide feedback/thoughts in #694

Matheus-Garbelini:
Hello, i can't compile your example with the branch you supplied. Could you try to update sming to the latest version, so we can compile your example with the updated Sming? Or could i just replace some Sming core files to match the modifications you did? if yes, wich files? Thanks

Issue #694, opened 2016-04-12
Feature: Http Multipart (File upload)


patrickjahns:
In the branch ["feature/http-multipart-post"](https://github.com/SmingHub/Sming/tree/Feature_HttpPost) I am currently extending the webserver to include multipart requests (file uploading). 

Please help to test the feature/branch and provide feedback in this issue

<hr>

Copy of the Initial PR:

Currently it provides functionality to:
- upload single files 
- upload multiple files at once
- upload files with additional vars

To enable uploading for a certain path, a delegate has to be provided:
`void addPath(String path, HttpPathDelegate callback, HttpUploadDelegate uploadcallback);`

The uploadcallback function is defined as following:

```
void functionName(HttpRequest&, HttpUpload&)
```

The function has thus access to the current request in the sense of all Header and Query Paramters (as well as path) can be accessed to decide if we accept the upload (or not)

In the pathdelegate information about the uploaded files is accessible by calling:

```
if(request.hasUpload()) {
    Vector<HttpUpload*> uploads = request.getUploads();
}
```

HttpUpload is defined as

```
struct HttpUpload {
    String filename;
    String name;
    String type;
    size_t totalSize;
    size_t curSize; // size of the current data block
    char* bufferdata; //pointer to the data
    file_t file;
    HttpUploadStatus status;
};
```

after a successfull upload `curSize` and `bufferdata` will be NULL and not accessible anymore (to avoid tampering with it
`type` contains the content-type field values
`name` is the variable name defined in the form field

If we receive a http-multipart-request without a filename or if the filename is empty - the data will be treated as normal http-post parameters and is accessible via `getPostParameter` - the limit `NETWORK_MAX_HTTP_PARSING_LEN` applies to all data that is not a file.

For testing see the working examples in `samples/HttpServer_Upload`

Please test it thoroughly and provide feedback. If you encounter any issues, it would be great if you document what has lead to the issue and provide a debug log.

Do you have any suggestion/remarks? 

**Known limitations/issues:**
- when providing a filename but no content - a file with the filename but 0 size will be created  
  (it seems that chrome sends the header again when one presses go-back to the fileupload page)  
- max lenght for filenames is 31chars (current spiffs setting)  
  how should we handle this use case? how to inform the user - automatically or should the "user of the framework" take care of this in the callback?

hreintke:
For all upload requests I get a Bad Request reply.
Error log :
```
path=/upload
host === 10.0.0.210
parsed
[...]

PR #697, opened 2016-04-13
Feature/http multipart post

patrickjahns:
Some fixes for issues noticed in #694

### DIFF
--- a/Sming/SmingCore/Network/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/HttpRequest.cpp
@@ -103,15 +103,24 @@ String HttpRequest::getContentType()
 
 HttpParseResult HttpRequest::parseHeader(HttpServer *server, pbuf* buf)
 {
-	int headerEnd = NetUtils::pbufFindStr(buf, "\r\n\r\n");
+	int urlEnd, headerEnd;
+	 headerEnd = NetUtils::pbufFindStr(buf, "\r\n\r\n");
 	if (headerEnd > NETWORK_MAX_HTTP_PARSING_LEN || headerDataProcessed > NETWORK_MAX_HTTP_PARSING_LEN \
 		|| (headerEnd != -1 && buf->tot_len > NETWORK_MAX_HTTP_PARSING_LEN))
 	{
 		debugf("NETWORK_MAX_HTTP_PARSING_LEN");
 		return eHPR_Failed_Header_Too_Large;
 	}
-	int urlEnd = 0;
-	tmpbuf += NetUtils::pbufStrCopy(buf, 0, buf->tot_len);
+
+	if(headerEnd == -1)
+	{
+		tmpbuf += NetUtils::pbufStrCopy(buf, 0, buf->tot_len);
+	}
+	else
+	{
+		tmpbuf += NetUtils::pbufStrCopy(buf, 0, headerEnd+4);
+	}
+	urlEnd = 0;
 	if (requestHeaders == NULL) {
 		// first time calling header
 		requestHeaders = new HashMap<String, String>();
@@ -196,7 +205,7 @@ HttpParseResult HttpRequest::parseHeader(HttpServer *server, pbuf* buf)
 			}
 
 		}
-		if (nextLine != -1) {
+		if (nextLine != -1 ) {
 			line = nextLine + 2;
 		}
 
@@ -263,6 +272,17 @@ HttpParseResult HttpRequest::parsePostData(HttpServer *server, pbuf* buf)
 	}
 }
 
+void printDebug(String s) {
+	int l = s.length();
+	int handled = 0;
+	while(handled < l) {
+		String part = s.substring(handled, handled+100);
+		Serial.printf("%s", part.c_str());
+		handled += 100;
+	}
+	Serial.printf("\r\n");
+}
+
 HttpParseResult HttpRequest::parseMultipartPostData(HttpServer *server, pbuf* buf)
 {
 	int start = 0;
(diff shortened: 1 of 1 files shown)


robotiko:
Tested all flows: 
it just seems to fail in extreme cases: 
-If file to upload is bigger than 1MB, it will give a successfully upload and list it in the file list but fails to open (Esp8266 1MB boundary??)
-Large file names trimm the name and extension but work ok and can be read. 
-reuploading a file will rewrite the content if there is enough space to upload it  (it doesn't check ahead that is a rewrite).

-In two cases I got file corruption of a previously and correctly uploaded content, after a new big upload (1MB boundary?). ** I will do further testing to track down this, as the only relevant issue **

[post log.txt](https://github.com/SmingHub/Sming/files/219139/post.log.txt)

Minor detail as example, any error is displayed in console but not on the upload page

patrickjahns:
> -If file to upload is bigger than 1MB, it will give a successfully upload and list it in the file list but fails to open (Esp8266 1MB boundary??)

That is a limitation of spiffs - while we can guess the probably free space - we can never surely tell if it will fit since spiffs also requires space for metadata
https://github.com/pellepl/spiffs/wiki/Using-spiffs#checking-filesystem

> Large file names trimm the name and extension but work ok and can be read. 
Also spiffs limitation - 31chars is the maximum (including the extension) - for this scenario it might be good to define the default behavior (show error page saying "filename too large")

> reuploading a file will rewrite the content if there is enough space to upload it (it doesn't check ahead that is a rewrite).
That is proper default behavior - it is up to the sming user to tell if the upload is accepted or not in the callback. 

> In two cases I got file corruption of a previously and correctly uploaded content, after a new big upload (1MB boundary?). ** I will do further testing to track down this, as the only relevant issue **

That would be good - I suspect it`s also related to spiffs  - it would be great if you can track this down. 
Thanks for including a log - would be great if you can limit it to the parts that caused issue - it`s hard for me to look at 27000 lines without knowing where the issue happened.

hreintke:
@patrickjahns : @robotiko : 
Quick first reply.
>If file to upload is bigger than 1MB, it will give a successfully upload and list it in the file list but fails to open (Esp8266 1MB boundary??)

>That is a limitation of spiffs - while we can guess the probably free space - we can never surely tell if it will fit since spiffs also requires space for metadata

Spiffs/filewrite should give an error return when a filewrite doesn't fit on the FS. That is independent on the size of the file currently being written. -> Also small files can get the FS filled.
It might be related to the possible mismatch in sizing of the spiffy created FS and the mounted FS in Sming (Solved in SmingRTOS :smile: ). In rboot application it is aligned by using mount_manual.


>Also spiffs limitation - 31chars is the maximum (including the extension) - for this scenario it might be good to define the default behavior (show error page saying "filename too large")

There is an application callback, if the filename is to long before the callback, the application can update it to a correct filename. If it is to long after the callback, the upload should be cancelled (skipped) and reported as an erroneous download when the httprequest is complete.

patrickjahns:
@hreintke 

> Spiffs/filewrite should give an error return when a filewrite doesn't fit on the FS. That is independent on the size of the file currently being written. -> Also small files can get the FS filled.
It might be related to the possible mismatch in sizing of the spiffy created FS and the mounted FS in Sming (Solved in SmingRTOS :smile: ). In rboot application it is aligned by using mount_manual.

That`s what I meant it`s a limitation/issue with spiffs - for the upload scenario if the file write doesn`t return an error or the free space is wrongly announced the upload part can`t know it

> There is an application callback, if the filename is to long before the callback, the application can update it to a correct filename. If it is to long after the callback, the upload should be cancelled (skipped) and reported as an erroneous download when the httprequest is complete.

more thoughts on this
- it is good that the application can intercept the behavior 
- filename length check will be skipped, if the callback handles the file writing and not the default handler
- i`d prefer cancelling the request and displaying a 4xx page 

different things that came to my mind:
It would probably be a good idea to provide access to httpresponse in the uploadhandler and have the upload handler return true/false (or a status code) to indicate if an error occured in the callback and if we should drop the request.

This is related to a different thought I was having: if we want to secure the upload endpoint with http basic auth - it is currently only possible after the upload finished and we are in the normal callback. But for security purposes we should intercept the upload after the headers have been parsed and before we actually open+write to a file

hreintke:
>i`d prefer cancelling the request and displaying a 4xx page

The http_response class should not write a response. It is up to the application to react to the "outside world" We just provide the information : The upload was aborted.

>This is related to a different thought I was having: if we want to secure the upload endpoint with http basic auth - it is currently only possible after the upload finished and we are in the normal callback

The httprequest is available in the httpupload callback. With that information the application can decide. (correct ?)

patrickjahns:
> The httprequest is available in the httpupload callback. With that information the application can decide. (correct ?)

Yes the request is available 

But my remarks are, with the current flow:
- C: send request with header (and maybe some data) 
- SMING: parse header
- SMING: parse body -> we find multipart
- SMING: hand over to application upload
- APP Upload callback: check for password/username in header -> not allowed and set to skip upload
- C: send more data until finished
- SMING: parse data until client finishes
- SMING: hand over to application callback
- APP: Request is not allowed - send not allowed response
- SMING: return the response

This way even if someone is not allowed to upload data - it is still forced to first accept all client packets before we actually react and close the connection

It would be better if the APP: has access to response object; sets the response to not allowed; tells Sming to process it further but send the response/error code at that very moment 

(Similar to the process we internally do when we receive a bad request)

slaff:
Post parameters with multipart encoding should be implemented with the upcoming Http refactoring (#1112).

PR #703, opened 2016-04-17
ignore vim swapfiles

zendes:


hreintke:
@zendes : 
Can you submit the same PR for SmingRTOS ?

PR #704, opened 2016-04-17
add make terminal option

zendes:
sometimes it's useful to be able to start the serial port monitor
without having to remember what com port or baud rate you've configured
for this project. this adds a simple makefile target that just kills &
starts the terminal in the same way make flash does, without flashing or
resetting the mcu.

i submitted this before, but it auto-closed when i deleted the branch on my side... (https://github.com/SmingHub/Sming/pull/640)

### DIFF
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -344,6 +344,11 @@ else
 endif
 	$(TERMINAL)
 
+terminal:
+	$(vecho) "Killing Terminal to free $(COM_PORT)"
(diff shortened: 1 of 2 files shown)


hreintke:
@zendes : 
I merged this PR but can you also make a PR for adding the make target for eclipse in the example applications.
In that same PR also remove the obsolete "flashonefile" and "flashboot" eclipse targets.

Combine the two PR's and submit for SmingRTOS ?

PR #705, opened 2016-04-17
make COM_SPEED_SERIAL from Makefile available inside apps

zendes:
the Makefiles define COM_SPEED_SERIAL and uses this to start the
terminal, but this value isn't passed to the code. inside user_config.h
there is SERIAL_BAUD_RATE that configures the port speed. this means
it's left up to the user to keep the two in sync, which is less than
ideal. this commit passes COM_SPEED_SERIAL to gcc as a define, and
allows users to use one value for both if they chooseto do so. this
won't break any existing code.

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -111,7 +111,7 @@ USER_LIBDIR = compiler/lib
 LIBS		= microc microgcc hal phy pp net80211 lwip wpa main
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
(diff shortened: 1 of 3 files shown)


zendes:
i previously created a PR (https://github.com/SmingHub/Sming/pull/637) that also updated the samples, but that got auto-closed cause i deleted the branch on my side. this PR is just for the makefile changes

hreintke:
@zendes : 
Can you submit the same PR for SmingRTOS ?

PR #713, opened 2016-04-24
Include directories in spiffy rom

Matzz:
According to spiffy documentation:

```
Presently, spiffs does not support directories. It produces a flat structure. Creating a file with path tmp/myfile.txt will create a file called tmp/myfile.txt instead of a myfile.txt under directory tmp.
```

spiffy don't support getting list of files in directory but is able to have files with slashes in names. There are a lot of cases where it is enough (eg separate javascript and css files on web server root). I modified spiffy.c to scan and include subdirectories.
I tested it on my esp (build and flashed on windows) and files are visible.
As I'm neither native c or c++ programmer please review my code.

### DIFF
--- a/Sming/spiffy/spiffy.c
+++ b/Sming/spiffy/spiffy.c
@@ -3,6 +3,7 @@
 #include <dirent.h>
 #include <spiffs.h>
 #include <sys/stat.h>
+#include <ftw.h>
 
 #define LOG_PAGE_SIZE       256
 #define SPI_FLASH_SEC_SIZE 4096
@@ -20,6 +21,10 @@ static u8_t spiffs_cache_buf[(LOG_PAGE_SIZE+32)*4];
 
 #define S_DBG
 //#define S_DBG printf
+
+#ifndef USE_FDS
+#define USE_FDS 15
+#endif
 
 static FILE *rom = 0;
 
@@ -147,12 +152,13 @@ int my_spiffs_format() {
 	return res;
 }
 
-int write_to_spiffs(char *fname, u8_t *data, int size) {
+int write_to_spiffs(const char *fname, u8_t *data, int size) {
 
 	int ret = 0;
 	spiffs_file fd = -1;
 
-	fd = SPIFFS_open(&fs, fname, SPIFFS_CREAT | SPIFFS_TRUNC | SPIFFS_RDWR, 0);
+	char *fnameOnSpiffy = strchr(fname, '/')+1;
+	fd = SPIFFS_open(&fs, fnameOnSpiffy, SPIFFS_CREAT | SPIFFS_TRUNC | SPIFFS_RDWR, 0);
 	if (fd < 0) {
 		printf("Unable to open spiffs file '%s', error %d.\n", fname, fd);
 	} else {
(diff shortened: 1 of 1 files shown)


hreintke:
@Matzz : 
You are updating the creating of the spiffs FE to be flashed.
Can you explain what the result of this update will be ?

On the patch : 
You are adding the ftw.h. Is that available on all versions and flavors of mingw ?
Did you verify how sming http/ftp/filesystem work with the new spiffs FE ?

raburton:
I did something very similar a while back after I rewrote spiffy, before noticing that directories weren't implemented in spiffs. My old patch doesn't apply cleanly now, but it looks like it was quite a bit simpler than this commit. It also used tfw which seems sensible, but I didn't test that too widely so can't answer that question.

Matzz:
@hreintke 
ftw.h is avaiable on my mingw which was installed with Sming (Windows7 64x). It also should be aviable on Linux - http://man7.org/linux/man-pages/man3/nftw.3.html
I haven't tested on other platofrms.

I'm able to serve files from subdirs via http. I'm also can read them in via serial handler. From spiffs point of view slashes in filenames are treaded just like any other characters.

Regarding what this patch exactly do - It scans selected directory and its subdirectories. All files are saved in spiffy rom. Name of the saved files do not contain root directory name. If you are burning d:/test :
```
d:\test
│   a.txt
│   b.txt
└───c
    │   d.txt
    │   e.txt
    ├───f
    │       g.txt
    └───h
```
On spiffy you will see:
```
a.txt
b.txt
c/d.txt
c/e.txt
f/g.txt
```
Empty directories are not included.

Matzz:
@raburton Please share it. It would be nice to have cleaner patch (via gist?)

raburton:
https://dl.dropboxusercontent.com/u/5500141/spiffy.c

Note there is a still a merge conflict in main() which I haven't had a look at do to lack of time. I'm not saying this is better than your version, just looked a bit simpler so have a look and see if there is anything of use in it...

hreintke:
@Matzz @raburton 
There is another remark/issue on the spiffy that is doesn't check for the 31 char limit on filenames within spiffs.
Can you include that check within this patch ?

patrickjahns:
@Matzz 

In regards to 
```
f/g.txt
```

it should actually be:
```
c/f/g.txt
``` 

Was that a typo? Or is it general implementaiton - if it is general implementation I`d suggest to keep the "folder tree"  intact.

hreintke:
@Matzz : 
- Not sure on the http upload/download filename conventions on directories but for sure the ftp server code within sming will break with this included. 

- Do we have to take care/have trouble with the linux/windows difference with regards to forward and backward slashes ?

hreintke:
ftp break fixed in upcoming ftp update on SmingRTOS.
Will not be backported to SmingNONOS

patrickjahns:
@hreinte @Matzz 

Uploading to directories will not be possible from client side. The application would need to alter the filename and write to a custom filename - but the http client can not supply names with "/" or "\" (host limitations)

Also how will sendFile from the httpserver send the file? will it send "a/b/c.txt" or will is just send "c.txt" ?    can please ensure that the client does receive proper filenames (otherwise downloading is not possible with the default name)

If this is breaking any feature in nonos I`d vouch against including it

hreintke:
@patrickjahns : 
First : Agree that we should not implement if it breaks other functionality.

- Uploading : If I read the "content-disposition" spec's correctly the http_client is required to remove all directory specifications from the filename. -> the proposed filename to the application will/should not supply a filename with "/' to the application. The application can update the filename to it's liking and then supply a filename with "/" included, which will be handled correctly within spiffs.
Is this correct ?

- Downloading : I did a very quick and non extensive check. : 
When requesting `http://my_esp/test/idx.html` the httpresponse correctly sends the "test/idx.html" file from spiffs.

Matzz:
@patrickjahns Yes it was a typo

Matzz:
I added check for file name length. If file name is too long file will be skipped and relevant information will printed.

slaff:
@Matzz I know that this PR was delayed for too long but if you still have interest in it I can merge it if you make the new feature to be optional. For example `make ENABLE_SPIFFS_TREE=1` should provide tree-like file names and just `make` should work as before.

mzakarczemny:
@slaff Ok I will try.

anakod:
Let's discuss it, I think what if it willn't break back compatibility (and it shouldn't because directories didn't work before anyway), we can merge it without additional compiler flags. If user don't want directories then he can just don't create folders. What do you think?

slaff:
@Matzz Feel free to reopen this PR if you have new useful changes to it. I will close it now due to inactivity.

Matzz:
@slaff Could you reopen it please? I know that it took ages but I added support of `ENABLE_SPIFFS_TREE`. Relevant commit is in https://github.com/Matzz/Sming/commit/b45dc843ddda815489afbe43deff9f10c3387222
I think when you will reopen this PR, this new commit will be visible here.

slaff:
@Matzz I cannot reopen it from the web interface. Can you try to create new PR with the changes?

Issue #712, opened 2016-04-24
HttpClient returns successfull if connection has timeout


patrickjahns:
I noticed that http client will return success, even if the request timed out.
See log:

```
Download file: app.min.js.gz 1
Download: http://patrickjahns.github.io:80/esp_rgbww_firmware/release/app.min.js.gz
+TCP connection
connect to: patrickjahns.github.io
realloc 86 -> 217
DNS record found: patrickjahns.github.io = 23.235.43.133
TcpConnection::connect result:, 0
OnConnected
TCP connected
onReadyToSendData: 0
TcpClient request completed
TCP sent: 88
onReadyToSendData: 2
Header pos: 591
Server === GitHub.com
Content-Type === application/octet-stream
Last-Modified === Sat, 23 Apr 2016 09:44:44 GMT
Access-Control-Allow-Origin === *
Expires === Sun, 24 Apr 2016 08:50:28 GMT
Cache-Control === max-age=600
X-GitHub-Request-Id === 17EB2B24:2087:19D83017:571C867C
Content-Length === 156513
Accept-Ranges === bytes
Date === Sun, 24 Apr 2016 11:12:13 GMT
Via === 1.1 varnish
Age === 0
Connection === close
X-Served-By === cache-ams4121-AMS
X-Cache === MISS
X-Cache-Hits === 0
X-Timer === S1461496333.757779,VS0,VE94
Vary === Accept-Encoding
X-Fastly-Request-ID === 759cf87c028ec58ef196ebec132bab38b0080a74
TCP received: 1460 bytes
onReadyToSendData: 1
-TCP connection

(....)

TCP connection closed by timeout: 70 (from 70)
Download file len written: 865, res^ 1
TCP connection closing
```

This is related to the wish in #626 to improve the handling in the http client.
I am not sure if fixing the problems is "better" or if we should keep the eyes open for a good low resource http client implementation to be ported to SMING

Suggestions?

patrickjahns:
I`ve just took the liberty and searched for some c(++) http libraries that might be worth having a look and porting:
- [http c client](https://github.com/reagent/http)
- [mbed HttpClient](https://developer.mbed.org/users/WiredHome/code/HTTPClient/)
- [libcurl] (https://curl.haxx.se/libcurl/)
- [cpr - curl for people](https://github.com/whoshuu/cpr)

CPR seems very promising, but my guess is, that it would be a rather large port

Any thoughts/comments on this?

patrickjahns:
[...]

PR #714, opened 2016-04-25
return TimeoutError when timeout occured

patrickjahns:
Fix for #712 

I`ve tested it with HttpClient and HttpServer and did not experience any different behavior. Any other remarks/comments when changing the return value to the proper one?


patrickjahns:
Just checked - this needs to be moved to RTOS as well when confirmed

Issue #715, opened 2016-04-27
rboot - makefile missing options


patrickjahns:
I have tried getting the temprom feature to work, but rboot never "enables" the rtc module

Serial output from booting

```
16:45:23.591> load 0x3ffe8000, len 660, room 12 
16:45:23.591> tail 8
16:45:23.591> chksum 0xcd
16:45:23.591> csum 0xcd
16:45:23.591> 
16:45:23.591> rBoot v1.3.0 - richardaburton@gmail.com
16:45:23.591> Flash Size:   32 Mbit
16:45:23.591> Flash Mode:   QIO
16:45:23.591> Flash Speed:  40 MHz
16:45:23.591> rBoot Option: Big flash
16:45:23.591> 
16:45:23.803> Booting rom 0.
```

Makefile-user.mk

```
# use rboot build mode
RBOOT_ENABLED = 1

# 
RBOOT_RTC_ENABLED = 1

# enable big flash support (for multiple roms, each in separate 1mb block of flash)
RBOOT_BIG_FLASH = 1

# two rom mode (where two roms sit in the same 1mb block of flash)
RBOOT_TWO_ROMS  = 0

# where does the filesystem reside
RBOOT_SPIFFS_0  = 0x100000
RBOOT_SPIFFS_1  = 0x300000 
```

Relevant section from makefile

```
make all V=1 
mkdir -p out/build/app
mkdir -p out/build//C/opt/Sming/Sming/rboot/appcode
mkdir -p out/firmware
OC out/build/libmain2.a
/usr/bin/make -C /C/opt/Sming/Sming/rboot
make[1]: Entering directory `/C/opt/Sming/Sming/rboot'
CC rboot-stage2a.c
xtensa-lx106-elf-gcc -Os -O3 -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals  -D__ets__ -DICACHE_FLASH -DBOOT_BIG_FLASH  -c rboot-stage2a.c -o /c/Users/patri/@projects/arduinoprojects/RGBCTRLfw/out/build/rboot-stage2a.o
LD /c/Users/patri/@projects/arduinoprojects/RGBCTRLfw/out/build/rboot-stage2a.elf
[...]

PR #716, opened 2016-04-27
allow enabling RBOOT_RTC

patrickjahns:
See #715 

@raburton 

I was not sure how to proceed with the other options I mentioned - if it is okay to also allow them via makefile-user.mk please let me know and I will extend the PR


Sming/rboot/Makefile
```diff
@@ -37,6 +37,9 @@ endif
 ifeq ($(RBOOT_INTEGRATION),1)
 	CFLAGS += -DRBOOT_INTEGRATION
 endif
+ifeq ($(RBOOT_RTC_ENABLED),1)
```
raburton: rBoot is an external component included in sming unmodified, please supply a patch upstream to rBoot and I'll get it merged

Sming/Makefile-rboot.mk
```diff
@@ -264,6 +264,7 @@ RBOOT_FW_BASE := $(abspath $(FW_BASE))
 export RBOOT_BIG_FLASH
 export RBOOT_BUILD_BASE
 export RBOOT_FW_BASE
+export RBOOT_RTC_ENABLED
```
raburton: you might want to add a use of this (commented out by default) to the rBoot sample app makefile with the other options, so people will know it's there

patrickjahns:
I will move the required changes upstream and see how I can integrate it with the examples.

patrickjahns:
- Moved rboot related changes upstream
- included the sample from @slaff for using temp_rom (hope that is okay with you?)
- allowed for the other options to be set (gpio/irom_chksum/config_chksum)

samples/Basic_rBoot/app/application.cpp
```diff
@@ -163,8 +217,21 @@ void init() {
 	Serial.begin(SERIAL_BAUD_RATE); // 115200 by default
 	Serial.systemDebugOutput(true); // Debug output to serial
 	
+	uint8 slot = 0;
+
+#ifdef BOOT_RTC_ENABLED
+	debugf("Temp Mode is enabled");
```
raburton: rtc being enabled doesn't necessarily mean you are in temp boot mode
also, currently running rom is printed further down this method already...

samples/Basic_rBoot/app/application.cpp
```diff
@@ -193,4 +260,4 @@ void init() {
 	Serial.println();
 	
 	Serial.setCallback(serialCallBack);
-}
+}
```
raburton: please leave the training newline

Sming/Makefile-rboot.mk
```diff
@@ -12,6 +12,9 @@
 RBOOT_BIG_FLASH  ?= 1
 RBOOT_TWO_ROMS   ?= 0
 RBOOT_RTC_ENABLED ?= 0
+RBOOT_GPIO_ENABLED ?= 0
+RBOOT_IROM_CHKSUM ?= 0
```
raburton: enabling this requires an extra option on the call to esptool2 that creates the rom image (to include the data in the checksum)

patrickjahns: Can you point me towards the right direction here, then I will update the example accordingly

raburton: You need to add `-iromchksum` to the parameters passed to esptool2 when the application rom is created if RBOOT_IROM_CHKSUM is enabled (set to 1). Probably want something like this in the makefile:
```
ifeq ($(RBOOT_IROM_CHKSUM), 1)
  RBOOT_E2_USER_ARGS += -iromchksum
endif
```

PR #721, opened 2016-04-30
RC-Switch library support with example

anakod:
Add https://github.com/sui77/rc-switch/ library support.
successfully checked on real hardware.

Future improvements: delegate for _code receiving completed_ event.

### DIFF
--- a/Sming/Libraries/RCSwitch/.gitignore
+++ b/Sming/Libraries/RCSwitch/.gitignore
@@ -1,0 +1,17 @@
+# Mac stuff
+.DS_Store
+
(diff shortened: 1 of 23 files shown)


hreintke:
@anakod : 
Would be good to include the delegate from the start. Can you add that ?

Did you test both interrupt and non-interrupt mode ?

anakod:
Delegate support is additional feature, and planned in the future.

hreintke:
@anakod : 

Did you test both interrupt and non-interrupt mode ?

PR #732, opened 2016-05-21
simple convert DateTime to ISO8601 string format

johndoe8967:
some IoT platforms require a iSO8601 time format (e.g. thingspeak)


hreintke:
@johndoe8967 : 
Thanks for your PR. Good extension.

Checking your code and the existing dateTime class I noticed that on some locations there is a buf[xx]. As you know this is using stack space.
The SmingRTOS is much more sensitive in stack usage than the SmingNONOS. 

Could you in your PR :  
- Update buf[..] to char* buf = new char[..] and add the required delete.
- Also do this for the already existing buf[..] 
- Also submit the PR to SmingRTOS.

Then we have both your extension and a generic improvement in one go.

johndoe8967:
of course I can do

I didn't do any stack measurement so I don't know if the stack size is smaller now because I don't know how to measure it :-)

ADiea:
@johndoe8967 to measure stack in general code can be something like this
```
void stackmeasure()
{
char ch[128];
debugf("stack ptr is %x\n", &ch);
stackmeasure()
}
```

calling it recursively will eventually crash, sifference between first call and last call ) * 128 is your initial stack space

In practice, for non-os, stack and heap share the same address space heap grows up from one end stack grows down from the other end so even if you will see that you have xxx amount of stack it's also important to keep in mind the heap you need

Hope this makes sense and helps
Cheers

anakod:
I think dynamic memory allocation in NonOS Sming isn't good idea: memory fragentation & code complexity grows.

Sming/Makefile-project.mk
```diff
@@ -139,7 +139,7 @@ EXTRA_INCDIR += $(SMING_HOME)/include $(SMING_HOME)/ $(SMING_HOME)/system/includ
 
 # libraries used in this project, mainly provided by the SDK
 USER_LIBDIR = $(SMING_HOME)/compiler/lib/
-LIBS		= microc microgcc hal phy pp net80211 lwip wpa main sming crypto pwm smartconfig $(EXTRA_LIBS)
+LIBS		= microc microgcc hal phy pp net80211 lwip wpa main wps sming crypto pwm smartconfig $(EXTRA_LIBS)
```
slaff: Why is `wps` added here? Is it needed for this change?

Sming/Makefile-project.mk
```diff
@@ -270,6 +270,8 @@ endef
 .PHONY: all checkdirs spiff_update spiff_clean clean
 
 all: checkdirs $(TARGET_OUT) $(SPIFF_BIN_OUT) $(FW_FILE_1) $(FW_FILE_2)
+	expr 0`cat count.txt` + 1 > count.txt
```
slaff: Why is that count added here?

slaff:
Rebased in #834.

PR #734, opened 2016-06-04
Added parameter to Station::config that allows setting of the WIFI pa…

slaff:
…ssword without persisting it in the flash.


anakod:
I think about implementing wifi_station_set_config_current method before, but currently I've disabled writing to flash in case if configuration same as current. Anyway your implementation looks nice too. How you think, in which cases this PR will provide additional benefits for user?

PR #741, opened 2016-06-07
Fix #740. Expose write methods from parent class Print

### LINKED ISSUE #740: HardwareSerial missing write methods from Print due to name hiding
Due to C++ name hiding write methods from Print class are hidden in HardwareSerial.
For example:
```C++
uint8_t  buffer[16];
memset(buffer,0, 16);
SerialPort.write(buffer, 16);
```

gives compile error ("no matching function for call to 'HardwareSerial::write(const char [16], int)").
Adding ` using Print::write` in HardwareSerial.h solves this (better could be refactor the whole thing).

pmst:
Adds `using Print::write` in HardwareSerial.h to expose write methods.

### DIFF
--- a/Sming/SmingCore/HardwareSerial.h
+++ b/Sming/SmingCore/HardwareSerial.h
@@ -93,6 +93,8 @@ public:
 	 */
 	size_t write(uint8_t oneChar);
 
+	using Print::write;
+
 	//void printf(const char *fmt, ...);
 	/** @brief  Configure serial port for system debug output
 	 *  @param  enabled True to enable this port for system debug output


slaff:
Related to issue #759.

Advantages
=========
Accepting this PR will allow sending binary data using UART. Use-Cases: Sending firmware from ESP8266 to other non-wifi MCU.

Disadvantages:
============
None, as far as I can tell.

@anakod @hreintke Do you have anything against merging this PR? 
If not I will merge it and port the changes also to SmingRTOS.

slaff:
@pmst Why don't you expose the actual class that you are inheriting from? For example instead of `using Print::write` to have `using Stream:write` ?

slaff:
Merged with this commit https://github.com/SmingHub/Sming/commit/c58e8fca806935b73cda2f54fa05661757c1b475

PR #748, opened 2016-06-22
Safe mprintf

ADiea:
mprintf: 
*reduce include overhead
*proper declare symbols with C linkage
*fix crash when size is smaller than arg_size for %d %u
*fix padding for %d and %f
*fix buffer overflow for %d %f
*add vsnprintf equivalent
*add putch equivalent


Sming/system/stringconversion.cpp
```diff
@@ -29,8 +34,9 @@ char* ltoa_w(long val, char* buffer, int base, int width)
 		width -= strlen(&buf[i+1]);
 		if(width > 0)
 		{
-			memset(buffer, ' ', width);
+			memset(buffer, pad, width);
 		}
+		else width = 0;
```
ADiea: fixes https://github.com/SmingHub/SmingRTOS/issues/129

Sming/system/stringconversion.cpp
```diff
@@ -64,17 +70,22 @@ char* ultoa_wp(unsigned long val, char* buffer, unsigned int base, int width, ch
 		{
 			memset(buffer, pad, width);
 		}
+		else width = 0;
```
ADiea: fixes https://github.com/SmingHub/SmingRTOS/issues/129

slaff:
Merged in #791.

PR #750, opened 2016-07-04
Adding libwps to Makefile to enable the PBC WPS function

johndoe8967:


slaff:
@johndoe8967 Isn't it enough for you to add `wps` to `EXTRA_LIBS` in your Makefile-user.mk ?
Do you think that `wps` will be used quite often in projects?

johndoe8967:
@slaff: sure it would be possible, maybe a small template sample would be enough to show the functionality.

anakod:
I think we have example how to add external library, but can't find it now. Anyway it isn't really clear.
@johndoe8967 may be better you can make PR to wiki documentation or PR with example of EXTRA_LIBS usage?

slaff:
Feel free to reopen this PR when the WPS sample is ready.

PR #755, opened 2016-07-15
fixed lowercase include path for ArduCAM

fodpeter:


harry-boe:
Just a note from the ArduCAM developer ..

We figured out that 3 out of 10 ArduCAM can not handle the default SPI speed set in the example app SPISettings Line 95

`    pinMode(CAM_CS, OUTPUT);
	digitalWrite(CAM_CS, HIGH);
	SPI.begin();
	SPI.beginTransaction(SPISettings(20000000, MSBFIRST, SPI_MODE0));
`
In case your Camera does not work try lower SPI Speeds like 12MHz

anakod:
I think we should merge it.

slaff:
Fixed with #779.

> We figured out that 3 out of 10 ArduCAM can not handle the default SPI speed set in the example app SPISettings Line 95

@harry-boe Can you add the SPI speed changes as a separate PR?

PR #763, opened 2016-08-04
Added access to nanosecond resolution RTC.

tht:
I'm working on a clock and need sub-second resolution to the RTC. This small change allow setting/getting of the RTC in nanosecond resolution. It is fully backwards compatible.

Why nanoseconds? For my specific use-case tens of milliseconds should do it but this adds additional multiplications to the codebase. By working in nanoseconds the number of multiplications for RTC access stays the same.


anakod:
Looking fine for me.

slaff:
The changes are sent as PR to SmingRTOS: https://github.com/SmingHub/SmingRTOS/pull/143.

PR #769, opened 2016-09-15
Rboot: Fixed issue with final bytes < 4 not being written on the flash memory.

slaff:
If the size of the last downloaded item is not divisible by 4 then the remaining final bytes are not written. This commit fixes this issue.

### DIFF
--- a/Sming/SmingCore/Network/rBootHttpUpdate.cpp
+++ b/Sming/SmingCore/Network/rBootHttpUpdate.cpp
@@ -50,6 +50,12 @@ void rBootHttpUpdate::updateFailed() {
 	if (updateDelegate) updateDelegate(false);
 }
 
+void rBootHttpUpdate::onItemDownloadCompleted(HttpClient& client, bool successful) {
+	if(successful) {
+		rboot_write_end(&rBootWriteStatus);
+	}
+}
+
 void rBootHttpUpdate::onTimer() {
 	
 	if (TcpClient::isProcessing()) return; // Will wait
@@ -80,7 +86,7 @@ void rBootHttpUpdate::onTimer() {
 	rBootHttpUpdateItem &it = items[currentItem];
 	debugf("Download file:\r\n    (%d) %s -> %X", currentItem, it.url.c_str(), it.targetOffset);
 	rBootWriteStatus = rboot_write_init(items[currentItem].targetOffset);
-	startDownload(URL(it.url), eHCM_UserDefined, NULL);
+	startDownload(URL(it.url), eHCM_UserDefined, HttpClientCompletedDelegate(&rBootHttpUpdate::onItemDownloadCompleted, this));
 }
 
 void rBootHttpUpdate::writeRawData(pbuf* buf, int startPos) {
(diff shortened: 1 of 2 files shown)


anakod:
How it worked before (without it) and how you test new code? Why additional function onItemDownloadCompleted needed (why we can't write these bytes from old method)?

raburton:
Roms are always a multiple of 4 bytes (as are spiffs afaik), so it has always worked fine. If you used the rBoot code to ota an arbitrary file to the flash then it could be any size and if not a multiple of 4 bytes the last 1-3 bytes could get lost.

slaff:
> If you used the rBoot code to ota an arbitrary file to the flash then it could be any size and if not a multiple of 4 bytes the last 1-3 bytes could get lost.

@raburton That is exactly what I did. 

> How it worked before (without it) and how you test new code? 

@anakod  It worked because the usual ROM and SPIFFS are multiple of 4 bytes. But for cases where the data length is not multiple of 4 it did not work.

> Why additional function onItemDownloadCompleted needed (why we can't write these bytes from old method)?

@anakod Because in the cases where downloaded data length is not multiple by 4 the remaining last bytes are never written to flash.

slaff:
@anakod @raburton 

How about merging this PR?

Advantages
=========
Allows downloading data with arbitrary size. Use-case: downloading firmware for another MCU connected to ESP8266.

Disadvantages:
============
None, that I know of.

If there are no comments in the coming days I will merge this PR and port the changes to SmingRTOS.

raburton:
It would be much better to create a PR in rBoot with the change in the API,
there is already an rboot_write_init function for starting the process,
ideally there would be a rboot_write_finish (or similar) that will handle
these last bytes if present, and then call that from Sming. That way useful
code is contributed back upstream and Sming isn't mixed between using the
rBoot api to do all the writing except for the last 1-3 bytes where it does
it's own flash write call. Also your code doesn't erase sectors, if those
bytes take you over the sector boundary. I would suggest the finish method
calls rboot_write_flash with a length of 4 byte and extra_bytes as the data
(ideally null/0xff pad unused bytes in extra_bytes so you don't get a
couple of bytes junk on the end).

hreintke:
@slaff : 
In SmingRTOS I made Sming functions for writing to flash. Those do not have the "4-byte issue" and do erase..
There is also a "high-level" interface -> Created a FlashStream that can be used in the application.
Currently used by FTP for download applications to ROM (FTP OTA).
Will be an easy update to HttpUpdate to use that too. 

That is for me the preferred solution for SmingRTOS.

slaff:
> ideally there would be a rboot_write_finish (or similar) that will handle these last bytes if present

@raburton Sounds good to me. Can you add these changes to rBoot so that we can integrate them in Sming?

anakod:
Agreed, changes in rboot API will be better. But this PR fine for me too. So as @raburton decided.

raburton:
rBoot api changes made upstream

slaff:
@raburton Thanks for making the changes. 

@raburton @anakod 
I updated the PR. Any recommendations for additional changes or it looks good to you?

raburton:
You've patched rBoot rather than updating the whole code, you should update the whole of rBoot (ideally in a separate PR) so you have a proper version instead of a custom version.

slaff:
> you should update the whole of rBoot 

@raburton I'll check if I can add rboot as a git submodule so that future updates are easy to perform.

PR #774, opened 2016-09-29
Feature/interupts fix

anakod:
1. Correct interrupts disabling implementation (based on http://www.esp8266.com/viewtopic.php?p=16758 forum topic)
2. Important fix: IRAM_ATTR added to prevent crash on few interrupt callback routines.


slaff:
@anakod This PR looks like a very nice interrupts improvement. Why don't you merge it with the rest of  the code?

hreintke:
@anakod @slaff 
What is the reason for implementing "local interrupt routine" instead of using the ones provided in the SDK.

What is the issue we are solving ?

anakod:
@hreintke, we already have Arduino compatible interrupts. But it wasn't implemented right. Also all functions what can be called from interrupts should have IRAM_ATTR. If not - program will crash sometimes.

anakod:
@slaff Agree with you, will merge.

PR #779, opened 2016-10-19
Fixed compilation of two samples:

slaff:
* ArduCam
* Basic_WebSkeleton


anakod:
I think everything is fine.

slaff:
@anakod 

> I think everything is fine.
Great. Meanwhile I will create another PR with changed .travis script to prevent accepting changes that do not compile. This way we can notice earlier if something is wrong.

anakod:
Yes, it will be great too.

slaff:
RTOS migration summary: 
* ArduCam is not present in SmingRTOS -> no need for migration.
* WebSkeletonApp already has the fix although in a different file (configuration.h instead of configuration.cpp). -> no need for migration

PR #780, opened 2016-10-20
Updated the travis script.

slaff:
If a file from samples/ is changed then that sample project will be recompiled.
If a file from Sming/ is changed then all sample projects will be recompiled.
These changes will prevent Sming from accepting/having code that does not compile.


slaff:
@anakod Shall I merge it? It is working quite well on [my "tasty" branch](https://travis-ci.org/slaff/Sming/jobs/168896658).

slaff:
@hreintke Are you interested in having these travis changes in SmingRTOS?

hreintke:
@slaff Yes, should go into SmingRTOS too.

PR #781, opened 2016-10-21
Set the irom0 segment to its max allowed size.

slaff:
The initial LD restrictions are too restrictive. Those changes to the linker script allow to use the full addressable space in ESP8266.

### DIFF
--- a/Sming/compiler/ld/eagle.app.v6.cpp.ld
+++ b/Sming/compiler/ld/eagle.app.v6.cpp.ld
@@ -5,7 +5,7 @@ MEMORY
   dport0_0_seg :                        org = 0x3FF00000, len = 0x10
   dram0_0_seg :                         org = 0x3FFE8000, len = 0x14000
   iram1_0_seg :                         org = 0x40100000, len = 0x8000
-  irom0_0_seg :                         org = 0x40209000, len = 0x78000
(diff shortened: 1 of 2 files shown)


slaff:
@anakod Any comments? If not this change will be part of the next merge.

hreintke:
@slaff 
This updates the Rboot ld, should the non-rboot also be fixed ?
Could/should be set depending on SPI_SIZE. I have a awk option on that but can't get it into makefile (getting errors). Can find it from history if needed

slaff:
@hreintke The non-rboot linker script is also updated `Sming/compiler/ld/eagle.app.v6.cpp.ld`. Concerning the SPI_SIZE - the linker script can be updated for ESPs with less than 1MB. For the rest - those settings should be just fine (as far as I know ESP does not allow you to address more than 1MB). Is that what you meant?

anakod:
Few questions: which benifits it provide? Larger flash code size?
It will work for old ESPs with 512k also, right?

I think I'm already increased this size some time befre, but not sure.

slaff:
> Few questions: which benifits it provide? Larger flash code size?

Exactly, one can use the maximum allowed flash size for an application. 

> I think I'm already increased this size some time befre, but not sure.

Increased Size != Max Size :). 

> It will work for old ESPs with 512k also, right?

Yes, as long as the application size does not grow bigger than (512 K - initial ESP boot bytes.)

PR #793, opened 2016-11-04
Better structure and usage of third-party repositories

slaff:
In the current Sming library we have local copies of the source code from other projects. Some of them are [rboot](https://github.com/raburton/rboot), [esp-gdbstub](https://github.com/espressif/esp-gdbstub), [spiffs](https://github.com/pellepl/spiffs), etc

And those local copies are containing either the same source code as the original one or a slightly modified one. 

This has two major disadvantages: 
1. We cannot easily update the local copy from the original source code. 
2. We cannot easily send improvements to those projects once we make local changes in our local copy.

Instead of having local copies of those projects we can use submodules and fetch the source code from the respective repositories. If we need local modifications than we can keep a <submodule-name>.patch file and apply it after fetching the submodule code. This will decrease our efforts and allow us to use the latest versions of those third-party projects.

This PR makes the needed changes and adds documentation describing the recommended way of using third-party git submodules (libraries, source code snippets, etc).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,9 @@
 [submodule "Sming/rboot"]
-	path = Sming/rboot
+	path = Sming/third-party/rboot
 	url = https://github.com/raburton/rboot.git
 [submodule "Sming.wiki"]
 	path = docs/wiki
 	url = https://github.com/SmingHub/Sming.wiki.git
+[submodule "Sming/third-party/esp-gdbstub"]
+	path = Sming/third-party/esp-gdbstub
+	url = https://github.com/espressif/esp-gdbstub.git
--- a/AddingExternalSource.md
+++ b/AddingExternalSource.md
@@ -1,0 +1,56 @@
+This document describes the recommended way to add source code from other `git` repositories.
+
+Introduction
+============
+In Sming we have source code from other repositories [rboot](https://github.com/raburton/rboot), [esp-gdbstub](https://github.com/espressif/esp-gdbstub), [spiffs](https://github.com/pellepl/spiffs), etc..
+
+Having local copies of those modules brings some disadvantages with it
+ 
+1. We cannot easily update the local copy from the original source code. 
+2. We cannot easily send improvements to those projects once we make local changes in our local copy.
+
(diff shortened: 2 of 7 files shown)


slaff:
@anakod How about releasing the current code as new Sming version and meging this PR to the new Sming version?

anakod:
@slaff I think it's right direction, but is it back compatible? Which issues we will face after switching to modular structure on your opinion?

slaff:
> Which issues we will face after switching to modular structure ...

@anakod  Although unlikely it is possible that a remote repository can disappear. In that case we should keep a copy of all submodules that we use. One possible strategy to keep a copy is just to fork the existing submodule and keep its fork.

anakod:
May be better create dependcies on exact submodule branch/version (not HEAD) and update it manually only to prevent unexpected breaking of compatiblity after changes in any submodule?

slaff:
@anakod Submodules point to a commit  in the remote repository. If you take a look here: https://github.com/slaff/Sming/tree/feature/modular-structure/Sming/third-party you will see the selected commits. Once there is a new change in the remote repository we can manually update the SHA to point to a newer commit.

anakod:
Ok, thats right. One more question: what about patches, will it be applied automatically?

slaff:
Yes, the make script is taking care of this.

anakod:
If you think what it's ready than we can merge it, of course.

slaff:
> If you think what it's ready than we can merge it, of course.

@anakod Yes, sure. But before I do this I would like to merge the documentation PR (https://github.com/SmingHub/Sming/pull/563) and then tag the code as 2.1.5 release. After that I will merge this PR as part of the 2.2 milestone.

anakod:
Agreed

slaff:
@hreintke Are you interested in porting this PR to SmingRTOS? The restructuring will help us have smaller differences between Sming and SmingRTOS when dealing with third-party code and hopefully easier and faster porting of features from Sming to SmingRTOS.

hreintke:
@slaff : 
I have no experience at all with working with submodules. What I've read up till now looks like a good idea.
SmingRTOS uses other version of at least SPIFFS and uses of the newer functionality to solve issues with filesystems.

slaff:
> I have no experience at all with working with submodules....

@hreintke Ok, then I will check in the coming week(s) how difficult it will be to port this PR and make it part of SmingRTOS.

PR #796, opened 2016-11-10
Feature/docker

slaff:
Rebase of #731


slaff:
Note to myself: Instead of building the toolchain we can fetch already pre-built toolchain for that image from bintray and unpack it in the right location, as we do for the travis tests.

slaff:
@hreintke Are you interested in having also docker in SmingRTOS?

PR #812, opened 2016-11-25
Adds / updates API documentation

riban-bw:
More APIdocumentation:
DateTime
Clock
Debug
NtpCient
RTC
SystemClock
Some elements of the SDK used by Sming that are insufficiently documented within the SDK source files.

(There are many more classes to document. This is another chunk of the ongoing effort to get each class documented.)

This pull request _should_ only add or update API documentation in header files. There _should_ be no change to the source code / API / implementation.


slaff:
@riban-bw Brilliant job! Thanks a lot for your efforts.

Issue #813, opened 2016-11-25
make Basic_HwPWM is failing on MacOS X


slaff:
Steps to reproduce:

```bash
cd /tmp
git clone https://github.com/SmingHub/Sming.git
cd Sming/Sming
export SMING_HOME=/tmp/Sming/Sming
make Basic_HwPWM
```

The compilation fails with 
```bash
Done
C+ app/application.cpp
LD out/build/app.out
/Users/travis/build/SmingHub/Sming/Sming/compiler/lib//libsming.a(user_main.o):(.text.user_init+0x4): undefined reference to `init()'
/Users/travis/build/SmingHub/Sming/Sming/compiler/lib//libsming.a(user_main.o): In function `user_init':
/Users/travis/build/SmingHub/Sming/Sming/appinit/user_main.cpp:29: undefined reference to `init()'
collect2: error: ld returned 1 exit status
make[1]: *** [out/build/app.out] Error 1
make: *** [Basic_HwPWM] Error 2
```

Workaround for the moment is to compile Basic_HwPWM in two steps, as shown below

```bash
cd /tmp
git clone https://github.com/SmingHub/Sming.git
cd Sming/Sming
export SMING_HOME=/tmp/Sming/Sming
make
cd ../samples/Basic_HwPWM
make
```

slaff:
Fixed with PR #818.

PR #818, opened 2016-11-28
Simplified the compilation of custom PWM and Basic_HwPWM sample.

slaff:
Related to #813.


slaff:
This fix should stop the "randomly" failing MacOS X build.

PR #819, opened 2016-11-28
Feature/websocket client ssl

slaff:
Added support for websocket communication over SSL.


avr39-ripe:
Try it on freebsd + gcc51 + esp12F - no luck. all compiles, links and flashes fine, but do not work on actual device. Nothing shown on console. I'v checkout your pr, in Sming make ENABLE_SSL=1 rebuild and in Websocket_sample make rebuild flash
Does it works for you on device?

avr39-ripe:
with recent changes it boots, first run of program IS working 10 message sent 10 received, but after disconnection and re-connection NO messages are received. Same code without SSL works just fine after MANY disconnects/re-connects

slaff:
> but after disconnection and re-connection NO messages are received. Same code without SSL works

In my tests I saw that sometimes the second reconnection works but the third and so on do not work as reported. It seems that the connections that don't work have SSL handshake that never receives any response back...
```
DNS record found: echo.websocket.org = 174.129.224.73
TcpConnection::connect result:, 0
OnConnected
staticOnConnected: useSSL: 1, Error: 0
SSL: Starting connection...
SSL: Switching to 160 MHz
SSL: handshake start (77025 ms)
ssl/tls1.c:630 malloc 6864, left 30360
please start sntp first !
ax_port_write: send_raw_packet length 92(92)

SSL: handshake is in progress...
TCP sent: 92
Len: 92, TCP: 1, Available Size: 0, Can Send: 0
Sending websocket message: Hello 0
Storing 14 bytes in stream
Polling tcp: 1, Can Send: 0, AvailableWriteSize: 0, BufSize: 2920
Sending websocket message: Hello 1
....
```

@avr39-ripe Until this is fixed I can suggest you the following: Merge those changes BUT without having ENABLE_SSL by default in the Websocket client sample. Later on it might be possible to workaround such cases by using SSL session resumption.

avr39-ripe:
Ok, I think it worth to merge as is without setting ENABLE_SSL by default.

PR #834, opened 2016-12-16
Pr732

slaff:
Rebase of #734. 

Notes: We need to evaluate the pros and cons of dynamic allocation for this specific case(introduced with **"Optimize stack usage"** commit  b9b7e69):
* Pros: Less initial heap usage.
* Cons: Heap Fragmentation.

### DIFF
--- a/Sming/Services/DateTime/DateTime.cpp
+++ b/Sming/Services/DateTime/DateTime.cpp
@@ -120,20 +120,24 @@ time_t DateTime::toUnixTime()
 
 String DateTime::toShortDateString()
 {
-	char buf[64];
+	char* buf = new char[64];
 	sprintf(buf, "%02d.%02d.%d", Day, Month + 1, Year);
-	return String(buf);
+	auto result = String(buf);
+	delete buf;
+	return result;
 }
 
 String DateTime::toShortTimeString(bool includeSeconds /* = false*/)
 {
-	char buf[64];
+	char* buf = new char[64];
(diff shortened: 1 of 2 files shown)


ADiea:
Given that this *could* be used quite frequently during application lifecycle(ex when data arrives from sensors and timestamp is generated) I would keep it on stack
But it all depends on application usage. It will not eat from initial RAM.

On the other hand, if we also m,erge the custom memory alloc ummalloc which does best fit allocation, allocating on heap might work. 
Bonus for using ummalloc: you can set custom callback in case mem alloc fails so you don't have to check always the pointer retrurned by new which will save you some code. (in this commit checking is not done)

slaff:
@ADiea thank you for the well-reasoned opinion. 

My plan is to release the next version before or on Christmas so I will wait until 21of December before I start with the final preparations.

slaff:
Custom Heap with UMM Malloc can also be part of the new release:  #835

ADiea:
Offtopic: @slaff thanks for latest work reviving sming nonos. 

I'm currently waiting for new pcbs and will start resynching my local branch of sming to the latest when I will have new hardware assembled. I think that the 3rd party additions is the most powerfull change that will provide better maintainability in the future!

slaff:
>  I think that the 3rd party additions is the most powerfull change that will provide better maintainability in the future!

@ADiea Glad to hear that! I am waiting also for great new additions from your local branch :)

PR #835, opened 2016-12-20
Ported the custom heap integration with umm_malloc from SmingTasty (h…

slaff:
…ttps://github.com/slaff/Sming) to main-line Sming.

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -142,6 +142,14 @@ else ifneq ($(SMING_RELEASE),1)
 	MODULES += gdb
 endif
 
+# => umm_malloc (custom heap allocation)
+ifeq ($(ENABLE_CUSTOM_HEAP), 1)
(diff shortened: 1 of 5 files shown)


Sming/Makefile
```diff
@@ -280,6 +288,11 @@ $(USER_LIBDIR)/libaxtls.a:
 	$(Q) $(MAKE) -C third-party/axtls-8266 -e V=$(V) BIN_DIR="$(SMING_HOME)/$(USER_LIBDIR)"
 endif
 
+$(USER_LIBDIR)/libmainmm.a: $(addprefix $(SDK_LIBDIR)/,libmain.a)
```
ADiea: should be guarded by  
ifeq ($(ENABLE_CUSTOM_HEAP), 1)
endif ?

slaff: That is not necessary, but will make things easier to follow/understand. So I will add the change in a minute...

PR #838, opened 2016-12-27
Doxygen check

riban-bw:
Check that doxygen is installed to avoid ugly errors when calling make docs (or make api).


Sming/Makefile
```diff
@@ -319,6 +320,9 @@ docs/wiki/Home.md:
 	$(Q) $(GIT) submodule update --init ../docs/wiki	
 
 docs/api/error.log:
+ifndef DOXYGEN
+	$(error doxygen not found - not building API docs)
+endif
```
slaff: Probably you need `else` here otherwise the code below will be executed anyway.

riban-bw: I thought that too but the `$(error...)` command throws a make error, hence terminating make.

PR #853, opened 2017-01-05
Update DS18S20.cpp and .h

Ogyb:
Added callback on end of conversion.
in DS18S20.h File
public section:

	void RegisterEndCallback(void (*)()); //End conversion "interrupt" function register
	void UnRegisterCallback(); //Delete "interrupt" [end conversion]
private

void (*ReadEndCallback)() = 0;

file DS18s20.h in separate pull...

### DIFF
--- a/Sming/Libraries/DS18S20/ds18s20.cpp
+++ b/Sming/Libraries/DS18S20/ds18s20.cpp
@@ -150,6 +150,10 @@ void DS18S20::StartReadNext()
    {
 	   debugx("  DBG: DS18S20 reading task end");
 	   InProgress=false;
+	   if(ReadEndCallback > 0)
+	   {
+		   ReadEndCallback();
+	   }
    }
 }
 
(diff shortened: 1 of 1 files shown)


Sming/Libraries/DS18S20/ds18s20.cpp
```diff
@@ -255,3 +259,12 @@ uint8_t DS18S20::GetSensorsCount()
 
 }
 
+void DS18S20::RegisterEndCallback(void (*Callback)())
```
slaff: @Ogyb Can you replace the callback with Delegates. This way we can have the old C style callback and also method calling from objects.

flexiti:
@Ogyb 
Not so fast please
Do not make changes that are not compatible with are already running programs! (Eg. sample from sming example, user programs etc.)
At most add new functionality only

slaff:
Please, update in one PR the header and the C++ files. A PR should contain a change that compiles and works on the device.

flexiti:
@Ogyb 

> Its compatible with are already running programs! :)
> I no delete any functionality, only add 2 function and 1 varible.

that's  good :) I saw that one line had left, and I did not notice that  is back
This is library so backwards compatibility is important  :)

Ogyb:
2 file in 1 pull.

PR #859, opened 2017-01-07
operator== to compare web sockets

johndoe8967:
in some web socket applications it is useful to relay an incoming message from a client to all other clients.

Without possibility to compare the sender with the clients list this is difficult.

Also the operator== is missing in the method
WebSocketsList.removeElement(const WebSocket & obj)


johndoe8967:
sorry, wrong branch to deliver easily

Issue #876, opened 2017-01-17
printf: width and precision not working for string output?


tius2000:
I tried to make CommandHandler output look nicer. However, It seems to me that the width and precision specifiers for printf (e. g. %16s, %10.10s, %-10.10s) are ignored for strings.

ADiea:
Hi @tius2000 indeed %s does not support yet width specifiers
Please see \Sming\Sming\system\m_printf.cpp for refference

tius2000:
Thank you very much @ADiea! I tried to add support for width and precision to m_printf.cpp: 

    diff --git a/Sming/system/m_printf.cpp b/Sming/system/m_printf.cpp
    index 58ada9d..62b0da8 100644
    --- a/Sming/system/m_printf.cpp
    +++ b/Sming/system/m_printf.cpp
    @@ -19,6 +19,8 @@ void (*cbc_printchar)(char ch) = uart_tx_one_char;
     
     #define is_digit(c) ((c) >= '0' && (c) <= '9')
     
    +#define MIN(a, b)   ( (a) < (b) ? (a) : (b) )
    +
     static int skip_atoi(const char **s)
     {
        int i = 0;
    @@ -148,11 +150,14 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args)
            width = 0;
            pad = ' ';
            base = 10;
    +        bool minus = 0;
     
            do
            {
    +            if ('-' == *fmt) minus = 1, fmt++;
    +            
                //skip width and flags data - not supported
    -			while ('+' == *fmt || '-' == *fmt || '#' == *fmt || '*' == *fmt || 'l' == *fmt)
    +			while ('+' == *fmt || '#' == *fmt || '*' == *fmt || 'l' == *fmt)
                    fmt++;
     
                if (is_digit(*fmt)) {
    @@ -179,20 +184,22 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args)
                *str++ = (unsigned char) va_arg(args, int);
                continue;
     
    -		case 's':
    +		case 's': {
                s = va_arg(args, char *);
     
    -			if (!s)
    -			{
    -				s = "<NULL>";
    -			}
    -			else
    -			{
    -				while (*s && (maxLen - (uint32_t)(str - buf) > OVERFLOW_GUARD))
    -					*str++ = *s++;
    -			}
    +			if (!s) s = "(null)";
    +            size_t len = strlen(s);
[...]

PR #880, opened 2017-01-20
m_printf: width and precision supported added for format %s

tius2000:
@ADiea, could you please review the patch? This is my very first PR and I hope I got it right ;-) 

See #876 for a small test case.


tius2000:
Thank you very much @ADiea!

PR #881, opened 2017-01-20
Stationclass 1701

tius2000:
StationClass: added return value to connect() and disconnect()


tius2000:
got the commits wrong ...

PR #886, opened 2017-01-23
Fix possible TcpConnection memory leak

gbl08ma:
I have finally made my Sming fork public and I'm slowly creating pull requests for my changes.

Some months ago, I had to deal with a memory leak in my project, which I traced back to the creation and immediate destruction of a `HttpClient`, and ultimately the `TcpConnection` it inherits from. The leaked memory was allocated by `tcp_new`. I figure nobody noticed this before because this is an extremely strange use case (in my case, it happens because some of my classes contain HTTP clients but don't always use them before getting destroyed).

I fixed the problem by only initializing the `TcpConnection` when it is actually used. Now, `tcp_new` is only called when `connect` or `TcpServer::listen` are called, not when the class is instantiated.

### DIFF
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -32,10 +32,12 @@ TcpConnection::~TcpConnection()
 	autoSelfDestruct = false;
 	close();
 
+#ifdef ENABLE_SSL
 	if(sslFingerprint) {
 		delete[] sslFingerprint;
 	}
 	freeSslClientKeyCert();
+#endif
 	debugf("~TCP connection");
 }
 
@@ -47,9 +49,9 @@ bool TcpConnection::connect(String server, int port, boolean useSsl /* = false *
 	ip_addr_t addr;
 
 	this->useSsl = useSsl;
+#ifdef ENABLE_SSL
 	this->sslOptions |= sslOptions;
 
-#ifdef ENABLE_SSL
 	if(ssl_ext == NULL) {
 		ssl_ext = ssl_ext_new();
 		ssl_ext->host_name = (char *)malloc(server.length() + 1);
@@ -84,7 +86,9 @@ bool TcpConnection::connect(IPAddress addr, uint16_t port, boolean useSsl /* = f
 		initialize(tcp_new());
 
 	this->useSsl = useSsl;
+#ifdef ENABLE_SSL
 	this->sslOptions |= sslOptions;
+#endif
 
 	return internalTcpConnect(addr, port);
 }
@@ -349,7 +353,7 @@ void TcpConnection::closeTcpConnection(tcp_pcb *tpcb)
 
 void TcpConnection::flush()
 {
-	if (tcp->state == ESTABLISHED)
+	if (tcp && tcp->state == ESTABLISHED)
 	{
 		//debugf("TCP flush()");
 		tcp_output(tcp);
@@ -632,6 +636,7 @@ void TcpConnection::staticDnsResponse(const char *name, ip_addr_t *ipaddr, void 
 	delete dlook;
 }
 
+#ifdef ENABLE_SSL
 void TcpConnection::addSslOptions(uint32_t sslOptions) {
 	this->sslOptions |= sslOptions;
 }
@@ -699,7 +704,6 @@ void TcpConnection::freeSslClientKeyCert() {
 	clientKeyCert.certificateLength = 0;
 }
 
-#ifdef ENABLE_SSL
 SSL* TcpConnection::getSsl() {
 	return ssl;
 }
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -80,6 +80,8 @@ bool TcpConnection::connect(String server, int port, boolean useSsl /* = false *
 
 bool TcpConnection::connect(IPAddress addr, uint16_t port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
 {
+	if (tcp == NULL)
+		initialize(tcp_new());
 
 	this->useSsl = useSsl;
 	this->sslOptions |= sslOptions;
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -19,7 +19,7 @@
 
 TcpConnection::TcpConnection(bool autoDestruct) : autoSelfDestruct(autoDestruct), sleep(0), canSend(true), timeOut(70)
 {
-	initialize(tcp_new());
+
 }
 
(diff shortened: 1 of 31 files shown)


ADiea:
Hi I think best approach would be to destroy it in the class destructor. It could be that the connection is needed in other cases than connect or listen.

gbl08ma:
Yes, I know, and at the time I tried to do just that. For some reason it didn't work (my memory it's quite fuzzy, I worked on this in September). Now that I'm looking at the source code again...

`~TcpConnection` calls `close()`, which does something to make LWIP close the connection, including "reset pointer to close connection on next callback". The problem, if I recall correctly, was that if the connection is never used, then there's no "next callback", and the LWIP side of things never gets properly freed. Perhaps `tcp_close` or `tcp_abort` should be used instead, but then I don't know why they weren't used in the first place. Perhaps there's something that depends on the connection not being closed immediately as the TcpConnection is destroyed?

Sming/SmingCore/Network/TcpServer.cpp
```diff
@@ -85,7 +85,7 @@ void TcpServer::setTimeOut(uint16_t waitTimeOut)
 bool TcpServer::listen(int port)
 {
 	if (tcp == NULL)
-		tcp = tcp_new();
+		initialize(tcp_new());
```
slaff: I am missing something. Where is the tcp initialization when the TcpClient is used to connect?

gbl08ma: It happens here, on `connect`: https://github.com/gbl08ma/Sming/blob/4fd2fe5356fd6cfb6431bf85e0276cbdb5de7133/Sming/SmingCore/Network/TcpConnection.cpp#L45

...but I just realized that I missed the overload that takes an IP address. It should check if `tcp == NULL` there too and initialize a tcp_new if yes. I'll add the two missing lines...

gbl08ma:
I still think it would be cleaner if we could just properly free the resources allocated by `tcp_new` on the destructor, even when the connection is never active. The way the connection is closed now, I think it requires that an attempt to connect (or even a successful connection?) be made for the memory to be freed. Hence why I moved `tcp_new` to the `connect` function, but it's not elegant.

slaff:
> ... but it's not elegant.

That is what bothers me too. Your PR will fix the rare cases when someone creates an instance of the TcpClient and deletes that instance without `connect`ing or `listen`ing. Which is ok.
**But** it still does not fix a more important problem which is: Why the destructor does not free the `tcp` memory . Please, investigate some time and see how to close and free the `tcp` in the destructor.

gbl08ma:
As I mentioned earlier, `TcpConnection::close()` does this strange thing where instead of simply calling LWIP's `tcp_close` (which, if not connected, would free the resources immediately, or if connected, put the connection in a closing state and free them once closed), it does
```
tcp_poll(tcp, staticOnPoll, 1);
tcp_arg(tcp, NULL); // reset pointer to close connection on next callback
```
I'm not sure what are the implications of this, but my first impression is that it allows the connection to close "asynchronously", i.e. at a later time after the TcpConnection is destroyed. There has to be a reason why it was made like this in the first place, instead of simply calling `tcp_close`. Perhaps it solved some obscure bug someone had run into, many years ago? Or it could be copied from low-quality Espressif examples?

If there's some kind of different behavior introduced by these two lines, it's possible there's code relying on it. Once I have time to work on this again, I guess I can try using `tcp_close` there instead of this strange `tcp_poll`/`tcp_arg` combination, but it will require extensive testing to make sure everything keeps working correctly.

slaff:
@anakod Can you help us understand why the [TcpConnection destructor is calling `close()`](https://github.com/SmingHub/Sming/blame/develop/Sming/SmingCore/Network/TcpConnection.cpp#L33) instead of closing `tcp` with calling `tcp_close(tcp)` ? See the above comment for details.

gbl08ma:
I mean, the problem is not that the destructor calls `close()`, since some other things besides the TCP stuff is freed there (including SSL stuff), the problem is the way `close()` is freeing (or not) the resources.

anakod:
@gbl08ma you are right, it was for asynchronous resource releasing, But unfortunately I forgot why it was important. May be async close important only in case when connection already done and on other cases we can close tcp handle directly. Anyway it's need deep testing. Your original idea with lazy call for tcp_new isn't so bad also in my opinion (an provide a little additional memory optimisation for some cases), but we should check what no direct access to this member can fire null reference exception.

hreintke:
This is the way to be used in raw LWIP to have the tcp layer closed correctly.
See LWIP doc's for more details.

slaff:
@gbl08ma please, rebase on develop and I will merge the PR if nothing else pops up.

gbl08ma:
I used the GitHub website to solve the conflict, so even though it was a
pretty simple change, I'm not sure it compiles. Either confirm it yourself
(or wait for CI), or wait for me to get access to my ESP8266 dev
environment.

Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -85,7 +85,8 @@ bool TcpConnection::connect(IPAddress addr, uint16_t port, boolean useSsl /* = f
 	if (tcp == NULL)
 		initialize(tcp_new());
 
-	this->useSsl = useSsl;
+  this->useSsl = useSsl;
```
slaff: Coding Style: Please, fix the line indentation.

Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -82,8 +82,11 @@ bool TcpConnection::connect(String server, int port, boolean useSsl /* = false *
 
 bool TcpConnection::connect(IPAddress addr, uint16_t port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
 {
+	if (tcp == NULL)
+		initialize(tcp_new());
+
+  this->useSsl = useSsl;
```
slaff: If I am not mistaken this line has to be inside of the #ifdef ENABLE_SSL block. Not before it.

gbl08ma:
Sigh, GitHub is far from perfect. In some hours I'll fix it with a good old
git and text editor combo...

gbl08ma:
I think that in my other PR about fixing non-SSL memory usage, I made
useSSL still react even on builds without SSL. I don't remember exactly
why, I think it was because of the "SSL disabled" warning. Anyway that's
why I put it out of the ifdef.

On Jan 30, 2017 17:18, "Gabriel Maia" <gbl08ma@gmail.com> wrote:

Sigh, GitHub is far from perfect. In some hours I'll fix it with a good old
git and text editor combo...

gbl08ma:
Now (after some confusion in my fork because apparently my local copy is configured so that `upstream` is the default remote instead of `origin`, need to reconfigure remotes so I don't mess up again), it is finally ready:

- Identation is fixed
- `this->useSsl = useSsl;` is exactly where it should be. `useSsl` is checked even on builds where SSL is disabled and thus should be outside of the #ifdef.

If you want you can merge this now, for the lack of a cleaner solution...

Issue #830, opened 2016-12-11
Increased TCP connection memory usage even with SSL disabled


gbl08ma:
Recently, SSL support was added to Sming. This is a good addition, and because it noticeably increases memory usage, it can be disabled at compilation time. However, I noticed that my programs would use more memory when linked against the develop branch, **even with SSL disabled**, so I decided to investigate. Turns out that some classes like TcpConnection just became much larger (in terms of memory usage) due to SSL, even when SSL support is disabled.

This is what the TcpConnection class holds:

```
protected:
	tcp_pcb *tcp = NULL;
	uint16_t sleep;
	uint16_t timeOut;
	bool canSend;
	bool autoSelfDestruct;
#ifdef ENABLE_SSL
	SSL *ssl = nullptr;
	SSLCTX *sslContext = nullptr;
	SSL_EXTENSIONS *ssl_ext=NULL;
#endif
	boolean useSsl = false;
	uint8_t *sslFingerprint=null;
	boolean sslConnected = false;
	uint32_t sslOptions=0;
	SSLKeyCertPair clientKeyCert;
	boolean freeClientKeyCert = false;
```

As you can see, there are lots of SSL-related stuff outside of the `ENABLE_SSL` preprocessor conditional. I tried to move these into the conditional, making the appropriate changes to TcpConnection.cpp, and quickly found out that lots of other classes (e.g. the HTTP and MQTT clients) expect some of the SSL stuff to exist even with SSL support disabled. A more thorough analysis and refactoring of the code would be needed to make non-SSL builds use as little memory as before SSL was added to Sming.

As things are now, each TCP connection consumes at least 18 more bytes of RAM than previously, with things that are useless because SSL is not enabled.

I have another idea: would it be possible to refactor SSL-enabled TCP connections into another class that inherits from TcpConnection? This would allow for non-SSL connections to use less RAM even on builds with SSL enabled. However, some more methods may need to become virtual and I'm not sure what impact this will have on performance.

slaff:
> However, I noticed that my programs would use more memory when linked against the develop branch, even with SSL disabled

Can you be more specific: which memory do you mean? Is it the flash memory, is the heap is it the IRAM size? Please, include in your answer also the tools/commands that you've used to measure that.


> A more thorough analysis and refactoring of the code would be needed to make non-SSL builds use as little memory as before SSL was added to Sming.

You are welcome to provide fixes that improve this :)

gbl08ma:
[...]

PR #888, opened 2017-01-23
Reduce TcpConnection heap usage on builds with SSL disabled

gbl08ma:
On issue #830, I described the problem: even when ENABLE_SSL is not defined, the TcpConnection class still contains some members related to SSL that are completely useless and wasted about 18 bytes of heap per connection.

My changes basically consist on moving the position of the `#ifdef`s so that all of the SSL stuff is disabled when Sming is compiled without SSL support.

It was really simple to do this, which makes me worried: am I missing the reason why it wasn't done like this from the beginning? My code seems to still work fine...

### DIFF
--- a/Sming/SmingCore/Network/HttpClient.h
+++ b/Sming/SmingCore/Network/HttpClient.h
@@ -61,11 +61,11 @@ public:
 
 	void reset(); // Reset current status, data and etc.
 
+#ifdef ENABLE_SSL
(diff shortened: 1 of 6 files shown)


slaff:
@gbl08ma Have you checked the changes on a real device? Can you test Basic_Ssl and Mqtt_Hello samples with ENABLE_SSL=1 please?

PR #893, opened 2017-01-25
rBootHttpUpdate: Hooks added for signed and/or encrypted OTA

tius2000:
_rboot_write_init()_, _rboot_write_flash()_ and _rboot_write_end()_ moved to virtual methods. This allows a subclass to implement hooks to download signed or encrypted firmware files. 

Hope I got it right ;-)

### DIFF
--- a/Sming/SmingCore/Network/rBootHttpUpdate.cpp
+++ b/Sming/SmingCore/Network/rBootHttpUpdate.cpp
@@ -60,7 +60,10 @@ void rBootHttpUpdate::onTimer() {
 	if (TcpClient::getConnectionState() == eTCS_Successful) {
 
 		//  always call writeEnd()
-		if (!writeEnd()) writeError = 1;
+		if (!writeEnd()) {
+			debugf("final checks failed!");
+            writeError = 1;
+        }    
 		
 		if (!isSuccessful()) {
 			updateFailed();
--- a/Sming/SmingCore/Network/rBootHttpUpdate.cpp
+++ b/Sming/SmingCore/Network/rBootHttpUpdate.cpp
@@ -51,9 +51,6 @@ void rBootHttpUpdate::updateFailed() {
 }
 
 void rBootHttpUpdate::onItemDownloadCompleted(HttpClient& client, bool successful) {
-	if(successful) {
-		rboot_write_end(&rBootWriteStatus);
-	}
 }
 
 void rBootHttpUpdate::onTimer() {
@@ -61,6 +58,9 @@ void rBootHttpUpdate::onTimer() {
 	if (TcpClient::isProcessing()) return; // Will wait
 	
 	if (TcpClient::getConnectionState() == eTCS_Successful) {
+
+		//  always call writeEnd()
+		if (!writeEnd()) writeError = 1;
 		
 		if (!isSuccessful()) {
 			updateFailed();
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Network/rBootHttpUpdate.h
```diff
@@ -50,6 +50,10 @@ class rBootHttpUpdate: private HttpClient {
 	void updateFailed();
 	void onItemDownloadCompleted(HttpClient& client, bool successful);
 
+    virtual void writeInit();
```
tius2000: I think you are right! I will move the methods to protected and test again.

tius2000:
Unfortunately, this is not so easy for me to provide a full working example at the moment. In order to get OTA running with the current SDK under Windows, I had to make changes in many other places. Therefore, my version is currently not portable.

However, using the hooks is quite easy. My own code currently looks like this:

    #define PREFIX_MAGIC    0x54, 0x49, 0x55, 0x53
    #define PREFIX_TYPE     0x00, 0x01
    #define PREFIX_SIZE     sizeof(_my_prefix)
    #define SIGNATURE_SIZE  64

    const u8 _my_prefix[6] = { PREFIX_MAGIC, PREFIX_TYPE };    

    typedef struct {
        u8  prefix[PREFIX_SIZE];
        u8  signature[SIGNATURE_SIZE];
    } MyHdr;

    //-----------------------------------------------------------------------------
    class MyUpdate : public rBootHttpUpdate {

    protected:
        virtual void writeInit();
        virtual bool writeFlash(const u8 *data, u16 size);
        virtual bool writeEnd();
        
    private:
        Sha256 *sha256;
        u8      hdr_len;
        MyHdr   hdr;
    };

    //-----------------------------------------------------------------------------
    void MyUpdate::writeInit() {
        rBootHttpUpdate::writeInit();
        sha256  = new Sha256;
        hdr_len = 0;
    }

    bool MyUpdate::writeFlash(const u8 *data, u16 size) {
        //  store header
        u8 missing = sizeof(hdr) - hdr_len;
        if (missing) {
            if (size < missing) missing = size;
            memcpy( &hdr, data, missing );
            size    -= missing;
            data    += missing;
            hdr_len += missing;
            
            //  check prefix
            if ( hdr_len >= PREFIX_SIZE ) {
                if ( memcmp(hdr.prefix, _my_prefix, PREFIX_SIZE) ) {
                    debugf("invalid prefix");
                    return 0;
                }
            }
        }

        //  update hash
        sha256->update(data, size);

        //  save data
        return rBootHttpUpdate::writeFlash(data, size);
    }

    bool MyUpdate::writeEnd() {
        if (!rBootHttpUpdate::writeEnd()) return 0;

        u8 hash[SHA256_BLOCK_SIZE];
        sha256->final( hash );

        bool sig_ok = /* add signature check here */;
        if (!sig_ok) {
            debugf("wrong signature");
            return 0;
        }
        return 1;
    }




Hope this helps?

slaff:
> Unfortunately, this is not so easy for me to provide a full working example

@tius2000 Ok, I will consider this PR for the next version but you should add that information/example to the WIKI or to the code documentation.

Sming/SmingCore/Network/rBootHttpUpdate.cpp
```diff
@@ -51,16 +51,16 @@ void rBootHttpUpdate::updateFailed() {
...
 }
 
 void rBootHttpUpdate::onTimer() {
 	
 	if (TcpClient::isProcessing()) return; // Will wait
 	
 	if (TcpClient::getConnectionState() == eTCS_Successful) {
+
+		//  always call writeEnd()
+		if (!writeEnd()) writeError = 1;
```
slaff: @tius2000 May be a debug message on write error will be helpful here.

tius2000:
I moved the sample code to the wiki (https://github.com/SmingHub/Sming/wiki/Signed-OTA). May be this is useful for someone.

Sming/SmingCore/Network/rBootHttpUpdate.h
```diff
@@ -57,6 +57,10 @@ class rBootHttpUpdate: private HttpClient {
 	rboot_write_status rBootWriteStatus;
 	uint8 romSlot;
 	otaUpdateDelegate updateDelegate;
+
+    virtual void writeInit();
```
slaff: Coding Style: Please, align the lines with the ones above.

tius2000: Well, they are aligned in my editor - you are right, need to change my tab settings ;-)

Also changed the base class to protected to allow better http error handling within derived classes.

The git workflow is still very new to me and I really appreciate any feedback!

PR #897, opened 2017-01-26
Fix WebSocket key compare for some servers

ADiea:
Sometimes server returns lowercase or different case headers, [allowed by RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2). Key will not match then. Happened on Heroku.

- compare key header lowercase for websocket
- small fix check TCP pointer causes reboot sometimes
- make Netutils::debugPrintTcpList static to call it without object instance


Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -347,7 +347,7 @@ void TcpConnection::closeTcpConnection(tcp_pcb *tpcb)
 
 void TcpConnection::flush()
 {
-	if (tcp->state == ESTABLISHED)
+	if (tcp && tcp->state == ESTABLISHED)
```
slaff: That is a change that needs to be moved to separate PR.

Sming/SmingCore/Network/NetUtils.cpp
```diff
@@ -180,3 +180,35 @@ void NetUtils::debugPrintTcpList()
 	if(prt_none) debugf("LWIP_DEBUG: none\r\n");
 	return;
 }
+
+//search lowercase
+char* NetUtils::strstrl(char* source, const char* sequence)
```
slaff: Why not just combine `strstr` and `tolower` ? You can see `Wiring/WString.cpp` for example usage.

Sming/SmingCore/Network/NetUtils.h
```diff
@@ -27,11 +27,12 @@ class NetUtils
 	static int pbufFindStr(pbuf *buf, const char* wtf, int startPos = 0);
 	static char* pbufAllocateStrCopy(pbuf *buf, int startPos, int length);
 	static String pbufStrCopy(pbuf *buf, int startPos, int length);
+	static char* strstrl(char* source, const char* sequence);
```
slaff: This method does not have place here.

Sming/SmingCore/Network/WebsocketClient.cpp
```diff
@@ -101,17 +103,35 @@ void WebsocketClient::onError(err_t err)
...
+	if(!endKey || endKey - buf > size)
+	{
+		debugf("wscli cannot find key(2) reason:%s", endKey ? "out of bounds":"NULL");
+		return false;
+	}
+	//terminate serverKey
+	*endKey = 0;
+	debugf("wscli search key %s", serverKey);
+
+
```
slaff: Coding Style: One empty new line as a separator should be enough.

Sming/SmingCore/Network/NetUtils.h
```diff
@@ -27,11 +27,12 @@ class NetUtils
...
 
 	static bool FixNetworkRouting();
 
 	// Debug
-	void debugPrintTcpList();
+	static void debugPrintTcpList();
```
slaff: Is this change related to Websocket Key comparison? If not - please move it to another PR.

ADiea:
Remade the PR, added indexOfIgnoreCase to String class

Sming/SmingCore/Network/TcpConnection.cpp
ADiea: I have rebased from l;atest dev and somehow it still sees this as modified although the modified version is already in develop

Sming/SmingCore/Network/WebsocketClient.cpp
```diff
@@ -99,10 +100,11 @@ void WebsocketClient::onError(err_t err)
 
 bool WebsocketClient::_verifyKey(char* buf, int size)
 {
-	String dd = String(buf);
-	uint8_t s = dd.indexOf("Sec-WebSocket-Accept: ");
+	String keyHeader("Sec-WebSocket-Accept: ");
+	String dd(buf);
```
slaff: This code assumes that the header will be in one tcp packet, which is wrong. It may cause broken validation.

```
switch (_mode)
		{
		case wsMode::Connecting:
			if (_verifyKey((char*)data, size) == true)
```

slaff: Coding Stype: Replace `dd` with `headerValue`, `s` with `startPos`, `t` with `endPos`.

Sming/SmingCore/Network/WebsocketClient.cpp
```diff
@@ -99,10 +100,11 @@ void WebsocketClient::onError(err_t err)
...
+	uint8_t s = dd.indexOfIgnoreCase(keyHeader);
```
slaff: `s` and `t` can have -1 as value. Replace the type from uint8_t to int.

Sming/SmingCore/Network/WebsocketClient.cpp
```diff
@@ -99,10 +100,11 @@ void WebsocketClient::onError(err_t err)
...
 	uint8_t t = dd.indexOf("\r\n", s);
```
slaff: For `s` and `t` - add checks if the return value is `-1` and return false in that case.

Sming/SmingCore/Network/WebsocketClient.cpp
slaff: Don't add new method(s) for something that can be achieved with the existing ones. For example you can do something like this:

```
bool WebsocketClient::_verifyKey(char* buf, int size)
{
	String keyHeader("sec-websocket-accept: ");
        String headerValue(buf);
	String headerValueLC = headerValue;
	headerValueLC.toLowerCase();
	int startPos = headerValueLC.indexOf(keyHeader);
	if(startPos == -1) {
		return false;
	}

        // ...
```

Sming/Wiring/WString.h
```diff
@@ -222,6 +222,8 @@ class String
     int indexOf(char ch, unsigned int fromIndex) const;
     int IRAM_ATTR indexOf(const String &str) const;
     int indexOf(const String &str, unsigned int fromIndex) const;
+    int IRAM_ATTR indexOfIgnoreCase(const String &s2) const;
```
slaff: Same as last comment.

Sming/SmingCore/Network/WebsocketClient.cpp
ADiea: Please keep in mind buf could be 1500bytes(MTU) packet You will create a new string  = heap + make tolower on it. I think this is not optimal that is why I created new search function to ignore case

ADiea: This is why in the first version I prefered to work directly with char* and not involve string copy constructor. We are playing with 1.5KB of heap alloc which you know it's not always available. I think I will revert to first solution of working directly on the buffer.

Sming/system/include/stringutil.h
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *  Contains utility functions for working with char strings.
+ *
+ *  Created on: 26.01.2017
+ *  Author: (github.com/)ADiea
+ */
+
+#ifndef INCLUDE_STRINGUTIL_H_
+#define INCLUDE_STRINGUTIL_H_
+
```
slaff: @ADiea Can you add also C linkage so that these functions are also available in C code?

ADiea: done

ADiea:
strstri unit test is [here](http://ideone.com/drrMPh)

not all squashes were sucessfull sorry for the big number of commits

slaff:
@ADiea  Thanks for your efforts and for the unit test. Keep up the good work!

PR #898, opened 2017-01-26
Fix TCPConnection::flush()

ADiea:
Null ptr sometimes reboots device.

### DIFF
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -347,7 +347,7 @@ void TcpConnection::closeTcpConnection(tcp_pcb *tpcb)
 
 void TcpConnection::flush()
 {
-	if (tcp->state == ESTABLISHED)
+	if (tcp && tcp->state == ESTABLISHED)
 	{
 		//debugf("TCP flush()");
 		tcp_output(tcp);


ADiea:
@avr39-ripe if I remember correctly you removed connection->flush() from WebSocket::send() you said it improves stability with cost of little latency.
Could you repeat your tests after this PR? This PR fixed a case where flush was using null TCP ptr => if your tests are passing we can readd flush to send() and improve latency again

avr39-ripe:
@ADiea reason, that I remove ftush() was caused by instantly heap reduction with in some cases lead to reboots, so this fix can't by design correct this behavior. By removing flush several subsequent websocket frames goes into one tcp packet (as far as I understand how it goes) so for first websocket frame some heap allocated fro tcp packet buffer, then first,second,third and so on we frame in SHORT time goes inside single TCP packet and when TCP packe is "full of data" it goes out. With flush() in websocket.send each ws frame needs its OWN TCP packet and each TCP packet needs some heap to hold its buffer. Then some time needed for memory to be free again after TCP packet is send by issuing flush(), so subsequent ws frame if they go one  by one with very short interval will "eat" all esp memory.
As compromise we can add some flag to send method (false by default) and in case of true issue flush () inside send() which can be used if latency IS MATTER and if there are not many subsequent ws frames.

ADiea:
I'll try to take a look in the future on ws code maybe we can optimize the heap issue somehow by passing dirrectly the packet buffer from tcp and avoid reallocation. something like this was done already for key comparing by not allocating Strings and diong the compare directly on the buffer. Maybe we can optimize more

PR #901, opened 2017-01-27
Reverted the previous commit (434ef0d60420940ede75bf895736a22bd8549f2…

slaff:
…3) because it broke the build

with esp-alt-sdk and esp-open-sdk on Linux and Mac OS X.
If there is a problem with the Windows built an issue needs to be provided with information before breaking the build process for other platforms.

### DIFF
--- a/Sming/Services/SpifFS/spiffs_config.h
+++ b/Sming/Services/SpifFS/spiffs_config.h
@@ -15,15 +15,13 @@
 #ifdef __ets__
 #include <user_config.h>
 #include "../system/flashmem.h"
-#include <stdint.h>
 #else
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


anakod:
Hello. I can't build Sming on Windows anymore :(

slaff:
@anakod Please, tell me what SDK are you using and what is the issue that you are getting. And we need to provide fix for that SDK.

anakod:
replied opn commit:

Can you look how declared ‘intptr_t’ in "/usr/include/unistd.h".
I think it should resolve issue.

Windows platform didn't have <stdint.h> file, I think. As result I can't compile spiffy utility anymore. It isn't issue with compiling Firmware for ESP, but it's issue with compiling command line utility. So I think, probably it shouldn't be depended on SDK version.

PR #903, opened 2017-01-27
Make default debug uart speed configurable.

ADiea:
For debug purposes it's better to use a bigger baudrate such as 921600.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -51,7 +51,7 @@ a [custom PWM library](https://github.com/StefanBruens/ESP8266_new_pwm) will be 
 - Custom Heap Allocation: If your application is experiencing heap fragmentation then you can try the Umm Malloc heap allocation. To enable it compile
 Sming with ENABLE_CUSTOM_HEAP=1. In order to use it in your sample/application make sure to compile the sample with ENABLE_CUSTOM_HEAP=1. Avoid enabling
 your custom heap allocation AND -mforce-l32 compiler flag.
-
+- Custom serial baud rate: By default serial debug interface baud is 115200. You can set COM_SPEED_SERIAL to alter that for ex: COM_SPEED_SERIAL=921600
 
 You can find more information about compilation and flashing process by reading esp8266.com forum discussion thread.
 
--- a/Readme.md
+++ b/Readme.md
@@ -51,7 +51,7 @@ a [custom PWM library](https://github.com/StefanBruens/ESP8266_new_pwm) will be 
 - Custom Heap Allocation: If your application is experiencing heap fragmentation then you can try the Umm Malloc heap allocation. To enable it compile
 Sming with ENABLE_CUSTOM_HEAP=1. In order to use it in your sample/application make sure to compile the sample with ENABLE_CUSTOM_HEAP=1. Avoid enabling
 your custom heap allocation AND -mforce-l32 compiler flag.
-- Custom serial baud rate: By default serial debug interface baud is 115200. You can set COM_SPEED_SERIAL to alter that for ex: COM_SPEED_SERIAL=921600
+- Custom serial baud rate: The default serial baud rate is 115200. If you want to change it to a higher baud rate you can recompile Sming and your application changing the COM_SPEED_SERIAL directive. For example COM_SPEED_SERIAL=921600
 
 You can find more information about compilation and flashing process by reading esp8266.com forum discussion thread.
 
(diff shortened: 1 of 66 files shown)


Sming/appinit/user_main.cpp
```diff
@@ -28,6 +28,8 @@ extern "C" void  __attribute__((weak)) user_init(void)
 // For compatibility with SDK v1.1
 extern "C" void __attribute__((weak)) user_rf_pre_init(void)
 {
+	uart_div_modify(UART_ID_0, UART_CLK_FREQ / SERIAL_BAUD_RATE);
```
ADiea: set uart speed earlier

Sming/include/user_config.h
```diff
@@ -6,7 +6,7 @@ extern "C" {
 #endif
 
 	// UART config
-	#define SERIAL_BAUD_RATE 115200
+	#define SERIAL_BAUD_RATE COM_SPEED_SERIAL
```
slaff: * Where is COM_SPEED_SERIAL defined? 
* Is `921600` good default speed? I would say no, because it is not supported by most of the serial adapters (http://digital.ni.com/public.nsf/allkb/D37754FFA24F7C3F86256706005B9BE7).  I would leave it to  `115200` which is the fastest `supported` speed by most of the devices.

ADiea: Default is now left at 115200. The fix just removes hardcoded value and takes into consideration COM_SPEED_SERIAL which can be given as make parameter to override

Makefile-project.mk
```
## COM port parameters
# Default COM port speed (generic)
COM_SPEED ?= 115200

# Default COM port speed (used for flashing)
COM_SPEED_ESPTOOL ?= $(COM_SPEED)

COM_SPEED_SERIAL  ?= $(COM_SPEED)
```

slaff: Looks good.

slaff:
@ADiea Please, check the compilation errors.

ADiea:
done thanks

Sming/appinit/user_main.cpp
```diff
@@ -12,7 +12,7 @@ extern void init();
 extern "C" void  __attribute__((weak)) user_init(void)
 {
 	system_timer_reinit();
-	uart_div_modify(UART_ID_0, UART_CLK_FREQ / 115200);
+
```
slaff: Coding Style: Please, remove the new line here.

Readme.md
```diff
@@ -51,7 +51,7 @@ a [custom PWM library](https://github.com/StefanBruens/ESP8266_new_pwm) will be
 - Custom Heap Allocation: If your application is experiencing heap fragmentation then you can try the Umm Malloc heap allocation. To enable it compile
 Sming with ENABLE_CUSTOM_HEAP=1. In order to use it in your sample/application make sure to compile the sample with ENABLE_CUSTOM_HEAP=1. Avoid enabling
 your custom heap allocation AND -mforce-l32 compiler flag.
-
+- Custom serial baud rate: By default serial debug interface baud is 115200. You can set COM_SPEED_SERIAL to alter that for ex: COM_SPEED_SERIAL=921600
```
slaff: May this should be improved a bit:

```
Custom serial baud rate: The default serial baud rate is 115200. If you want to change it to a higher baud rate you can recompile Sming and your application changing the COM_SPEED_SERIAL directive. For example COM_SPEED_SERIAL=921600
```

slaff:
@ADiea Another thing comes to my mind. In all samples under `include/user_config.h` we have 
```
// UART config
        #define SERIAL_BAUD_RATE 115200
```

may be those lines should be removed from all samples because they will conflict with setting SERIAL_BAUD_RATE from the command line, env or in the Makefile-user.mk during compilation.

ADiea:
I think maybe i should guard with ifndef serial baud rate

Readme.md
ADiea: Very soon we will have to write a simple html file with javascript like a tool to check what options you need and it will give you the make line because the number of make switches is growing...

ADiea:
About the issue with samples containing user_config.h also. It will be included from the sample and the speed will be hardcoded. The user_config.h from sming will not be included anymore because it is guarded by the same defines
```
#ifndef __USER_CONFIG_H__
#define __USER_CONFIG_H__
```
so no compilation errors will occur. 

I will modify all sample files to follow the COM_SPEED_SERIAL makefile parameter to be able to set it at make time as described in Readme.md. Default of course will remain 115200 for max compatibility..

Sming/appinit/user_main.cpp
slaff: @ADiea please, remove the newline between system_timer_reinit and cpp_core_initialize. After that rebase on develop and ping me to merge the PR.

slaff:
Will be merged with that rebase #918.

PR #905, opened 2017-01-29
fix lwip windows

ADiea:
Custom LWIP was not building on windows because make call did not specify target all result was: nothing to do

Lwip patch also contains pull requests [tcp fixes](https://github.com/pfalcon/esp-open-lwip/pull/5) and [fix warnings](https://github.com/pfalcon/esp-open-lwip/pull/6)

Also small cosmetisations, fixed excessive printing when VERBOSE=0 is passed to make.

Please also test on Linux/BSD and give feedback it does not break anything


ADiea:
sming\Sming\third-party\esp-open-lwip\lwip\app\dhcpserver.c has CRLF line-endings it should be converted to LF only linux style or else patches will not work the same way ... I will remove the patch for this file because it just fixes a warning

ADiea:
cleaned up patch, defined sint16_t waiting for new travis output

PR #908, opened 2017-01-30
Unified the PROGMEM handling.

slaff:


ADiea:
Hi @slaff I am also working to transfer my current sollution for debugging and preparing a PR to make all debugf use static const char* defined in irom section it's here [https://github.com/ADiea/Sming/commit/33e28bc911aa494a7720d06547a358cec4214344](https://github.com/ADiea/Sming/commit/33e28bc911aa494a7720d06547a358cec4214344)

Your PR defines PROGMEM but then probably you have to use PROGMEM before every char* and still have to use pgm_read_byte to read it and be carefull of alignment right? Did I understand correctly? Or is you PR doing more than this it is automating more?

slaff:
@ADiea I am trying to migrate all "tasty" features back into the upstream "develop". And this was one of them. Please, send me your solution as PR and I guess we can decide which one to choose (or may be merge them together).

slaff:
> .. Your PR defines PROGMEM but then probably you have to use PROGMEM before every char* ..

@ADiea  PROGMEM is only for static constant data. That data can be safely stored in the flash and read whenever needed. This will give us additional free heap.

slaff:
If the current compiler supports mforce-l32 then we should define the FakePgm functions to be the real ones. The compiler will take care to do the read/write magic for us.

PR #913, opened 2017-01-30
Ogyb thing speak

Ogyb:
ThingSpeak libary.

Usage:
```
ThingSpeak thingSpeak;
int fieldValue = 9;

thingSpeak.setWriteAPIKey("WriteApiKey");
thingSpeak.setField(1, String(fieldValue));` //set field1 of int value
thingSpeak.setField(2, String(90));`
thingSpeak.setField(3, "60");
thingSpeak.sendFields();
```

### DIFF
--- a/Sming/Libraries/ThingSpeak/README.md
+++ b/Sming/Libraries/ThingSpeak/README.md
@@ -1,7 +1,6 @@
-# Sming 
-## ThingSpeak driver libary
-### Usage
-#### Declatration
+# Sming - ThingSpeak driver libary
+## Usage
+### Declatration - init
 
 First way set write APIKey in the constructor.
 ```C++
@@ -11,6 +10,52 @@ second way is set in the function:
 
 ```C++
 ThingSpeak thingSpeak; //free constructor
-thingSpeak.setWriteAPIKey( APIKey ); // APIKey - string varible of Write API TS Key
+thingSpeak.setWriteAPIKey( APIKey ); // APIKey - String varible of Write API TS Key
 ```
-You can use **setWriteAPIKey** whenever you need, while changing API by user etc.
+You can use `setWriteAPIKey` whenever you need, while changing API by user etc.
+
+### Usage in loop
+#### Add a field value
+If you are adding value to the specific fields you can use `setField(iField, value)`
+* `iField` is field id (1-8)
+* `value` is value of field data (int, float, String)
+
+Or if you want to add a field sequence use `setNextField(value)`
+* `value` is value of field data (int, float, String)
+
+this function add a value of next field (1, 2, 3, 4, ...) if you use first `setField` and add a field to a specific number, eg 4
+function `setNextField` add value to next field (5)
+
+#### Send values for server
+for send all added values and clear fields to next mesures use `sendFields()`
+
+#### Clear fields
+to clear all field value use `cleanFields()`
+
+### Example
+
+```C++
+#include <SmingCore/SmingCore.h>
+#include <Libraries/ThingSpeak/ThingSpeak.h>
+
+Timer procTimer;
+ThingSpeak thingSpeak("YouTsAPIKey");
+uint16_t value;
+
+void readData() //Timered sensor read theard
+{
+	value = system_adc_read();      // Read value from ESP ADC
+	thingSpeak.setField(1, value);  // Add value to TS libary
+	thingSpeak.sendFields();        // Send all added values
+}
+
+void init()
+{
+	Serial.begin(SERIAL_BAUD_RATE);           // 115200 by default
+	Serial.systemDebugOutput(true);           // Allow debug output to serial
+	WifiStation.enable(true);                 // Enable WiFi
+	WifiStation.config(WIFI_SSID, WIFI_PWD);  // You pass to WiFi
+
+	procTimer.initializeMs(20 * 1000, readData).start(); 
+}
+```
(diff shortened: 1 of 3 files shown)


anakod:
> thingSpeak.setField(1, String(fieldValue));` //set field1 of int value

I think it isn't clear. Why you don't allow to user pass numbers directly? You can create overloads in your class for that cases.

anakod:
Also will be very good to have short descriptions (docs) for main methods, because it can be not very obvious at first look what user should call "writeFields" to send his request.

Ogyb:
Ok, i add overloads for other datatype (int float...)

I can write an example? or add comments to .h file?

slaff:
> I can write an example? or add comments to .h file?

@Ogyb Both would be great. But at least try to add documentation in the header file. Follow the guidelines from [here](
https://github.com/SmingHub/Sming/wiki/Documenting-the-API)

Ogyb:
maybe rename "writeFields" to "sendFields"

Ogyb:
overload function with template or normal overload?
```
template <typename T>
uint8_t setField(uint8_t iField, T value);
```
or
```
uint8_t setField(uint8_t, int);
uint8_t setField(uint8_t, unsigned int);
uint8_t setField(uint8_t, long);
...
```
?

anakod:
> overload function with template or normal overload?
Second option should be more simple solution for this case. And you don't need int, unsigned int, long and etc. Only different kind of types: long, float(or double), String.

Ogyb:
Add documentation and template for other datatype;)

Docs is ok? (in .h file)

slaff:
@Ogyb The PR is improving but I am still missing information what is this about. Can you add documentation comment(s) in the header file explaining a bit more or at least a link to the API documentation URL?

Ogyb:
i create markdown file;)

Sming/Libraries/ThingSpeak/README.md
```diff
@@ -0,0 +1,61 @@
+# Sming - ThingSpeak driver libary
```
slaff: The text needs style and syntax corrections. May be @riban-bw can help you here?

Sming/Libraries/ThingSpeak/ThingSpeak.cpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+#include "ThingSpeak.h"
+
+ThingSpeak::ThingSpeak()
+{
+	cleanFields();
+}
+
+ThingSpeak::ThingSpeak(String APIKey)
```
slaff: Coding Style: Rename APIKey to apiKey

Sming/Libraries/ThingSpeak/ThingSpeak.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+	writeAPIKey = APIKey;
```
slaff: Why is the class property named writeAPIKey and not just apiKey? Are there other API keys? If not the above code can be changed to
```
this->apiKey = apiKey
```

Sming/Libraries/ThingSpeak/ThingSpeak.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+}
+
+void ThingSpeak::setWriteAPIKey(String APIKey)
+{
+	writeAPIKey = APIKey;
+}
+
+//Used only for string value! Else data type use template (see .h file)
+template <> uint8_t ThingSpeak::setField(uint8_t iField, String value)
```
slaff: Please, rename `iField` to `id`

Sming/Libraries/ThingSpeak/ThingSpeak.cpp
```diff
@@ -0,0 +1,94 @@
...
+	if(httpClient.isProcessing())
+	{
+		return ERR_IS_PROCESING;
+	}
+	return httpStatus;
+}
+
+void ThingSpeak::cleanFields()
+{
+	for(uint8_t iField = 0; iField < TS_MAX_FIELD; iField++)
```
slaff: In the loop below you will end up with an array containing the old ids with empty values. Wouldn't it be better to completely clean the fields?

Sming/Libraries/ThingSpeak/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Usage
```
slaff: Before even starting to talk about usage you can add two or three sentences explaining what is ThingSpeak. You can quote the information from their website (https://mathworks.com/help/thingspeak/getting-started-with-thingspeak.html) "ThingSpeak is an IoT platform that enables you to collect, store, analyze, visualize, and act on data from sensors or actuators, such as Arduino®, Raspberry Pi™, BeagleBone Black, and other hardware. For example, with ThingSpeak you can create sensor-logging applications, location-tracking applications, and a social network of things with status updates, so that you could have your home thermostat control itself based on your current location."

Sming/Libraries/ThingSpeak/README.md
riban-bw: The Sming API documentation specifically excludes upstream libraries. Such libraries should be appropriately documented but we avoided imposing Sming coding style on external projects. We may want to include the libraries documentation in a Sming API documentation module in the future and having consistent coding style is a bonus. I do not see libraries as part of the core Sming code.

slaff:
The current PR can be an example of how to improve the initial code and refactor it. It also shows how you learned to use C++ templates.

Apart from this it has, IMHO, little value to a general IoT framework like Sming because it can be replace with these lines:
```
HttpClient client;
String postMessage = "http://api.thingspeak.com/update?key=" + apiKey;
postMessage += "&field1=" + fieldValue[0]; //...
httpClient.downloadString(postMessage,onDataSentCallback);
```

@anakod What is your opinion?

PR #915, opened 2017-01-30
Include sming sdk into docker image

nevers:
First and foremost, you should know this project is awesome. Keep up the good work!

I've noticed it was a bit time-consuming to setup the docker environment because everyone has to build it on their own. Building takes quite some time and when finished, it doesn't yet include the prebuilt SDK (as is documented). I understand this is feasible when developing the framework itself, but from an framework-user point of view, this is somewhat tedious.

This patch will embed the full prebuilt SDK into the docker image and allow users to simply start building and flashing their own projects (or the samples). This assumes of course that you push this image to the central docker registry.

I've added a short README.md that explains how to correctly build the image because you now have to build the docker container from the root of the repo. If you like this patch, I'll send another patch for updating the wiki.


slaff:
Take a look at that line: https://github.com/SmingHub/Sming/blob/develop/docker/sming-ide.yml#L6. Sming is actually exported volume and you should be able to work on the files in Docker and on your host machine. 

> This patch will embed the full prebuilt SDK into the docker image and allow users to simply start building and flashing their own projects (or the samples).

Did you commit that change, because I do not see it in your PR?

nevers:
I realise that you've added a host volume to expose the framework into the docker image. Therefore, when running the image using the docker-compose file you provided one may access sming through this volume. 

This works fine, but it has a few downsides I'm trying to solve:
1. The sming framework is not bound to the docker image version. This could cause inconsistencies for the unexpecting user. For example, a user might use an old version of the docker image that is lacking some important dependencies and run it with the latest version of the sming framework. Because the user is missing the required dependencies his build will fail. Every time the Dockerfile is changed, all users must rebuild the docker image to have the latest version. People will forget to do this.
2. Building the sming docker image takes a lot of time. Instead of having all users building the image themselves (taking the risk that they might do it in a wrong way) you could just push the docker image (you personally verified to be working) to the central docker registry. Users can then just pull the image, which then includes the full sming framework and start working in a matter of minutes.
3. A user has to cd /opt/sming/Sming && make, before being able to work on their own projects. This is somewhat tedious.
4. Simplification of the user development workspace: when the docker image already contains the (prebuilt) sming framework, users don't have to pull the source code on their machines anymore. Users would then only need a simple docker-compose.yml at the root of their project (which pulls in the sming framework from the docker registry) and exposes their workspace from the host to the image. 

Taking into account the above points, the docker-compose.yml would only be useful as an example on how to setup the workspace of a framework user, not building the docker image. I've adapted the docker-compose.yml to better reflect this. Please also have a look at my changes to the docker quickstart wiki pages here: https://github.com/nevers/sming-wiki/compare/include-sming-sdk-into-docker-image
 
Hope this makes sense to you. Having an official docker image for sming that includes the full framework would really lower the threshold for people to start using it.

nevers:
>> This patch will embed the full prebuilt SDK into the docker image and allow users to simply start building and flashing their own projects (or the samples).

> Did you commit that change, because I do not see it in your PR?

Please have a look at all the changes. For your convenience, this is where the sming framework is added to the image:
https://github.com/nevers/Sming/blob/include-sming-sdk-into-docker-image/docker/Dockerfile#L153

nevers:
For your info:
- I've built and pushed a docker image for sming, taking into account the feedback in this issue.  You could try pulling nevers/sming-ide:3.0.2-dev and follow my changes in the docker quickstart wiki here: https://github.com/nevers/sming-wiki/compare/include-sming-sdk-into-docker-image
- You can still export your local git checkout of the sming framework into /opt/sming (like before in your compose file) in order to mount it over the prebuilt sming dir in the image. This way you can continue using your current development scenario.

anakod:
Another idea - nothing prevent us to have two different Docker images: with embeded full framework for quick start and basic docker template with external framework which user can build\replace manually.

anakod:
> You can still export your local git checkout of the sming framework into /opt/sming (like before in your compose file) in order to mount it over the prebuilt sming dir in the image. This way you can continue using your current development scenario.

This probably fine too.

nevers:
Good idea. Having 2 docker images might require a bit more maintenance though, as there will be two images per release. Beyond that remark, however, I don't really have a preference. As long as you push the image(s) to the central docker registry and it contains the full, pre-built framework, you're a hero.

slaff:
@nevers I'll start reviewing your work next week. So far I am having the following major issues with the current(the one in the repository) docker image that need to be addressed in the next release:
* The docker container runs as root user. This is the same root user as on the host system. The Dockerfile needs to be changed to use the `espbuild` user. 
* Docker image generation takes quite some time: It will be great if we have official DockerHub image and a docker-compose.yaml file that uses our official image and allows mounting host directories, as suggested by you.

slaff:
Will be replace with that PR #955.

PR #917, opened 2017-01-30
Fix lwip compilation on windows

ADiea:
On windows lwip compilation is skipped because target all is not specified.


slaff:
Much cleaner patch than the one related to the same issue. The code also compiles on my old Windows with UDK. Once Travis reports that everything is fine I will test it on FreeBSD-current and prepare for merging the PR.

ADiea:
yes removed patch hunks for the other files but will give some warnings of implicit function declarations...

ADiea:
on old udk memalyzer path is without ESP8266 you will get error on compiling a sample, on new UDK the path is the correct one in Makefile-windows.mk (with ESP8266)

PR #918, opened 2017-01-30
Rebase of PR #903

slaff:


anakod:
This PR breaks our compatibility :(

Before samples working as expected without first line:
``    `Serial.begin(SERIAL_BAUD_RATE); // 115200 by default`

Now them output garbage without it (wrong speed from Sming core lib by default, I belive). I've catched this issue with my Yeelight pull-request.

slaff:
@ADiea Can you, please, check the problem that **anakod** is reporting?

ADiea:
Could it be that some samples redefine this function?
```
extern "C" void __attribute__((weak)) user_rf_pre_init(void)
```
In that case probably it's safer to put the line
```
uart_div_modify(UART_ID_0, UART_CLK_FREQ / SERIAL_BAUD_RATE);
```
back in user_init() function although it's a little later..

ADiea:
Also the way of setting this in user_init or user_rf_pre_init does not take into account multiple UART support by @slaff, maybe we can find a better way?  Do we know from compile time which uart will be used for debug?

slaff:
> Do we know from compile time which uart will be used for debug?

We [rely](https://github.com/SmingHub/Sming/blob/develop/Sming/system/m_printf.cpp#L20) on the `uart_tx_one_char` function from Espressif. For now all supported SDKs direct the output to UART0.

PR #926, opened 2017-02-03
Added CI tests for SDK 2.0 on Linux with esp-open-sdk.

slaff:
We can declare SDK 2.0 as supported and start tackling issues as soon as they pop up.


anakod:
Is it good to have much SDKs in CI? May be drop support old SDK and switch to two versions: 1.5 and 2.0? Faster builds/checks, less bugs.
Also on Windows with UDK it's probably anyway working only with last toolchain because of MemAnalyzer path changes.

slaff:
> May be drop support old SDK and switch to two version

Good point. For the moment also the MacOS X builds aren't giving us any valuable information. Maybe we should test only on Linux with esp-alt-sdk v1.5 and esp-open-sdk v2.0. Any comments on this?

PR #927, opened 2017-02-03
Automatic store all debugf format strings in flash

ADiea:
Sometimes a section type conflict can arrise when using a debugf in the same file both in a virtual methid in a class and in another method of another class. In this case the only solution I have found is to use debugf_RAM which has the same effect as old debugf stores format string in RAM. Tried several tricks ex construct subsections irom0.text.0...n  for each string based on __COUNTER__ or __LINE__ but gcc will ignore the strings.


ADiea:
There are now 4 debug functions similar to printf:
* debugf - legacy function redefined as debug_e
* debug_e - debug error for error messages
* debug_w - debug warning for warning messages
* debug_i - debug info for information messages
The debug level can be set via make flag DEBUG_VERBOSE_LEVEL (0=error, 1=warning, 2=info, 3=debug)
These functions store the format string in flash and reads it at runtime from flash into a 256 byte buffer allocated on stack. 

There is an additional debugf_RAM function to be used instead of debug_x when there is a section type conflict (rare cases). Probably will erase this macro and suggest using directly m_printf which is the same.

ADiea:
where to put all the above documentation? Readme.md seems to general for this kind of info.

ADiea:
@anakod what about the other: debugf_e/w/i ? I like the _x because it's easier to see error/warn etc. I will remove debugfram and just use m_printf

anakod:
I'm personally like debug_w more than debugf_w, I think - it's better for reading.

ADiea:
For ex for HttpServer_WebSockets
+3.3kB free RAM but also +8kB in flash image because of the repeated expansion of debug_e.
I am trying to optimize code if possible to have less impact on flash...
```
17:41:34 **** Build of configuration Build (GNU) for project sming ****
make rebuild ENABLE_SSL=1 ENABLE_CUSTOM_LWIP=1 ENABLE_CUSTOM_HEAP=1 DISABLE_SPIFFS=1 V=0 ICACHE_FLASH 
C+ app/CUserData.cpp
C+ app/application.cpp
AR out/build/app_app.a
LD out/build/app.out

#Memory / Section info:
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8384|     900
    rodata|           ReadOnly Data (RAM)|    3FFE8390|    3FFEAFF4|   11364 (-3380)
       bss|      Uninitialized Data (RAM)|    3FFEAFF8|    3FFF1DF8|   28160
      text|            Cached Code (IRAM)|    40100000|    40105E3A|   24122
irom0_text|           Uncached Code (SPI)|    4020A000|    402526DE|  296670 (+8128)
Total Used RAM : 40424
Free RAM : 41496
Free IRam : 8664
# Generating image...
Generate firmware images successully in folder out/firmware.
Done
(!) Spiffs support disabled. Remove 'DISABLE_SPIFFS' make argument to enable spiffs.
make: *** No rule to make target `ICACHE_FLASH'.  Stop.
```

anakod:
I just will show my solution (I didn't commit it to repository when checking it some time ago). May be it will help you resolve issue with conflicts:

> Tried several tricks ex construct subsections irom0.text.0...n for each string based on COUNTER or LINE but gcc will ignore the strings.

```
// http://stackoverflow.com/a/35441900
#define MACROCAT2(x,y) x##y
#define MACROCONCAT(x,y) MACROCAT2(x,y)
#define MACROQUOT(x) #x
#define MACROQUOTE(x) MACROQUOT(x)
#define SET_SECT() __attribute__((section(MACROQUOTE(MACROCONCAT(.irom.debug.,__COUNTER__)))))

#undef assert
#define debugf(fmt, ...) do {    \
    SET_SECT() \
    static const char flash_str[] STORE_ATTR = fmt"\r\n";    \
    char buf[257]; \
    ets_strcpy(buf, flash_str); \
    m_printf(buf, ##__VA_ARGS__); \
    } while(0)

#define assert(condition) if (!(condition)) SYSTEM_ERROR("ASSERT: %s %d", __FUNCTION__, __LINE__)

#define SYSTEM_ERROR(fmt, ...) debugf("ERROR: " fmt, ##__VA_ARGS__)
```


LD script:
```
    out/build/app_app.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
  *libsming.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
```

First stackoverflow link should be useful - looks like solution with __COUNTER__  works as expected.

ADiea:
Already tried subsection solution from stackovf but without the LDscript modofocations, will retest again now with modified ldscript thanks

anakod:
I think you can combine both our solutions.

ADiea:
@anakod i still cannot make the linker to include irom.debug.* symbold in final output
```
    *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom.text .irom.debug.*)
	out/build/app_app.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
  *libsming.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
  *libsmingssl.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
```

If I build with section irom.text I see the symbols included but without there are no more log_* symbols (the static global char* strings) still _don't know how it still links with no error_!

ADiea:
This is the map file. I have also concatenated the filename, but it's the same with only counter
[firmware.map.txt](https://github.com/SmingHub/Sming/files/752556/firmware.map.txt)

ADiea:
Hmm still symbols are there when i look into the .out file (xtensa-lx106-elf-readelf)
[symbols_obj.txt](https://github.com/SmingHub/Sming/files/752561/symbols_obj.txt)

ADiea:
Almost ready with this. Some issues remain with printing filename and line number. Subsections seem to work and no more compile errors

avr39-ripe:
Nice feature! But may be we should think about #ifdef all debugf_* if SMING_RELEASE=1 to completely remove unnecessary debug stuff from release image and free heap/flash ?

ADiea:
@avr39-ripe it's included in debug_progmem.h

```
+//This enables or disables logging
 +//Can be overridden in Makefile
 +#ifndef DEBUG_BUILD
 +	#ifdef SMING_RELEASE
 +		#define DEBUG_BUILD 0
 +	#else
 +		#define DEBUG_BUILD 1
 +	#endif
 +#endif
```

then at the end of file 

```
+#else /*DEBUG_BUILD*/
 +	#define debug_e
 +	#define debug_w
 +	#define debug_i
 +	#define debug_d
 +#endif /*DEBUG_BUILD*/
```

avr39-ripe:
Wow! I missed this! Great! :) Waiting for this to be merged ASAP :)

slaff:
@avr39-ripe The earliest possible date is the start of the next release. Which is next week, if things go according to the plan. That PR needs to be tested extensively and we should describe the pros and cons of it. Which means that we would like you to test it as much as you can even before we merge it in the `develop` branch and send us your feedback ;-)

Sming/system/include/esp_systemapi.h
```diff
@@ -38,7 +39,14 @@
...
 #else
-#define debugf(fmt, ...) m_printf(fmt"\r\n", ##__VA_ARGS__)
+//(!)Important: Normally debugf is defined as debug_e.
+//There are rare cases when expanding debug_e will cause a section type conflict
+// this is most likely because you are using debugf in a virtual member
+// of a class and a member of another class in the same file.
+//
+//If you get a section type conflict, please use m_printf instead of debugf
+// which will store the format string in RAM.
+#define debugf debug_e
```
anakod: May be decrease old debugf level to debug_i, because it isn't error messages really.

anakod: And mark it as deprectated after that :)

ADiea: I was planning another 2 PRs
1. replace all debugf with corresponding d/i/w/e functions. Ex in Websocket code there are some debugf commented (debug) and the rest of them some are not useful. Some are errors that should stay.
2. modify debug function in axtls and in open_lwip -> they also take up a lot of RAM

ADiea: Will remove redundant comments

ADiea:
rebased on current develop

Sming/Makefile
```diff
@@ -109,6 +109,12 @@ COM_SPEED_ESPTOOL ?= $(COM_SPEED)
 # Default COM port speed (used in code)
 COM_SPEED_SERIAL  ?= $(COM_SPEED)
 
+#Default debug print format is not to show filename and line for every debug line
+DEBUG_PRINT_FILENAME_AND_LINE ?= 0
+
+#Defaut debug verbose level is INFO, where DEBUG=3 INFO=2 WARNING=1 ERROR=0 
+DEBUG_VERBOSE_LEVEL = 2
```
ADiea: has to be ?=

Sming/Makefile-project.mk
```diff
@@ -18,6 +18,12 @@ COM_SPEED_ESPTOOL ?= $(COM_SPEED)
 # Default COM port speed (used in code)
 COM_SPEED_SERIAL  ?= $(COM_SPEED)
 
+#default debug print format is not to show filename and line for every debug line
+DEBUG_PRINT_FILENAME_AND_LINE ?= 0
+
+#Defaut debug verbose level is INFO, where DEBUG=3 INFO=2 WARNING=1 ERROR=0 
+DEBUG_VERBOSE_LEVEL = 2
```
ADiea: has to be ?=

ADiea:
Hi from  my part PR is finished please test.
Added the debug_x functions to Basic_Serial
Output with DEBUG_PRINT_FILENAME_AND_LINE=1 is:
```
[application:91] (INFO) message printed on UART0

[application:93] (WARNING) message printed on UART0

[application:95] (ERROR) message printed on UART0

[application:98] (ERROR) message printed with debugf on UART0

[application:114] (DEBUG) print me again on UART0


== PRINTF TEST START ==

Float numbers display test:
Pi with 2 decimals: 3.14 and with 4 decimals: 3.1416
Pi without specifying precision(default 9): 3.141592654
A number with leading fraction zeroes like 1.050250=1.05025
...
```

ADiea:
After this one is merged I will start a new PR to replace most of the debugf with coresponding error/warning/debug message. Also enable logging for lwip and ssl libraries

ADiea:
Ready for merge

slaff:
@anakod Can you review the changes?

Sming/Makefile-windows.mk
```diff
@@ -9,7 +9,7 @@ SDK_BASE	?= $(ESP_HOME)/ESP8266_SDK
 SDK_TOOLS	 ?= $(ESP_HOME)/utils
 
 # Other tools mappings
-ESPTOOL		 ?= $(SDK_TOOLS)/esptool.exe
+ESPTOOL		 ?= $(SDK_TOOLS)/ESP8266/esptool.exe
```
slaff: Smuggling changes in PRs is not good! The change above has nothing to do with the main purpose of that PR.

ADiea: Needed for correctly flashing. Since flashing is part of testing I considered this blocking issue and needed urgent fix...

slaff: I have to double check but as far as I can see SDK_TOOLS is used only for the tools directory, which means that in you case where it is under SDK_TOOL/ESP8266 you can define it directly in your environment without the need to change it in the Sming source code (as @anakod correctly noticed ( #938) and complained about it ). 

@ADiea Do you know if there is a way to get the UDK version? We could add conditions that set correctly the SDK_TOOLs directory for UDK on Windows.

ADiea: Could not find anything in the registry. Also no env variable. We could use perl/awk to parse changelog for the first occurence of regex of version string but it's a little dirty...

Created [issue](https://github.com/CHERTS/esp8266-devkit/issues/38)

ADiea: Maybe then remove this addition and also the one on memanalyzer and define it by chocolate, and manually for advanced users ?

slaff: > Maybe then remove this addition and also the one on memanalyzer and define it by 

@ADiea That would be great. Can you please create a PR with those changes?

Another question: Do you know how we can use the latest version of Sming in `chocolate`?

ADiea: No, I manually installed it...

anakod: > you can define it directly in your environment without the need to change it in the Sming source code
Yes, but current version didn't work under any UDK what is wrong. SO some solution was needed anyway, I think.

Lat's contnue discuss this issue in related topic: https://github.com/SmingHub/Sming/issues/938

slaff:
@ADiea your changes break the compilation of C++ files if they have dash (-) in their name. In order to reproduce it do the following:

```
cd Basic_rBoot
make clean
mv app/application.cpp app/app-lication.cpp
make
```

It might be possible that also other allowed characters in the file name break the compilation. Please, prepare a new PR with fixes.

ADiea:
Problem is gnu-make knows nothing about regexp... i wanted to match only alphanumeric,  I wanted to avoid subst(subst(...)

PR #928, opened 2017-02-03
Initialize default baud rate.

slaff:
If the weak function `user_rf_pre_init` is not used we need to be able to define the default baud rate.
Fixes issues with PR #903.


slaff:
@anakod Can you check if this PR is fixing the problem that you have with the UART output?

anakod:
Yes, for me looks like this PR completely resolve issue with default baund rate.

PR #931, opened 2017-02-06
AM2321 I2C sensor support (with example)

anakod:
Example attached. Tested on hardware.


anakod:
Hello. I prefer to merge it to develop. I don't like to create separate branch to use this code with my other updates.
And I belive this PR can't break anything because it change only separated lib & example :)

slaff:
> And I belive this PR can't break anything because it change only separated lib & example :)

You are personally responsible if that code does not compile on  some OSes or SDKs ;)

PR #933, opened 2017-02-07
Custom LWIP compilation: Added full path to the xtensa bin files.

slaff:

PR #939, opened 2017-02-10
* Fix whitespace issue

anakod:
Fix windows compilation issues: https://github.com/SmingHub/Sming/pull/933


anakod:
Also my git client show me one untracked file:
https://gyazo.com/b906a1590f9eeb2da2280f87fe1dc467

Should it be added to gitignore\pushed to repository? Or I'm doing something wrong?

slaff:
If you want add `Sming/compiler/lib/lib*.a` so that the library files are ignored too.

slaff:
@anakod Check this PR #940 for ignoring files that should not be in the repo or we do not care if they are changed.

PR #946, opened 2017-02-12
Fix debug messages when rBoot is enabled

gbl08ma:
Apparently, the people in charge of #927 forgot to apply the changes necessary for things to work on ROMs compiled with rBoot in mind. This resulted in all Sming debug messages appearing as garbage on such ROMs:

garbage debug messages

This pull request fixes the issue by applying the necessary changes to the rBoot-specific Makefile and also to the example rBoot linker script.

In addition, I changed some things in the rBoot-specific Makefile to make it more similar to the common one. I did not port the new memory information output as it seemed to require quite a bit more effort than my time allowed, and I guess that can be postponed to a later pull request...

This has been tested and confirmed to work for my projects only. I can't guarantee that other configurations will still work, but I don't see why not.

### DIFF
--- a/Sming/Makefile-rboot.mk
+++ b/Sming/Makefile-rboot.mk
@@ -184,6 +184,8 @@ ENABLE_CUSTOM_HEAP ?= 0
 LIBMAIN = main
 LIBMAIN_SRC = $(addprefix $(SDK_LIBDIR)/,libmain.a)
 
+USER_LIBDIR = $(SMING_HOME)/compiler/lib/
+
 ifeq ($(ENABLE_CUSTOM_HEAP),1)
 	LIBMAIN = mainmm
 	LIBMAIN_SRC := $(USER_LIBDIR)lib$(LIBMAIN).a
@@ -197,7 +199,6 @@ ifeq ($(RBOOT_BIG_FLASH),1)
 else
 	LIBMAIN_DST = $()
 endif
-# libraries used in this project, mainly provided by the SDK
 
 LIBLWIP = lwip
 ifeq ($(ENABLE_CUSTOM_LWIP), 1)
@@ -215,7 +216,6 @@ ifeq ($(ENABLE_CUSTOM_PWM), 1)
 endif
 
 # libraries used in this project, mainly provided by the SDK
-USER_LIBDIR = $(SMING_HOME)/compiler/lib/
 LIBS		= microc microgcc hal phy pp net80211 $(LIBLWIP) wpa $(LIBMAIN) $(LIBSMING) crypto $(LIBPWM) smartconfig $(EXTRA_LIBS)
 
 # compiler flags using during compilation of source files
--- a/Sming/Makefile-rboot.mk
+++ b/Sming/Makefile-rboot.mk
@@ -51,6 +51,12 @@ COM_SPEED_ESPTOOL ?= $(COM_SPEED)
 # Default COM port speed (used in code)
 COM_SPEED_SERIAL  ?= $(COM_SPEED)
 
+#default debug print format is not to show filename and line for every debug line
+DEBUG_PRINT_FILENAME_AND_LINE ?= 0
+
+#Defaut debug verbose level is INFO, where DEBUG=3 INFO=2 WARNING=1 ERROR=0 
+DEBUG_VERBOSE_LEVEL ?= 2
+
 ## Flash parameters
 # SPI_SPEED = 40, 26, 20, 80
 SPI_SPEED ?= 40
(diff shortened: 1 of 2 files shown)


ADiea:
Good job and sorry for the missing code!

Sming/Makefile-rboot.mk
```diff
@@ -248,11 +253,6 @@ ifeq ($(ENABLE_SSL),1)
 	CXXFLAGS += $(AXTLS_FLAGS)	
 endif
 
-ifeq ($(ENABLE_CUSTOM_LWIP), 1)
```
slaff: Why is that remove?!

Sming/Makefile-rboot.mk
```diff
@@ -173,31 +179,7 @@ endif
 
 EXTRA_INCDIR += $(SMING_HOME)/include $(SMING_HOME)/ $(LWIP_INCDIR) $(SMING_HOME)/system/include $(SMING_HOME)/Wiring $(SMING_HOME)/Libraries $(SMING_HOME)/SmingCore $(SMING_HOME)/Services/SpifFS $(SDK_BASE)/../include $(THIRD_PARTY_DIR)/rboot $(THIRD_PARTY_DIR)/rboot/appcode $(THIRD_PARTY_DIR)/spiffs/src
 
-USER_LIBDIR  = $(SMING_HOME)/compiler/lib/
```
slaff: Why was that paragraph placed in another location?!

Sming/Makefile-rboot.mk
```diff
@@ -153,8 +159,8 @@ TARGET		= app
 
 THIRD_PARTY_DIR = $(SMING_HOME)/third-party
 
-LIBSMING = sming
 SMING_FEATURES = none
+LIBSMING = sming
```
slaff: Why is that change here?!

gbl08ma: "In addition, I changed some things in the rBoot-specific Makefile to make it more similar to the common one."
If you looked at the diff between `Makefile-project.mk` and `Makefile-rboot.mk` before and after this commit, instead of simply looking at the commit diff, perhaps you would understand the end result better.

By reducing the amount of differences between these two files, it becomes easier to correctly apply changes to both at the same time, and perhaps even merge parts of them into a single file one day.

Sming/Makefile-rboot.mk
gbl08ma: I simply moved it to the same place where it was in `Makefile-project.mk`, but on second analysis it seems that perhaps its location in `Makefile-project.mk` is wrong too, since I see references to `USER_LIBDIR` before it is set. Furthermore, now there's a duplicate line, `# libraries used in this project, mainly provided by the SDK`, which wasn't there before. I'll fix this and move the `USER_LIBDIR` line back to before all its usages, but we should check if `Makefile-project.mk` isn't wrong too.

Sming/Makefile-rboot.mk
gbl08ma: Because `LWIP_INCDIR` is already being added to `EXTRA_INCDIR` on line 180 of my `Makefile-rboot.mk` (or line 174 of yours). While apparently the path is not the same (it doesn't end in `/include`), if you look at `Makefile-project.mk`, you'll see that these lines are not there, either, and only the path that doesn't end in `include` is ever added to `EXTRA_INCDIR` (or is it wrong too? doubt so, as that would have broken custom LWIP for everyone, I think).

Furthermore, the path it was adding to `EXTRA_INCDIR` was missing the `$(SMING_HOME)` prefix and I doubt it would work when the makefile is run from the project directory.

Sming/Makefile-rboot.mk
slaff: @gbl08ma Please, make only one logical change per PR. It makes things easier for us to review and maintain the code. If there are other issues in the makefile(s) then make a second, third, etc PR. Therefore I would like to ask you to remove all changes that are not related to "Fix debug messages when rBoot is enabled". If you cannot do it then I will have to rebase this PR later this week.

gbl08ma: The reason why there are all these changes is because I was having trouble bringing in the changes related to the debug messages with the rBoot makefile being so different from the default one. So I basically had to diff them and go change by change applying (or not) each one.

I wouldn't mind trying again but this requires testing on the hardware to see if it's fixed, and I no longer have the ESP8266 I was using with me, so this may have to wait some days.

slaff:
... Will be merged with #947. Thanks for noticing the issues and reporting them.

anakod:
@gbl08ma 
> By reducing the amount of differences between these two files, it becomes easier to correctly apply changes to both at the same time, and perhaps even merge parts of them into a single file one day.

May be we can go step forward and use single Makefile with different options for rboot and usual build?

ADiea:
like ENABLE_REBOOT=1 ? or it's not that simple? Never used rboot until now but planning to...

PR #949, opened 2017-02-13
Fix SystemClockClass::setTime to work accordingly to timezone settings

ScruggRoman:
Wrong condition didn't allow to set right time to the RTC.

### DIFF
--- a/Sming/SmingCore/SystemClock.cpp
+++ b/Sming/SmingCore/SystemClock.cpp
@@ -21,7 +21,7 @@ DateTime SystemClockClass::now(TimeZone timeType /* = eTZ_Local */)
 void SystemClockClass::setTime(time_t time, TimeZone timeType /* = eTZ_Local */)
 {
 	bool timeSet =
-	(timeType == eTZ_UTC) ?	RTC.setRtcSeconds((time + (timezoneDiff * SECS_PER_HOUR))) : RTC.setRtcSeconds(time);
+	(timeType == eTZ_Local) ?	RTC.setRtcSeconds((time + (timezoneDiff * SECS_PER_HOUR))) : RTC.setRtcSeconds(time);
 	debugf("time updated? %d", timeSet);
 	status = eSCS_Set;
 }


slaff:
LGTM. @anakod Any comments on that PR?

anakod:
That's OK I think.

KarolBedkowski:
Hi,
I have problem with this change.
Assuming system clock using local time - :::now(eTZ_local) returns time from rtc without TZ apply. - ::setTime(.., eTZ_Local)  should not apply TZ offset. 

So IMHO either system clock should keep time in UTC (and ::now should be changed; also timeType param make some confusion), or this path should be reverted...

ScruggRoman:
@KarolBedkowski. Hi, could you explain in more detail, what's wrong?
At this moment I call SystemClock.setTimeZone( desired_timezone ), then set the time (for example with NTP ). When I call ::now( eTZ_UTC ) or ::now( eTZ_Local ) in the both cases I get right time. In case eTZ_Local - with applied desired_timezone...

KarolBedkowski:
Hi,
Based on samples/DS3232RTC_NTP_Setter:
`SystemClock.setTime(timestamp, eTZ_UTC); //System timezone is LOCAL so to set it from UTC we specify TZ`

Assuming system clock is in local time - `setTime` timeType parameter describe source (times to set) time zone. Setting it from UTC (`timeType=eTZ_UTC`) should be corrected by `timezoneDiff`. When setting system time from current (local) time (`timeType=eTZ_LOCAL`) time should be stored as is.

After this patch - setting time work inversely. 
Setting time in UTC and `timeType=eTZ_UTC` - rtc clock is set to utc time.

When `SystemClockClass::now(timeType=eTZ_Local)` I expect local time  and there is no tz correction, time is returned directly from rtc. When `timeType=eTZ_UTC` - tz correction is apply.

If system clock (rtc) work in local time (as in samples and `SystemClockClass::now()`) this patch broke setting time.

Of course, we may assume, that system clock should work in UTC. In this case - new `setTime` method is ok, but `now()` method for local time return time in UTC, so it's broken.

avr39-ripe:
I think we must revie all time-aware code to make it consistent. My suggestion that time in rtc must be in UTC and on-demand it should apply tz-correction

PR #955, opened 2017-02-14
Docker: Faster build time: Fetch pre-compiled toolchain instead of compilin…

slaff:
…g your own.

* SDK 2.0: Using Espressif's stock SDK 2.0.
* Simplified usage and generation.


slaff:
@nevers When you have time can you check if this PR is working for you? I've spend quite some time to make the generation as slim and fast as possible. It also uses the idea that you've mentioned to have the sming git repo inside of the container, and if needed, to allow mounting from the host.

nevers:
I'll try and have a look at it tomorrow.

nevers:
I found moment to look into this, still today. 

I must say this is a major leap forward: you've put the most time consuming bits into a nice separate, pre-built/published container and  this saves a lot of time. I can confirm that c9 works properly, and that I can compile the all sample code.

Some remaining questions:
- Why don't you run make from within the Dockerfile? You could spare the users of your framework the trouble.
- Why is docker-compose still building the Dockerfile? Is this because you want to avoid pushing a docker container for each Sming release?

Suggestion - why not altering the Dockerfile such that:
- it checks out a very specific commit/release of the sming repo 
- builds the full repo
- push this image to the central docker hub (with a tag that corresponds to the same release number as in git)
- update docker-compose.yml to depend on this specific image

This would solve point 1, as I explained in https://github.com/SmingHub/Sming/pull/915 

Anyways, thanks for looking into this, great work.

slaff:
> I found moment to look into this, still today.
Thanks for your time. I appreciate it.

> Why is docker-compose still building the Dockerfile? Is this because you want to avoid pushing a docker container for each Sming release?

The main reason is that if we push something we should support it.  I would like to concentrate our efforts on the framework. If there is a need for container-per-release then other volunteers should easily generate one. 

> Why don't you run make from within the Dockerfile? You could spare the users of your framework the trouble.

I thought of this image as something that developers will deal with and test. And as such compilation is an important part. 

What we should do is change the welcome screen to have "Getting Started" instructions that explain the user what to do next. And I will need some help here ;)

anakod:
> The main reason is that if we push something we should support it. I would like to concentrate our efforts on the framework. If there is a need for container-per-release then other volunteers should easily generate one.

May be we can automated that process? Or somebody can help with support? It shouldn't prevent end user from building new image with any Sming version, but will be very nice to have some pre-build docker images with major releases also.

PR #963, opened 2017-02-16
Initial version of a WIFI HotSpot example.

slaff:


Sming/third-party/.patches/esp-open-lwip.patch
```diff
@@ -24,10 +24,10 @@ index e69de29..0c5dff9 100644
 +
```
slaff: > This file shouldn't be here?

I don't understand. Without this file LWIP will not be patched. Can you elaborate a bit what shouldn't be there?

slaff:
See https://github.com/martin-ger/esp_wifi_repeater for more advanced  solution.

PR #964, opened 2017-02-16
Do makefile reload only once.

slaff:
Prevent full build when only spiffy compilation is required.


ADiea:
Hei @slaff so could this help in using the -j option of make?

slaff:
> Hei @slaff so could this help in using the -j option of make?

@ADiea Can you explain first what are the issues that you experience using the -h option, so that I can better understanding about the problem?

ADiea:
Our makefile recipes do not seem to work with -j optoin of make. Using j option enables make on parallel jobs to better use multiple cores of the CPU(fatser build time) ex make -j3

I was now reading [this](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Parallel-Make.html) so I think that it's hard to make it portable anyway so probably we should not take care of it at this point.

PR #965, opened 2017-02-16
Added NodeJS's http parser.

slaff:
For a long time we were missing first-class http parser.That PR changes this by modifying the NodeJS parser to work for ESP8266 and Sming.  Once ready this change will allow us:
* To process all type of headers
* Be able to process the response body as stream of data.
* Be able to handle multiple transfer encodings (even chunked).
* Have the fastest HttpClient on an ESP8266 device :)

This PR is here because that change needs a lot of testing before being merged. On my humble tests with Basic_Ssl it was blazingly fast, but I would prefer if you try it for yourself. This change also may break something and your precious bug report will help us fix it.

There are some improvements that will be added soon:
- [x]  Moving static constants from DRAM to Flash (better memory usage) for compilers that support the mforce-l32 flag.
- [X] Pass the data to the parser without intermediate buffer.

And issues that need to be tackled:
- [x] Strange Fatal exception 20 when used in applications with big code base. 

With your help we can have the best http client on ESP8266!


israellot:
Good one. I've used this parser on my esp-ginx project. Blazing fast and reliable.

anakod:
I'm like idea of fast and memory effective parser too. @slaff did you check compatibility with old Sming samples?

slaff:
> did you check compatibility with old Sming samples?

All samples work like charm. It would be great if you can also check in your applications/samples if it works as expected.

I would have merged it long ago if it wasn't for two things that make me hesitant. 
1. This parser takes 1.5 K from the free heap. I was thinking about having directives to enable/disable parser features and make it take even less. But that is not a big issue.
2.  The big issue is a problem that occurs in applications having a lot of code. When I use the http parser in them  I get strange Fatal exception 20 for returning to inaccessible memory address and I could not find a cure for that. (Same parser, same URL, small app works as expected.) I guess I am also missing the Xtensa knowledge to understand why that strange problem is happening. My suspicion is that it is related to the memory layout and allowed address difference for rjump. Once I have more time I will investigate it further. If you have an idea about that problem it will be great to share.

anakod:
>  I get strange Fatal exception 20 for returning to inaccessible memory address and I could not find a cure for that

I can't help you with that exception, but I know man who can help:
https://github.com/jcmvbkbc - he is great specialist in Xtensa. You can ask him, he already helped us before. He working on compilers optimisation and low level solutions for a long time.

Link to one of original posts: http://www.esp8266.com/viewtopic.php?f=9&t=224

slaff:
@israellot @anakod @avr39-ripe @riban-bw  Guys, I think I was able to sort out all issues preventing me from merging the http_parser in our HttpClient. Before I do the merge can you test it on your devices and/or your applications (if you don't have one use the Basic_Ssl app or any of the other HttpClient_* samples.) ? Thanks in advance.

avr39-ripe:
@slaff Sorry to say, but my app crashes with this pr. As soon as it try to use httpClient it hangs for a while, and them esp module just reboots. Same app works fine with recent develop of Sming

slaff:
> As soon as it try to use httpClient it hangs for a while, and them esp module just reboots.

The same issue that I am having. Sad :( It would be great if we can figure out what is causing it...

slaff:
> Sorry to say, but my app crashes with this pr.

@avr39-ripe Can you try again? I added more changes and now it should be rock-solid. Here is also a list of commands for all of you who want to try it before we merge it:

```
cd $SMING_HOME
git fetch origin pull/965/head:pr/better-http-parser
git checkout pr/better-http-parser
make dist-clean
```

avr39-ripe:
@slaff I'v checked my app again:
1. it do not crash ;)
2. it work.. but..
3. just one http request is successful, all other http requests failed ;( with old http_client + old parser all works fine :(

slaff:
> just one http request is successful, all subsequent http requests failed ;( with old http_client + old parser all works fine :(

Strange. Can you paste the code with which you test?

avr39-ripe:
Not easy "paste" code. I use TempsensorsHttp Class (https://github.com/avr39-ripe/TyTherm/blob/newAppClass/lib/tempsensor/tempsensors.cpp) to get data from another esp8266 (here is server code https://github.com/avr39-ripe/TyTherm/tree/newAppClass) TempsensorsHttp Class make http request each 4 second to get "fresh" data. With current httpClient it works good.

slaff:
That code is already battle-tested and part of the new [Http refactoring](https://github.com/slaff/Sming/issues/5)  in  [branch](https://github.com/slaff/Sming/tree/feature/better-http-client)  that will land in sming 3.5.0 or directly 4.0.0 ...

PR #966, opened 2017-02-17
TcpClient: setReceiveCallback and setCompleteCallback

tius2000:
Sometimes it is desirable to change the callbacks for TcpClient later. For example, this allows to setup a TcpServer with separate processing for each client like this:

    
    class CliTcp {
        public:
            CliTcp(TcpClient *client) {
                _client = client;
                client->setReceiveCallback( TcpClientDataDelegate(&CliTcp::_on_receive, this) );
                client->setCompleteCallback( TcpClientCompleteDelegate(&CliTcp::_on_complete, this) );
            }
    
            ...
    
        private:
            bool _on_receive(TcpClient&, char *data, int size) {
                ...
            }
    
            //  self destruct on close
            void _on_complete(TcpClient& client, bool) {
                client.setReceiveCallback( 0 );
                client.setCompleteCallback( 0 );
                delete this;
            }
            
            TcpClient *_client;
    };
    
    static void _on_connect(TcpClient* client) {
        new CliTcp(client);
    }
     
    void telnet_init() {
        auto srv = new TcpServer(_on_connect, 0, 0);
        srv->listen(TELNET_PORT);
    }  

This approach seems cleaner to me than using a local lookup table (like TelnetServer does to identify the first and only client).

### DIFF
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -180,7 +180,7 @@ void TcpClient::pushAsyncPart()
 		delete stream; // Free memory now!
 		stream = NULL;
 	}
-} 
+}
 
 err_t TcpClient::onSent(uint16_t len)
 {
@@ -199,7 +199,7 @@ err_t TcpClient::onSent(uint16_t len)
 
 	return ERR_OK;
 }
- 
+
 void TcpClient::onError(err_t err)
 {
 	state = eTCS_Failed;
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -180,7 +180,7 @@ void TcpClient::pushAsyncPart()
 		delete stream; // Free memory now!
 		stream = NULL;
 	}
-}
+} 
 
 err_t TcpClient::onSent(uint16_t len)
 {
@@ -199,7 +199,7 @@ err_t TcpClient::onSent(uint16_t len)
 
 	return ERR_OK;
 }
-
+ 
 void TcpClient::onError(err_t err)
 {
 	state = eTCS_Failed;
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Network/TcpClient.h
```diff
@@ -54,6 +54,13 @@ class TcpClient : public TcpConnection
 	__forceinline bool isProcessing()  { return state == eTCS_Connected || state == eTCS_Connecting; }
 	__forceinline TcpClientState getConnectionState() { return state; }
 
+	inline void setReceiveCallback( TcpClientDataDelegate on_receive ) {
```
slaff: Please, replace `inline` with `__forceinline` to be consistent with the other inline methods.

Sming/SmingCore/Network/TcpClient.h
```diff
@@ -54,6 +54,13 @@ class TcpClient : public TcpConnection
...
+		receive = on_receive;
+	}
+	inline void setCompleteCallback( TcpClientCompleteDelegate on_complete ) {
```
slaff: That function is duplicating already existing `void TcpClient::setCompleteDelegate(TcpClientCompleteDelegate completeCb)`.

tius2000: Oops, you are right @slaff - I missed that somehow becaused it is not mentioned in the api documentation ...

So I will add a new method _setReceiveDelegate()_ only? Should both methods be `__forceinline` to improve consistency? Or should the new method go to _TcpClient.cpp_ as _setCompleteDelegate()_?

slaff: > I missed that somehow becaused it is not mentioned in the api documentation ...

You have a perfect chance to fix that by adding proper documentation comments :)

> Or should the new method go to TcpClient.cpp as setCompleteDelegate()?

I would prefer that.

tius2000: Ok (-:
But I need to learn how to use doxygen first, might take a day or two ;-)

Sming/SmingCore/Network/TcpClient.cpp
```diff
@@ -180,7 +180,7 @@ void TcpClient::pushAsyncPart()
 		delete stream; // Free memory now!
 		stream = NULL;
 	}
-}
+}
```
slaff: Please, fix the spacing here. Look at the diff for all spacing issues.

PR #968, opened 2017-02-18
refactor linker script files

tius2000:
The recent changes to the linker script files broke compatibility with custom rboot linker scripts. Having _rboot.rom0.ld_ now is a great thing, but does not cover all cases (e. g. special flash layouts for 1M devices). 

Moving everything from _eagle.app.v6.cpp.ld_ except MEMORY to a separate script file allows very simple custom linker scripts without breaking compatibility with later Sming upgrades. Hope this is ok?


Sming/Makefile-rboot.mk
```diff
@@ -443,13 +443,13 @@ $(RBOOT_ROM_1): $(TARGET_OUT_1)
 
 $(TARGET_OUT_0): $(APP_AR)
 	$(vecho) "LD $@"
-	$(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) -L$(BUILD_BASE) $(RBOOT_LD_0) $(LDFLAGS) -Wl,--start-group $(APP_AR) $(LIBS) -Wl,--end-group -o $@
+	$(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) -L$(BUILD_BASE) -L$(SMING_HOME)/compiler/ld $(RBOOT_LD_0) $(LDFLAGS) -Wl,--start-group $(APP_AR) $(LIBS) -Wl,--end-group -o $@
```
ADiea: this change + the one above is equivalent to original code.

Sming/Makefile-rboot.mk
```diff
@@ -29,7 +29,7 @@ RBOOT_ROM_0      ?= rom0
 RBOOT_ROM_1      ?= rom1
 RBOOT_SPIFFS_0   ?= 0x100000
 RBOOT_SPIFFS_1   ?= 0x300000
-RBOOT_LD_0 ?= $(SMING_HOME)/compiler/ld/rboot.rom0.ld
+RBOOT_LD_0 ?= rboot.rom0.ld
```
slaff: Will it be possible with those changes to have, if needed, application specific rboot.rom.ld or eagle.app.v6.cpp.include.ld that differs from framework one?

Sming/compiler/ld/eagle.app.v6.cpp.include.ld
```diff
@@ -0,0 +1,217 @@
+/*
```
slaff: Wait a second :) How are the normal, non-rboot, applications going to be linked if the memory section is gone? Maybe you should have: one rboot file with memory section, one non-rboot with memory section and one common ld file to be included from the other two.

tius2000: Hi @slaff, the normal applications will use _eagle.app.v6.cpp.ld_ as before. But you are right, this should also include  _eagle.app.v6.cpp.include.ld_ to avoild code duplication.

Sming/Makefile-rboot.mk
tius2000: Yes it will. It allows you to have a different _MEMORY_ section in your projects linker files and still use all the other settings provided by the framework. I will try to provide an example of a Makefile soon that uses this and and other features to automatically do most of the memory calculations and generate all the required files.

Sming/compiler/ld/eagle.app.v6.cpp.include.ld
slaff: Looks better now. Please, make sure that your changes are also passing our CI tests.

tius2000: btw, are there any advantages of the espressif bootloader? Would it make sense to to support rboot only to simplify the framework?

tius2000: Well, the tests seem to fail because they do not see the new file _eagle.app.v6.cpp.include.ld_, how can I fix this? Should I split the PR (1st PR: add _eagle.app.v6.cpp.include.ld_,, 2nd PR change _eagle.app.v6.cpp.ld_)?

slaff: > Well, the tests seem to fail because they do not see the new file eagle.app.v6.cpp.include.ld

Not really. You should add the compiler/ld directory also for normal projects. A sample patch is given below.

```patch
diff --git a/Sming/Makefile-project.mk b/Sming/Makefile-project.mk
index 4e37697..503471a 100644
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -365,7 +365,7 @@ spiff_update: spiff_clean $(SPIFF_BIN_OUT)
 
 $(TARGET_OUT): $(APP_AR)
        $(vecho) "LD $@"        
-       $(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) $(LD_SCRIPT) $(LDFLAGS) -Wl,--start-group $(LIBS) $(APP_AR) -Wl,--end-group -o $@
+       $(Q) $(LD) -L$(USER_LIBDIR) -L$(SDK_LIBDIR) -L$(SMING_HOME)/compiler/ld $(LD_SCRIPT) $(LDFLAGS) -Wl,--start-group $(LIBS) $(APP_AR) -Wl,--end-group -o $@
 
        $(Q) $(STRIP) $@
 ```

slaff: > btw, are there any advantages of the espressif bootloader? Would it make sense to to support rboot only to simplify the framework?

Can you post this as discussion in the issues? If no one really needs the espressif bootloader we get rid of it and can easily simplify and unify our build system :)

tius2000: Dear @slaff, please put the PR on hold. Further test showed that there might be some impacts that I do not fully understand. I don't know why, but the new _rboot.rom0.ld_ seems not to be completely equivalent the old one. In one of my projects I get garbled sming debug output, possible caused by a wrong linker setting. I already did some research, but have not found the reason yet. Better safe than sorry ;-)

slaff: >  In one of my projects I get garbled sming debug output, possible caused by a wrong linker setting.

Make sure to initialize the flash memory when changing memory layouts or/and SDKs: 
https://github.com/SmingHub/Sming/wiki/Application%20prints%20random%20symbols%20and%20restarts

anakod: @tius2000 I think this is your issue: https://github.com/SmingHub/Sming/pull/946
So please check your makefile. Problem only with debug output?

tius2000: Dear @anakod, thank you very much for the hint! I will try this as soon as I'm back on my desk on Friday.

slaff: Please, rename this file `eagle.app.v6.cpp.include.ld` to  `common.ld`.  And that one `eagle.app.v6.cpp.ld` to standalone.rom.ld. And apply the changes to the make files. IMHO it makes it easier to figure out what are the different LD script for.

tius2000: Hi @anakod, I think you are right. I was not able to reproduce the problem, but it looked exactly like the screenshot you provided in #946.

PR #970, opened 2017-02-20
Revert "Fix SystemClockClass::setTime to work accordingly to timezone settings"

ScruggRoman:
Reverts SmingHub/Sming#949
removing changes made without the need


anakod:
Yes, we should at least add comment here, because otherwise issue will repeats in future.

PR #978, opened 2017-02-22
[Debug Strings] Don't use filename for naming variables or sections

ADiea:
A filename can have all sorts of illegal characters for naming a C/++ variable. Since make's subs function does not know regex it's not practical to replace all possible characters that could be used in a fname.

Apparently all compilers do not have an issue is there are multiple static variables with the same name in different scopes and after analysing the symbols it looks like the compiler will rename each variable to avoid linker conflict. So it turns out that using the filename is not needed anymore! Kept it in the makefile for displaying when needed but removed the subst calls. 

Not yet tested on real hw, just compiled and analysed symbols. WIP


Sming/system/include/debug_progmem.h
```diff
@@ -63,17 +63,17 @@ extern "C" {
 //Next special version of printf from FakePgmSpace is called to fetch and print the message
 #if DEBUG_PRINT_FILENAME_AND_LINE
 #define debug_e(fmt, ...) \
```
ADiea: Also need to remove helper macros that are not used anymore

samples/Basic_Serial/app/application.cpp
```diff
@@ -82,20 +82,20 @@ void init()
 	 * Functions debugf, debug_d, debug_i, debug_w, and debug_e store the format string
 	 * in flash so that the RAM is freed for more important information.
 	 *
-	 * Another useful feature is printing the filename and line number of every degub line
+	 * Another useful feature is printing the filename and line number of every debug line
 	 * This will require extra space on flash and can be enabled
 	 * using make parameter PRINT_FILENAME_AND_LINE=1*/
 
-	debug_d("(DEBUG) message printed on UART0");
+	debug_d("(DEBUG) message printed on UART0\n");
```
slaff: Why you suddenly changed the behaviour of debugf ? Is there a reason for this?

Sming/system/include/debug_progmem.h
```diff
@@ -63,17 +57,16 @@ extern "C" {
...
-	({static const char TOKEN_PASTE2(log_,CUST_FILE_BASE,__LINE__)[] \
+	({static const char log_string[] \
 	__attribute__((aligned(4))) \
-	__attribute__((section(MACROQUOTE(MACROCONCAT(.irom.debug.,__COUNTER__,CUST_FILE_BASE))))) = "[" GET_FNAME2(CUST_FILE_BASE) ":%d] " fmt "\r\n"; \
-	printf_P_stack(TOKEN_PASTE2(log_,CUST_FILE_BASE,__LINE__), __LINE__, ##__VA_ARGS__);})
+	__attribute__((section(MACROQUOTE(MACROCONCAT(.irom.debug.,__COUNTER__,__LINE__))))) = "[" MACROQUOTE(CUST_FILE_BASE) ":%d] " fmt; \
```
slaff: Why the \r\n is removed? Why you suddenly changed the behaviour of debugf ? Is there a reason for this?

samples/Basic_Serial/app/application.cpp
ADiea: In most debugf statements \n is appended to the format string. By applying it also by default in debug_progmem.h we get to the situation of double \n which makes reading the console harder.

So \\n was removed from the definition of debug_e and the user is responsible to put \n at the end if he wants (flexibility)
Alternative is to modify all debugf in the project to remove \n from the end of the format string

Sming/system/include/debug_progmem.h
ADiea: See above. This generates extra \n in most of the cases. It is better to leave flexibility for the user to include or exclude the \n as s/he wishes.

slaff: Please, change
```
= "[" MACROQUOTE(CUST_FILE_BASE) ":%d] " fmt; \
```
to
```
= "[" MACROQUOTE(CUST_FILE_BASE) ":%d] " fmt "\n"; \
```
And remove the new lines from debugf() in Basic_Serial. The change with the new debugf behaviour can be suggested as separate PR. 

The rest of the changes seem to be ok and fix the issue with  non-alpha characters in file name.

PR #979, opened 2017-02-22
Update axtl to latest 993a29f2b26cefeb80189451b88c29b31e84e640

ADiea:
Patches apply clean, builds with no error. 
Needed for further debugf patches for RAM optimization


slaff:
Wouldn't it be better to wait also for your patches to get merged in axtls-8266?

ADiea:
Yes I think so but I don't know how much it will take, I have asked the maintainer...

ADiea:
I alkso have more stuff to do on axtls like redefining printf to debugf but this will be done as a patch unfortunately

Issue #971, opened 2017-02-20
Discussion: Drop support for espressif bootloader?


tius2000:
Are there any advantages of the espressif bootloader? Would it make sense to to support rboot only to simplify the framework?

slaff:
If no one really insists on supporting Espressif's bootloader we can go even further:
* If bootloaded is needed then only rBoot will be used
* We can offer also images without bootloader, thus freeing space on Flash and memory.

ADiea:
+1 to drop support (lathough not used rboot yet, ut if you force me...i'll have to learn :) )

slaff:
Moving to only one bootloader, namely the superb rBoot, will help us get rid of:
- [ ] Different LD scripts
- [ ] Different project makefiles
- [ ] Different HttpFirmwareUpdate classes

Adding applications without bootloader, where no OTA or ROM switching is required, will
- [ ] Increase the amount of free space and heap available to the application
 

So I can see a lot of advantages for the code quality and for the users of Sming.

**Is anyone here using the Espressif Bootloader to do Over-The-Air (OTA) firmware updates?**

avr39-ripe:
+1 to drop original bootloader
+1 to all @slaff remarks

slaff:
> Drop support for espressif bootloader

@anakod @piperpilot @raburton What is your opinion on this?

tius2000:
@slaff: great - also the option of having no bootloader at all!

raburton:
I don't think I realised we supported the espressif bootloader anyway, or if I did I'd forgotten. In what way is it currently supported?

piperpilot:
I use only rboot...I tried the espressif bootloader and it sucked.  I +1 the switch to rboot or nothing.

rosiste:
+1 to drop original bootloader

slaff:
Ok, starting from version 3.2.0 (#977) we will drop support for Espressif's boolader and use only rBoot as bootloader. Plus, hopefully, add support for standalone applications.

Any final thoughts before we close the discussion?

tius2000:
I recently started to create completely refactored makefile for rboot-only projects with many additions (address calculation, automatic linker script creation, dependency detection etc.). However, it is still work in progress. It has a different name, so there are no collisions with the current makefiles. 

Should I create a PR for that when it is finished?

slaff:
> Should I create a PR for that when it is finished?

Yes, please.

anakod:
[...]

PR #985, opened 2017-02-25
completely refactored makefile

tius2000:
I recently started to create completely refactored makefile _sming.mk_ for my rboot projects with many additions:

* roms with and without bootloader
* automatic flash address caclulation 
* automatic linker script creation
* dependency tracking for spiffs sources
* dependency tracking for all header files included in cpp and c sources
* memanalyzer for rboot builds added
* automatic patch of _esp_init_data_default.bin_ for _system_get_vdd33()_ 
* improved terminal handling for windows
* allow duplicate source and header file names (in different directories)
* EXTRA_SRC and EXTRA_OBJ options added
* assembler support added

When reorganizing the file, I tried hard to group things that belong together. It is feature complete now, any comments are very welcome! It has a different name, so there are no collisions with the current makefiles. To use it, simple include it from your project makefile (see example). 

OS-specific files for Windows, Linux, Darvin and FreeBSD are provided, but only the Windows has been tested so far (appart from the Travis build test).

It is up to the maintainers, whether this could be a starting point for a new unified makefile or might just serve as an inspiration for further development of the existing makefiles.


anakod:
Refactoring looks nice and more clear. But what about back compatibility?

Sming/sming.mk
```diff
@@ -0,0 +1,593 @@
...
+WIFI_PWD        ?= ""
+
+ifneq ($(WIFI_SSID), "")
+	DEFINES += WIFI_SSID WIFI_PWD
+endif
+
+#==============================================================================
+#   do flash address calculations
+#------------------------------------------------------------------------------
+hexcalc          = $(shell $(PERL) -e"printf '0x%06X', $(1)")
```
slaff: It would be better to depend on python here instead of asking the users of Sming to install also perl on their machines.

ADiea:
+1 looking nice!

Sming/sming.mk
tius2000: As far as I know, python is not a requirement for the windows make process as _esptool.py_ and _memanalyzer.py_ are provided in compiled form within the udk. I do not have any experience with python, but the following command should work:

    hexcalc = $(shell $(PYTHON) -c"print(format(int($(1)),'\#08x'))")

However, I do not know a replacement for this line

    $(PERL) -ple "s{(^\s*irom0_0_seg *: *).*}{\\1org = $(IROM0_ORG0), len = $(IROM0_SIZE)}" $< >$@

so any help is welcome!

tius2000:
Thank you @anakod! I tried hard not to break things and keep backward compatibility. However, only rboot roms are supported so far and much more testing will be required.

tius2000:
Hi @slaff, I like the idea of creating roms without bootloader! However, I'm not familiar with that option. Is there a sample linker file (e. g. for _Basic_Blink_)? Is there anything else that needs to be changed beside the linker file? I will try to include that option, too.

Sming/sming.mk
slaff: > esptool.py and memanalyzer.py are provided in compiled form

Thanks for bringing to my attention that we are using esptool.exe. I will have to check if that is compiled python file or some other esptool. If it is the latter we should move to esptool.py to have all the latest advantages from it.

riban-bw:
> However,  I do not know a replacement for this line
>`$(PERL) -ple "s{(^\s*irom0_0_seg *: *).*}{\\1org = $(IROM0_ORG0), len = $(IROM0_SIZE)}" $< >$@`
If you explain what this perl code does then someone may be able to offer a python alternative. There are fewer developers who know both perl and python than know just one!

tius2000:
@riban-bw: This line is just a simple search and replace with a regular expression, something a *nix user probably would use _sed_ for. However, I wanted to minimize the number of required tools for windows builds and used _perl_ instead.

slaff:
>  something a *nix user probably would use sed for.

Then, please, replace it with `sed`. Sed is also part of cygwin/mingw for Windows.

tius2000:
I removed the dependency on Perl and changed the directory handling to allow projects with multiple directories with duplicate names for source and header files. A new option _EXTRA_SRC_ was added.

samples/Basic_Blink/Makefile
```diff
@@ -1,24 +1,42 @@
...
+GET_VDD33		= 0
+
+#==============================================================================
+#	now let sming do all the work ...
+#------------------------------------------------------------------------------
+#	make sure that SMING_HOME is set and the path compatible with make 
 ifndef SMING_HOME
-$(error SMING_HOME is not set. Please configure it in Makefile-user.mk)
-endif
-ifndef ESP_HOME
```
anakod: You remove SMING_HOME from this file? Where it should be declared?

anakod:
Now we have 3 main makefiles:
1. Makefile (to build Sming itself)
2. Makefile-project.mk (to build classic projects)
3. Makefile-rboot.mk (to build rboot projects)
+ we have 4 additional system related Makefile (Win, Linux, MacOS, FreeBSD)

I don't think what we should add additional 4th Makefile. It should replace some of previous files instead.
You plan to implement only rboot build or classic projects also? Because I'm sure will be right to have single universal makefile instead of two independend for both project types.

samples/Basic_Blink/Makefile
tius2000: IMHO it is a good idea not to have settings regarding the build enviroment withing the project files. That is why I think it is better to set _SMING_HOME_ as as a enviroment variable. The makefile than can find and include _sming.mk_ and all plattform specific settings (e. g. _windows.mk_).

tius2000:
Hi @anakod, thank you for your feedback. You are right, we should not have more makefiles but less. I did not plan to include support for the Espressif bootloader (classic projects) unless there is a real need (see discussion #971). 

However, I really like @slaff's idea of builds without any bootloader and would like to include that as an option. Perhaps someone can give me a hint where to find the correct linker and flash settings for this?

Of course, this PR may also just serve as an inspiration for further development of the existing makefiles.

slaff:
> However, I really like @slaff's idea of builds without any bootloader and would like to include that as an option. Perhaps someone can give me a hint where to find the correct linker and flash settings for this?

@tius2000 If you compile the current Basic_Blink sample, or any other non-rboot projects,  you will actually create a standalone application. The application comes with two ROM files, the one starting at 0x000 contains the .text .data .rodata sections (IRAM and DRAM) while the other one contains libraries/SDK code (SPI Flash )

```
esptool2 -quiet -bin -boot0 out/build/app.out out/firmware/0x00000.bin .text .data .rodata
# --boot0 above is for single, non-bootloaded apps.
esptool2 -quiet -lib out/build/app.out out/firmware/0x0a000.bin 
```

anakod:
> I did not plan to include support for the Espressif bootloader

That's right, we decide remove support for old Espressif bootloader.

>  (classic projects)

Classic project - I mean project *without* bootloader - single application splitted to two binaries as @slaff described above (or both binaries can be joined to single).

**ANOTHER OPTION**
Also we can drop support for applications without rboot bootloader at all. We can consider this option but:
1, We should provide solution compatible with old small 512 KB flash also
2. rBoot add additional delays in startup process. It can be really critical for battery powered deveices
3. In my opinion better to keep both options, it give us more flexibility

anakod:
@tius2000 , this information can be helpful also: https://github.com/SmingHub/Sming/pull/968

tius2000:
@anakod: I created #968 to simplify the use of project specific linker files. When using _sming.mk_ this is no longer required, as the required linker files are created on demand (as long as only _irom0_0_seg_ needs to be changed).

anakod:
WOW, it's your pull request :) I'm glad what finally we have teammate who knows makefiles and building process very well, because our old makefiles was bad and unreadable!

tius2000:
Thank you @anakod! In fact, I do not understand much about it. That is why I simply started to sort things and put label on them ;-)

slaff:
>  I do not understand much about it.  ...

That is not a very good sign... Anyway I will try to review the changes later this week or the next one. We must not forget our main goal:
- Less efforts in adding changes to project and library makefiles.

IMHO, this implies merging the Makefile-project.mk and Makefile-rboot.mk into one file WITHOUT sacrificing functionality and WITHOUT causing "pain" to our users.

tius2000:
That's great, @slaff - if you don't like it, just drop it. But I would be glad if you can give me some feedback anyway.

I recently added support for assembler files it will commit change that later today.

anakod:
I don't think what current solution with two different makefiles is good, even if we skip question about code quality\comments.We have too many related issues during last month (just for example after moving debugf to irom and other). 

> WITHOUT causing "pain" to our users.

Yes, it's first priority.

slaff:
Some thoughts on the refactoring:

I) We have to give a chance to the users to adapt to the new makefile re-organization.
Therefore the following could help:
- At the moment old and new makefiles are present. Please, replace the content of Makefile with the content from sming.mk and delete sming.mk. 
- Rename Linux.mk to Makefile-Linux.mk and remove the old Makefile-linux.mk file. Apply the same to the other OS makefiles.
- For Makefile-rboot.mk and Makefile-project.mk should be present but having the following content:

1.) Info message `$(info "...")` saying that the Makefile-{rboot|project}.mk will be removed in the next versions of Sming
2.) Internally include the new makefile.

II) Issues
- Assember .S files are included always.
```
AS system/irq_check.s
AS gdb/gdbstub-entry.s
```

While this is true for `irq_check.s` `gdbstub-entry.s` should be included ONLY when ENABLE_GDB=1 is given.

- Compilation of the library with `make -f sming.mk` finishes with error: 
```
sed -r "s/(^\s*irom0_0_seg *: *).*/\\1org = 0x4020A000, len = 0x0F6000/" /home/slavey/dev/esp8266.dev.box/dev/Sming/Sming//compiler/ld/rboot.rom0.ld >out/build/rom0.ld
LD out/build/app_0.out
out/build/app_app.a(user_main.o):(.text.user_init+0x4): undefined reference to `init()'
out/build/app_app.a(user_main.o): In function `user_init':
/home/slavey/dev/esp8266.dev.box/dev/Sming/Sming/appinit/user_main.cpp:70: undefined reference to `init()'
collect2: error: ld returned 1 exit status
make: *** [out/build/app_0.out] Error 1
```
That has to be fixed.

III) Makefile not in synch with the latest changes. I will write shortly  comments next to the line numbers.

Sming/Windows.mk
```diff
@@ -0,0 +1,34 @@
...
+# 	default serial settings
+#------------------------------------------------------------------------------
+COM_PORT		?= COM3
+
+#==============================================================================
+# 	tool settings
+#------------------------------------------------------------------------------
+ESPTOOL		 	?= $(SDK_TOOLS)/esptool.exe
+GET_FILESIZE 	?= stat --printf="%s"
+MEMANALYZER  	?= $(SDK_TOOLS)/memanalyzer.exe $(OBJDUMP).exe
```
slaff: For Windows we also use the python version of memanalyzer.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
+#   sming.mk
+#
+#   to do:
```
slaff: At some point take care to remove all comments from the top of the makefile and add them to the commit history.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+MODULES         ?= app
+
+#   your extra include directies (default: none)
+EXTRA_INCDIR    ?=
+
+#   your extra libraries (default: none)
+EXTRA_LIBS      ?=
+
+#   your extra source files (default: none)
+EXTRA_SRC       ?=
```
slaff: We have already variable called `MODULES`.Please, don't introduce yet another variable for the same thing.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+else
+	UNAME := $(shell uname -s)
+	include $(SMING_HOME)/$(UNAME).mk
+endif
+
+#==============================================================================
+#   default serial settings
+#------------------------------------------------------------------------------
+#   use the maximum speed supported by your board to get things done faster
+COM_SPEED_ESPTOOL   ?= 921600
```
slaff: Use the maximum supported speed from most devices - 115200. Not only from some (http://digital.ni.com/public.nsf/allkb/D37754FFA24F7C3F86256706005B9BE7). change the COM_SPEED_SERIAL too. Otherwise we will be overwhelmed with issues saying that flashing aborted, etc.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+
+CC              := $(XTENSA_TOOLS)/xtensa-lx106-elf-gcc
+CXX             := $(XTENSA_TOOLS)/xtensa-lx106-elf-g++
+AR              := $(XTENSA_TOOLS)/xtensa-lx106-elf-ar
+LD              := $(XTENSA_TOOLS)/xtensa-lx106-elf-gcc
+OBJCOPY         := $(XTENSA_TOOLS)/xtensa-lx106-elf-objcopy
+OBJDUMP         := $(XTENSA_TOOLS)/xtensa-lx106-elf-objdump
+STRIP           := $(XTENSA_TOOLS)/xtensa-lx106-elf-strip
+
+PERL            ?= perl
```
slaff: Get rid of Perl. Use stock bash commands and if bash cannot help only then use python.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+XXD             := xxd
+
+ifeq ($(MANUAL_RESET),1)
+	RESET_ESP       := $(PAUSE5) "*** PLEASE RESET ESP8266 NOW ***"
+endif
+
+export COMPILE  := gcc
+export PATH     := $(XTENSA_TOOLS):$(PATH)
+
+ifeq ("$(PYTHON)","")
```
slaff: Get rid of Perl. Use stock bash commands and if bash cannot help only then use python.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+    $(info FLASH MEMORY MAP)
+    $(info )
+    $(info MEM_SIZE        = $(MEM_SIZE))
+    $(info ESP_PARM_ADDR   = $(ESP_PARM_ADDR))
+    $(info ESP_INIT_ADDR   = $(ESP_INIT_ADDR))
+    $(info RF_CAL_ADDR     = $(RF_CAL_ADDR))
+    ifneq ($(USER_FLASH_SIZE),0)
+        $(info USER_FLASH_ADDR = $(USER_FLASH_ADDR) ($(USER_FLASH_SIZE) bytes))
+    endif
+    ifneq ($(SPIFF_SIZE),0)
```
slaff: We need calculations for RBOOT and non-RBOOT projects.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+EXTRA_INCDIR    += $(THIRD_PARTY_DIR)/spiffs/src
+EXTRA_INCDIR    += $(THIRD_PARTY_DIR)/rboot $(THIRD_PARTY_DIR)/rboot/appcode
+
+#==============================================================================
+#   intermediate files
+#------------------------------------------------------------------------------
+#   name for the target project
+TARGET          = app
+
+APP_AR          := $(BUILD_BASE)/$(TARGET)_app.a
```
slaff: Here we should tackle RBOOT AND non-RBOOT projects.

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+
+$(BUILD_DIR) $(FW_BASE):
+	$(Q) mkdir -p $@
+
+dirs: $(BUILD_DIR) $(FW_BASE)
+
+#==============================================================================
+#   (re-)compile sming framework if required
+#------------------------------------------------------------------------------
+LIBSMING        := sming
```
slaff: Why use ':=' instead of '=' ? Is there any reason?

Sming/sming.mk
```diff
@@ -0,0 +1,662 @@
...
+OBJ             := $(addprefix $(BUILD_BASE)/,$(addsuffix .o,$(basename $(SRC)))) $(EXTRA_OBJ)
+
+$(APP_AR): $(OBJ)
+	$(vecho) "AR $@"
+	$(Q) $(AR) cru $@ $^
+
+#==============================================================================
+#   create custom linker scripts
+#------------------------------------------------------------------------------
+IROM0_ORG0      := $(call hexcalc, 0x40200000 + $(ROM0_ADDR))
```
slaff: Here we should tackle RBOOT and non-RBOOT projects.

Sming/sming.mk
tius2000: ok

Sming/sming.mk
tius2000: I thought that _MODULES_ is used to specify directories? _EXTRA_SRC_ is used to specify single files.

Sming/sming.mk
tius2000: yeah, you are probably right ;-)

Sming/sming.mk
tius2000: I'm not sure, whether I understand this. There are calculations in both branches, but those lines are for debug info only.

Sming/sming.mk
tius2000: Do we need a different library name for non-rboot?

Sming/sming.mk
tius2000: Well, I always prefer ':=' if possible. It has less confusing effects (more conventional logic, no "backward assignments") for occasional make users.

Sming/sming.mk
slaff: > EXTRA_SRC is used to specify single files.

Ok, then leave it there.

Sming/sming.mk
tius2000: This is done by the different values for _ROM0_ADDR_. Does it not work for you?

Sming/sming.mk
slaff: No, better use the same library name, and same application name.

tius2000:
Dear @slaff, thank you very much for reviewing! I agree, there is still a long way before that should be merge. 

However, I was wondering, whether marking the new makefile as "experimental" in the beginning might be an option. That way we will not break things, but users willing to try the new options can give it try. You could also place a hint with _$(info)_ within the old makefiles to motivate the user to give it a try.

Sming/Windows.mk
tius2000: This will make python a new required for windows builds. Is there any advantage of the python version?

Sming/sming.mk
tius2000: I will check, whether it is possible to use standard bash commands instead.

anakod: I think in most cases python should be already available. Current Makefile require python as dependency, am I right? If yes - we can skip additional code branches and make things more simple. Python is good for calculation and other things like that in my opinion.

anakod:
I'm prefer to replace old Makefiles with your version, but please take attention to II) and III) points in @slaff comment above.

Sming/sming.mk
tius2000: No, currently neither python nor perl are required for windows builds or standard tools on windows machines. I will try to remove dependencies on both.

tius2000: done

tius2000:
Well, IMHO top priority should be **not to break any existing projects**. As I do not think that we can fix any possible incompatibility for every existing project before the first release, it might be a good idea to move to the new makefile in four phases.

phase 1:
Use the new name and mark the new makefile as "experimental". Include a message in the old makefiles to tell the users that there is something new.

phase 2:
Fix and improve the new makefile based on user feedback.

phase 3:
Mark the old makefiles as "deprecated" and inform the users, that they should switch to the new make process. 

phase 4:
Remove the old makefiles or replace them with a stub pointing to the new makefile.

This way we do not break things and our users keep complete control of there projects. What do you think?

tius2000:
> II) Issues
> 
>     Assember .S files are included always.
> 
> AS system/irq_check.s
> AS gdb/gdbstub-entry.s
> 
> While this is true for irq_check.s gdbstub-entry.s should be included ONLY when ENABLE_GDB=1 is given.

The whole gdb directory structure looks very confusing to me. _makefile-rboot.mk_ references _$(THIRD_PARTY_DIR)/gdbstub_ and I kept this. However, this directory does not exist on my system. I see copies of _gdbstub-entry.s_ in _Sming/gdb_ and _Sming/third-party/esp-gdbstub_. This is why the compilation goes wrong.

Which copy should be removed? Which directory should be compiled when ENABLE_GDB is set?

tius2000:
> Compilation of the library with make -f sming.mk finishes with error

The new makefile is meant as a replacement for _makefile-rboot.mk_ and _makefile-project.mk. It is not meant also to replace _Makefile_ at the moment. However, this might be a task for a later phase.

slaff:
> The new makefile is meant as a replacement for makefile-rboot.mk and _makefile-project.mk

Perfect. Make sure to add an info message guiding our users.

> Which directory should be compiled when ENABLE_GDB is set?

If you compile the library / or a project with  ENABLE_GDB=1 then 
1) there are new compile flags appended to the current ones.
2) The `$(THIRD_PARTY_DIR)/gdbstub` directory is added to the directories that include header files
3) The `$(THIRD_PARTY_DIR)/gdbstub` directory is added to the directories that contain source code and all (*.c, *.cpp AND *.s) files in it should be compiled.
```
else ifeq ($(ENABLE_GDB), 1)
	CFLAGS += -Og -ggdb -DGDBSTUB_FREERTOS=0 -DENABLE_GDB=1
	MODULES		 += $(THIRD_PARTY_DIR)/gdbstub
	EXTRA_INCDIR += $(THIRD_PARTY_DIR)/gdbstub
	STRIP := @true
else
```

tius2000:
> If you compile the library / or a project with ENABLE_GDB=1 then
> 
>there are new compile flags appended to the current ones.
>The $(THIRD_PARTY_DIR)/gdbstub directory is added to the directories that include header files
>The $(THIRD_PARTY_DIR)/gdbstub directory is added to the directories that contain source code and all (*.c, *.cpp AND *.s) files in it should be compiled.

I kept this from _makefile-rboot.mk_. However, the [path seems not to be correct](https://github.com/SmingHub/Sming/pull/985#issuecomment-286401061). Is there something wrong with my setup?

slaff:
So far in the latest `develop` we have the following implemented:

- [x] roms with and without bootloader
- [x] automatic flash address calculation + automatic linker script adjustment per application
- [x] dependency tracking for all header files included in cpp and c sources ( that includes dependency tracking for spiffs sources )
- [x] memanalyzer for rboot builds added

What is still missing is that one:
- [ ] automatic patch of esp_init_data_default.bin for system_get_vdd33()
@tius2000 Is that still needed ? Can you refresh my memory and tell me why was that added and do we need to have it for all applications?

tius2000:
Hi @slaff, if you want to use _system_get_vdd33()_, you need to patch _esp_init_data_default.bin_, otherwise you will get wrong values.

slaff:
> you need to patch esp_init_data_default.bin, otherwise you will get wrong values.

We can think about patching the physical params by wrapping around `register_chipv6_phy`. This way we will have better control over the phy data and be better prepared for changes in the phy params in new SDKs.

PR #986, opened 2017-02-26
Add flash debug strings to third party libs

ADiea:
Free up additional RAM by enabling flash storage for lwip,axtls and umm_malloc
Enable debug in lwip and set it default to log tcp reset errors


ADiea:
Seems there are some issues with espconn will investigate

Sming/third-party/.patches/axtls-8266.patch
```diff
@@ -114,13 +114,10 @@ diff -Nuar a/replacements/mem.c b/replacements/mem.c
 +
 +void ax_port_free(void* ptr) {
 +    free(ptr);
-+    uint32_t *p = (uint32_t*) ptr;
-+    size_t size = p[-3];
-+    if (size >= 1024)
-+        DEBUG_TLS_MEM_PRINT("free %d, left %d\r\n", p[-3], system_get_free_heap_size());
++    DEBUG_TLS_MEM_PRINT("free %d, left %d\r\n", p[-3], system_get_free_heap_size());
```
slaff: `p` is not defined here? That code shouldn't compile.

Sming/third-party/.patches/axtls-8266.patch
```diff
@@ -196,6 +222,24 @@ index 0cf2e4c..78576f7 100644
      if (IS_SET_SSL_FLAG(SSL_SENT_CLOSE_NOTIFY))
          return SSL_CLOSE_NOTIFY;
  
+@@ -2226,8 +2229,15 @@ void DISPLAY_STATE(SSL *ssl, int is_send, uint8_t state, int not_ok)
+     if (!IS_SET_SSL_FLAG(SSL_DISPLAY_STATES))
+         return;
+ 
+-    printf(not_ok ? "Error - invalid State:\t" : "State:\t");
```
slaff: I would prefer if that block stays as it was. The changes are only coding style ones and axTLS maintainers should be the one caring about coding style. We care here only about making thing more with Sming and work in the best possible way.

Sming/third-party/.patches/axtls-8266.patch
```diff
@@ -459,3 +503,16 @@ index 4972119..f6f44f8 100644
...
+diff --git a/crypto/crypto_misc.c b/crypto/crypto_misc.c
+index f166bb8..0ffe2aa 100644
+--- a/crypto/crypto_misc.c
++++ b/crypto/crypto_misc.c
+@@ -290,7 +290,7 @@ EXP_FUNC void STDCALL print_blob(const char *format,
+     va_list(ap);
+ 
+     va_start(ap, size);
+-    sprintf(tmp, "%s\n", format);
++    snprintf(tmp, sizeof(tmp), "SSL: %s\n", format);
```
slaff: What is the reason for that change?

Sming/third-party/.patches/umm_malloc.patch
```diff
@@ -16,12 +16,12 @@ index fcc4fb1..661b94a 100644
...
 +#ifdef __ets__
 +// ESP specific code
++#include "debug_progmem.h"
 +extern void ets_intr_lock();
 +extern void ets_intr_unlock();
-+extern void ets_printf(const char*, ...);
 +
-+#define printf ets_printf
-+#define puts(...) ets_printf("%s",...)
++#define printf debug_i
```
slaff: I would be extremely cautious in adding DRAM magic to code related to such an important stuff as low level memory allocation. Was that change tested extensively on a device with a lot of memory allocation / freeing operations? Is there a slow down? Is the code functioning without segfaults?

Sming/third-party/.patches/axtls-8266.patch
ADiea: This code was used to try and guess the length of the freed chunk. If freed chunlk was over 1024 => printf.
This is however not portable, when using for ex umm_malloc size = p[-3] will not return you the size because umm_malloc does not store size, but only next free block.

Sming/third-party/.patches/axtls-8266.patch
ADiea: This was necessary because argument of printf would not expand to a string at preprocessing time and will cause expansion of debug_i to fail...

Sming/third-party/.patches/axtls-8266.patch
ADiea: This was included for robustness. Not directly part of PR's scope I can remove it.

slaff: If you want send it as PR to the upstream project.

Sming/third-party/.patches/umm_malloc.patch
ADiea: The debug_i macro uses printf_P_stack which does not involve any heap alloc... But I also see your concern. Will remove this and make a separate PR Then we can have better confidence. Will then test on my application which does SSL / websockets etc and measure some statistics.

Sming/third-party/.patches/axtls-8266.patch
slaff: Looking at the [patch file](https://github.com/ADiea/Sming/blob/c12c6bd84d3e316ce5966d0648dca2bc881e2dfa/Sming/third-party/.patches/axtls-8266.patch) I see

```C
+void ax_port_free(void* ptr) {
+    free(ptr);
+    DEBUG_TLS_MEM_PRINT("free %d, left %d\r\n", p[-3], system_get_free_heap_size());
+}
```

In that function `p` is used but it is not declared. So, if I am not totally mistaken, this code should not compile at all. May be you meant to replace `p` with `ptr` casted to uint32_t?

Sming/third-party/.patches/axtls-8266.patch
slaff: > This was necessary because argument of printf would not expand to a string at preprocessing time

Hm... that is not a good sign. Is there a way to make the expansion and preprocessing work as expected?

ADiea: actually it could be fixed by parenthesis ?! will see, thx for review

ADiea: I had no success. It will expand to ```char x[] = (condition ? "aa" : "bb") "\r\n"``` which will fail to be pasted together.

Sming/third-party/.patches/axtls-8266.patch
ADiea: You are right, removed it completely. We have no way to determine size of allocated chunk in free() in a portable way. Instead we'll print the address freed.

ADiea:
@slaff do you have other concerns here? Except the conditional operator in a printf (found no solution for preprocessor eval & pasting in this case) it's ok to merge from my tests. I also excluded debug_i from ummalloc for safety although prints were happening only on error cases but will stress test more

slaff:
@ADiea I would like to test it a bit better on my devices. Unfortunately I do not have the time right now. It would be great if you can ask some of the other active developers/users of Sming to test before I jump in.

Sming/Makefile
```diff
@@ -177,6 +177,12 @@ ENABLE_ESPCONN ?= 0
 ifeq ($(ENABLE_CUSTOM_LWIP), 1)
 	THIRD_PARTY_DATA += third-party/esp-open-lwip/Makefile.open
 	EXTRA_INCDIR += third-party/esp-open-lwip/include
+	EXTRA_CFLAGS_LWIP = -I../../system/include -I../../Wiring
+	ENABLE_LWIPDEBUG ?= 1
```
slaff: Can you set the default for ENABLE_LWIPDEBUG to 0? Otherwise LWIP may start printing a lot of data.

Sming/Wiring/FakePgmSpace.cpp
```diff
@@ -1,4 +1,4 @@
-
+#include "WiringFrameworkDependencies.h"
```
slaff: I am just curious: what is the reason for moving that include from the header file to here?

Sming/third-party/.patches/axtls-8266.patch
```diff
@@ -1,40 +1,7 @@
-diff -Nuar a/replacements/libc.c b/replacements/libc.c
```
slaff: The libc file is completely gone. Is that correct?

Sming/third-party/.patches/axtls-8266.patch
slaff: For that line: can you just change it from:

```
printf(not_ok ? "Error - invalid State:\t" : "State:\t");
```
to

```
printf("%s", (not_ok ? "Error - invalid State:\t" : "State:\t") );
```

That change does not add new lines and will be easier to merge into `upstream` or other axtls clones.

Sming/third-party/.patches/esp-open-lwip.patch
```diff
@@ -125,6 +142,24 @@ index eaa8dd6..6a08192 100644
  #endif
  #endif
  
+@@ -1823,7 +1823,7 @@
+  * debug messages of certain types.
+  */
+ #ifndef LWIP_DBG_TYPES_ON
+-#define LWIP_DBG_TYPES_ON               LWIP_DBG_OFF
++#define LWIP_DBG_TYPES_ON               LWIP_DBG_ON
```
slaff: Leave the debugging to be OFF by default. If someone wants to have it ON he/she should still be able to set it by using the ENABLE_LWIPDEBUG=1

Sming/third-party/.patches/esp-open-lwip.patch
```diff
@@ -60,6 +60,7 @@ index 1bc584f..0bfc424 100644
 +lwip/app/espconn_tcp.o \
 +lwip/app/espconn_udp.o \
 +lwip/app/espconn_mdns.o \
++lwip/api/err.o \
```
slaff: Just for my understanding: in which case is `api/err.c` needed?

Sming/third-party/.patches/axtls-8266.patch
ADiea: Not completely agree with you here. To me the effort of 2 minutes more is worth the extra 50 bytes of saved RAM because your proposed change will store "%s" in flash and the other const char strings in RAM...

Sming/Makefile
ADiea: Sure I can. Right now it prints only errors that are related to tcp reset which I find very helpful and are not shoiwing up very often but of course I can disable it maybe better for more people this way

Sming/Wiring/FakePgmSpace.cpp
ADiea: This was originally in the header. When including debug_progmem.h in external projects like third party thi would start including all sming header uselessly and generate all sorts of issues. This header is actually needed by the source file => we keep header file clean for including in external thirtparty projects

Sming/Makefile
slaff: > it prints only errors

Then leave it as it is.

Sming/third-party/.patches/axtls-8266.patch
ADiea: No need for it anymore. m_printf is used by including debug_progmem.h

Sming/third-party/.patches/axtls-8266.patch
slaff: >  is worth the extra 50 bytes of saved RAM

Ok, then leave the change as it is.

Sming/third-party/.patches/esp-open-lwip.patch
ADiea: When compiling with ENABLE_ESPCONN esconn code uses a function defined in err.c to map an int error code to a string. Absence of err.o would generate linker symbol not found err

Sming/third-party/.patches/esp-open-lwip.patch
ADiea: Will disable

Sming/Wiring/FakePgmSpace.cpp
slaff: > external projects like third party thi would start including all sming header uselessly and generate all sorts of issues.

Ok, so it is the same reason that I was thinking about.

slaff:
@ADiea Sorry for the delay, but the earliest I can test it on a real device will be later today or early tomorrow. I can see two major tests: with debug ON and OFF. I will let you know If I find some issues .

Sming/third-party/.patches/esp-open-lwip.patch
ADiea: This is a second step to enable debugging in LWIP. It is still dependent on ENABLE_LWIPDEBUG in Sming/Makefile. In other words if ENABLE_LWIPDEBUG is 0 debugging will be disabled completely This has to remain like this to enable all debug messages. Please see macro definition of LWIP_DEBUGF in (lwip/include/lwip/debug.h)

ADiea:
@slaff thanks very much! I have disabled lwip logging by default in Makefile. If you want to see more lwip logs from specific areas like dhcp you can enable them in lwiopts.h the same like tcp_reset.

Issue #987, opened 2017-02-27
Replace memanalyzer with python script


slaff:
The current [memanalyzer](https://github.com/Sermus/ESP8266_memory_analyzer) tool is available on platforms that have .NET support which increases the dependency of external tools and libraries. Because we already have dependency on python (esptool.py depends on it) and it is available on all supported OSes (Windows, FreeBSD, Linux, MacOS X) it would be a good idea to create a python script that produces output as  `memanalyzer` and allows us to do more fine grained calculations as   free heap, heap start address, stack start address, available memory space between heap and stack, etc. 

```
#Memory / Section info:
------------------------------------------------------------------------------
   Section|                   Description| Start (hex)|   End (hex)|Used space
------------------------------------------------------------------------------
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8374|     884
    rodata|           ReadOnly Data (RAM)|    3FFE8378|    3FFEA0B8|    7488
       bss|      Uninitialized Data (RAM)|    3FFEA0B8|    3FFF0F70|   28344 (+8)
      text|            Cached Code (IRAM)|    40100000|    40105D46|   23878
irom0_text|           Uncached Code (SPI)|    40209000|    40244AFA|  244474 (+32)
Total Used RAM : 36716
Free RAM : 45204
Free IRam : 8908
------------------------------------------------------------------------------
```

Plus we will have consistent output for all supported OSes. As a start we can use  PR #753  from @cmarrin.

PR #988, opened 2017-02-28
Unified the memanalyzer output on all supported OSes.

slaff:
Implements #987


slaff:
> Why not use the same for windows that way we have a unified solution and don't need mema~.exe anymore

It is already used also in Windows :)

slaff:
@ADiea Thanks for noticing. That #989 should add memanalyzer.py also to Windows.

PR #1006, opened 2017-03-04
Smart config 1004

### LINKED ISSUE #1004: smartConfig does not work
The smartConfig sample (and default smartConfig function) does not work. It gets to the SC_STATUS_LINK stage but does not complete.

riban-bw:
Fixes #1004 smartConfig.


riban-bw:
Sorry - applied to SmingHub - should have applied to Sming. I will close this PR and (attempt to) do it right!

PR #1014, opened 2017-03-07
Guard CommandExecutor and CommandHandler

ADiea:
This functionality is sometimes not used but because it is present in places like websock and http server it occupies RAM and heap

Without CommandExecutor, compiling sample HttpServer_WebSockets
```
#Memory / Section info:
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8384|     900
    rodata|           ReadOnly Data (RAM)|    3FFE8390|    3FFEA264|    7892 (-956)
       bss|      Uninitialized Data (RAM)|    3FFEA268|    3FFF1100|   28312 (-144)
      text|            Cached Code (IRAM)|    40100000|    40105E46|   24134
irom0_text|           Uncached Code (SPI)|    4020A000|    40254163|  303459 (-3247)
Total Used RAM : 37104
Free RAM : 44816
Free IRam : 8652
```


anakod:
I prefer use full names to keep things clear, it shouldn't be problem to write it :)

DISABLE_COMMAND_EXECUTOR or DISABLE_CMD_EXECUTOR

anakod:
Sorry! Have checked current Makefile code, if we already have "ENABLE_SSL" than we should add new flags in same manner, I think: ENABLE_XXX.
So ENABLE_COMMAND_EXECUTOR=0 or ENABLE_CMD_EXECUTOR=0

Sming/Makefile-rboot.mk
```diff
@@ -188,7 +191,7 @@ EXTRA_INCDIR += $(SMING_HOME)/include $(SMING_HOME)/ $(LWIP_INCDIR) $(SMING_HOME
 USER_LIBDIR  = $(SMING_HOME)/compiler/lib/
 
 # compiler flags using during compilation of source files
-CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL) $(USER_CFLAGS)
+CFLAGS		= -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL) $(USER_CFLAGS) -DDISABLE_CMD_EXEC=$(DISABLE_CMD_EXEC)
```
slaff: The same directive should be also in the Makefile-project.mk file.

ADiea: Basically this is only used for sming building so app should not care for this flag. But will add it anyway for completion.

slaff: Ok, then just remove it from here. Rename also the directive as anakod suggested: ENABLE_COMMAND_EXECUTOR and take care to change the preprocessor if-conditions accordingly.

PR #1033, opened 2017-03-13
Initial public version of the asynch AtClient command handler.

slaff:
Useful for communicating with AT modems, AT webcams or other devices that use AT command set.

### DIFF
--- a/Sming/SmingCore/AtClient.cpp
+++ b/Sming/SmingCore/AtClient.cpp
@@ -1,0 +1,168 @@
+/*
+ * AtClient.cpp
+ *
+ *  Created on: Feb 23, 2017
+ *      Author: slavey
+ */
+
+#include <AtClient.h>
+
+#ifndef debugf
+#define debugf(fmt, ...)
+#endif
+
+AtClient::AtClient(HardwareSerial* stream) : stream(stream) {
+	this->stream->setCallback(StreamDataReceivedDelegate(&AtClient::processor, this));
+}
+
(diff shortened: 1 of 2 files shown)


ADiea:
Hint: Try to use String& function parameters to avoid copying and malloc-ing new data for every string object

anakod:
> Hint: Try to use String& function parameters to avoid copying and malloc-ing
"const String&" is the best option for input (readonly) parameters for compatibility.

slaff:
> no code for receiving and parsing response?

I am not sure that I understood you. There is actually code that is taking care for the received data: `AtClient::processor`. In it there is a basic parsing of command responses. If something special is needed then there are two callbacks where the user can implement his/her logic - one is called `onReceive` and will be triggered every time data is received. In this case it will be up to the developer to process the response. Useful in the cases where you have binary data coming back. Or something different than the usual `short-word-response-\r-\n`. The other call back is `onComplete` and it assumes that the response contains end marker \r\n. Once that end is reached the whole response will be passed back to the onComplete callback. In there the developer can process for example the data and decide what to do with it. Example: an AT command to return the current MAC address `AT+CIPAPMAC_CUR?` will get a response `+CIPAPMAC_CUR:"62:01:94:80:e0:f3"` and then the developer can process the response using `onComplete` callback. 

Think of AtClient as the base class for other "higher-level" classes: on top of the AtClient and its "low-level" functions one can build a more specific AT command client that is targeted to a specific device. And in there the response processing can be built-in.

PR #1042, opened 2017-03-21
[Fixes #994] unify delegate class name styles

### LINKED ISSUE #994: Unify style for delegate class names
I just found out that there are 2 styles (maybe more?, I just started learning Sming) used to format delegate class names, there is eg. `ConnectionDelegate` and `onStationDisconnectDelegate`. I think it would be better to maintain consistent coding style across framework.

zgoda:
Both samples that use `WifiEvents` compile, flash and run fine - these are `Websocket_Client` and `Basic_WebSkeletonApp`. There are no samples that use rBoot OTA update and I was not able to test this code on live system.

This change will break user code that uses `WifiEvents` and creates delegate instances directly. Code that relies on internal delegates created in `WifiEvents` class is not affected.

### DIFF
--- a/Sming/SmingCore/Network/rBootHttpUpdate.cpp
+++ b/Sming/SmingCore/Network/rBootHttpUpdate.cpp
@@ -35,11 +35,11 @@ void rBootHttpUpdate::switchToRom(uint8 romSlot) {
 	this->romSlot = romSlot;
 }
 
-void rBootHttpUpdate::setCallback(otaUpdateDelegate reqUpdateDelegate) {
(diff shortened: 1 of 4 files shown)


avr39-ripe:
@zgoda I think you should provide patches for either samples that use WifiEvents for this change to be consistent

zgoda:
These samples do not use delegate classes directly, they're used only internally in `WifiEvents` class.

avr39-ripe:
@zgoda You're right, it was my old code where I do not yet use classes, so no mention of on* delegates, sorry.. I use on* delegates of WifiEvents in my new code :)

PR #1043, opened 2017-03-21
Fix crash at PROBEREQRECVED event if only STADISCONNECTED delegate defined

pief:
WifiEvents' event handler for EVENT_SOFTAPMODE_PROBEREQRECVED
incorrectly tested for definition of a onSOFTAPDisconnect delegate where it
actually calls the onSOFTAPProbeReqRecved delegate, so when a delegate
for EVENT_SOFTAPMODE_STADISCONNECTED was defined but not for
EVENT_SOFTAPMODE_PROBEREQRECVED, the ESP would crash.


pief:
Replacing with PR with proper branch name

PR #1044, opened 2017-03-22
Make WifiStation.enable(false) disable waitConnection's timer

pief:
Disabling WifiStation did not prevent a delegate installed via a call to
waitConnection() from being called after the timeout.


pief:
Replacing with PR with proper branch name

PR #1046, opened 2017-03-22
Make WifiStation.enable(false) disable waitConnection's timer

pief:
WifiEvents' event handler for EVENT_SOFTAPMODE_PROBEREQRECVED
incorrectly tested for definition of a onSOFTAPDisconnect delegate where it
actually calls the onSOFTAPProbeReqRecved delegate, so when a delegate
for EVENT_SOFTAPMODE_STADISCONNECTED was defined but not for
EVENT_SOFTAPMODE_PROBEREQRECVED, the ESP would crash.

### DIFF
--- a/Sming/SmingCore/Platform/Station.cpp
+++ b/Sming/SmingCore/Platform/Station.cpp
@@ -33,7 +33,10 @@ void StationClass::enable(bool enabled, bool save)
 		mode = wifi_get_opmode_default() & ~STATION_MODE;
 	else
 		mode = wifi_get_opmode() & ~STATION_MODE;
-	if (enabled) mode |= STATION_MODE;
+	if (enabled)
+		mode |= STATION_MODE;
+	else if (connectionTimer)
+		delete connectionTimer;
 	if (save)
 		wifi_set_opmode(mode);
 	else
--- a/Sming/SmingCore/Platform/Station.h
+++ b/Sming/SmingCore/Platform/Station.h
@@ -84,6 +84,7 @@ public:
 	~StationClass();
 
 	/**	@brief	Enable / disable WiFi station
+	 *	@note	Disabling WiFi station will also disable and clear the handler set with <i>waitConnection</i>.
 	 *	@param	enabled True to enable station. False to disable.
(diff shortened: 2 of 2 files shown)


avr39-ripe:
I think waitConnection stuff should be abandoned, it do not works reliable, there is WifiEvents, and they make its job rather good. But fix itself is good as I can see.

slaff:
> it do not works reliable, there is WifiEvents, and they make its job rather good.

@avr39-ripe Make a PR request for this.

avr39-ripe:
@slaff I think I should make Issue first, to decide and VOTE for removal of waitConnection stuff

pief:
Removal of waitConnection would be a separate issue, yes, thus my fix would still apply in the mean time, as we'll certainly not want to remove waitConnection in a minor release?

slaff:
I am hesitant to merge this PR because I have seen developers using the following code
```
WifiStation.enable(true);
WifiStation.waitConnection(connectOk, 30, connectFail);

// ... 
WifiStation.enable(false);

//.. and later on 
WifiStation.enable(true); // and here the developers expected that the connectOk or connectFail should work...
```
If we merge this PR there should be documentation to the `enable(flag)` method  that warns the developers that once disabled ( = enable(false) ) the callbacks should be attached again via a new call to `waitConnection`.

avr39-ripe:
@slaff, too many work for feature considered deprecated IMHO..

pief:
I'd have no trouble adding a commit to update the documentation if 3.2.0 is still some weeks away, especially as waitConnection is not mentioned to be deprecated anywhere.

avr39-ripe:
@pief Indeed not marked yet, you're right, but is it really worth it? Try out new, and proper way of doing things, see #1047 for details. Of course if you insist fixing waitConnection you can provide PR and it will be merged I think :)

slaff:
> I'd have no trouble adding a commit to update the documentation

@pief That would be great. Thanks in advance.

pief:
Here we go. Proposing to merge this for a 3.1.3. For 3.2.x, I support @avr39-ripe's proposal of dropping waitConnection in favor of WifiEvents.

PR #1055, opened 2017-03-29
HardwarePWM: Fix period unit in comments to uS instead of nS

pief:
1000 ms are 1000 * 0,001s = 1s which gives a frequency of 1/1 = 1 Hz
instead of 1 KHz. Now either the period or the frequency is wrong. Turns
out it's the period as the ESP8266 Technical Reference associates
1000 uS with 1 KHz.


slaff:
> ESP8266 Technical Reference associates
1000 uS with 1 KHz.

@pief  Would it be possible to add also link in the changed comment pointing to the technical reference that you are mentioning?

pief:
Like this?

PR #1060, opened 2017-04-04
Fix i2c master by properly setup SDA pin as INPUT_PULLUP

avr39-ripe:
Fixes issue with some i2c setup where INPUT_PULLUP is mandatory (missing SDA/SCL pull up resistors).
There is new version of pinMode after Sming 3.1.0, so to be consistent with old behavior, SDA pin MUST be INPUT_PULLUP, not only INPUT and digitalWrite(SDA, HIGH). This was tested with MCP23017 as both input and output port expander connected to esp8266 without SDA/SCL resistors.


slaff:
@avr39-ripe Do we need that change or this is no longer needed if #1061 is merged?

avr39-ripe:
this applicable only to current i2cMaster, no need for new version, as this file is deleted and replaced by new i2c HAL

PR #1061, opened 2017-04-05
Ported from esp8266/Arduino i2c HAL and Wire library

avr39-ripe:
Ported from esp8266/Arduino i2c HAL, faster by direct pin manipulation
not via digitalWrite, some additional features of i2c bus implemented
Wire library implement some additional features such setClock.
Looks like being fully compatible with former implementation of both i2c HAL and Wiring lib.
Tested with DS3232RTC_NTP_Setter, LiquidCrystal_44780 samples and my app with MCP23017


Sming/SmingCore/Wire.cpp
```diff
@@ -1,145 +1,249 @@
...
+  // // !!! this will kill any pending pre-master sendTo() activity
+  // txBufferIndex = 0;
+  // txBufferLength = 0;
+  // // alert user program
+  // user_onRequest();
 }
 
-TwoWire Wire = TwoWire(I2C_DEFAULT_SCL_PIN, I2C_DEFAULT_SDA_PIN);
+void TwoWire::onReceive( void (*function)(int) ){
+  //user_onReceive = function;
```
slaff: Why the onReceive callback stuff is commented out?

Sming/SmingCore/Wire.h
```diff
@@ -1,60 +1,91 @@
...
+
+    inline size_t write(unsigned long n) { return write((uint8_t)n); }
+    inline size_t write(long n) { return write((uint8_t)n); }
+    inline size_t write(unsigned int n) { return write((uint8_t)n); }
+    inline size_t write(int n) { return write((uint8_t)n); }
+    using Print::write;
 };
 
+#if !defined(NO_GLOBAL_INSTANCES) && !defined(NO_GLOBAL_TWOWIRE)
 extern TwoWire Wire;
```
slaff: It would be better, memory wise, to avoid creating global objects.  Therefore I would prefere the global instance to be created only if requested (in the code above it is created by default).

Sming/SmingCore/Wire.cpp
avr39-ripe: This is PORT of the library with as minimum changes as possible. May be we should import it as is, try how it works and then add more useful stuff to it as delegates etc??

Sming/SmingCore/Wire.h
avr39-ripe: Creating default Wire object is "good?? old Arduino practice" so I think we should not invent wheel and disable this. IMHO.

ADiea: Yes, also if no-one uses the Wire object, linker will not include it in final binary so no memory impact

Sming/SmingCore/Wire.h
```diff
@@ -1,60 +1,91 @@
...
+
+    static uint8_t transmitting;
+    static void (*user_onRequest)(void);
+    static void (*user_onReceive)(int);
+    static void onRequestService(void);
+    static void onReceiveService(uint8_t*, int);
+  public:
+    TwoWire();
+    void begin(int scl, int sda);
+    void pins(int scl, int sda);
```
zgoda: This method is deprecated in https://github.com/esp8266/Arduino/blob/master/libraries/Wire/Wire.h#L54

avr39-ripe: Yes, deprecated, but it widely used in our samples and it do not hurts, so IMHO it should be included. Moreover, I changed SCL/SDA definition order from original SDA,SCL to used now in samples//libs/etc SCL/SDA.

zgoda: This will require porting for more hw libs as the argument order is reversed - or writing compat layers on top. Keeping compatibility with Arduino will make drop-in replacement for Wire.h. Hard to say what will be better although I don't mind writing some code to make existing lib working.

avr39-ripe: Yes, we should decide either rewrite current samples/libs to use SDA/SCL pin order or rewrite every next lib with Wire. Even after my change of pin order I think it was mistake. Better change pin order in known list of our samples/libs and our own code and then have compatibility with Arduino. Comments are welcome! Thank you @zgoda to point this issue!

Issue #1048, opened 2017-03-22
sprintf_P broken


zgoda:
This is how Basic_ProgMem fails in current HEAD:

```
--- Miniterm on /dev/ttyUSB0  115200,8,N,1 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
tail 12
chksum 0x19
ho 0 tail 12 room 4
load 0x3ffe8000, len 900, room 12 
tail 8
chksum 0xf4
load 0x3ffe8390, len 2772, room 0 
tail 4
chksum 0xa1
csum 0xa1
��{�d�l␛�␄␄�d`␂��r�l�$␓�␄␌␄l ␃��{�l␄��␀�␌␄ld`␃{$�> Read bytes from PROGMEM:
1 2 3 4 5 6 7 8 
> Copy a string from PROGMEM:
assertEqualsString: Demo != Demo�� @p��?
Demo�� @p��?
assertEqualsString: DemoDemo != Demo�� @p��?Demo
assertEqualsString: Demo 1 != (...)
> Copy bytes from PROGMEM:
1 2 3 4 5 
> ESP8266EX Memory Layout:
> 0x3FFE8000 ~ 0x3FFFBFFF - User data RAM, 80kb. Available to applications.
> 0x40200000 ~ ...        - SPI Flash.
> demoRam array address: 0x3FFE89E9 is in the RAM
> demoPgm array address: 0x40234AEC is in the Flash
```

The code looks like a test case so to me if is clearly failing.

And `sprintf_P` always returns string `"(...)"` regardless of format string and params.

slaff:
@ADiea when you are back can you look at that issue?

ADiea:
Next week i plan to start sming-ing again. Will have a look

zgoda:
The change from #1062 has no impact on this issue.

ADiea:
@zgoda I think PR #1065 will fix this can you check please?

zgoda:
Okay, will try this and report back.

ADiea:
because stuff like 
assertEqualsString: Demo != Demo�� @p��?
could mean that zero character at the end was not present perhaps

zgoda:
With this fix:

```
Read bytes from PROGMEM:
1 2 3 4 5 6 7 8 
> Copy a string from PROGMEM:
Demo
assertEqualsString: Demo 1 != (...)
> Copy bytes from PROGMEM:
[...]

PR #1068, opened 2017-04-07
[Fixes #1065] Copy terminating null byte in strcpy_P

### LINKED ISSUE #1065: strcpy_P: terminating null byte not copied
IMHO _strcpy_P_ should also copy the terminating null byte.

zgoda:
So it behaves exactly as regular `strcpy`. Thanks @tius2000 for working fix in code.

This is related to #1048 as well and fixes 2 of 3 failing tests.

### DIFF
--- a/Sming/Wiring/FakePgmSpace.cpp
+++ b/Sming/Wiring/FakePgmSpace.cpp
@@ -31,8 +31,7 @@ extern "C" size_t strlen_P(const char * src_P)
 
 extern "C" char *strcpy_P(char * dest, const char * src_P)
 {
-	int len = strlen_P(src_P);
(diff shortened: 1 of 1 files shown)


ADiea:
Also strncpy needs fix
For sprintf_P 
I think
'''__result = m_snprintf(s, len_P, __localF, ##__VA_ARGS__); \
'''
Should pass len_P+1

tius2000:
I'm not sure @ADiea, whether memcopy_P will always improve performance, as you need strlen_P before and the string is always iterated twice that way. Did you benchmark this for different string lengths? Perhaps there could be a 2nd method for aligned strings?

ADiea:
@tius2000 i think you are right did not pay attention to strlen

PR #1071, opened 2017-04-10
WifiStation.waitConnection is marked as obsolete and will be removed …

slaff:
…in coming versions.

Use WifiEvents instead.
The Basic_Wifi example is changed and should be used as a starting point for the migration of your own code.

### DIFF
--- a/samples/Basic_WiFi/app/application.cpp
+++ b/samples/Basic_WiFi/app/application.cpp
@@ -28,17 +28,17 @@ void listNetworks(bool succeeded, BssList list)
 }
 
 // Will be called when WiFi station was connected to AP
-void connectOk()
+void connectOk(IPAddress ip, IPAddress mask, IPAddress gateway)
 {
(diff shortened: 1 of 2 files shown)


samples/Basic_WiFi/app/application.cpp
```diff
@@ -28,17 +28,17 @@ void listNetworks(bool succeeded, BssList list)
 }
 
 // Will be called when WiFi station was connected to AP
-void connectOk()
+void connectOk(IPAddress ip, IPAddress mask, IPAddress gateway)
 {
 	debugf("I'm CONNECTED");
-	Serial.println(WifiStation.getIP().toString());
+	Serial.println(ip.toString().c_str());
```
zgoda: Is there any reason to convert to C String here?

slaff: You are right. The standard string should work just fine `size_t Print::println(const String &s)` . Thanks for pointing this out.

PR #1072, opened 2017-04-10
[Rel #1071] more samples converted to WifiEvents

zgoda:
This is related to PR #1071 - all samples are converted to use `WifiEvents`. The PR is massive, but all network related examples have been converted.

### DIFF
--- a/samples/Arducam/app/application.cpp
+++ b/samples/Arducam/app/application.cpp
@@ -257,9 +257,14 @@ void StartServers()
 
 
 // Will be called when WiFi station was connected to AP
-void connectOk()
+void connectOk(String ssid, uint8_t ssid_len, uint8_t bssid[6], uint8_t channel)
 {
 	Serial.println("I'm CONNECTED");
+}
+
+// Will be called when station is fully operational
+void gotIP(IPAddress ip, IPAddress netmask, IPAddress gateway)
+{
 	StartServers();
 }
 
@@ -280,8 +285,8 @@ void init()
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiAccessPoint.enable(false);
 
-	// Run our method when station was connected to AP
-	WifiStation.waitConnection(connectOk);
+	WifiEvents.onStationConnect(&connectOk);
+	WifiEvents.onStationGotIP(&gotIP);
 
 	// setup the ArduCAM
 	initCam();
--- a/samples/Arducam/app/application.cpp
+++ b/samples/Arducam/app/application.cpp
@@ -256,12 +256,6 @@ void StartServers()
 }
 
 
-// Will be called when WiFi station was connected to AP
-void connectOk(String ssid, uint8_t ssid_len, uint8_t bssid[6], uint8_t channel)
-{
-	Serial.println("I'm CONNECTED");
-}
-
 // Will be called when station is fully operational
 void gotIP(IPAddress ip, IPAddress netmask, IPAddress gateway)
 {
@@ -285,7 +279,6 @@ void init()
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiAccessPoint.enable(false);
 
-	WifiEvents.onStationConnect(connectOk);
 	WifiEvents.onStationGotIP(gotIP);
 
 	// setup the ArduCAM
--- a/samples/Arducam/app/application.cpp
+++ b/samples/Arducam/app/application.cpp
@@ -285,8 +285,8 @@ void init()
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiAccessPoint.enable(false);
 
-	WifiEvents.onStationConnect(&connectOk);
-	WifiEvents.onStationGotIP(&gotIP);
+	WifiEvents.onStationConnect(connectOk);
+	WifiEvents.onStationGotIP(gotIP);
 
 	// setup the ArduCAM
 	initCam();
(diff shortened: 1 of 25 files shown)


samples/Arducam/app/application.cpp
```diff
@@ -280,8 +285,8 @@ void init()
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiAccessPoint.enable(false);
 
-	// Run our method when station was connected to AP
-	WifiStation.waitConnection(connectOk);
+	WifiEvents.onStationConnect(&connectOk);
```
slaff: Is there a reason to the callback params by reference? If I am not mistaken delegates should take care to handle those cases.

zgoda: I am not sure.

slaff: > I am not sure.

Let's say that I would prefer it to be without reference. And thanks for your great help!

samples/Echo_Ssl/app/application.cpp
```diff
@@ -109,19 +109,24 @@ bool onReceive(TcpClient& tcpClient, char *data, int size) {
...
-	client->connect(IPAddress(SERVER_IP), 4444, true);
+	debugf("Connected.");
 }
 
-void connectFail()
+void connectFail(String ssid, uint8_t ssid_len, uint8_t bssid[6], uint8_t reason)
 {
 	debugf("I'm NOT CONNECTED!");
-	WifiStation.waitConnection(connectOk, 10, connectFail); // Repeat and check again
+	WifiStation.disconnect();
```
slaff: Another thing here - you should not disconnect and then connect.

avr39-ripe: +1 for not to disconnect, esp sdk reconnects itself.

zgoda: This is how it's done in `samples/Websocket_Client`, the only example that uses `WifiEvents` atm. So how this should be handled?

slaff: The websocket client should be fixed. But that is another thing. For now please take care to make those changes related to your PR.

zgoda: If SDK reconnects on connection lost then is there any need to call `WifiStation.connect()` in disconnect handler? I suppose not but please verify.

avr39-ripe: This, disconnect/connect was my fault, I'm copy/paste it from my application Class and forgot to remove it completely. IMHO we should do nothing, but some debug message in STADisconnect-equivalent event handler. In my app class I start access point (if not started already) to enable user change ssid/pwd if no connection to predefined ssid but in generic there is no means to do so.

zgoda: Okay, thank you very much.

avr39-ripe:
One more note, IMHO in generic examples there is no need to separate onStationConnect and onStationGotIP event handlers, just use onStationGotIP as "final connection event handler". In my application class I use separate handlers to handle case when after power cycle of hardware router/access point there is no dhcp server running on router/AP and esp already connects to it and can't get IP from it by dhcp server timeout of esp-sdk. So in onStationConnect I start timer that run disconnect/connect to such router/AP again after 35 seconds. Adn is onStationGotIP I stop this timer. But again, no need of separation in generic.

zgoda:
I plan to do general cleanup of examples, one by one. Replace obsolete calls with current, replace `debugf()` calls in client code with plain printing to serial, better diagnostic messages, turn off system debug output to serial, make the examples more concise and focused on topic, etc. While lack of documentation is pain, good examples may be the most valuable source of knowledge.

zgoda:
And back to `WifiEvents`, it deserves its own example.

slaff:
>  plan to do general cleanup of examples, one by one. Replace obsolete calls with current, replace debugf() calls in client code with plain printing to serial, better diagnostic messages, turn off system 

Before you fall into that trap do one thing at a time. And that one thing here is to finish that PR. 

> make the examples more concise and focused on topic, etc.

I completely agree with you. At the moment we have a lot of kitchen-sink type of samples. And the focus is lost, which is bad for the developer planning to use the code, plus it makes it difficult for us to maintain those examples during refactoring.

> And back to WifiEvents, it deserves its own example.

Please, do not introduce another sample. Use the Basic_Wifi sample.

samples/Basic_Neopixel/app/application.cpp
```diff
@@ -188,7 +190,9 @@ void init()
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiStation.enable(true);
 	WifiAccessPoint.enable(false);
-	WifiStation.waitConnection(connect_Ok, 20, connect_Fail);
+	WifiEvents.onStationConnect(connect_Ok);
```
slaff: Talking about 'focus': Please, remove the WifiEvents.onStationConnect callback and move its code to the callback for `WifiEvents.onStationGotIP(got_IP)` .  Example usage of `WifiEvents.onStationConnect` can be added to Basic_Wifi sample, which is the one dedicated to Wifi connectivity and Wifi Events handling.

zgoda:
I think basic wifi is too basic to showcase AP events.

zgoda:
Okay, I think it's better now. There is still place for improvement (eg. `onStationDisconnect` does nothing but log message to serial) but I would leave this to general samples cleanup tasks. At this stage we have all (soon to be) obsoleted `WifiStation` features removed.

slaff:
> At this stage we have all (soon to be) obsoleted WifiStation features removed.

@zgoda Thanks again for your help. I will leave the PR in review status for 2-3 days and if there are no complaints it will become part of the develop branch by the end of the week.

zgoda:
Thanks, I will squash all these commits into one to limit clutter in develop.

slaff:
>  I will squash all these commits into one to limit clutter in develop.

No need. I will squash it during merge.

PR #1073, opened 2017-04-10
Revert "Fix i2c master by properly setup SDA pin as INPUT_PULLUP"

slaff:
Reverts SmingHub/Sming#1060

In some of my tests that commit caused instability during uart reading/writing operations. For the moment a better fix will be to document that behaviour instead of always setting the INPUT_PULLUP.


avr39-ripe:
@slaff how can I reproduce your tests? Why setting sda pin to INPUT_PULLUP affect uart? Do you use default scl / sda 0/2 pins or non-default??

slaff:
> how can I reproduce your tests? 

@avr39-ripe Forgot to mention that I am also using those pins and reading from flash. What I will do is revert that change and test later this week the other PR that you have proposed. If there is not instability there then it will be merged.

Issue #1066, opened 2017-04-07
m_vsnprintf does not return size of required buffer


tius2000:
It seems to me, that _m_vsnprintf_ does not return the size of the required buffer if _maxLen_ is too small. _Print::printf(...)_ does handle this by increasing the size by one until the string fits, but this **very** inefficient.

ADiea:
it's difficult to know the exact size needed. If you run out of buffer you cannot print more. There is actually an overflow guard and if current size < maxSize-overflow_guard then maxlength is returned.
I agree that increasing by 1 byte is inefficient I would increase by 16 for ex. This way we also make fewer mallocs => better control of fragmentation

tius2000:
Hi @ADiea, I will have a look on it try to find a solution.

tius2000:
PR #1077 should fix this, perhaps you can take a look @ADiea? Unfortunately I had to change the existing logic, hope I got it right.

PR #1077, opened 2017-04-11
fix m_vsnprintf return value

tius2000:
Changed the logic of m_vsnprintf() to return the correct value if the buffer is not large enough (see #1066 ). The overflow guard is no longer required. Unfortunately this required  some changes to the existing logic, I hope I did not break anything. Here is a minimal test application:

```
/*
	test return value of snprintf
	
	expected result:
	
		>123<      5
		>1234<     6
		>12345<    7
		>123456    8
		>123456    9
		% A 777 0x033A > 42< -1 > 3.14< >  123< >123  <
*/	

#include <SmingCore.h>

static void _test(const char *s) {
	char buf[8];
	size_t r = m_snprintf( buf, sizeof(buf), ">%s<", s );
	debugf("%-10.10s %d", buf, r);
}




void init() {
	Serial.begin(74880); 
	debugf("\n\ntest m_snprintf\n");
	
	//	test return values
	_test("123");
	_test("1234");
	_test("12345");
	_test("123456");
	_test("1234567");

	//	minimal format testing
	m_printf(
		"%% %c %o 0x%04X >%3u< %d >%5.2f< >%5.5s< >%-5.5s<\n", 
		65, 0777, 0x33A, 42, -1, 3.14159, "123", "123"
	);
}
```

### DIFF
--- a/Sming/system/m_printf.cpp
+++ b/Sming/system/m_printf.cpp
@@ -116,7 +116,7 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args) {
 	};
 
 	while (*fmt) {
-		//	copy verbatim text
+		//	copy verbatim text 
 		if (*fmt != '%')  {
 			add(*fmt++);
 			continue;
@@ -124,29 +124,25 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args) {
 		fmt++;
 
 		const char*	s;							// source for string copy
-		char	tempNum[40];					// buffer for number conversion
-
+		char 	tempNum[40];					// buffer for number conversion
+		
 		//	reset attributes to defaults
 		bool	minus		= 0;
-		uint8_t	base		= 10;
-		int8_t	precision	= -1;
-		int8_t	width		= 0;
-		char	pad			= ' ';
-
-		//	process flags
-	DO_FLAGS:
-		switch (*fmt) {
-			case '-':
-				minus = 1;
-
-			case '+':
-			case ' ':
-			case '#':
-				fmt++;
-				goto DO_FLAGS;
+		uint8_t	base 		= 10;
+		int8_t	precision 	= -1;
+		int8_t	width 		= 0;
+		char	pad 		= ' ';
+		
+		while (char f = *fmt) {
+			if (f == '-')			minus = 1;
+			else if (f == '+')		;			// ignored
+			else if (f == ' ')		;			// ignored
+			else if (f == '#')		;			// ignored
+			else 					break;
+			fmt++;
 		}
 
-		//	process padding
+		//	process padding 
 		if (*fmt == '0') {
 			pad = '0';
 			fmt++;
@@ -157,21 +153,21 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args) {
 			width = skip_atoi(&fmt);
 		}
 
-		//	process precision
+		// 	process precision
 		if( *fmt == '.' ) {
 			fmt++;
 			if ( is_digit(*fmt) ) precision = skip_atoi(&fmt);
 		}
-
-		//	ignore length
+		
+		//	ignore length 
 		while (*fmt == 'l' || *fmt == 'h' || *fmt == 'L') fmt++;
-
-		//	process type
+		
+		//	process type	
 		switch (char f = *fmt++) {
 			case '%':
 				add('%');
 				continue;
-
+			
 			case 'c':
 				add( (unsigned char) va_arg(args, int) );
 				continue;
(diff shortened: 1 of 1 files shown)


tius2000:
Well, C is not my primary language, I will give it a try to reduce the number of gotos ;-)

Sming/system/m_printf.cpp
```diff
@@ -100,160 +94,134 @@ int m_vprintf ( const char * format, va_list arg )
...
+            case 'd':
+            case 'i':
+                s = ltoa_wp(va_arg(args, int), tempNum, base, width, pad);
+                break;
+
+            case 'f':
+                s = dtostrf_p(va_arg(args, double), width, precision, tempNum, pad);
+                break;
+
+            case 'o':
```
ADiea: I would rewrite as
```
case 'o':
case 'x':
case 'X':
if('o' == f)
base=8;
else
base=16;
case 'u':

```

tius2000: You are very strict regarding gotos ;-)

I changed it a bit to get rid of the fall-through, too.

ADiea:
Have no other remarks. If "unit tests" in basic_string example work ok it's fine with me. Thanks again!

Sming/system/m_printf.cpp
```diff
@@ -11,39 +11,33 @@ Descr: embedded very simple version of printf with float support
 
 #define MPRINTF_BUF_SIZE 256
 
-#define OVERFLOW_GUARD 24
-
 static void defaultPrintChar(uart_t *uart, char c) {
-	return uart_tx_one_char(c);
+    return uart_tx_one_char(c);
```
slaff: The are multiple white-space changes that are not related to this PR. Please, revert them.

tius2000: Sorry, if this is a FAQ - is there a guide regarding source code formatting for Sming (or even better, a command line tool that working with windows to do the correct formatting)? 

I tried to follow the rule 38 [here](http://geosoft.no/development/cppstyle.html#White%20Space): _Special characters like TAB and page break must be avoided.
These characters are bound to cause problem for editors, printers, terminal emulators or debuggers when used in a multi-programmer, multi-platform environment._ 

That's why I converted all tabs to space before committing.

slaff: > That's why I converted all tabs to space before committing.

Changing of formatting and coding style can happen in another PR. This PR is only related to `m_vsnprintf` changes.

tius2000:
* support for format and data strings in PROGMEM added
* format specifier '%S' added for compatibility (behaves like '%s)
* _m_vprintf_ now allocates less memory on the stack unless it is actually required

There is no longer a need for separate *_P methods, this should make it easier to solve the conflicts created by the macro _printf_P_ in FakePgmSpace.h.

Sming/system/m_printf.cpp
```diff
@@ -68,27 +65,23 @@ int m_snprintf(char* buf, int length, const char *fmt, ...)
...
-		return 0;
-	}
-
-	char buf[MPRINTF_BUF_SIZE], *p;
+	size_t size = INITIAL_BUFFSIZE - 1;
+
+	//	need to retry if size is not big enough
+	while (1) {
+		char buffer[size + 1];
+		auto sz = m_vsnprintf(buffer, sizeof(buffer), fmt, va);
```
slaff: Please, change `auto` to `size_t`.

Sming/system/m_printf.cpp
```diff
@@ -68,27 +65,23 @@ int m_snprintf(char* buf, int length, const char *fmt, ...)
...
-	m_vsnprintf(buf, sizeof(buf), format, arg);
-
-	p = buf;
-	while (p && n < sizeof(buf) && *p)
-	{
-		cbc_printchar(cbc_printchar_uart, *p);
-		n++;
-		p++;
+		const char *p = buffer;
+		while (char c = *p++) cbc_printchar(cbc_printchar_uart, c);
```
slaff: Format the while with opening and closing curly braces. 

```
while (char c = *p++)  {
     cbc_printchar(cbc_printchar_uart, c);
}
```

Sming/system/m_printf.cpp
```diff
@@ -117,143 +110,122 @@ int m_printf(const char* fmt, ...)
...
-	}
-
-	*str = '\0';
-	return str - buf;
+    size_t size = 0;
+    auto add = [&](char c) {
+        if (++size < maxLen) *buf++ = c;
+    };
+
+    while ( char f = pgm_read_byte(fmt) ) {
```
slaff: If you read every byte using `pgm_read_byte` that will cause overhead. Is that really needed in a very core component? What are the advantages?

tius2000: You a right, this will make every call to *printf a tiny bit slower. The code size is almost identical. I see the following advantages:

- it will reduce total flash size and complexity as a many additional methods are no longer needed
- all classes and methods using _m_printf_ will benefit immediately
- life will be much easier for users as they do not need to care which method to call and which format specifier to use
- format strings should go to flashmem anyway to reduce ram size, it is faster than the existing implementation for format strings in flashmem
- printf is mostly used for serial output where execution speed does not matter that much
- for time critical operations direct c string manipulation should be used instead of the printf functions

IMHO the main advantage is that it eliminated a major pitfall for users working with progmem.

tius2000:
I'm quite new to GitHub - how can I create a PR that depends on another PR?

slaff:
> I'm quite new to GitHub - how can I create a PR that depends on another PR?

Create a branch for PR 1. Work on it until you are ready. Then create another branch for PR1 that is based on PR1.
```
git check PR1
# ... work...
# .... ready
git branch PR2
git checkout PR2
# additional work 
```

If there are changes in PR1 you can keep PR2 up to date using the following commands:

```
git checkout PR1 
# .. make sure that you have the latest code for PR1
git checkout PR2
git merge PR1
# git mergetool # if needed
git rebase PR1
# git mergetool # if needed
git push -f 
``` 

That should be it.

tius2000:
I created a new PR #1096 that way, hope I got it right!

PR #1078, opened 2017-04-12
POSIX shell, seriously?

l29ah:


tools/decode-stacktrace.py
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/python2
```
slaff: @l29ah Please, replace /usr/bin/python2 with /usr/bin/python.

avr39-ripe: even more proper way is to use 
#!/usr/bin/env python
invocation of python. It is more platform independent then hardcoded path

l29ah:
On Tue, Apr 11, 2017 at 10:50:15PM -0700, slaff wrote:
> @l29ah Please, replace /usr/bin/python2 with /usr/bin/python.

/usr/bin/python is python3 on some of the modern systems, and the script is incompatible with it.

-- 
()  ascii ribbon campaign - against html mail
/\  http://arc.pasp.de/   - against proprietary attachments

slaff:
>  and the script is incompatible with it.

I am not sure about that. Can you tell me with which python3 version did you try and it did not work?

tools/decode-stacktrace.py
slaff: Yes, please go for "#!/usr/bin/env python".

l29ah:
On Wed, Apr 12, 2017 at 05:36:39AM -0700, slaff wrote:
> I am not sure about that. Can you tell me with which python3 version did you try and it did not work?

3.4.6.

-- 
()  ascii ribbon campaign - against html mail
/\  http://arc.pasp.de/   - against proprietary attachments

slaff:
> 3.4.6.

Can paste the error output here?

slaff:
@l29ah Please, change the name of the PR to some more descriptive as "Changed Python Interpreter". Then change the first line to `#!/usr/bin/env python` and finally paste the errors that you are experiencing with Python 3.4.6 plus information about the OS and python interpreter itself.

slaff:
This change will be merged with PR #1088.

Issue #1076, opened 2017-04-11
Liberal use of String class


zgoda:
I'm just a noob when it comes to embedded programming so bear with me if I am asking funny questions.

I've been told many times that dynamic memory allocation in embedded programming is root of all evil and everytime you use `String` instance a sweet little chimpanzee is killed. Now I am looking at Sming code and I see `String` everywhere, passed by value to functions, concatenated many times, some modules are just interfaces to SDK functions only to return `String` like in javascript.

What's the key here? Does Sming have that magic memory allocator that squashes heap every so often? Or what?

ADiea:
You are right... if you have time please find as many places where String is passed as value and could be passed as ref/ptr and make a PR it will improve fragmentation.
Also all core functions in sming should NOT use String at all on the long term.

When you create a string object internally malloc is used to reserve string_length bytes of RAM on heap. When a string is destructed free is used to unreserve the RAM, but strings are tipically small and create holes in RAM known as fragmentation. 
For instance you start with 32 bytes of free heap. If you need 16 bytes it will give you first 16 bytes and mark next 16 bytes as free. then a string comes in of 4 bytes and next 4 bytes are given to the string remaining 10 bytes free. then a 10Byte string comes in place and allocates the last 10 bytes. Then the first 16 bytes and the last 10 bytes are freed as not needed but the 4 byte string is there ight in the middle
<16B free> <4B string> <10Bytes free>
Now if you need 20 bytes even if you have 16+10=26 free bytes reported by system_free_heap() you will have nowhere 20 continuous bytes because the 4 bytes are right in the middle.

Now, if you were to use stack memory to work with strings like this
```
void fct(char* buffer)
{
///
int len = strlen(buffer);
if(len < 128)
{
 char buf[len]; //stack ponter is increased by len
//work on buf
}
else
{
debugf("string too long, could cause stack overflow");
}
} //stack pointer is decreased by len
```
Now the memory is used by function and on fct return all memory is reclaimed and can be used by caller function etc
With stack you have to be carefull not to allocate too much as this coupled with too deep funciton call tree can cause stack overflow (no more memory available) or in our system it can cause to write over the heap space, because we increase heap from bottom to top of free ram and the stack from top to bottom so there is posibillity of meetoing in the center and we currently have no protection for that.

I hope I have provided enough details if not let me know and I can explain more. 
Bottom line is we should use malloc as little as possible and if you already constructed a string try to pass it by refference always when possible.

[...]

PR #1080, opened 2017-04-13
[Rel #1076] use C strings where possible in StationClass

zgoda:
* const C strings as args for `config()` and `setHostname()`
* Overloaded methods returning C strings
* Overloaded method returning byte array for MAC

Code redundancy will be removed in next stage, I don't want to introduce too many changes to class implementation at once. Suggestions for improvement are very welcome.

### DIFF
--- a/Sming/SmingCore/Platform/Station.cpp
+++ b/Sming/SmingCore/Platform/Station.cpp
@@ -290,7 +290,7 @@ uint8 StationClass::getChannel()
  
 String StationClass::getPassword()
 {
-	char buf[63];  // passphrase is 8-63 ASCII chars (to make 256 bits PSK)
+	char buf[64];  // passphrase is 8-63 ASCII chars (to make 256 bits PSK)
 	return String(getPassword(buf, sizeof(buf))); 
 }
 
--- a/Sming/SmingCore/Platform/Station.cpp
+++ b/Sming/SmingCore/Platform/Station.cpp
@@ -49,12 +49,12 @@ bool StationClass::isEnabled()
 	return wifi_get_opmode() & STATION_MODE;
 }
 
-bool StationClass::config(String ssid, String password, bool autoConnectOnStartup /* = true*/, bool save /* = true */)
+bool StationClass::config(const char * ssid, const char * password, bool autoConnectOnStartup /* = true*/, bool save /* = true */)
 {
 	station_config config = {0};
 
-	if (ssid.length() >= sizeof(config.ssid)) return false;
-	if (password.length() >= sizeof(config.password)) return false;
+	if (sizeof(ssid) >= sizeof(config.ssid)) return false;
+	if (sizeof(password) >= sizeof(config.password)) return false;
 
 	bool enabled = isEnabled();
 	bool dhcp = isEnabledDHCP();
@@ -63,14 +63,14 @@ bool StationClass::config(String ssid, String password, bool autoConnectOnStartu
 	bool cfgreaded = wifi_station_get_config(&config);
 	if (!cfgreaded) debugf("Can't read station configuration!");
 
-	if (strncmp(ssid.c_str(), (char*)config.ssid, sizeof(config.ssid))!=0
-		|| strncmp(password.c_str(), (char*)config.password, sizeof(config.password))!=0 || config.bssid_set)
+	if (strncmp(ssid, (char*)config.ssid, sizeof(config.ssid))!=0
+		|| strncmp(password, (char*)config.password, sizeof(config.password))!=0 || config.bssid_set)
 	{
 		memset(config.ssid, 0, sizeof(config.ssid));
 		memset(config.password, 0, sizeof(config.password));
 		config.bssid_set = false;
-		strcpy((char*)config.ssid, ssid.c_str());
-		strcpy((char*)config.password, password.c_str());
+		strcpy((char*)config.ssid, ssid);
+		strcpy((char*)config.password, password);
 
 		noInterrupts();
 
@@ -90,12 +90,12 @@ bool StationClass::config(String ssid, String password, bool autoConnectOnStartu
 			if (!enabled) enable(enabled);
 			return false;
 		}
-		debugf("Station configuration was updated to: %s", ssid.c_str());
+		debugf("Station configuration was updated to: %s", ssid);
 
 		interrupts();
 	}
 	else
-		debugf("Station configuration is: %s", ssid.c_str());
+		debugf("Station configuration is: %s", ssid);
 	if (!dhcp) enableDHCP(dhcp);
 	if (!enabled) enable(enabled);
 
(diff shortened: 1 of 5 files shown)


slaff:
@zgoda  please, make sure that the CI tests are passing.

zgoda:
Sorry, forgot to check if all examples are OK.

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -49,12 +49,12 @@ bool StationClass::isEnabled()
...
 }
 
-bool StationClass::config(String ssid, String password, bool autoConnectOnStartup /* = true*/, bool save /* = true */)
+bool StationClass::config(const char * ssid, const char * password, bool autoConnectOnStartup /* = true*/, bool save /* = true */)
 {
 	station_config config = {0};
 
-	if (ssid.length() >= sizeof(config.ssid)) return false;
-	if (password.length() >= sizeof(config.password)) return false;
+	if (sizeof(ssid) >= sizeof(config.ssid)) return false;
```
avr39-ripe: sizeof(ssid) is used incorrectly here. sizeof(char ) = 4 on esp8266 always. Strings cant be easily converted to char..:(

ADiea: should be strlen(ssid) instead of sizeof here.

samples/MeteoControl/app/application.cpp
```diff
@@ -54,7 +54,7 @@ void init()
 	lcd.createChar(6, icon_cross);
 	lcd.createChar(7, icon_check);
 
-	WifiStation.config(ActiveConfig.NetworkSSID, ActiveConfig.NetworkPassword);
+	WifiStation.config(ActiveConfig.NetworkSSID.c_str(), ActiveConfig.NetworkPassword.c_str());
```
slaff: That change breaks the compatibility with already existing code. Also there should be quantitative measurement that proves that this effort is worth it.

Isn't it possible to have similar improvements by changing ?

```
StationClass::config(String ssid, String password, bool autoConnectOnStartup /* = true*/, bool save /* = true */)
```

to 

```
`StationClass::config(const String& ssid, const String& password, bool autoConnectOnStartup /* = true*/, bool save /* = true */) 
```

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -49,12 +49,12 @@ bool StationClass::isEnabled()
...
+	if (sizeof(password) >= sizeof(config.password)) return false;
```
ADiea: should be strlen(password). sizeof (char*) is 4 bytes it does not reflect the length of the string

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -142,16 +142,26 @@ void StationClass::enableDHCP(bool enable)
...
 	return (String) wifi_station_get_hostname();
 }
 
+char * StationClass::getHostname(char * s, size_t bufSize)
+{
+	char * buf;
+	buf = wifi_station_get_hostname();
+	size_t bufLen = strlen(buf);
+	strncpy(s, buf, bufSize - 1);
+	s[bufLen] = '\0';
```
ADiea: should be 
```
size_t copiedLen = strncpy(s,buf, bufSize-1);
s[copiedLen] = 0;
```
This way you make sure that you don't write out of s[bufSize]

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -241,6 +265,21 @@ String StationClass::getSSID()
...
+{
+	station_config config = {0};
+	if (!wifi_station_get_config(&config))
+	{
+		debugf("Can't read station configuration!");
+	}
+	else
+	{
+		debugf("SSID: %s", (char *)config.ssid);
+		strncpy(s, (char *)config.ssid, bufSize - 1);
```
ADiea:
```
size_t lenCopied = strncpy(...)
s[lenCopied] = 0;
```

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -268,6 +305,22 @@ String StationClass::getPassword()
...
+	station_config config = {0};
+	if (!wifi_station_get_config(&config))
+	{
+		debugf("Can't read station configuration!");
+	}
+	else
+	{
+		debugf("Pass: %s", (char *)config.password);
+		strncpy(s, (char *)config.password, bufSize - 1);
+		s[strlen((char *)config.password)] = '\0';
```
ADiea:
```
size_t lenCopied = strncpy(...)
s[lenCopied] = 0;
```

samples/MeteoControl/app/application.cpp
zgoda: Okay, I will leave these intact and provide overloaded versions that take C strings as arguments.

Sming/SmingCore/Platform/Station.h
```diff
@@ -195,10 +215,25 @@ class StationClass : protected ISystemReadyHandler
 	 */
 	String getSSID();
```
ADiea: is this still implemented?

Sming/SmingCore/Platform/Station.h
```diff
@@ -195,10 +215,25 @@ class StationClass : protected ISystemReadyHandler
...
+	 *  @param s buffer to fill with SSID
+	 *  @param bufSize buffer size
+	 *	@retval	char* pointer to buffer holding SSID
+	 */
+	char * getSSID(char * s, size_t bufSize);
+
 	/**	@brief	Get WiFi station password
 	 *	@retval	String WiFi station password
 	 */
 	String getPassword();
```
ADiea: is this still implemented?

Sming/SmingCore/Platform/Station.cpp
zgoda: `config.ssid` is array of `uint8_t` so `sizeof(config.ssid)` is OK here.

Sming/SmingCore/Platform/Station.cpp
zgoda: `strncpy()` returns pointer to char buffer, not the number of copied characters

Sming/SmingCore/Platform/Station.h
zgoda: Yes. I would not bother if was not. ;)

samples/MeteoControl/app/application.cpp
zgoda: > Also there should be quantitative measurement that proves that this effort is worth it.

As you probably know, there's no way to measure heap fragmentation on ESP. At least not yet.

slaff: > As you probably know, there's no way to measure heap fragmentation on ESP. At least not yet.

Not really. Umm-Malloc can give you a nice table with allocated and free memory.  It is possible to disable umm-malloc's allocation functions and just use the stats. It requires work, but it is not impossible.

zgoda: This https://github.com/rhempel/umm_malloc/blob/master/src/umm_info.c#L20 ?

Sming/SmingCore/Platform/Station.cpp
zgoda: Because:

> If the end of the source C string (which is signaled by a null-character) is found before num characters have been copied, destination is padded with zeros until a total of num characters have been written to it.

and:

> No null-character is implicitly appended at the end of destination if source is longer than num.

I'm writing '\0' at the end of provided buffer in the case when `bufSize' is smaller than `sizeof(config.ssid)` and `config.ssid` would not fit entirely. Otherwise buffer will be padded with zeros anyway.

ADiea: Looks good now thx

zgoda:
This PR and next in the series will make #1074 obsolete.

Sming/SmingCore/Platform/Station.cpp
```diff
@@ -254,18 +288,26 @@ uint8 StationClass::getChannel()
 	return wifi_get_channel();
 }
  
-
-
 String StationClass::getPassword()
 {
+	char buf[63];  // passphrase is 8-63 ASCII chars (to make 256 bits PSK)
```
zgoda: Should be 64

PR #1081, opened 2017-04-14
Extend POST parameter handling to PUT requests as well

pief:
For some reason, we didn't handle parameters supplied with PUT requests as we did with POST requests. This fixes that.


slaff:
Post parameters for non GET or HEAD requests should be implemented with the upcoming Http refactoring (#1112)

PR #1086, opened 2017-04-18
[Rel #1076] C string version of `connect()` method

zgoda:
I did not touch existing `connect()` that takes `String` but it calls `TcpConnection::connect()` with C string as argument already.

### DIFF
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -53,6 +53,14 @@ TcpClient::~TcpClient()
 	}
 }
 
+bool TcpClient::connect(const char * server, int port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
+{
+	if (isProcessing()) return false;
+
+	state = eTCS_Connecting;
+	return TcpConnection::connect(server, port, useSsl, sslOptions);
+}
+
 bool TcpClient::connect(String server, int port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
 {
 	if (isProcessing()) return false;
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -63,10 +63,7 @@ bool TcpClient::connect(const char * server, int port, boolean useSsl /* = false
 
 bool TcpClient::connect(String server, int port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
 {
-	if (isProcessing()) return false;
-
-	state = eTCS_Connecting;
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Network/TcpClient.cpp
```diff
@@ -53,6 +53,14 @@ TcpClient::~TcpClient()
 	}
 }
 
+bool TcpClient::connect(const char * server, int port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
+{
+	if (isProcessing()) return false;
+
+	state = eTCS_Connecting;
+	return TcpConnection::connect(server, port, useSsl, sslOptions);
```
ADiea: This might actually call the present function again making stack ovf by recursion. Compiler will probably choose const char* version and not string version. Also even if you make it call the string version you are doing the isprocessing checks again... i just see no added benefit here maybe if it was the other way aroung e.g. the string fct would call the const char* fct. Do you agree? (Sorry for typing badly i'm on mobile phone)

zgoda: I don't know the reason existing code passes cstr. It does not make sense to me but again, I'm not sure. I think it should be internally only 2 methods, one that gets hostname and another that takes IP address, and again, internally the one that takes hostname should do DNS lookup and call the IP one. To keep existing class interface there should be one more method that takes String object and calls the cstr version. I've ran couple of examples that use client class and did not observe any instabilities but this was limited test.

Anyway, this may be worth cleaning up to have concise and consistent class interface.

zgoda: I changed TcpClient code to call cstr version from String version of connect() and left firmware running for couple of hours calling simple tcp echo server every 10 seconds, no instabilities observed:

```
diff --git a/Sming/SmingCore/Network/TcpClient.cpp b/Sming/SmingCore/Network/TcpClient.cpp
index 638373f..1d7a2eb 100644
--- a/Sming/SmingCore/Network/TcpClient.cpp
+++ b/Sming/SmingCore/Network/TcpClient.cpp
@@ -63,10 +63,7 @@ bool TcpClient::connect(const char * server, int port, boolean useSsl /* = false
 
 bool TcpClient::connect(String server, int port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
 {
-       if (isProcessing()) return false;
-
-       state = eTCS_Connecting;
-       return TcpConnection::connect(server.c_str(), port, useSsl, sslOptions);
+       return TcpClient::connect(server.c_str(), port, useSsl, sslOptions);
 }
 
 bool TcpClient::connect(IPAddress addr, uint16_t port, boolean useSsl /* = false */, uint32_t sslOptions /* = 0 */)
```

ADiea:
Looks ok

PR #1089, opened 2017-04-18
MeteoControl_mqtt sample cleanup

zgoda:
* added config defines guards
* serial printout cleanups
* translated serial printout to ASCII/English
* removed unused config values
* removed unneccessary WifiEvents handlers, use `System.onReady`
* removed empty CPP file

### DIFF
--- a/samples/MeteoControl_mqtt/app/application.cpp
+++ b/samples/MeteoControl_mqtt/app/application.cpp
@@ -24,9 +24,9 @@ void init()
 
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiStation.enable(true);
+	WifiEvents.onStationGotIP(gotIP);
 	WifiAccessPoint.enable(false);
 	WDT.enable(false);	//disable watchdog
-	System.onReady(systemReady);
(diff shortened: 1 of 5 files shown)


samples/MeteoControl_mqtt/app/application.cpp
```diff
@@ -26,11 +24,9 @@ void init()
 
 	WifiStation.config(WIFI_SSID, WIFI_PWD);
 	WifiStation.enable(true);
-	//WifiStation.setIP(ESP_IP);
 	WifiAccessPoint.enable(false);
 	WDT.enable(false);	//disable watchdog
-	WifiEvents.onStationDisconnect(connectFail);
-	WifiEvents.onStationGotIP(gotIP);
+	System.onReady(systemReady);
```
slaff: @zgoda Please, change System.onReady back to WifiEvents.onStationGotIP. In this example we only have Station connecting to AP.

slaff: @zgoda You forgot to change also that line in you last commit.

PR #1091, opened 2017-04-18
Cleaned up Nokia 5110 LCD sample code

zgoda:
* do not use `delay()`, `Timer` is the way to go
* use default SPI pins on NodeMCU and WeMos D1 mini

### DIFF
--- a/samples/ScreenLCD_5110/app/application.cpp
+++ b/samples/ScreenLCD_5110/app/application.cpp
@@ -2,43 +2,38 @@
 #include <SmingCore/SmingCore.h>
 #include <Libraries/Adafruit_PCD8544/Adafruit_PCD8544.h>
 
-// Software SPI (slower updates, more flexible pin options):
-	// pin 7 - Serial clock out (SCLK)
-	// pin 6 - Serial data out (DIN)
-	// pin 5 - Data/Command select (D/C)
-	// pin 4 - LCD chip select (CS)
-	// pin 3 - LCD reset (RST)
-Adafruit_PCD8544 display = Adafruit_PCD8544(14, 13, 12, 5, 4);
+// GPIO13/D7 - Serial clock out (SCLK)
+// GPIO12/D6 - Serial data out (DIN)
+// GPIO14/D5 - Data/Command select (D/C)
+// GPIO5/D1 - LCD chip select (CS)
+// GPIO4/D2 - LCD reset (RST)
+Adafruit_PCD8544 display = Adafruit_PCD8544(13, 12, 14, 5, 4);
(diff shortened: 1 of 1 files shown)


samples/ScreenLCD_5110/app/application.cpp
```diff
@@ -2,43 +2,38 @@
...
-	    display.println("Sming");
-	    display.setTextSize(2);
-	    display.println("Example!");
-	    display.display();
+void displayTest()
+{
+	display.begin();
+	display.setContrast(10);
+	display.display(); // show splashscreen
+	displayTimer.initializeMs(5000, displayTest2).start();
```
slaff: You've changed the delay from 200 to 5000 ms. Is there a reason for it?

zgoda: No specific reason, just personal preference.

samples/ScreenLCD_5110/app/application.cpp
```diff
@@ -2,43 +2,38 @@
...
-	    display.println("Sming");
-	    display.setTextSize(2);
-	    display.println("Example!");
-	    display.display();
+void displayTest()
+{
+	display.begin();
+	display.setContrast(10);
+	display.display(); // show splashscreen
+	displayTimer.initializeMs(2000, displayTest2).start();
```
slaff: The original delay is 200 ms. 200 ms != 2000 ms. Also for delays below 900ms I would prefer to leave the delay() function and do not introduce timers in that sample.

zgoda: 200 ms is 0.2 seconds, this will be hardly visible at all, as flashing at best.

slaff: @zgoda Please, change the code to use `delay(2000)` and remove the use of timers. Since this #1154 PR is merged delays bigger than 1000ms should not cause reboots due to WDT.

samples/ScreenLCD_5110/app/application.cpp
```diff
@@ -2,43 +2,37 @@
...
-	    display.println("Sming");
-	    display.setTextSize(2);
-	    display.println("Example!");
-	    display.display();
+void displayTest()
+{
+	display.begin();
+	display.setContrast(10);
+	display.display(); // show splashscreen
+	delay(2000);
```
slaff: @zgoda And don't forget to combine the displayTest  and displayTest2 in one function :)

zgoda: Geez... :D

PR #1092, opened 2017-04-18
Cleaned up DS RTC example

zgoda:
* cleaned up comments
* removed unnecessary code
* cleaned up serial printout

### DIFF
--- a/samples/DS3232RTC_NTP_Setter/app/application.cpp
+++ b/samples/DS3232RTC_NTP_Setter/app/application.cpp
@@ -47,11 +47,10 @@ void init()
 	// Station - WiFi client
 	WifiStation.config(WIFI_SSID, WIFI_PWD); // Put you SSID and Password here
 	WifiStation.enable(true);
+	WifiEvents.onStationGotIP(gotIP);
 
 	// set timezone hourly difference to UTC
 	SystemClock.setTimeZone(2); // GMT+2
-
-	WifiEvents.onStationGotIP(gotIP);
 
 	printTimer.initializeMs(2000, onPrintSystemTime).start();
 }
--- a/samples/DS3232RTC_NTP_Setter/app/application.cpp
+++ b/samples/DS3232RTC_NTP_Setter/app/application.cpp
@@ -5,9 +5,6 @@
 #include <Libraries/DS3232RTC/DS3232RTC.h>
 
 
-// Connect DS3232 / DS3231 to GND, VCC, GPIO0 - SCL, GPIO2 - SDA or any other free GPIO and change below Wire.pins(0,2)
-// to you SCL and SDA pins
-
 // If you want, you can define WiFi settings globally in Eclipse Environment Variables
 #ifndef WIFI_SSID
 	#define WIFI_SSID "ENTER_YOUR_SSID" // Put you SSID and Password here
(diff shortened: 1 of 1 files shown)


samples/DS3232RTC_NTP_Setter/app/application.cpp
```diff
@@ -33,34 +32,26 @@ void onNtpReceive(NtpClient& client, time_t timestamp) {
...
 	// Station - WiFi client
 	WifiStation.config(WIFI_SSID, WIFI_PWD); // Put you SSID and Password here
 	WifiStation.enable(true);
-	WifiEvents.onStationGotIP(gotIP);
-	WifiEvents.onStationDisconnect(connectFail);
 
 	// set timezone hourly difference to UTC
 	SystemClock.setTimeZone(2); // GMT+2
 
+	System.onReady(systemReady);
```
slaff: `WifiEvents.onStationGotIP(gotIP);` cannot be replaced with `System.onReady(systemReady);`. System.onReady sets callback to be triggered when the ESP system is ready. And `ready` != `STA connection got IP`

zgoda: Many examples use onReady to initiate networked operations. Are these examples all wrong?

slaff: > Many examples use onReady to initiate networked operations. 

It depends on your definition of "initiate networked operations".  System.onReady callbacks are  triggered after the TCP stack is ready. So far that should be included in "initiate networked operations". But that is different than connecting to an AP and getting IP address from the DHCP.   

> Are these examples all wrong?

Without knowing which examples you mean I cannot say for sure. Please, post here one or two names of samples that you've been referring to.

samples/DS3232RTC_NTP_Setter/app/application.cpp
```diff
@@ -33,34 +32,26 @@ void onNtpReceive(NtpClient& client, time_t timestamp) {
 	Serial.printf("Time synchronized: %s\n", SystemClock.getSystemTimeString().c_str());
 }
 
-void gotIP(IPAddress ip, IPAddress netmask, IPAddress gateway)
+void gotIP(IPAddress ip, IPAddress netmask, IPAddress gw)
```
slaff: Please, don't shorten the name of the variable `gateway` to `gw`.

samples/DS3232RTC_NTP_Setter/app/application.cpp
```diff
@@ -38,29 +37,21 @@ void gotIP(IPAddress ip, IPAddress netmask, IPAddress gateway)
...
-	debugf("I'm NOT CONNECTED!");
-}
-
 void init()
 {
 	Serial.begin(SERIAL_BAUD_RATE);
-	Serial.systemDebugOutput(true); // Allow debug print to serial
 	Serial.println("Sming DSRTC_NTP_SETTER started!");
-//Uncomment proper line, or set right pins by hand
 	Wire.pins(0, 2); //Change to your SCL - 0,SDA - 2 GPIO pin number
```
avr39-ripe: I think we shoudn't use Wire.pins(0, 2) by default, it is already default value, so better to comment out Wire.pins occurrences with comments how to set non-default scl/sda pins.

samples/DS3232RTC_NTP_Setter/app/application.cpp
```diff
@@ -38,29 +37,21 @@ void gotIP(IPAddress ip, IPAddress netmask, IPAddress gateway)
...
 	// Station - WiFi client
 	WifiStation.config(WIFI_SSID, WIFI_PWD); // Put you SSID and Password here
 	WifiStation.enable(true);
-	WifiEvents.onStationGotIP(gotIP);
-	WifiEvents.onStationDisconnect(connectFail);
 
 	// set timezone hourly difference to UTC
 	SystemClock.setTimeZone(2); // GMT+2
 
+	WifiEvents.onStationGotIP(gotIP);
```
slaff: Please,  move `WifiEvents.onStationGotIP ...` to be directly after `WifiStation.enable(tr...`.

Issue #1067, opened 2017-04-07
possible extension: PROGMEM methods for Print.cpp


tius2000:
I recently added some missing methods to Print.cpp:

```
//	allow the user to implement methods with variable number of arguments
size_t	vprintf(const char *, va_list va);

//	allow the user to save some ram by moving strings to PROGMEM
size_t	write_P(const char* data, size_t size);
size_t	write_P(const char* data);
size_t	printf_P(const char* data, ...);
size_t	vprintf_P(const char *, va_list va);
```

I can contribute the implementation if there is any interest. However, there is a conflict with the macro _printf_P_ in FakePgmSpace.h :-(

I do not fully understand the implications, but IMHO it would be better to use inline functions instead of macros in FakePgmSpace.h - what do you think?

ADiea:
Probably your implementation of printf_P is the same so should be no issue replacing printf_P macro with inline function. For debug strings printf_P_stack is used which does no malloc and is safer.

tius2000:
I created a PR (#1094), perhaps you can have a look @ADiea? Stack space is allocated on demand, so malloc is not needed. However, I do not know how to fix the conflict with the macro _printf_P_ in FakePgmSpace.h :-(

ADiea:
Reviewed the PR. I think only way to avoid name collision is to make printf_P a function in FakePgmSpace...  
Or to rename the printf_P function in Print class. Or to rename printf_ in Fakepgmspace.h to something like PRINTF_P (macro case) but please make sure it is not used in other parts of the code. Also this last option will break user code if someone is relying on printf_P as a global symbol.

slaff:
That should be solved in the latest `develop` code.

PR #1094, opened 2017-04-18
PROGMEM methods for Print.cpp

tius2000:
* missing PROGMEM methods for class Print added
* dynamic stack allocation for _printf_ moved to new method _vprintf_ 

However, I have no idea how to fix the conflict with the new macro _printf_P_ in FakePgmSpace.h :-(

(see #1067)

### DIFF
--- a/Sming/Wiring/Print.cpp
+++ b/Sming/Wiring/Print.cpp
@@ -18,6 +18,7 @@
 
 #include "Print.h"
 #include "WiringFrameworkIncludes.h"
+
 /*
 || @description
 || | Virtual method - may be redefined in derived class (polymorphic)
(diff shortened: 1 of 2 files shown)


Sming/Wiring/Print.cpp
```diff
@@ -197,31 +198,63 @@ size_t Print::println(const Printable &p)
...
-			{
-				write(tempBuff,sz);
-			}
+	size_t size = INITIAL_PRINTF_BUFFSIZE - 1;
+	
+	//	need to retry until size is big enough as m_vsnprintf is currently broken
+	//	(returns maxLen instead of required size)
+	while (1) {
+		char buffer[size + 1];
+		auto sz = m_vsnprintf(buffer, sizeof(buffer), fmt, va);
```
ADiea: > I guess this PR should wait for #1077. Then you shuld have directly the correct size needed +1 ending char. So I would not use while(1) and wait for the PR

Sming/Wiring/Print.cpp
```diff
@@ -197,31 +198,63 @@ size_t Print::println(const Printable &p)
...
-	} while (retry);
+		size = sz;	
+	}
+}
+
+size_t Print::printf(const char *fmt, ...)
+{
+	va_list va;
+	va_start(va, fmt);
+	auto cnt = vprintf(fmt, va);
```
ADiea: Please only use auto when hiding an ugly/templetized type specifier. In this case it is size_t. IMHO it's faster to read and debug if you know the types.

Sming/Wiring/Print.cpp
tius2000: You are right @ADiea. I just added automatic PROGMEM support to #1077, so this PR might even no longer be needed.

Sming/Wiring/Print.cpp
tius2000: ok

Sming/Wiring/Print.h
```diff
@@ -26,12 +26,14 @@
 
 #define INITIAL_PRINTF_BUFFSIZE 128
 
+#undef printf_P
```
ADiea: You should remove this here. If you do not want name collision you have 2 options: namespaces or explicit calling e.g. this.printf_P

tius2000: You are right, this is ugly and will cause problems. Therefore I would suggest the following procedure:

* wait for PR #1077 
* optional: implement _printf_P()_ etc. as functions instead a macros in FakePgmSpace (can be changed to an inline wrapper for _printf()_ if #1077 is accepted)
* remove _printf_P()_ from this PR as it is no longer needed
* add printf_P() and println_P() instead

ADiea: Yes, I guess if printf / m_printf will be progmem safe, then no need for _P variant. For printf I do not care about speed so much, but sprintf I would keep it fast

PR #1096, opened 2017-04-19
m_printf: PROGMEM support added

tius2000:
(moved from #1077 to a new PR)

* support for format and data strings in PROGMEM added
* format specifier '%S' added for compatibility (behaves like '%s)
* _m_vprintf_ now allocates less memory on the stack unless it is actually required

There is no longer a need for separate *_P methods, this should make it easier to solve the conflicts created by the macro _printf_P_ in FakePgmSpace.h.

_Q: If you read every byte using pgm_read_byte that will cause overhead. Is that really needed in a very core component? What are the advantages?_

You a right, this will make every call to *printf a tiny bit slower. The code size is almost identical. I see the following advantages:

- it will reduce total flash size and complexity as a many additional methods are no longer needed
- all classes and methods using _m_printf_ will benefit immediately
- life will be much easier for users as they do not need to care which method to call and which format specifier to use
- format strings should go to flashmem anyway to reduce ram size 
- it is faster and uses less memory than the existing implementation for format strings in flashmem
- printf is mostly used for serial output where execution speed does not matter that much
- for time critical operations direct c string manipulation should be used instead of the printf functions

IMHO the main advantage is that it eliminated a major pitfall for users working with progmem.

### DIFF
--- a/Sming/system/m_printf.cpp
+++ b/Sming/system/m_printf.cpp
@@ -116,7 +116,7 @@ int m_vsnprintf(char *buf, size_t maxLen, const char *fmt, va_list args) {
 	};
 
 	while (*fmt) {
-		//	copy verbatim text
+		//	copy verbatim text 
 		if (*fmt != '%')  {
 			add(*fmt++);
 			continue;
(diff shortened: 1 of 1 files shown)


ADiea:
The issue is that also sprintf is affected and sprintf should be fast... We could make m_printf progmem safe by memcopy-ing the format into a stack buffer and then passing it to m_vsnprintf maybe? Like this? Would it be ok for you?

(not tested)
```
int m_printf(const char* fmt, ...)
{
	int n=0;

	if(!fmt)
		return 0;

+ size_t sizeFmt = strlen_P(fmt) + 1;
+ char safeFmt[sizeFmt];
+ memcpy_P(safeFmt, fmt, sizeFmt);

	va_list args;
	va_start(args, fmt);

	n = m_vprintf(fmt, args);

	va_end(args);

	return n;
}
```

ADiea:
So, if m_printf will be progmeme safe it will be possible to say
```
m_printf("RAM string");
m_printf((PSTR) "flash string");
```

Also debug_e implementation will be simplified to m_printf((PSTR) format, __VA_ARGS__); or something similar. This is to avoid double usage of pgm_read_byte

tius2000:
Well, I think would be nice to have all *printf functions (also sprintf) progmem safe and avoid replicated code. However, if you think that speed might be an issue for sprintf we should benchmark it and possible provide both variants. Do you have an application or function in mind that depends on the speed of sprintf?

PR #1098, opened 2017-04-19
Mount spiffs or file operations on index.html fail miserably

zgoda:
The code later tries to access file `index.html` and fails with `open errno -10024` (fileExists) and `write errno -10024` (fileSetContent) if spiffs is not mounted.


slaff:
Note to myself: this example does not need spiffs at all. The index.html content is not needed and can be replaced with generated on-the-fly stream data.

PR #1112, opened 2017-05-02
Refactored HttpClient, HttpServer and WebsocketConnection.

slaff:
- Better Code Design (There were some places that deserved to be called ["big-ball-of-mud"](https://en.wikipedia.org/wiki/Big_ball_of_mud) ).
- Faster and powerful HttpParser used in both HttpClient and HttpServer.
- Support for pipelining in both HttpClient and HttpServer.
- Support for connection reuse for both HttpClient and HttpServer.
- SSL session resumption support for HttpClient and HttpServer.
- Added ResourceTree to HttpServer to allow more flexible definition of resources.
- Added streaming support to both Http and WebSocket processing.
- rBootHttpUpdate should stop now on failure as early as possible.
- Changed the ContentType code to allow easier definition of new mime types.

## Backwards-Incompatible changes
- WebSocket is renamed to WebSocketConnection to reflect better its meaning and intended use.
- Removed the tightly coupled websocket methods inside the HttpServer. You can use WebSocketResource and add it to the HttpServer resource tree to achieve the same results.
- Moved writeInit(), writeFlash(const u8 *data, u16 size) and writeEnd() methods from rBootHttpUpdate to rBootItemOutputStream.
- TemplateFileStream::setVarsFromRequest needs to be refactored completely. The current code couples TemplateFileStream to HttpRequest, which is too restrictive. TemplateFileStream::setVars method should be introduced that accepts HashMap<String,String>
- HttpRequest::getRequestMethod() is removed. Use HttpRequest::method instead.
- CommandExecutor with WebsocketConnection won't work.


Vity01:
God bless the guy (@slaff) who rewrote the HttpClient.
It was the weakest part of the Sming. I can't wait for 3.3.0.

Vity01:
@slaff  Do you close/clean/free input stream for body in HttpRequest somewhere or the caller is responsible?

slaff:
@Vity01 If you use the HttpClient then at the end of an HTTP request the request data is freed in [HttpConnection::staticOnMessageComplete](https://github.com/SmingHub/Sming/blob/develop/Sming/SmingCore/Network/Http/HttpConnection.cpp#L189) and similar thing is happening for the HttpServer in [HttpServerConnection::staticOnMessageComplete](https://github.com/SmingHub/Sming/blob/develop/Sming/SmingCore/Network/Http/HttpServerConnection.cpp#L141).

Vity01:
@slaff thanks for the info, maybe the line comment might be helpful also for other developers
Thanks again for the great job.

lucas-bremond:
Great job!

Just out of curiosity, what is the rationale behind:

```cpp
aRequest.getRequestMethod() == RequestMethod::GET
```

being changed into:

```cpp
aRequest.method == HTTP_GET
```

?

It's shorter indeed, but seems more brittle to me.
Was it performance driven?

PR #1135, opened 2017-05-25
add Timer::loop() for better class derivation from Timer class

it-guru:
Derivations from Timer class can not add own loop methods. With this patch, you can create own derivations from Timer and place direct your own loop method.

### DIFF
--- a/Sming/SmingCore/Timer.cpp
+++ b/Sming/SmingCore/Timer.cpp
@@ -181,13 +181,13 @@ void Timer::processing(void *arg)
 					ptimer->stop();
 			}
 		}
-      ptimer->loop();
+		ptimer->tick();
 	}
 
 }
 
 
-void Timer::loop()
+void Timer::tick()
 {
 	if (callback)
 	{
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Timer.cpp
```diff
@@ -182,15 +181,23 @@ void Timer::processing(void *arg)
...
+{
+	if (callback)
+	{
+		callback();
+	}
+	else if (delegate_func)
+	{
+		delegate_func();
+	}
+   else{
```
slaff: @it-guru Please, fix the line indentation here.

Sming/SmingCore/Timer.cpp
```diff
@@ -182,15 +181,23 @@ void Timer::processing(void *arg)
...
-		{
-			ptimer->delegate_func();
-		}
+      ptimer->loop();
 	}
 
 }
+
+
+void Timer::loop()
```
slaff: @it-guru Please, rename the 'loop' method to 'tick'. Otherwise the Arduino people will be confused with the purpose of that one.

it-guru:
OK, i have fix/change the two deficiencies established by you.

PR #1144, opened 2017-06-03
CPU Frequency for SSL connection is only changed when it's lower than 160MHz

ZyperPL:
There was a problem with CPU frequency changing to 80MHz after SSL connection even if it was 160MHz before.
So now, if the user set CPU frequency to 160MHz there will be no debug messages and it will be changed to 80MHz.


slaff:
The change is not complete. You have to update also the code here: https://github.com/ZyperPL/Sming/blob/937278bca18577d0bb7509596b55d7cdc8ca188c/Sming/SmingCore/Network/TcpConnection.cpp#L550

ZyperPL:
Hm.. You're right @slaff . I've looked into that and I think it's impossible to know user defined frequency at this point. 
There is extra field somewhere needed to store CpuFrequency before SSL connection. I don't know where to put it actually. What do you think?
Also this "fix" might need more work than I thought. There should be "switching back" before some return statements. Because I'm not sure if we are switching back when there's an error.

PR #1145, opened 2017-06-04
support for WPS configuration

it-guru:
I have extend the StationClass by methods to enable WPS use. Due huge memory use of the wps lib (from SDK), i have add a make switch ENABLE_WPS=1 .
Tested with my access point and all looks well.otn

With WPS a sensor can be secure integrated in WLAN without any additional Apps on the handy. All you have to do ist activating WPS button on your access point.


it-guru:
It seems, my git handling is not perfect.

PR #1146, opened 2017-06-06
Fix 64 character WPA2 PSK password.

BotoX:
See: https://github.com/esp8266/Arduino/issues/1921
and https://github.com/esp8266/Arduino/commit/1b8f6d2e8e2dc7008660732b96197fd1ae1a1439

### DIFF
--- a/Sming/SmingCore/Platform/Station.cpp
+++ b/Sming/SmingCore/Platform/Station.cpp
@@ -54,7 +54,7 @@ bool StationClass::config(String ssid, String password, bool autoConnectOnStartu
 	station_config config = {0};
 
 	if (ssid.length() >= sizeof(config.ssid)) return false;
-	if (password.length() >= sizeof(config.password)) return false;
+	if (password.length() > sizeof(config.password)) return false;
 
 	bool enabled = isEnabled();
 	bool dhcp = isEnabledDHCP();
@@ -70,7 +70,10 @@ bool StationClass::config(String ssid, String password, bool autoConnectOnStartu
 		memset(config.password, 0, sizeof(config.password));
 		config.bssid_set = false;
 		strcpy((char*)config.ssid, ssid.c_str());
-		strcpy((char*)config.password, password.c_str());
+		if (password.length() == 64)
+			memcpy((char*)config.password, password.c_str(), 64);
+		else
+			strcpy((char*)config.password, password.c_str());
 
 		noInterrupts();
 


Sming/SmingCore/Platform/Station.cpp
```diff
@@ -70,7 +70,10 @@ bool StationClass::config(String ssid, String password, bool autoConnectOnStartu
 		memset(config.password, 0, sizeof(config.password));
 		config.bssid_set = false;
 		strcpy((char*)config.ssid, ssid.c_str());
-		strcpy((char*)config.password, password.c_str());
+		if (password.length() == 64)
+			memcpy((char*)config.password, password.c_str(), 64);
```
ADiea: The issue with this is that station_config::password is exactly 64 chars so you will have no ending zero... IMHO the correct fix would be to declare uint8 password[__65__]; in esp_sta.h, and signal a bug to espressif.

zgoda:
I've been there.

https://stackoverflow.com/a/21554182/12138

Sming/SmingCore/Platform/Station.cpp
zgoda: WPA passphrases are 8-63 ASCII characters by 802.11 spec so this is not a bug.

slaff:
@BotoX Please, fix the issue that Codacy is [reporting](https://www.codacy.com/app/slaff2/SmingOfficial/pullRequest?prid=694956).

slaff:
@zgoda @ADiea Any comments from your side before I merge this PR?

zgoda:
This change does not fix any problem because there is still 63 characters limit in SDK for both AP and station. Did anybody made a test how the application on esp behaves if wifi has 64 characters passphrase? Will it connect at all?

BotoX:
It works fine for me in station mode, my home network access point is using a 64 character passphrase.
I did not try to create an access point on the ESP with 64 character passphrase however.

zgoda:
Ok, none of my access points I have at home allows passphrases longer than 63 chars so I have no chance to test this.

ADiea:
To me still not clear what happens here if you use the 64th position for passphrase and no space left for zero ending the string. Maybe the sdk has special way of reading from this array?

zgoda:
Tape side A: ESP8266 core for Arduino gets this wrong, against IEEE 802.11 spec and against vendor published SDK.
Tape side B: we want compatibility with ESP8266 core for Arduino.

My guess is SDK just copies array using `memcpy`, not string with `strcpy` - all implementations of PBKDF2 I know of use byte arrays, not strings so this assumption seems plausible. Things may work by accident for border case of 64 byte passphrase, until somebody decides to make string from byte array somewhere in AP software effectively replacing last byte with zero and then things will go boom. So as with all network related programming I'd go traditional route of "be relaxed when receiving and strict when sending". If my assumption about SDK is right then SDK does this exactly.

Issue #1140, opened 2017-05-30
Issues with building sming_developer ENABLE_SSL=1


icodk:
Cloning developer branch, running make clean && make get the following errors:
`Services/Yeelight/YeelightBulb.cpp:9:23: fatal error: TcpClient.h: No such file
or directory`
I don't need this library anyway so I commented out 
`//#include "../Services/Yeelight/YeelightBulb.h"`
in SmingCore.h and removed the library from C:\tools\Sming\Sming\Services
started make and got a bit farther to the point that all was compiled except spiffy with the error 

>The program can't start because libwinpthread-1.dll is missing from your computer
> Try reinstalling the program to fix this problem

So I  C:\tools\Sming\Sming>cd spiffy and run make clean && make
Everything compiled nicely.
Now I tried to do ` C:\tools\Sming\Sming>make clean && make ENABLE_SSL=1`
and got the following error:
`make[2]: Leaving directory `/C/tools/sming/sming/third-party/esp-open-lwip'
/bin/sh: xtensa-lx106-elf-gcc: command not found
make[2]: xtensa-lx106-elf-gcc: Command not found`
and

 `make[2]: xtensa-lx106-elf-gcc: Command not found
make[2]: *** [crypto/aes.o] Error 127`

What do I do wrong ?

riegaz:
> Cloning developer branch, running make clean && make get the following errors:
Services/Yeelight/YeelightBulb.cpp:9:23: fatal error: TcpClient.h: No such file or directory
I don't need this library anyway so I commented out
//#include "../Services/Yeelight/YeelightBulb.h"
in SmingCore.h and removed the library from C:\tools\Sming\Sming\Services
started make and got a bit farther to the point that all was compiled except spiffy with the error

I created a pull request. https://github.com/SmingHub/Sming/pull/1148

PR #1148, opened 2017-06-11
TcpClient.h is already within SmingCore.h

riegaz:


slaff:
I would prefer to leave the source code as it is. The change that you have suggested is fixing an issue in your local setup. I would recommend you to make sure that your local environment is set correctly.

If this was an issue in the code then all our CI tests should have failed, which is not the case.

riegaz:
https://github.com/SmingHub/Sming/issues/1140 says the same. Also I just followed the steps from the provided how-to guide. Also the TcpClient.h is not within the same directory.

PR #1153, opened 2017-06-13
Add compatibility with Espressif SDK 2.1.0 published in git repository

avr39-ripe:
Add compatibility with ESP SDK 2.1.0 by removing extra extern declaration of SDK stuff and by including proper SDK headers where needed

### DIFF
--- a/Sming/Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
+++ b/Sming/Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
@@ -157,7 +157,9 @@ void Adafruit_ILI9341::begin(void) {
 	transmitCmdData(0xE1, data, 15);    	//Set Gamma
 
 	transmitCmd(0x11);    	//Exit Sleep
-	os_delay_us(120000);
+//	os_delay_us(120000);
+	os_delay_us(60000);
+	os_delay_us(60000);
 
 	transmitCmd(0x29);    //Display on
 	transmitCmd(0x2c);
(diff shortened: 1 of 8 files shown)


Sming/Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
```diff
@@ -157,7 +157,8 @@ void Adafruit_ILI9341::begin(void) {
 	transmitCmdData(0xE1, data, 15);    	//Set Gamma
 
 	transmitCmd(0x11);    	//Exit Sleep
-	os_delay_us(120000);
```
slaff: @avr39-ripe Can you remove that change from the PR. As far as I can see it has nothing to do with SDK 2.1.0.

slaff:
@avr39-ripe Why not mandatory, it would be very helpful if we can add also SDK 2.1.0 in our CI system.

Sming/Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
avr39-ripe: it is mandatory change. without this this library do not compile any more. Old os_delay_us declaration was uint32_t and new, from esp sdk 2.1.0 is uint16_t, so I made such "strange" change here.

avr39-ripe:
@slaff do not know nothing about how to add sdk 2.1.0 to CI :( can you help with it?

Sming/Libraries/Adafruit_ILI9341/Adafruit_ILI9341.cpp
zgoda: More code should be reviewed then to check if delays do not excede 65,535 limit.

slaff:
Will be merged with rebase PR #1264

PR #1182, opened 2017-07-12
Make WebsocketClient ping, pong, disconnect compatible to RFC 6455 (W…

jochenjagers:
…ebSocket Protocol)

ping, pong and disconnect are control frames sent from client to server.
As described in chapter 5.2 of RFC 6455 mask-bit must be set for all frames sent to the server.
Therefore byte 2 of these frames are changed from 0x00 to 0x80 and all frames are extended by 4 byte mask key (all zero)


slaff:
@avr39-ripe Can you, please,  review that PR ?

slaff:
@jochenjagers How can I test this? Was there a Websocket server that complained about it?

jochenjagers:
My Websocket Server is based on websocketpp (https://github.com/zaphoyd/websocketpp) and closes the connection after receiving a wrong ping message. Sming WebsocketClient sends a ping after a successfull websocket handshake. So it ends up in an endless connect/disconnect loop.

slaff:
The websocket RFC allows sending of ping and pongs without payload and mask. And that is what the current code is doing. What you suggest is to add sending ping/pong without payload with a mask and the mask to be 0x00.  While that may fix an issue that you are experiencing with your server it is not the best way to fix this issue.

What will be better is to improve the sendPing and sendPong methods to allow specifying payload and mask. 

I will close the PR and will submit another PR with the advanced sendPing and sendPong that you can test.

PR #1187, opened 2017-07-13
HttpServer tuning to allow serving of more requests.

slaff:
- Workaround for issue with incorrect size of files (Disadvantages: The loading process looks slower because the browser waits for the TIMEOUT in order to determine the size)

- TcpClient: Added reporting for condition that shouldn't happen and is not handled optimally.
- TcpServer: Decreased the minimum required heap size.
- LWIP: Decreased the TCP_MSL to 2 seconds. That will lead to faster garbage collection of tcp connections in TIME_WAIT state and will release resources faster.


slaff:
@piperpilot Can you try this branch with your shiny new esp-open-sdk compiler and your project? On my test project the results seem quite promising.

PR #1189, opened 2017-07-13
Handle incomplete frame to prevent WebsocketClient ending in an endle…

jochenjagers:
…ss loop

When a WebsocketClient receives an incomplete frame it hangs in an endless loop until the watchdog resets the system.
Marking the frame as incomplete and returning false resolves this issue.
But the data received with this frame is lost. I will create another pull-request for temoprary saving the data and combining them with the next frame(s) for correct evaluation


slaff:
@avr39-ripe  Can you look at that PR too?

PR #1203, opened 2017-07-26
Small fixes before the final 3.3.0

slaff:
- ugly fix for systems that don't have GNU `find` installed by default (example: Windows)
- small fix for passing by reference of the SSL fingerprint and key-certificate pair.


kwis2:
Technical question
I am wondering why there is a lot of problems with Sming compile. I think I found one reason.  Sming work with third-party modules, sometimes not newest ones. With master branch there are proper modules included with Sming, while with develop branches modules are linked to git (current) versions. 
Would it be possible to include proper modules with all branches?

slaff:
> while with develop branches modules are linked to git (current) versions.

@kwis2 Not really. It think that was explained multiple times but will explain it one more time. 

1. When you clone the develop branch you must NOT clone it recursively.

```
cd <dev-folder>
git clone https://github.com/SmingHub/Sming.git
export SMING_HOME=<dev-folder>/Sming # if you use Windows then: set %SMING_HOME%=<dev-folder->\Sming
```

2. When you want to compile the code the build system will take care to get the needed version of the third-party module. The only thing that needs to be done is to type

```
cd $SMING_HOME
make
```

If you manually update some of the git-submodules then it is your own fault. If git is difficult for you then work only with the `master` branch where the third-party modules are fetched and batched for you. And, of course, if you use the master branch you will get updates only when there is a new version. Take today for example :)

kwis2:
@slaff let me explain. With windows the only way I am able to make Sming to work is to download zip file. In all master branch there are files in third-pary folder, while in develop versions third-party folder is empty and I have to git clone. 
Additionally check git third-party folders. With master branch there are files, while in develop there are git links

PR #1213, opened 2017-07-31
Arduino compatibility improvements.

slaff:
- Fixed the registry length to be 32 bit (closes #398)
- Updated  Adafruit st7735 to its latest version ( the "speed-up" improvement is not re-applied yet )
- Updated Adafruit_SSD1306
- Added more small changes that should improve compatibility with Arduino Libraries.


slaff:
@harry-boe  Can you check if your Adafruit 1.4" TFT display with ST7735 chipset is still working with that PR ? If you have Linux/MacOS the steps to do this are:

```
cd $SMING_HOME
git stash
git fetch origin pull/1213/head:arduino-1213
git checkout arduino-1213
make dist-clean
make
```

And then compile and test your favourite sample(s).

PR #1226, opened 2017-08-20
Enable pull-up resistor in OneWire library

max-weller:
Enable the internal pull-up resistor for the GPIO used by OneWire library. OneWire requires by its design a pull-up resistor, and if the internal one is not enabled an external resistor would be required.

### DIFF
--- a/Sming/Libraries/OneWire/OneWire.cpp
+++ b/Sming/Libraries/OneWire/OneWire.cpp
@@ -124,7 +124,7 @@ OneWire::OneWire(uint8_t pin)
 
 void OneWire::begin()
 {
-	pinMode(pin, INPUT);
+	pinMode(pin, INPUT_PULLUP);
 	noPullup(pin);
 	bitmask = PIN_TO_BITMASK(pin);
 	baseReg = PIN_TO_BASEREG(pin);
@@ -136,7 +136,7 @@ void OneWire::begin()
 void OneWire::begin(uint8_t pinOneWire)
 {
 	pin = pinOneWire;
-	pinMode(pin, INPUT);
+	pinMode(pin, INPUT_PULLUP);
 	noPullup(pin);
 	bitmask = PIN_TO_BITMASK(pin);
 	baseReg = PIN_TO_BASEREG(pin);
--- a/Sming/Libraries/OneWire/OneWire.h
+++ b/Sming/Libraries/OneWire/OneWire.h
@@ -112,7 +112,7 @@
 #define IO_REG_TYPE						uint16_t
 #define IO_REG_ASM
 #define DIRECT_READ(base, mask)         (digitalRead(mask) ? 1 : 0)
-#define DIRECT_MODE_INPUT(base, mask)   pinMode(mask, INPUT)
+#define DIRECT_MODE_INPUT(base, mask)   pinMode(mask, INPUT_PULLUP)
 #define DIRECT_MODE_OUTPUT(base, mask)  pinMode(mask, OUTPUT)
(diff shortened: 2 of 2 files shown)


slaff:
@ADiea @avr39-ripe Can you check this PR?

Sming/Libraries/OneWire/OneWire.cpp
```diff
@@ -124,7 +124,7 @@ OneWire::OneWire(uint8_t pin)
 
 void OneWire::begin()
 {
-	pinMode(pin, INPUT);
+	pinMode(pin, INPUT_PULLUP);
 	noPullup(pin);
```
ADiea: This looks like it will disable the pullup. To me this looks put here on purpose maybe the Arduino way is not to enable pullups?. Also be aware that if you enable pullups and also have pullups they will go in parallel making the equivalent smaller in value that together with line capacitance might not be appropiate for given sck frequency. 
I would not touch lib code if it is working an tested as it probably is consistent with Arduino way. After onewire::begin you can manually set the pins to pullup in your client code.

ADiea:
I am not sure what is the default Arduino behavior here but I guess the default way is to disable pullups because there is a specifically disable pullups in the begin(). 

I would not enable them by default to not interfere with already present external pullups. If your application requires pullups internally I would advise of setting sda/sck pullups after ::begin() function is called. This will ensure everyone else is not affected by the change :)
Best regards

Sming/Libraries/OneWire/OneWire.h
```diff
@@ -112,7 +112,7 @@
 #define IO_REG_TYPE						uint16_t
 #define IO_REG_ASM
 #define DIRECT_READ(base, mask)         (digitalRead(mask) ? 1 : 0)
-#define DIRECT_MODE_INPUT(base, mask)   pinMode(mask, INPUT)
+#define DIRECT_MODE_INPUT(base, mask)   pinMode(mask, INPUT_PULLUP)
```
max-weller: This is the really interesting change

max-weller:
Hi @ADiea,
it is not possible to do it only in my application, without changing the library. The really important change is the one I highlighted above. The OneWire protocol is implemented by the library switching between Input and Output Mode all the time. If I would just change it to INPUT_PULLUP after my call to Onewire::begin(), it would be overwritten in no time. The change in the begin function is not important, the change in the headers is.

max-weller:
Ok, I've read up on the topic and I agree now that it's probably not good to change it unconditionally:
* For 1-Wire, [Dalllas recommends](https://datasheets.maximintegrated.com/en/ds/DS18B20.pdf) a 4k7 pull-up resistor
* [ESP8266 has 30k-100k pull-up resistor](http://bbs.espressif.com/viewtopic.php?t=1079)

However, it works pretty well with the internal pull-up if the sensor is close to the controller, therefore I'd like to change the PR to make it configurable.
Is that something you'd merge?

uzi18:
Please understand,  sometimes You need strong pullup with external
transistor,  so dont force it in library.
Maybe add it as parameter in constructor.

21.08.2017 12:04 PM "Max Weller" <notifications@github.com> napisał(a):

max-weller:
Thinking more about this, I don't see a problem with just enabling the internal pull-ups all the time. If a "stronger" pull-up is needed in an application, we can just put an external pull-up resistor. It will be in parallel to the internal pull-up, lowering the resistance, which should not hurt.

Therefore I think it would help in many applications where the sensor is close to the ESP, making the external pull-up unnecessary, while being no problem in the applications where an external pullup is used.

PR #1228, opened 2017-08-21
Fix python versions

max-weller:
Fix python calls to always use /usr/bin/env, and specify version where necessary (otherwise it won't work on distros like ArchLinux where python3 is the default)
Also fixes a small bug in decode-stacktrace which made it AFAIK incompatible with both python versions.


slaff:
Will be merged with rebase #1229.

PR #1243, opened 2017-09-15
Allow HttpResponse::sendString to be used multiple times for a single…

### LINKED ISSUE #1240: Regression: only last call to HttpResponse::sendString is used
With Sming 3.2.0, I used the sendString method multiple times in my handler to build up the response, like follows:

```
void Info::onHttp_Info(HttpRequest &request, HttpResponse &response)
{
//...
    response.sendString(this->getInfo());

    char s[128];
    for (int i = 0; i < this->device->services.count(); i++) {
        sprintf(s, "Service: %s\n", this->device->services[i]->getName() );
        response.sendString(s);
    }
//...
}
```

With Sming 3.3.0, this stopped working as expected. Only the string passed to the last invocation of sendString is sent to the client, the earlier strings are discarded.

slaff:
… response.

Fixes #1240.

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpResponse.cpp
+++ b/Sming/SmingCore/Network/Http/HttpResponse.cpp
@@ -55,22 +55,20 @@ HttpResponse* HttpResponse::setHeader(const String& name, const String& value)
 
 bool HttpResponse::sendString(const String& text)
 {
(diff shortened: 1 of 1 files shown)


slaff:
@max-weller Can you check if that PR is fixing the issue with sendString?

Issue #1168, opened 2017-06-24
Build system issues


Informatic:
- [x] `make clean` does not properly clean additional modules set in `Makefile-user.mk` variable `MODULES` (only removes `out/build/app*` files)
- [ ] Current build system does not use dependency files (`.d`), so that changes in headers do not cause rebuilds of source files

slaff:
@Informatic Please, submit PR with the proposed changes.

Informatic:
Yes, I (hopefully) will, in following week/weeks... :)

Informatic:
Ok, It seems like the first one is caused by me using relative path (`../base`) in `MODULES`, so that objects instead of being stored in `out/build/MODULENAME`, wind up in `out/build/../MODULENAME`. Sadly, I don't have any idea how to solve that without moving build directories over to module directories...

Informatic:
Ok, well, using `$(realpath ../base)` in `MODULES`, as used in rboot appcode, seems to have done the trick

PR #1248, opened 2017-09-30
Makefile: add make dependencies support

Informatic:
This allows for automatic rebuilds when header files are changed.

This solves #1168. Right now only `Makefile-project.mk` and `Makefile-rboot.mk` support this. This might be preferable to be implemented in main `Makefile` as well.


slaff:
> This might be preferable to be implemented in main Makefile as well.

@Informatic Please, add it also there as it will be the most used place during the development of the library itself.

slaff:
Merged with rebase #1257

PR #1269, opened 2017-10-22
spiffs - Detect rBoot integration and replace default mount and format functions...

jochenjagers:
by the manual ones with correct spiffs position and size from Makefile-rboot.mk

There is still an issue with the FTPServer as it can call spiffs_format. This call is compiled on libary compile time without any knowlege about rBoot rather than on appilication compile time.
So the use of "FSFORMAT" via the FTP-Server is impossible as long as the application is compiled with rBoot support.

### DIFF
--- a/Sming/Services/SpifFS/spiffs_sming.c
+++ b/Sming/Services/SpifFS/spiffs_sming.c
@@ -142,7 +142,7 @@ static void spiffs_mount_internal(spiffs_config *cfg)
   //debugf("%X", dat);
 }
 
-void spiffs_mount()
+void spiffs_mount_default()
 {
   spiffs_config cfg = spiffs_get_storage_config();
   spiffs_mount_internal(&cfg);
@@ -165,7 +165,7 @@ void spiffs_unmount()
 }
 
 // FS formatting function
-bool spiffs_format()
+bool spiffs_format_default()
 {
   spiffs_unmount();
   spiffs_config cfg = spiffs_get_storage_config();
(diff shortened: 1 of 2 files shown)


anakod:
@jochenjagers I think it will be more "right C\C++ way" solution to check RBOOT defines directly inside spiffs_mount() function code and set related addresses for spiffs_mount_internal, instead of replacing function defination with external macros (and same for format function). Also it will be more simple to read and debug.

jochenjagers:
@anakod I don't like this macros either. My first idea was to integrate the rboot information in the function code of spiffs_get_storage_config. 
But the RBOOT defines are project specific. There is no infotmation about it when compiling the sming library. 
I would also prefer a more "right C/C++ way" but the only idea i have for this, is moving the implementaion of spiffs_get_storage_config function from the library to the project code or add some kind of callback function to overwrite the values defined in the library. This would also solve the problem of using spiffs functions in other parts of the library e. g. the FTPServer. 
I will think about it and try to find a better solution.

slaff:
@jochenjagers My 2 cents - mark all functions in the library, that should behave differently when rBoot is used as `weak`. 

Example: `void __attribute__((weak)) spiffs_mount()`

This way they will be in the final sming*.a library. 

Now during compilation and linking of an rBoot app compile the same functions with the new code (adjusted during compile time), DON'T mark the new function as `weak` and leave the linker to select the right function version during linking time. (Ex: if we have a weak spiffs_mount and a non-weak one the linker will choose the latter)

This way they correct `spiffs_mount` will be used in the final application.

slaff:
The implementation of the hints above are in that PR #1292. The new PR lacks thorough testing so help is welcomed :)

PR #1270, opened 2017-10-24
Initial test code for improved sendPing and sendPong.

slaff:
Related to #1182.

### DIFF
--- a/Sming/SmingCore/Network/Http/Websocket/WebSocketConnection.cpp
+++ b/Sming/SmingCore/Network/Http/Websocket/WebSocketConnection.cpp
@@ -125,7 +125,9 @@ int WebSocketConnection::staticOnControlBegin(void* userData, ws_frame_type_t ty
 		return -1;
 	}
 
-	connection->controlFrameType = type;
+	connection->controlFrame.type = type;
+	connection->controlFrame.payload = NULL;
+	connection->controlFrame.payloadLegth = 0;
 
 	if (type == WS_FRAME_CLOSE) {
 		connection->close();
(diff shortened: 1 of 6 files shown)


jochenjagers:
I think, that optional masking of control frames is no good idea. The websocket-standard (RFC 6455) says clearly that every message send from client to server must be masked. There is no good reason to leave the mask out.

From RFC 6455:
> Mask:  1 bit
> Defines whether the "Payload data" is masked.  If set to 1, a
> masking key is present in masking-key, and this is used to unmask
> the "Payload data" as per Section 5.3.  **All frames sent from
> client to server have this bit set to 1.**

slaff:
The RFC is a big vague IMHO. I would have expected something like " All frames sent from
client to server **MUST** have this bit set to 1.".

But if that is really the requirement then actually we should remove the mask parameter from the method signature and directly generate random mask inside the sendPing and sendPong methods.

@jochenjagers Just to make sure that we are on the right track, can you find how it is implemented in popular websocket clients?

jochenjagers:
@slaff i will check some clients next week

slaff:
@jochenjagers Any news on that?

jochenjagers:
@slaff sorry for my late respone. I was very busy the last three weeks. I will check other clients this weekend

jochenjagers:
I checked several client implementations

Python:

ws4py: Client masks all frames (client/__init__.py line 125)
websocket_client: Client masks all frames (websocket/_abnf.py line 200)

C++:

websocketpp: Client masks all frames (websocketpp/processors/hybi13.hpp line 605)

JS:

WebKit/JavaScriptCore: Client masks all frames (Source/WebCore/Modules/websockets/WebSocket.cpp line 643)
node.js / ws : Client masks all frames by default (mask for ping/pong can be disabled, but is enabled by default)

All clients I checked, masking the frames send to the server. Nearly none has an option to disable this.
I think, we can also omit this possibility as there is no need for it.

I can check more next weekend if you are not convinced :)

slaff:
> I can check more next weekend if you are not convinced :)

@jochenjagers Thanks for the research so far. There is no need to invest more time. Can you get the source code from this PR, clean up using `make dist-clean`, compile and check if it is working with your websocket server ?

PR #1273, opened 2017-10-31
Fix Memory Leak in Mqtt

frankdownunder:
Added missing memory deallocations.  Only exhibits the memory leak when your connection fails and constantly reties to connect.


frankdownunder:
@slaff My first Pull Request ever, and it failed.  Did I do it incorrectly?

slaff:
@frankdownunder Welcome to the contributors :) 

> Did I do it incorrectly?

As far as I can see everything is fine. We would be glad to see other PRs from you.

PR #1280, opened 2017-11-08
Add method to stop HttpSever

bbourdel:
Done : Add function close() to HttpServer class.
It closes the connection when the last TCP connection is closed.

To do :  free all RAM (heap) taken when HTTP server is created (about 50% is freed now, 50 other % is still losed).

### DIFF
--- a/Sming/SmingCore/Network/HttpServer.cpp
+++ b/Sming/SmingCore/Network/HttpServer.cpp
@@ -17,12 +17,14 @@
 
 HttpServer::HttpServer()
 {
+	active = true;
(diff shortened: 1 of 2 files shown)


slaff:
@bbourdel There should be code that prevents opening new connections if the server is not active. Also the code relies on the fact that all connections will close themselves. If you have a Websocket connection then the waiting can be quite long. It would be better to stop accepting new connections, get all open connections, close them first, and when all are reported closed to shutdown the server.

slaff:
The changes from this PR will be merged with rebase #1284.

PR #1281, opened 2017-11-08
fix/MemoryLeak(Heap) during TCP Client connection and delete

bbourdel:
Not really sure of this fix. 

But on my side, it's help to solve Memory Leak when I call httpClient.cleanup(). (About 50Bytes losed each time)

### DIFF
--- a/Sming/SmingCore/Network/HttpClient.cpp
+++ b/Sming/SmingCore/Network/HttpClient.cpp
@@ -115,17 +115,47 @@ void HttpClient::freeSslSessionPool() {
 			free(sslSessionIdPool[key]->value);
 		}
 		free(sslSessionIdPool[key]->value);
+		free(sslSessionIdPool[key]);
 	}
 	sslSessionIdPool.clear();
 }
 #endif
 
+
+void HttpClient::freeRequestQueue() {
+	for(int i=0; i< queue.count(); i ++) {
+		debugf("~RquestQueue");
+		String key = queue.keyAt(i);
+		RequestQueue* rq = queue[key];
+		HttpRequest* req_to_del = rq->dequeue();
+		while(req_to_del != NULL){
+			debugf("~Request");
+			delete req_to_del;
+			req_to_del = rq->dequeue();
+		}
+		queue[key]->flush();
+		delete queue[key]; // Delete the FIFO list of request (does it delete request too ?)
+	}
+	queue.clear();
+}
+
+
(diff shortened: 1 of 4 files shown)


Sming/SmingCore/Network/TcpClient.cpp
```diff
@@ -46,8 +46,11 @@ TcpClient::TcpClient(TcpClientDataDelegate onReceive)
 
 TcpClient::~TcpClient()
 {
-	delete stream;
-	stream = NULL;
+	if (stream != NULL)
```
slaff: That change is not needed. `delete` works nicely with pointers that have NULL value.

Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -36,6 +36,12 @@ TcpConnection::~TcpConnection()
 		delete[] sslFingerprint.pkSha256;
 	}
 	freeSslClientKeyCert();
+	if(ssl_ext != NULL) {
```
slaff: That is not the right place to put that change. It is also not needed because SSL closing will free internally also ssl_ext.

Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -115,17 +115,47 @@ void HttpClient::freeSslSessionPool() {
...
 }
 #endif
 
+
+void HttpClient::freeRequestQueue() {
+	for(int i=0; i< queue.count(); i ++) {
+		debugf("~RquestQueue");
+		String key = queue.keyAt(i);
+		RequestQueue* rq = queue[key];
+		HttpRequest* req_to_del = rq->dequeue();
```
slaff: Coding Style: Naming of variables is incorrect. Please use camelCase.

Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -115,17 +115,47 @@ void HttpClient::freeSslSessionPool() {
...
+		free(sslSessionIdPool[key]);
 	}
 	sslSessionIdPool.clear();
 }
 #endif
 
+
+void HttpClient::freeRequestQueue() {
+	for(int i=0; i< queue.count(); i ++) {
+		debugf("~RquestQueue");
```
slaff: Remove all `debugf` calls that you have added. We have more than enough debug data.

slaff:
Will be merged with Rebase #1283.

PR #1282, opened 2017-11-08
Add Adafruit_BME280 Library

mschader:
This will add the Adafruit_BME280 Library to Sming to have compatibility for the Sensor.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -45,3 +45,6 @@
 	path = Sming/Libraries/Adafruit_SSD1306
 	url = https://github.com/adafruit/Adafruit_SSD1306.git
 	ignore = dirty
+[submodule "Sming/Libraries/Adafruit_BME280_Library"]
+	path = Sming/Libraries/Adafruit_BME280_Library
+	url = https://github.com/adafruit/Adafruit_BME280_Library.git
--- a/Sming/Libraries/.patches/Adafruit_BME280_Library.patch
+++ b/Sming/Libraries/.patches/Adafruit_BME280_Library.patch
@@ -1,0 +1,37 @@
+diff --git a/Adafruit_BME280.cpp b/Adafruit_BME280.cpp
+index 6ed7dcf..cc8a41f 100644
+--- a/Adafruit_BME280.cpp
++++ b/Adafruit_BME280.cpp
+@@ -14,9 +14,8 @@
+   Written by Limor Fried & Kevin Townsend for Adafruit Industries.
+   BSD license, all text above must be included in any redistribution
+  ***************************************************************************/
+-#include "Arduino.h"
+-#include <Wire.h>
+-#include <SPI.h>
++#include "../../SmingCore/Wire.h"
++#include "../../SmingCore/SPI.h"
+ #include "Adafruit_BME280.h"
+ 
+ /***************************************************************************
+diff --git a/Adafruit_BME280.h b/Adafruit_BME280.h
+index 61aeeed..b9c3e7f 100644
+--- a/Adafruit_BME280.h
++++ b/Adafruit_BME280.h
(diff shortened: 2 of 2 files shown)


slaff:
@mschader Can you add this library as submodule? Use something like this:

```
cd $SMING_HOME/..
git submodule add https://github.com/adafruit/Adafruit_BME280_Library.git \
Sming/Libraries/Adafruit_BME280_Library
```

And ONLY if there is a need for local changes then add those changes to Sming/Libraries/.patches/Adafruit_BME280_Library.patch.

mschader:
@slaff Done so.

.gitmodules
```diff
@@ -45,3 +45,6 @@
 	path = Sming/Libraries/Adafruit_SSD1306
 	url = https://github.com/adafruit/Adafruit_SSD1306.git
 	ignore = dirty
+[submodule "Sming/Libraries/Adafruit_BME280_Library"]
+	path = Sming/Libraries/Adafruit_BME280_Library
+	url = https://github.com/adafruit/Adafruit_BME280_Library.git
```
slaff: @mschader Please, add also that option to the submodule:
`ignore = dirty`

Sming/Libraries/.patches/Adafruit_BME280_Library.patch
```diff
@@ -0,0 +1,37 @@
...
+-#include <Adafruit_Sensor.h>
+-#include <Wire.h>
++#include "Adafruit_Sensor.h"
++#include <../../SmingCore/Wire.h>
+ 
+ /*=========================================================================
+     I2C ADDRESS/BITS
+     -----------------------------------------------------------------------*/
+-    #define BME280_ADDRESS                (0x77)
++    #define BME280_ADDRESS                (0x76)
```
slaff: @mschader I have no idea about this sensor but I am curious to know what is the reason for changing the address here?

Sming/Libraries/.patches/Adafruit_BME280_Library.patch
```diff
@@ -0,0 +1,37 @@
...
+index 61aeeed..b9c3e7f 100644
+--- a/Adafruit_BME280.h
++++ b/Adafruit_BME280.h
+@@ -23,13 +23,13 @@
+  #include "WProgram.h"
+ #endif
+ 
+-#include <Adafruit_Sensor.h>
+-#include <Wire.h>
++#include "Adafruit_Sensor.h"
```
slaff: Where is `Adafruit_Sensor.h` coming from? The compilation fails due to that missing file. It looks to me that also this `https://github.com/adafruit/Adafruit_Sensor` should be added as a submodule.

Sming/Libraries/.patches/Adafruit_BME280_Library.patch
slaff: Just for my understanding why is the address value changed here?

.gitmodules
slaff: @mschader Please, add also the following line to the file:
`ignore = dirty`

Sming/Libraries/.patches/Adafruit_BME280_Library.patch
```diff
@@ -0,0 +1,37 @@
...
++#include <../../SmingCore/Wire.h>
```
slaff: Have you tried using <Wire.h> as in the original code?

Sming/Libraries/.patches/Adafruit_BME280_Library.patch
mschader: The I2C Address can differ between chip models. I bought several from different resellers and tested them and found none using address 0x77. Thats why i decided to change the default to 0x76.

PR #1284, opened 2017-11-08
Allow shutting down of TcpServers

slaff:
Rebase of #1280.

### DIFF
--- a/Sming/SmingCore/Network/HttpServer.cpp
+++ b/Sming/SmingCore/Network/HttpServer.cpp
@@ -113,7 +113,7 @@ void HttpServer::setDefaultResource(HttpResource* resource)
 void HttpServer::shutdown()
 {
 	active = false;
-	for(int i; i < connections.count(); i++) {
+	for(int i=0; i < connections.count(); i++) {
 		HttpServerConnection* connection = connections[i];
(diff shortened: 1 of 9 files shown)


slaff:
@bbourdel Can you check if this PR is working for you? The way to shutdown the server is to call `HttpServer::shutdown()`.

bbourdel:
@slaff Tested and OK !! :-)
Thanks for your completion.

Sming/SmingCore/Network/HttpServer.cpp
```diff
@@ -85,21 +93,43 @@ void HttpServer::setDefaultHandler(const HttpPathDelegate& callback)
...
+	}
+}
+
+void HttpServer::onConnectionClose(TcpClient& connection, bool success)
+{
+	connections.removeElement((HttpServerConnection*)&connection);
+	totalConnections = connections.count();
+	if(totalConnections == 0 && !active){
+		debugf("Shutting down the Http Server");
+		delete this;
```
slaff: Note to myself: That will cause a race condition - HttpServer will start freeing the resources and meanwhile the HttpServerConnections will try to use those resource in their destructors... A timer with 3 - 4 seconds delay should fix this...

PR #1288, opened 2017-11-14
fix some ssl memory leaks

ericogr:
I've got memory leak using mqtt with ssl enabled. It occurs when broker restart (using mqtt.connect(...) again to reconnect)

adding ssl_ctx_free(sslContext) when closing and onError apparently worked but I'm not shure if it can cause side effects or best way to do it

### DIFF
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -418,8 +418,8 @@ err_t TcpConnection::staticOnConnected(void *arg, tcp_pcb *tcp, err_t err)
 #endif
 			debugf("SSL: handshake start (%d ms)", millis());
 
-			if(con->ssl != NULL) {
-				ssl_free(con->ssl);
+			if(con->sslContext != NULL) {
+			    ssl_ctx_free(con->sslContext);
 			}
 
 			con->sslContext = ssl_ctx_new(SSL_CONNECT_IN_PARTS | sslOptions, 1);
--- a/Sming/SmingCore/Network/TcpConnection.cpp
+++ b/Sming/SmingCore/Network/TcpConnection.cpp
@@ -165,7 +165,7 @@ void TcpConnection::onError(err_t err)
 #ifdef ENABLE_SSL
 	if(ssl) {
 		sslConnected = false;
-		ssl_free(ssl); // ssl_free frees internally also the SSL context and the SSL extension data
+		ssl_ctx_free(sslContext);
 		sslContext=nullptr;
 		sslExtension = NULL;
 		ssl=nullptr;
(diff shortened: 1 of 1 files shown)


Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -165,7 +165,8 @@ void TcpConnection::onError(err_t err)
 #ifdef ENABLE_SSL
 	if(ssl) {
 		sslConnected = false;
-		ssl_free(ssl); // ssl_free frees internally also the SSL context and the SSL extension data
+		ssl_free(ssl);
+		ssl_ctx_free(sslContext);
```
slaff: Hm.. looking at the axtls code for `ssl_ctx_free` it seems that the function internally calls `ssl_free`. Excerpt:

```C
EXP_FUNC void STDCALL ssl_ctx_free(SSL_CTX *ssl_ctx)
{
    SSL *ssl;
    int i;

    if (ssl_ctx == NULL)
        return;

    ssl = ssl_ctx->head;

    /* clear out all the ssl entries */
    while (ssl)
    {
        SSL *next = ssl->next;
        ssl_free(ssl);
        ssl = next;
    }

// ...
```

Which means that probably just replacing `ssl_free(ssl);` with `ssl_ctx_free(sslContext);` should work.

ericogr: that's true, I'll try :)

Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -165,7 +165,8 @@ void TcpConnection::onError(err_t err)
 #ifdef ENABLE_SSL
 	if(ssl) {
 		sslConnected = false;
-		ssl_free(ssl); // ssl_free frees internally also the SSL context and the SSL extension data
+		ssl_free(ssl);
```
slaff: @ericogr Please, delete that line. The line below should take care to do the right free-ing.

Sming/SmingCore/Network/TcpConnection.cpp
```diff
@@ -297,8 +298,8 @@ void TcpConnection::close()
 #ifdef ENABLE_SSL
 	if (ssl != nullptr) {
 		debugf("SSL: closing ...");
-//		ssl_ctx_free(sslContext);
 		ssl_free(ssl);
```
slaff: @ericogr Same as above - delete that line above. The line below should take care to do the right free-ing.

Sming/SmingCore/Network/TcpConnection.cpp
slaff: @ericogr May be one last thing - add the following comment behind the `ssl_ctx_free(sslContext);` // ssl_ctx_free frees the context. Internally it calls ssl_free which frees SSL extension data and SSL itself.

PR #1300, opened 2017-12-14
Modified libemqtt & MqttClient to allow the publishing of binary messages

MayaPosch:
During local testing it was found that the Sming-based firmware was not able to submit binary MQTT messages across the network, or only part of the binary message. The problem was localised in the libemqtt library, which assumes that any incoming message string is ASCII or similar, and uses strlen() to determine the length. Any null-byte will thus terminate the message at that point.

By changing the MqttClient class to submit the message length (String::length()) and modifying the libemqtt publish() method to use this new parameter, this issue is prevented.

(My apologies for the previous broken PR. Hadn't fully tested it before committing)

### DIFF
--- a/Sming/Services/libemqtt/libemqtt.c
+++ b/Sming/Services/libemqtt/libemqtt.c
@@ -400,13 +400,12 @@ int mqtt_ping(mqtt_broker_handle_t* broker) {
 	return 1;
 }
 
-int mqtt_publish(mqtt_broker_handle_t* broker, const char* topic, const char* msg, uint16_t mlength, uint8_t retain) {
-	return mqtt_publish_with_qos(broker, topic, msg, mlength, retain, 0, NULL);
+int mqtt_publish(mqtt_broker_handle_t* broker, const char* topic, const char* msg, uint16_t msglen, uint8_t retain) {
+	return mqtt_publish_with_qos(broker, topic, msg, msglen, retain, 0, NULL);
 }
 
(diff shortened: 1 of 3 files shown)


Sming/Services/libemqtt/libemqtt.c
```diff
@@ -400,13 +400,13 @@ int mqtt_ping(mqtt_broker_handle_t* broker) {
...
-	return mqtt_publish_with_qos(broker, topic, msg, retain, 0, NULL);
+int mqtt_publish(mqtt_broker_handle_t* broker, const char* topic, const char* msg, uint16_t mlength, uint8_t retain) {
+	return mqtt_publish_with_qos(broker, topic, msg, mlength, retain, 0, NULL);
 }
 
-int mqtt_publish_with_qos(mqtt_broker_handle_t* broker, const char* topic, const char* msg, uint8_t retain, uint8_t qos, uint16_t* message_id) {
+int mqtt_publish_with_qos(mqtt_broker_handle_t* broker, const char* topic, const char* msg, uint16_t mlength, uint8_t retain, uint8_t qos, uint16_t* message_id) {
 	uint16_t topiclen = strlen(topic);
-	uint16_t msglen = strlen(msg);
+	uint16_t msglen = mlength;
```
slaff: @MayaPosch Please, remove line 409 completely and instead name the function parameter msglen. Apply the name change also to the header file.

MayaPosch:
Figured I would also update the documentation comments in the libemqtt header while I was at it. Shame if they got out of date :)

PR #1301, opened 2017-12-18
sed -r not working on MacOS has to be replaced by sed -E

johndoe8967:
after replacing sed the new offset calculation not working, irom0_0_seg is placed on 0x7000

### DIFF
--- a/Sming/compiler/ld/standalone.rom.ld
+++ b/Sming/compiler/ld/standalone.rom.ld
@@ -5,7 +5,7 @@ MEMORY
   dport0_0_seg :                        org = 0x3FF00000, len = 0x10
   dram0_0_seg :                         org = 0x3FFE8000, len = 0x14000
   iram1_0_seg :                         org = 0x40100000, len = 0x8000
-  irom0_0_seg :                         org = 0x4020a000, len = (1M - 0x0a000)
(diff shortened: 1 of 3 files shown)


Sming/compiler/ld/standalone.rom.ld
```diff
@@ -5,7 +5,7 @@ MEMORY
   dport0_0_seg :                        org = 0x3FF00000, len = 0x10
   dram0_0_seg :                         org = 0x3FFE8000, len = 0x14000
   iram1_0_seg :                         org = 0x40100000, len = 0x8000
-  irom0_0_seg :                         org = 0x4020a000, len = (1M - 0x0a000)
```
slaff: @johndoe8967 Please, remove that change from the PR. It is not related to the issue that you are solving.

johndoe8967: I think it is, because otherwise I get seg faults, but I removed the change
-> so on MacOS Sming is still not working

slaff:
>  so on MacOS Sming is still not working

A proper fix could be adding gnu-sed installation to the MacOS X wiki 

```
brew install gnu-sed
```

and change the `Makefile-macos.mk` to point SED to `gsed`.

The current solution that you have suggested is not working on MacOSX and breaks all other platforms.

Issue #1304, opened 2017-12-22
Makefile - can not have same filenames in different locations


cr1st1p:
I have one library (added via MODULES, in my Makefile) that has a base Controller.cpp file - in one namespace.
Then, in my application, I have app/Controller.cpp , in a different namespace (a class  derived from the library one)
The problem is that current Makefile will compile the library one, but not also the app one.
This is because of the vpath setting and implicit .cpp rules, which makes Make find that 
*build/app/Controller.o dependency is on my module/Controller.cpp instead of app/Controller.cpp

There is a mention about this type of issue also in http://www.conifersystems.com/whitepapers/gnu-make/ , chapter "Pattern Rules and Search Paths"

cr1st1p:
I'm on my way to fix this. Bumped also on a problem with the archive program ('ar') which also has problem with same filenames.

cr1st1p:
Forgot to mention - Pull request is #1305

slaff:
Fixed with that rebase #1335

PR #1305, opened 2017-12-23
Fix makefiles for when you have same filenames. (#1304)

cr1st1p:
One fix is strictly regarding Makefile's logic.
Removed vpath setting + different way of generating
the implicit rules.

Another fix is about 'ar' being too dumb and not being able to
cope too well with identical file names. One item, part of the
fix, is to remove the library file before generating it.


slaff:
@cr1st1p Can you please rebase your changes against the latest `develop` version and make sure that you don't introduce changes related to spaces and indentation?

PR #1306, opened 2017-12-24
Adds Cygwin32 build support

riban-bw:
This PR adds support for Cygwin (32-bit) on Windows build hosts.

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -38,7 +38,7 @@ GIT ?= git
 #Detect if Cygwin and / or Windows
 ifeq ($(OS),Windows_NT)
   # Windows detected
-ifneq (a,$(shell echo "a"))
+ifeq (,$(wildcard /usr/bin/cygwin1.dll))
   UNAME := Windows
 else
   UNAME := Cygwin
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -35,36 +35,48 @@ COM_SPEED = 115200
 # Git command
 GIT ?= git
 
-#Detect if Cygwin and / or Windows
+# Detect OS and build environment
+UNAME := $(shell uname -s)
+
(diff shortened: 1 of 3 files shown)


slaff:
@riban-bw Can you please check why your changes are breaking the Windows CI system (https://ci.appveyor.com/project/slaff/sming-sb483/build/1.0.546)?

riban-bw:
I am trying to figure out why chocolaty install fails but I don't have the facility to install via choco. I have run up an Azure Windows 10 instance but it fails to install Sming - 7za hangs - so I am struggling. I will continue to investigate but anyone with a Choco install who can take a look would be beneficial.

riban-bw:
Fixed the build issues. It looks like Windows CI system (AppVeyor) is using msys environment (which I may also try to configure locally - but previous attempts have failed) which was not distinguishable from cygwin using the test I had implemented. Now checking for cygwin1.dll in /usr/bin. I also had to fix the double slash filter I had added to two of the makefiles in which I had added a spurious space.
I also checked with AppVeyor and Travis building 2.1.0 and that worked fine but I have not left the 2.1.0 config in appveyor.yml as that should be done under a separate PR.
I had to make similar changes to several Makefiles - I suggest we consider some consolidation of Makefiles to reduce data redundancy (duplication) but again, the subject of a separate PR.
If this PR is accepted I will add instructions to the Wiki describing how to configure Sming for Windows using Cygwin.

riban-bw:
Hold off - there are some issues with running Cygwin compiler in Windows environment, e.g. compiling within an IDE. It is to do with Make using Windows style paths for includes but Unix style paths for other commands and particularly the failure of the sed command. I am investigating...

riban-bw:
Okay - I think this is good. Please review.

PR #1318, opened 2018-01-29
Improvements to the InfraRed libraries

slaff:


samples/IR_lib/app/application.cpp
```diff
@@ -4,37 +4,40 @@
...
 
 void init()
 {
 	Serial.begin(SERIAL_BAUD_RATE); // 115200 by default
-	Serial.println("Setting up...");
-	irrecv.blink13(1);
+	Serial.println("Setting up...");	
 	irrecv.enableIRIn(); // Start the receiver
-	irTimer.initializeMs(1000, receiveIR).start();
+        irTimer.initializeMs(30, receiveIR).start();
```
slaff: - Add correct indentation
- A timer that is below 50 ms may cause watchdog resets.

Sming/Makefile
```diff
@@ -181,6 +181,10 @@ EXTRA_INCDIR    = include system/include Wiring Libraries SmingCore $(SDK_BASE)/
 THIRD_PARTY_DATA = third-party/rboot/Makefile
 EXTRA_INCDIR += third-party/rboot third-party/rboot/appcode 
 
+# => IRremote8266
+MODULES += Libraries/IR/src
```
slaff: Things to improve in our build system: 
- The Arduino libraries c*, h files can be stored in the main root folder of the library or under `src`. We should change the build system to take care also for the second case.

.gitmodules
```diff
@@ -61,3 +61,11 @@
...
 	url = https://github.com/espressif/ESP8266_NONOS_SDK.git
 	ignore = dirty
+[submodule "Sming/Libraries/RingBufCPP"]
+	path = Sming/Libraries/RingBufCPP
+	url = https://github.com/wizard97/Embedded_RingBuf_CPP.git
+        ignore = dirty
+[submodule "Sming/Libraries/IR"]
+	path = Sming/Libraries/IR
+	url = https://github.com/markszabo/IRremoteESP8266.git
+        ignore = dirty
```
slaff: - Better indentation needed.

.gitmodules
```diff
@@ -61,3 +61,11 @@
 	path = Sming/third-party/ESP8266_NONOS_SDK
 	url = https://github.com/espressif/ESP8266_NONOS_SDK.git
 	ignore = dirty
+[submodule "Sming/Libraries/RingBufCPP"]
+	path = Sming/Libraries/RingBufCPP
+	url = https://github.com/wizard97/Embedded_RingBuf_CPP.git
+        ignore = dirty
```
slaff: - Better indentation needed.

Sming/Libraries/.patches/IR.patch
```diff
@@ -0,0 +1,511 @@
...
++#include <user_interface.h>
+ }
+ #include <Arduino.h>
+-#endif
++
+ #include <algorithm>
+ #include "IRremoteESP8266.h"
++#include "Print.h"
+ 
+-#ifdef UNIT_TEST
```
slaff: Is there a reason for the big amount of changes in that file? I would prefer if those changes are improvements to be submitted directly to the upstream project and to keep here only the small changes that make the code compile.

Sming/Makefile
slaff: @horde3d With commit 58f0214cfbd33d6aca021103bdeed1e3d1a1544c that line is no longer needed. Please, rebase your fork and push back the changes. If you are not sure how then take a look at our https://github.com/SmingHub/Sming/blob/develop/CONTRIBUTING.md

horde3d:
I agree that the changes to the IR lib would be better placed in the IR libraries repository. But I'm not sure if I will find the time for adjusting the code very soon. Some of m changes are probably not that easy to integrate, as the IRRemote library is not providing support for external libraries like the RingBuf library. I'm also not sure if it's possible to have a class member as an IRQ callback without the support of the SmingCore library. As mentioned I'm new to the ESP/Arduino programming.

Another problem with the WEST resets seems to be solved. After upgrading the ESP SDK from 1.5.4 to 2.1 the problem was gone. It was also gone after disabling WiFi, so maybe something improved in the SDK regarding WiFi handling. The timer was not the problem. Even if I disabled it completely, the wdt resets occurred with 1.5.4 when I added data to the ringbuffer in the IRQ callback. But as it is now solved with the newer SDK I don't care about it anymore

samples/IR_lib/app/application.cpp
slaff: @horde3d Can you please fix the indentation on the line above?

Sming/Libraries/.patches/RingBufCPP.patch
```diff
@@ -0,0 +1,22 @@
...
++        if (_numElements < MaxElements) {
+             _buf[_head] = obj;
+             _head = (_head + 1)%MaxElements;
+             _numElements++;
+@@ -53,7 +53,7 @@ bool pull(Type *dest)
+ 
+     RB_ATOMIC_START
+     {
+-        if (!isEmpty()) {
++        if(_numElements) {
```
slaff: @horde3d Why is the change above needed? I don't see an apparent reason for it so I would be glad if you can share more details?

Sming/Libraries/.patches/RingBufCPP.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/RingBufCPP.h b/RingBufCPP.h
+index 3b5f092..8df2c0d 100644
+--- a/RingBufCPP.h
++++ b/RingBufCPP.h
+@@ -28,7 +28,7 @@ bool add(const Type &obj)
+     bool ret = false;
+     RB_ATOMIC_START
+     {
+-        if (!isFull()) {
++        if (_numElements < MaxElements) {
```
slaff: @horde3d Why is the change above needed? I don't see an apparent reason for it so I would be glad if you can share more details?

horde3d:
The change in the RingBufCPP was a leftover from my attempts to analyse the WDT resets. Just wanted to optimize the function call which is probably senseless as the compiler should optimize it anyway. But I was really desperate when trying to fix the WDT resets.

I normally don't use the example, that's why I didn't fixed the example application so far. Will fix it soon

slaff:
@horde3d Please, rebase on the latest develop version.  At the moment I see a lot of changes https://github.com/SmingHub/Sming/pull/1318/files which are not directly related to your patch. If you need help with rebasing in the [CONTRIBUTING.md document](https://github.com/SmingHub/Sming/blob/develop/CONTRIBUTING.md) you will find a a section related to rebasing **"Rebase if needed"**.

horde3d:
Should be now rebased. Git is still a bit of a mistery for me as a Mercurial user

slaff:
> that's why I didn't fixed the example application so far. Will fix it soon

@horde3d The rebase looks good. How about the example? Is it working now?

horde3d:
It's working, although it's value is of course limited (why should one want to resend what you receive). And polling with a timer interval of one second might be a bit slow as this results in quite a bit of delay until a received code is decoded. Probably even if I'm now able to store several codes in the ring buffer it is too small if one only empties it only every second. But for demonstration purposes it should be Ok.

slaff:
> And polling with a timer interval of one second might be a bit slow

Can you make it 200ms and give it a try?

horde3d:
Sure. It works also with 200ms timer.
I'm not sure if there are some related changes in Sming that caused an issue I just found in the JVC send method. While I definetly had a working sample in the past, it first did not work when I've reduced the timer to 200ms. The problem was that as soon as I tried to send a JVC command, the application stopped. The reason was that the check for maximum in the IRremoteESP8266 library has a signedness problem. I currently fixed it by converting the value to int. But I just found out, that the problem has been also fixed upstream. So I may have to pull a later version of the IRremoteESP8266 library in.

slaff:
@horde3d do you think your PR is ready for testing from the other Sming users? Are the any pending things/issues?

horde3d:
Actually I haven't had the time to do many tests, but at least with a WROOM-02 ESP8266 and a simple breadboard circuit with an IR emitter and receiver it seems to work so far.

PR #1336, opened 2018-03-02
Fix/dht

frankdownunder:
The two new source files come from the Arduino libraries which according to the documentation are also compatible with ESP32.  The following changes were made

- added initialisation in new constructor
- removed the AUTO_DETECT option as testing showed it to be unreliable in my setup
- added new exampel code using a Timer

Tested on Wemos D1 mini DHT22 shield.

I have not (yet) removed the original source codes

### DIFF
--- a/Sming/Libraries/DHT/DHTesp.cpp
+++ b/Sming/Libraries/DHT/DHTesp.cpp
@@ -6,9 +6,9 @@
   - Auto detect sensor model
   - Very low memory footprint
   - Very small code
-  
+
   https://github.com/beegee-tokyo/DHTesp
-  
+
   which I think was previously :
   https://github.com/beegee-tokyo/arduino-DHTesp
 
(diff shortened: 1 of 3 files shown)


frankdownunder:
@slaff   OK I agree a submodule is better than having yet another copy of the code.
But,  I raised this issue  [AUTO_DETECT doesnt work after a "cold" start]( https://github.com/beegee-tokyo/DHTesp/issues/5) and now that leaves us with a decision - do we wait until the AUTO_DETECT problem is fixed by beegee@giesecke.tk? or make a patch to remove the AUTO_DETECT feature ( since it isnt reliable) or write a comment in the sample
Also, there are uninitialised variables in that code - do we worry about them also?

slaff:
@frankdownunder As with the other third party modules you can add a patch file that is responsible for the local modifications. The patch must have the same name as the directory where the module will be stored. You can check here for an example: https://github.com/SmingHub/Sming/tree/develop/Sming/Libraries/.patches.   You can use this document as a starting guide: https://github.com/SmingHub/Sming/blob/develop/AddingExternalSource.md. For the Arduino libraries due to the fact that they adhere to a more standardised folder structure you do not need to make any changes to the makefiles.

slaff:
@frankdownunder Here is a rebase (https://github.com/slaff/Sming/tree/rebase/pr1336) of your PR with changes relates to using a submodule. Can you please test that code on your hardware because I do not have DHT* sensors? You can test it using the following commands:

```
cd /tmp
git clone -b rebase/pr1336  https://github.com/slaff/Sming.git
cd Sming/Sming
export SMING_HOME=`pwd`
cd ../samples/Humidity_DHT22
make
make flash
```

You will notice that the following changes from your PR were removed:
- added initialisation in new constructor
- removed the AUTO_DETECT option as testing showed it to be unreliable in my setup

It would be much better for the DHTesp project and everyone using it if you submit and discuss those changes directly in the upstream project. The fewer modifications we have related to submodules the better.

frankdownunder:
@slaff  I have been doing more research to make sure I/we picked the right library to replace the one we have at the moment.

A question for you before we go ahead and add this library in. 
Is it philosophically "better" to use interrupts?  The DHT22 sends 5 characters (5 * 120us) after a bit of a handshake (approx 100us), and most libraries it would seem, just use a sort of polling technique to read them.
Further, during the time it takes to read those bytes, interrupts are actually disabled, as the code is timing critical, which strikes me as a bit unfriendly.
The main use case for the DHT22 is in a thermostat; I would have thought a thermostat is a fairly simple device - no requirement for it to be super-responsive.  

In addition to the four "official" Arduino libraries, we have this one, which uses interrupts:
	https://github.com/niesteszeck/idDHTLib
I havent had time to evaluate it, but I suspect it is using a better technique.
Thoughts?

BTW, I only have a DHT22 to test with.

frankdownunder:
@slaff I made some progress on the interrupt version, but I need to attend to some other targets first, and I will come back to it when I have more time.  
So I came back to testing the DHTesp library which you have added as a submodule.  I have tested it a lot, and it mostly works, BUT I am not prepared to say it is reliable.  Sometimes I get:
OK      Humidity: 2.00% Temperature: 1.00*C     HeatIndex-2.8
I can assure you in my part of the world it never ever gets that cold and dry.
I have researched whether the lack of initialisation in the constructor causes the problem or whether AUTO_DETECT might be implicated. Ive also looked at whether a "warm" start was helpful,  but nothing conclusive. Maybe I will instrument the code to try to find out what causes the failures. 

So... in the mean time, this library is an improvement on the original since it MOSTLY works.

PR #1349, opened 2018-03-18
fixed buggy reading of http body from stream resulting in incomplete …

verybadsoldier:
…body returned when size >1024 byte

For me sending HTTP POST requests with bodies greater than 1024 byte led to bodies cut to 1024.

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpRequest.cpp
+++ b/Sming/SmingCore/Network/Http/HttpRequest.cpp
@@ -157,7 +157,7 @@ String HttpRequest::getBody()
 	if(stream->available() != -1 && stream->getStreamType() == eSST_Memory) {
 		MemoryDataStream *memory = (MemoryDataStream *)stream;
 		char buf[1024];
-		for(int i=0; i< stream->available(); i += 1024) {
+		while(stream->available() > 0) {
 			int available = memory->readMemoryBlock(buf, 1024);
 			memory->seek(max(available, 0));
 			ret += String(buf, available);


Sming/SmingCore/Network/Http/HttpRequest.cpp
```diff
@@ -157,7 +157,7 @@ String HttpRequest::getBody()
 	if(stream->available() != -1 && stream->getStreamType() == eSST_Memory) {
 		MemoryDataStream *memory = (MemoryDataStream *)stream;
 		char buf[1024];
-		for(int i=0; i< stream->available(); i += 1024) {
+		while(stream->available()) {
```
slaff: Can you add a comparison for the value to be bigger than 0?

slaff:
> For me sending HTTP POST requests with bodies greater than 1024 byte led to bodies cut to 1024.

@verybadsoldier Can you paste in the comments also a sample application that demonstrates the issue?

verybadsoldier:
I don't have a minimal example at hand so I can only refer to our project where this problem popped up after upgrading from `3.5.0` to `3.5.1`:
https://github.com/verybadsoldier/esp_rgbww_firmware/blob/master/app/webserver.cpp#L264

When doing a `HTTP POST` with a body length of, for example, `1123` bytes then the function `request.getBody()` returns a `String` object with a length of only `1024` bytes.

According to my understanding, inside of the function `HttpRequest::getBody()` a call to `stream->available()` will return the number of bytes currently left for processing. So the loop will start, consume the first `1024` bytes, so variable `i` will be `1024`. So when checking the next `for`-loop iteration then `i` will be `1024` but a call to `stream->available()` will only return `99` (`1123 - 1024`) and the loop will exit, leaving the remaining `99` bytes unprocessed in the stream.

Also added `> 0`.

slaff:
Great, thanks!

PR #1356, opened 2018-04-04
Include directories in spiffy rom - reopened

onzo-mateuszzakarczemny:
I'm reopening #713 Finally I had time to implement `ENABLE_SPIFFS_TREE` flag which was requested in the previous PR. Sorry for the long delay.  If you have more requests for change, this time I should have more time. 

Original description
-------------
> ```
> Presently, spiffs does not support directories. It produces a flat structure. Creating a file with path tmp/myfile.txt will create a file called tmp/myfile.txt instead of a myfile.txt under directory tmp.
> ```
> spiffy doesn't support getting list of files in directory but is able to have files with slashes in names. There are a lot of cases where it is enough (eg separate javascript and css files on web server root). I modified spiffy.c to scan and include subdirectories. I tested it on my esp (build and flashed on windows) and files are visible. As I'm neither native c or c++ programmer please review my code.


slaff:
@onzo-mateuszzakarczemny Please, fix fist the reported issues on Codacy and in AppVeyor.

onzo-mateuszzakarczemny:
@slaff I'm not sure how to fix AppVeyor issue.
```spiffy.c:6:17: fatal error: ftw.h: No such file or directory```
It was working in previous old PR. Seems that there is no ftw on windows. Any idea how to fix that? Is there any way to add this library?

PR #1362, opened 2018-04-12
Moved RF24 to a submodule.

slaff:


slaff:
For the moment the upstream seems older than our version. If that changes we can re-open this PR.

PR #1384, opened 2018-05-16
SmingCore/Timer: Mark 'const' methods

ololoshka2871:
Timer methods `isStarted()`, `getIntervalUs()` and `getIntervalMs()` are marked with **const** modifier.


slaff:
@ololoshka2871 What is the issue that this PR solves?

PR #1389, opened 2018-05-21
Add support for Nextion displays.

### LINKED ISSUE #1372: Add support for Nextion displays.
Hi,
Thanks for such a great framework!
I've got this Nextion Display and I have managed to implement a few functions to use it (clear screen, show some text etc). I would like to contribute to Sming and Nextion displays could be a good option. 

Currently I am caught between writing own library from scratch or patching ITead's arduino library (using the patch mechanism in Sming).

Please share your opinions (even coding standards etc for Sming).

Thanks
Vijay

vijaygill:
Fixes #1372.

Hi @slaff,
Please review the PR and let me know.
About the proof, do I need to post a video of the example (added to samples - Nextion_Button) somewhere?

Regards
Vijay


slaff:
A short demo video on youtube will be enough.

slaff:
An one more thing: please, move the library from Sming/third-party to Sming/Libraries and remove the changes in the makefile. I will review the PR more carefully later this week.

vijaygill:
@slaff - A few questions:

1.  what is the criteria of putting third-party code in "third-party" or "Libraries".
2. Does patch system work for the code in "Libraries" too?
3. How do I include code for compilation if I cannot make changes to Make file?

Thanks
Vijay

slaff:
> what is the criteria of putting third-party code in "third-party" or "Libraries".

- [x] If this is an Arduino Library (at least the name says so) 
- [ ] and follows the Arduino standard (a file called library.properties is present in the root of the library) which unfortunately for this library is not the case. This means that the patch system should create a dummy `library.properties` file.
+ [x] and has all source code and headers either in its root or in its src/ sub-folder

>   Does patch system work for the code in "Libraries" too?

Yes, there is also `.patches` folder under Libraries

>  How do I include code for compilation if I cannot make changes to Make file?

The build system will include it automatically. Actually you will have less efforts to integrate it compared to "third-party".

vijaygill:
@slaff thanks for your reply. I will make the changes. I am not much of a c++ developer, so I just followed the patterns there to make things work :).
I will make changes and ask for another review.

Vijay

vijaygill:
Hi @slaff, I've made the changes and build works and example works too. I will post video on YouTube tomorrow as it's too late here (00:49 in Ireland).
I noticed one slight problem that my patch seems to be applied at-least twice and shows error second time but make moves on to build code successfully. Is there something else I need to look at?

Vijay

slaff:
> I noticed one slight problem that my patch seems to be applied at-least twice and shows error second time 

@vijaygill Please, add in the patch the creation of a file called `library.properties` in the root folder of the library that has a very simple content with the following line in it:

name=ITEAD Arduino_Nextion Library

> @slaff - here is the link to video on YouTube: Nextion display with ESP8266 using Sming

Awesome work! It would be nice if you can add the link with a bit of text as a comment block inside the sample application.cpp.  

Once those two small changes are ready I will merge the PR.

vijaygill:
Hi @slaff - added comment and updated patch to create library.properties in ITEADLIB_Arduino_Nextion submodule directory.

.gitmodules
```diff
@@ -77,3 +77,6 @@
 	path = Sming/Libraries/DHTesp
 	url = https://github.com/beegee-tokyo/DHTesp.git
 	ignore = dirty
+[submodule "Sming/third-party/ITEADLIB_Arduino_Nextion"]
+	path = Sming/Libraries/ITEADLIB_Arduino_Nextion
+	url = https://github.com/itead/ITEADLIB_Arduino_Nextion.git
```
slaff: @vijaygill Add one more line here with content:
ignore = dirty

similar to the other repositories.

vijaygill:
Hi @slaff, done.

PR #1410, opened 2018-07-01
Fix/mqtt protocol

frankdownunder:
We allow a URL of the form "mqtts://user:password@server:port"
but we (currently) never check the protocol bit


Sming/SmingCore/Network/MqttClient.cpp
```diff
@@ -77,6 +77,10 @@ bool MqttClient::setWill(const String& topic, const String& message, int QoS, bo
 bool MqttClient::connect(const URL& url, const String& clientName, uint32_t sslOptions)
 {
 	this->url = url;
+	if (not (url.Protocol == "mqtt" || url.Protocol == "mqtts")) {
+		debug_e("Invalid protocol");
```
slaff: @frankdownunder Maybe the error message should be a bit more descriptive and explain that only mqtt and mqtts protocols are allowed.

PR #1438, opened 2018-09-16
Progmem string enhancements

mikee47:
This PR differs from my description in the original V4 proposal. Using functions to return strings never really sat well with me, felt pretty naff but it got the job done. The new approach uses a FlashString structure which works much better. There's also a tweak to ArduinoJson.

I also talked about the mforce-l32 GCC compiler mod. Has anyone actually compiled Sming with this mod?


slaff:
> I also talked about the mforce-l32 GCC compiler mod. Has anyone actually compiled Sming with this mod?

I use it and once upon a time I thought of it as the magical fix for correct placement of strings. But there are some disadvantages like: Available only on some platforms; Using mforce with custom heap allocation based on ummalloc caused exceptions, etc.. The best would be to have a better compiler that takes better care to analyze and put static data in the right place but that is a wishful thinking for the moment.

mikee47:
Be aware I'm having issues testing this... I'll post an update later on

mikee47:
This probably needs a new sample application, or extending another (perhaps Basic_ProgMem?) for testing. Because of the macro renaming I'd got confused and ended up calling memcpy_aligned on a RAM buffer, so of course it went pop!

I've taken steps to make things a bit clearer, but it is somewhat complicated.

I originally called the structure FlashStringBuffer but then renamed to just FlashString, hence the changes from fsb -> fstr. I've used FSTR as a close parallel to PSTR, which I guess needs to be retained for Arduino compatibility.

The whole point of this PR is to make it as simple as possible to use, so thoughts on changes are definitely welcome! Sorry for the chopping and changing; I wanted to get an initial commit out there so you'd have something to look at.

Sming/SmingCore/Debug.cpp
```diff
@@ -18,20 +18,21 @@ DebugClass::~DebugClass()
 void DebugClass::initCommand()
 {
 #if ENABLE_CMD_EXECUTOR
-	commandHandler.registerCommand(CommandDelegate("debug", "New debug in development", "Debug",
+	auto debug = F("Debug");
```
slaff: Why you add a separate variable for the debug text that is used only once? Maybe for consistency it would be better to write `CommandDelegate(debug, F("New debug in development"), F("Debug"),`

Sming/Services/CommandProcessing/CommandHandler.cpp
```diff
@@ -15,6 +16,8 @@
 CommandHandler::CommandHandler()
 {
 	 registeredCommands = new HashMap<String, CommandDelegate>;
```
slaff: The indentation needs fixing.

Sming/SmingCore/Debug.cpp
mikee47: Sorry, that's an error - I mistook both strings to be the same, but case is different.

Sming/Services/CommandProcessing/CommandHandler.cpp
mikee47: Do coding standards apply to 'Services'? I left the indentation as I found it...

mikee47:
Another thought. I've been checking this against the Arduino code, and a signficant difference is in the implementation of F(). For Arduino, it provides a flash_string_t, but I recall changing it to a String to avoid the un-necessary strlen_P() call. I think most Arduino code will rely on the implicit conversion to a String anyway, but we could always call it something else, e.g. FS() to differentiate.

mikee47:
This PR includes (almost) all of the changes to WString. I've omitted the renaming of member variables with an underscore as that could be done as a separate step. For other modules it would help to do that first, perhaps as a separate commit; it would simplify the substantive changes and make problems easier to spot.

Sming/Services/CommandProcessing/CommandHandler.cpp
slaff: Line 18 and the following lines should have the same indent. In that specific case it seems that line 18 already had a wrong indentation. Just for consistency can you change line 18 to have the same indentation as line 19?

Sming/Services/CommandProcessing/CommandHandler.cpp
```diff
@@ -179,46 +184,50 @@ void CommandHandler::processCommandOptions(String commandLine  ,CommandOutput* c
...
+			if (commandToken[1] == _F("help"))
 			{
 				printUsage = true;
 			}
-			if (commandToken[1] == "verbose")
+			if (commandToken[1] == _F("verbose"))
 			{
 				commandHandler.setVerboseMode(VERBOSE);
-				commandOutput->printf("Verbose mode selected\r\n");
+			commandOutput->print(_F("Verbose mode selected\r\n"));
```
slaff: Indentation needs adjusting to be the same as the line above.

Sming/Services/CommandProcessing/CommandHandler.cpp
```diff
@@ -179,46 +184,50 @@ void CommandHandler::processCommandOptions(String commandLine  ,CommandOutput* c
...
 		case 3 :
-			if (commandToken[1] != "prompt")
+			if (commandToken[1] != _F("prompt"))
 			{
 				errorCommand = true;
 				break;
 			}
 			commandHandler.setCommandPrompt(commandToken[2]);
-			commandOutput->printf("Prompt set to : %s\r\n",commandToken[2].c_str());
+		commandOutput->print(_F("Prompt set to : "));
```
slaff: Indentation of lines [212 - 214] needs adjusting to be the same as the line above.

slaff:
@mikee47 Is the PR ready for testing on a real device?

> probably needs a new sample application, or extending another (perhaps Basic_ProgMem?)

Updating/Extending [Basic_ProgMem](https://github.com/SmingHub/Sming/tree/develop/samples/Basic_ProgMem) is the way to go. It will save us a lot of questions from users asking us what is the recommended way to use those new ProgMem enhancements.

mikee47:
@slaff I'm wary of diverging from the Arduino standards too much, but at the same time I think FlashString does bring some useful additions. I've done a quick'n'dirty timing check loading a 44-character flash string 200 times:
1. Load FlashString into stack buffer...T - Elapsed: 469
2. Load FlashString into String...T - Elapsed: 3090
3. Load PSTR into String...T - Elapsed: 5174

Times in microseconds, on standard 80MHz ESP12F; the difference between (2) and (3) will be for the strlen_P() call. Also gives another reason to use the stack!

I'll get started on [Basic_ProgMem](https://github.com/SmingHub/Sming/tree/develop/samples/Basic_ProgMem), hopefully have something out in the next day or two.

I'm finding this PR submission process quite useful as it's helping make further improvements to the main application I'm using to test all this; it won't build directly against the develop branch (yet) so I'm getting better at the whole 3-way-merge thing.

So yes, ready for testing on real hardware!

slaff:
> I'll get started on Basic_ProgMem, hopefully have something out in the next day or two.

Waiting for the Basic_ProgMem changes in order to merge the PR.

slaff:
Another thing that needs fixing is the coding style of the final sample. Just type:
```
cd %SMING_HOME%
make cs
```
And if this does not work for you then make sure to read the release notes related to coding style:
https://github.com/SmingHub/Sming/releases/tag/3.6.1

mikee47:
Whoops, forgot to set formatter in Eclipse. Hopefully all sorted now.

slaff:
@mikee47 Thanks again for the awesome PR! As a next step can you create a new PR with the HardwareTimer changes that were initially part of this PR?

PR #1440, opened 2018-09-18
Allow the use of multiple DS18S20 sensors

kmihaylov:
DS18S20 files changed so that if there are multiple instances of the DS18S20 class, each own has local instance of the OneWire class. This is to prevent false readings in case of using multiple DS18S20 sensors on different pins.

### DIFF
--- a/Sming/Libraries/DS18S20/ds18s20.h
+++ b/Sming/Libraries/DS18S20/ds18s20.h
@@ -127,6 +127,8 @@ private:
 
 
 	float celsius[MAX_SENSORS], fahrenheit[MAX_SENSORS];
+public:
+	OneWire* ds;
 };
 
 /** @} */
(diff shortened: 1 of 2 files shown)


Sming/Libraries/DS18S20/ds18s20.h
```diff
@@ -41,7 +41,7 @@ class DS18S20
 public:
     /** @brief  Instantiate a DS18S20 object
     */
-    DS18S20();
+    DS18S20(uint8_t);
```
slaff: The default value for the constructor should be here and not in the cpp file. You should move `#define DS1820_WORK_PIN 2 ` here, change the signature of the constructor to be 
```
DS18S20(uint8_t workPin = DS1820_WORK_PIN);
```

And in the CPP file to have the following for the constructor:

```
DS18S20::DS18S20(uint8_t workPin /* = DS1820_WORK_PIN */)
{
	ds = ...
```

PR #1442, opened 2018-09-20
Add ElapseTimer class, update HardwareTimer

mikee47:
Change hardware timer frequency calculation to account for current divisor setting; not safe to use hard-coded value.
Add usToTimerTicks() and timerTicksToUs() functions.


Sming/SmingCore/ElapseTimer.h
```diff
@@ -0,0 +1,41 @@
+/****
```
slaff: The ElapseTimer class has to prove itself. Move it for the moment as part of the sample. Later one we can make it part of the core if ElapseTimer is used a lot in other parts of the framework.

mikee47: I uses this in other places as a profiling aid, not so much in release code. It would be helpful to me if it could be more accessible; could it go in Services, perhaps?

slaff: > could it go in Services, perhaps?

Which reminds me that we should refactor Services and name it to something like "Extras" + Move libbase64 to third-party + move WebHelers to SmingCore, etc.

Ok, please move ElapsedTime to Services for now.

mikee47: I'll stick it in a folder called 'Profiling' ?

slaff: Ok, make sure the Profiling is also checked for coding style issues.

PR #1445, opened 2018-09-28
Allow callback execution for DS18S20 even in absence of connected sensor

kmihaylov:
Good day. I moved out
```	
numberOfread = 0;
StartReadNext();
```
from the else body in DS18S20::DoSearch(void) in order to force the program to execute the callback even if no DS1820 sensor was found. This way sensors can be attached or detached at any time. It is responsibility of the callback to distinguish if there is usable temperature reading or no.

Otherwise there is a risk of false initialization of the device with no sensor found (even if one is attached), if the 1-wire line is noisy.

All other changes to the file are blank lines removal and code formatting according to the Coding style rules (automatic code formatting from Eclipse with .clang-format).

### DIFF
--- a/Sming/Libraries/ArduinoJson/include/ArduinoJson/JsonObjectKey.hpp
+++ b/Sming/Libraries/ArduinoJson/include/ArduinoJson/JsonObjectKey.hpp
@@ -17,6 +17,28 @@ class JsonObjectKey {
   JsonObjectKey(const char* key) : _value(key), _needs_copy(false) {}
   JsonObjectKey(const String& key) : _value(key.c_str()), _needs_copy(true) {}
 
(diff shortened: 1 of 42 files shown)


kmihaylov:
Should be OK now, formatting preserved.

slaff:
At the current moment Sming and also the Arduino libraries assume that all devices are connected before starting to use them. If we want to allow some sort of hot-plug then we need to create the needed infrastructure for that, which can, but is not limited to:
* adds interrupts that get triggered if there is detected change on PIN, or/and combination of PINS
* allows setting of callbacks to be triggered when a pin change is detected

Therefore that PR will be closed for now and can be re-opened later once we have that infrastructure.

PR #1446, opened 2018-09-28
Microsoft Visual SourceSafe status files added to .gitignore

tius2000:
Microsoft Visual SourceSafe creates status files named "vssver2.scc" that should be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,6 @@ nbproject
 GTAGS
 GRTAGS
 GPATH
+
+# Microsoft Visual SourceSafe files
 vssver2.scc
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ nbproject
 GTAGS
 GRTAGS
 GPATH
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -14,3 +14,4 @@ nbproject
 GTAGS
 GRTAGS
 GPATH
+vssver2.scc
```
slaff: @tius2000 Please put a comment line above this one saying "Microsoft Visual SourceSafe files". Ex:

```
# Microsoft Visual SourceSafe files
vssver2.scc
```

slaff: It will be great if you can add also comments for the other type of files. For example Eclipse IDE files, NetBeans IDE files, etc.

tius2000: I'm not sure about the other file types @slaff, so I just added a comment line for Visual SourceSafe.

slaff: Great! Thanks!

PR #1450, opened 2018-09-30
Remove m_print dependency on serial port.

mikee47:
* There's no apparent reason for the dependency.
* Change output callback to handle multiple characters instead of just one (efficiency).
* Output routines return character count, so caller can determine if output was actually performed. This is for consistency with return values from m_printf, etc. and also in support of a forthcoming uart driver update which includes extended transmit buffering and interrupt servicing.
* Change default behaviour for m_printf() so it doesn't output anything, i.e. no callback set.
* Add m_printHex() function, mainly as a debugging aid.

### DIFF
--- a/Sming/SmingCore/HardwareSerial.h
+++ b/Sming/SmingCore/HardwareSerial.h
@@ -197,9 +197,20 @@ public:
 	 *  @param  oneChar Character to write to the serial port
 	 *  @retval size_t Quantity of characters written (always 1)
 	 */
-	size_t write(uint8_t oneChar);
+	size_t write(uint8_t oneChar) override
+	{
+		return uart_write_char(uart, oneChar);
+	}
 
-	using Stream::write;
+	/** @brief  write multiple characters to serial port
+	 *  @param buffer data to write
+	 *  @param size number of characters to write
+	 *  @retval size_t Quantity of characters written, may be less than size
+	 */
+	size_t write(const uint8_t* buffer, size_t size) override
+	{
+		return uart_write(uart, buffer, size);
+	}
 
 	/** @brief  Configure serial port for system debug output and redirect output from debugf
 	 *  @param  enabled True to enable this port for system debug output
--- a/Sming/SmingCore/HardwareSerial.h
+++ b/Sming/SmingCore/HardwareSerial.h
@@ -197,17 +197,19 @@ public:
 	 *  @param  oneChar Character to write to the serial port
 	 *  @retval size_t Quantity of characters written (always 1)
 	 */
-	size_t write(uint8_t oneChar) override
+	virtual size_t write(uint8_t oneChar)
 	{
 		return uart_write_char(uart, oneChar);
 	}
+
+	using Stream::write;
 
 	/** @brief  write multiple characters to serial port
 	 *  @param buffer data to write
 	 *  @param size number of characters to write
 	 *  @retval size_t Quantity of characters written, may be less than size
 	 */
-	size_t write(const uint8_t* buffer, size_t size) override
+	virtual size_t write(const uint8_t* buffer, size_t size)
 	{
 		return uart_write(uart, buffer, size);
 	}
--- a/Sming/system/include/m_printf.h
+++ b/Sming/system/include/m_printf.h
@@ -62,21 +62,17 @@ static inline size_t m_puts(const char* str)
 }
 
 #ifdef __cplusplus
-#define M_PRINTHEX_BYTESPERLINE = 16
-#else
-#define M_PRINTHEX_BYTESPERLINE
-#endif
 
 /** @brief output a block of data in hex format
- *  @param tag brief name to display with the data block
+ *  @param tag brief name to display with the data block. Specify nullptr if not required.
  *  @param data
  *  @param len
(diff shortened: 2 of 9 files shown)


mikee47:
@slaff For info, I've got more changes to HardwareSerial than just the System.deferCall thing - there's a whole update to the uart driver.

Sming/system/include/m_printf.h
```diff
@@ -9,16 +9,66 @@ Descr: embedded very simple version of printf with float support
...
+ *  @retval number of characters written, which may be less than the requested size
+ *  @note data does not need to be nul terminated and may contain any 8-bit values including nul
+ */
+typedef size_t (*nputs_callback_t)(void* param, const char* str, size_t length);
+
+/** @brief set the character output routine
+ *  @param callback
+ *  @param param
+ */
+void m_setPuts(nputs_callback_t callback, void* param);
```
slaff: @mikee47 When you have time can you check if there is any performance impact if in here we use std::function instead of good-old C callback? The std::function implementation will allow us to call not only functions and anon. functions but also object methods.

Sming/system/include/m_printf.h
```diff
@@ -9,16 +9,66 @@ Descr: embedded very simple version of printf with float support
...
+#endif
+
+/** @brief output a block of data in hex format
+ *  @param tag brief name to display with the data block
+ *  @param data
+ *  @param len
+ *  @param bytesPerLine If non-zero, data will be output in block separated by carriage return
+ *  @note intended for debugging
+ */
+void m_printHex(const char* tag, const void* data, size_t len, size_t bytesPerLine M_PRINTHEX_BYTESPERLINE);
```
slaff: Just thinking out loud: Maybe we should get rid of the HexDump class and start using only this function to reduce the redundancy.

Sming/system/uart.cpp
```diff
@@ -156,23 +156,27 @@ void uart_stop_isr(uart_t* uart)
...
-        return;
+        return 0;
     }
     while((USS(uart->uart_nr) >> USTXC) >= 0x7f);
     USF(uart->uart_nr) = c;
+    return 1;
 }
 
-void uart_write(uart_t* uart, const char* buf, size_t size)
+size_t uart_write(uart_t* uart, const void* buf, size_t size)
```
slaff: What is the reason behind changing the `const char* buf` to `const void* buf` and casting it after that? Is there any useful use-case for it?

mikee47: Because serial port can be used for non-ascii data it means caller doesn't need to cast. Also no cast required to deal with signed/unsigned char arrays.

For example:

struct SomethingStruct mydata = {.... };
uart_write(uart, &mydata, sizeof(mydata)); // No cast required

Sming/system/include/m_printf.h
mikee47: @slaff Runs fine, new commit so you can have a look.

Sming/system/include/m_printf.h
mikee47: I see that Hexdump outputs addresses, where as m_printHex doesn't. It could be added as another parameter. Also, if we restricted use of m_printHex to C++ code it would simplify the header, and we'd have the option of overloading it for more complex use cases.

```

/** @brief output a block of data in hex format
 *  @param tag brief name to display with the data block
 *  @param data
 *  @param len
 *  @param addr Prefix lines with addresses starting at the given value, use -1 if not required.
 *  @param bytesPerLine If non-zero, data will be output in block separated by carriage return
 *  @note intended for debugging
 */
void m_printHex(const char* tag, const void* data, size_t len, int addr = -1, size_t bytesPerLine = 16);
```

Sming/SmingCore/HardwareSerial.h
```diff
@@ -197,9 +197,20 @@ class HardwareSerial : public Stream
 	 *  @param  oneChar Character to write to the serial port
 	 *  @retval size_t Quantity of characters written (always 1)
 	 */
-	size_t write(uint8_t oneChar);
+	size_t write(uint8_t oneChar) override
+	{
+		return uart_write_char(uart, oneChar);
+	}
 
-	using Stream::write;
```
slaff: This change breaks the compatibility because 
size_t write(const char *str);
size_t write(const char *buffer, size_t size);

are not available any longer. Make sure to add back `using Stream::write;`

Sming/system/include/m_printf.h
slaff: Can add in this PR also the suggested changes?

mikee47: OK, I've revised m_printHex and tweaked it to get a more typical hex / ASCII output.
Tested using this snippet:

```
	static DEFINE_FSTR(test, "This is a some test data \1\2\3 Not all ASCII\0\0\0Anything, really\0Can I go home now?");
	LOAD_FSTR(data, test);
	m_printHex("TEST1", data, test.length());
	m_printHex("TEST2", data, test.length(), 0);
	m_printHex("TEST3", data, test.length(), 0xFFFFF9, 8);
	m_printHex(nullptr, data, test.length(), 0xFFFFF9, 0);
```

PR #1452, opened 2018-10-01
Fix/string concat

### LINKED ISSUE #1451: Strange bug with Strings
@mikee47 Strangely enough the following code produces incorrect result:

```
http_method_str(request->method) + String(" ") + request->uri.getPathWithQuery() + " HTTP/1.1\r\n"
```

Instead of the complete HTTP request line I am getting just "HTTP/1.1\r\n". The issue is happening after your recent PRs were merged (related to the String class). Can you check if you can reproduce it or it is just me and my compiler with `mforce-l32`?

mikee47:
Fix inconsistent behaviour in String::concat(). Fixes #1451

* Result String should only be invalidated if a memory allocation error occurs
* An invalid string is considered equivalent to an empty string

### DIFF
--- a/Sming/Wiring/WString.cpp
+++ b/Sming/Wiring/WString.cpp
@@ -276,8 +276,8 @@ bool String::concat(const String &s)
 bool String::concat(const char *cstr, unsigned int length)
 {
   unsigned int newlen = len + length;
-  if (!cstr) return false;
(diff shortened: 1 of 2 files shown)


mikee47:
@slaff Think I've made a mess of this PR, didn't base on develop. Do you want it redone or can you handle?

PR #1455, opened 2018-10-03
Optimise HashMap to use pass-by-reference where possible.

mikee47:
* When used with Strings, most operations result in duplicate objects being created. Changing methods to use references (const or otherwise) avoids this and improves performance considerably for lookups and modifications.
* Requires change to comparator callback arguments - should provide significant performance improvement as callback invoked multiple times during a search - see HttpHeaders class.
* Change use of int to unsigned where appropriate
* Initialise member variables directly, rather than in constructor (except for _nil_ - could be a primitive or an object)

### DIFF
--- a/Sming/Wiring/WHashMap.h
+++ b/Sming/Wiring/WHashMap.h
@@ -117,6 +117,11 @@ class HashMap
     || | An indexer for accessing and assigning a value to a key
     || | If a key is used that exists, it returns the value for that key
     || | If there exists no value for that key, a nil value is returned
+    || |
+    || | Note that if the HashMap object is not const, the non-const version
+    || | of this operator will be called which will create a default value
+    || | for this key. If that behaviour is not desired, then check for the
+    || | existence of the key first, using either contains() or indexOf().
     || #
     ||
     || @parameter key the key to get the value for
--- a/Sming/Wiring/WHashMap.h
+++ b/Sming/Wiring/WHashMap.h
@@ -15,6 +15,17 @@
 ||
 */
 
+/*
+ * @author: 3 Oct 2018 - mikee47 <mike@sillyhouse.net>
+ *
+ * Modified to use references (const or otherwise) to avoid object copies when used with classes, e.g. String.
+ *
+ * Note that if the value is a primitive then setNullValue should be called to provide a default value to be
+ * used when adding a new unspecified entry, or if a key value is not present. This should not be necessary
+ * for object values as the default constructor will be used.
+ *
+ */
+
 #ifndef HASHMAP_H
 #define HASHMAP_H
 
@@ -24,7 +35,7 @@ template<typename K, typename V>
 class HashMap
 {
   public:
-    typedef bool (*comparator)(K, K);
+    typedef bool (*comparator)(const K&, const K&);
 
     /*
     || @constructor
@@ -33,13 +44,8 @@ class HashMap
     ||
     || @parameter compare optional function for comparing a key against another (for complex types)
     */
-    HashMap(comparator compare = 0)
+    HashMap(comparator compare = nullptr) : cb_comparator(compare)
     {
-      cb_comparator = compare;
-      currentIndex = 0;
-      size = 0;
-      keys = NULL;
-      values = NULL;
     }
 
     ~HashMap()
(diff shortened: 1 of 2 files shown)


Sming/Wiring/WHashMap.h
```diff
@@ -199,86 +218,85 @@ class HashMap
...
-          }
-        }
-        else
-        {
-          if (key == *keys[i])
-          {
-            return true;
-          }
-        }
+        *keys[i - 1] = *keys[i];
```
slaff: Can you try the following test code with your new modifications?

```
	HttpHeaders headers;
	headers["X-Auth-Header"] = "Value Data";
	headers["x-Auth-header"] = "Value Data New";

	debugf("Header: %s", headers["X-AUTH-HEADER"].c_str());

	headers.remove("x-auth-header");

	debugf("Headers count: %d", headers.count());

```

Expected result:
```
Header: Value Data New
Headers count: 0
```

mikee47: That is weird. The comparator doesn't get set... not entirely sure why but I'm happier having a default constructor anyway. Thanks for spotting that one!

slaff: Looks better now.

> I think the compiler decided to call the HashMap constructor with the default of nullptr instead of the inherited default constructor. Compiler bug? Or am I missing something?

My guess is that it has nothing to do with the compiler but a confirmation will require disassembling of the compiled binary and analyzing  it. Anyway you can try reverting the constructor code back to `HashMap(comparator compare=0)` instead of ``HashMap(comparator compare=nullptr)  +  comparator cb_comparator = nullptr` and check if it works.

slaff: > I think the compiler decided to call the HashMap constructor with the default of nullptr instead of the inherited default constructor. 

That can be checked easier with a debugger.

Sming/Wiring/WHashMap.h
```diff
@@ -91,16 +116,18 @@ class HashMap
...
-    || | If there exists no value for that key, the key is added
+    || | If there exists no value for that key, a nil value is returned
     || #
     ||
     || @parameter key the key to get the value for
     ||
     || @return The const value for key
     */
-    const V& operator[](const K key) const
+    const V& operator[](const K& key) const
```
slaff: Another test comes to my mind:
```
       HttpHeaders headers;

	debugf("Non-Existent Header: [%s]", headers["Non-Existent"]);

	headers["X-Auth-Header"] = "Value Data";
	headers["x-Auth-header"] = "Value Data New";

	debugf("Header: %s", headers["X-AUTH-HEADER"].c_str());

	headers.remove("x-auth-header");

	debugf("Headers count: %d", headers.count());
```

If I am not mistaken the expected output with those changes should be:
```
Header: Value Data New

Headers count: 0
```
right? 
If the answer is no - then add a comment describing this
if the answer is yes - check if you are getting the expected result and if not make sure to fix it :)

mikee47: Answer is no, because headers is not a const object.

```
	HttpHeaders headers;

	const HttpHeaders& constHeaders = headers;
	debugf("Non-Existent Header (const): \"%s\"", constHeaders["Non-Existent"].c_str());
	debugf("  count: %d", headers.count());

	debugf("Non-Existent Header (non-const): \"%s\"", headers["Non-Existent"].c_str());
	debugf("  count: %d", headers.count());

	headers["X-Auth-Header"] = "Value Data";
	debugf("set(\"X-Auth-Header\")");
	debugf("  count: %d", headers.count());

	headers["x-Auth-header"] = "Value Data New";
	debugf("set(\"x-Auth-Header\")");
	debugf("  count: %d", headers.count());

	debugf("Header[\"X-AUTH-HEADER\"]: \"%s\"", headers["X-AUTH-HEADER"].c_str());
	debugf("  count: %d", headers.count());

	headers.remove("x-auth-header");
	debugf("headers.remove(\"x-auth-header\")");
	debugf("  count: %d", headers.count());
```

```
1916918089 Non-Existent Header (const): ""
1916918089   count: 0
1916918089 Non-Existent Header (non-const): ""
1916918089   count: 1
1916918089 set("X-Auth-Header")
1916918089   count: 2
1916918089 set("x-Auth-Header")
1916918089   count: 2
1916918089 Header["X-AUTH-HEADER"]: "Value Data New"
1916918089   count: 2
1916918089 headers.remove("x-auth-header")
1916918089   count: 1
```

slaff: Perfect, thanks!

PR #1456, opened 2018-10-04
Optimise url escaping / unescaping

mikee47:
http-parser - patch update
* place tokens[] into flash memory and use get_token() function instead - doesn't rely on mforce32 compiler support to save RAM
* Use unhex() function in stringutils.h, comment out unhex[] array
* Comment-out http_errno_name() - replaced with httpGetErrnoName() defined in HttpCommon.cpp. Note the replacement returns a String, not a const char*
* Replace m_printf() with SYSTEM_ERROR()

WebHelpers/escape
* Use hexchar() function instead of static array - saves a little RAM but also safer and avoids code duplication, at the expense of a little speed
* Use unhex() function defined in stringutil.h - also used by http-parser
* Add must_escape() function to remove duplication of code
* Add uri_escape() and uri_unescape() methods to work with Strings, simplifying user code and taking care of buffer allocations

HttpBodyParser
* Replace magic numbers -1 and -2 with definitions
* Simplify and eliminate String copy operations using uri_unescape_inplace() and revised HashMap to modify string content in-situ

UrlEncodedOutputStream
* Simplify

HttpCommon
* Add httpGetErrnoName(), httpGetErrnoDescription() and httpGetStatusText() functions - saves RAM by storing text strings in flash

HttpServerConnection
* getStatus() method removed, superseded by httpGetStatusText() function. The _message_ parameter is now a const String&, instead of const char*. As the default value is nullptr, if () statement works as expected.

### DIFF
--- a/Sming/Services/WebHelpers/escape.cpp
+++ b/Sming/Services/WebHelpers/escape.cpp
@@ -6,13 +6,10 @@
 #include <ctype.h>
 #include <math.h>
 
-/* WARNING : none of the code has been tested properly */
-
-static const char tohex[] = "0123456789ABCDEF";
-
+// Append str to dest with checks
 static unsigned safe_append(char *dest, size_t len, const char *str) {
 	unsigned len2;
-	len2=strlen((const char*)str);
+	len2=strlen(str);
 	if(len2>len) {
 		return 0; /* refuse to append */
 	}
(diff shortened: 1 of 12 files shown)


slaff:
@mikee47 As much as I like this PR I would like to ask you to split it in two PRs:
* One containing the url-un/escaping changes only.
* Another one on top of the first one containing the other changes in this PR.

The reason for more focused PRs is that if it is needed I can revert easily a problematic PR. In your case this PR contains two useful improvements but if one of them does not work and the other one is great we will have to revert both or make hot-fixing like the string-concat issue #1452.

mikee47:
@slaff I knew I was trying it on a bit, being lazy! I have reprimanded myself and shall start afresh...

PR #1457, opened 2018-10-04
Improve uri escaping / unescaping

mikee47:
http-parser - patch update
* Use unhex() function in stringutils.h, comment out unhex[] array
* Replace m_printf() with SYSTEM_ERROR()

WebHelpers/escape
* Use hexchar() function instead of static array - saves a little RAM but also safer and avoids code duplication, at the expense of a little speed
* Use unhex() function defined in stringutil.h - also used by http-parser
* Add must_escape() function to remove duplication of code
* Add uri_escape() and uri_unescape() methods to work with Strings, simplifying user code and taking care of buffer allocations

HttpBodyParser
* Replace magic numbers -1 and -2 with definitions
* Simplify and eliminate String copy operations using uri_unescape_inplace() and revised HashMap to modify string content in-situ

UrlEncodedOutputStream
* Simplify


Sming/Services/WebHelpers/escape.cpp
```diff
@@ -219,16 +159,21 @@ char *uri_unescape(char *dest, size_t dest_len, const char *src, int src_len)
...
-			case '<': ret+=4; break; /* &lt; */
-			case '>': ret+=4; break; /* &gt; */
-			case '&': ret+=5; break; /* &amp; */
-			case '"': ret+=6; break; /* &quot; */
-			case '\'': ret+=6; break; /* &apos; */
-			default:
-				ret++;
-		}
+	for(ret=0;len && *s;len--, s++) {
+		char c = *s;
```
slaff: What is the advantage of replacing the switch statement with if statement?

Sming/third-party/.patches/http-parser.patch
```diff
@@ -153,41 +153,48 @@ index 5b5657b..e7a2b3d 100644
...
-         0,       0,       0,       0,       0,       0,       0,       0,
- /*   8 bs     9 ht    10 nl    11 vt    12 np    13 cr    14 so    15 si   */
-@@ -217,7 +227,7 @@ static const char tokens[256] = {
+@@ -217,6 +227,7 @@ static const char tokens[256] = {
         'x',     'y',     'z',      0,      '|',      0,      '~',       0 };
  
  
--static const int8_t unhex[256] =
-+static const int8_t unhex[256] PROGMEM_L32 =
++/*
```
slaff: Remove the commented code.

Sming/Services/WebHelpers/escape.cpp
```diff
@@ -237,28 +182,47 @@ unsigned html_escape_len(const char *s, size_t len) {
...
-			case '\'':
-				i+=safe_append(dest+i,len-i, "&apos;");
-				break;
-			default:
-				dest[i++]=*s;
-		}
+	unsigned i=0;
+	for(;i<len && *s;s++) {
+		char c = *s;
+		if (c == '<')
```
slaff: Advantages of replacing switch with if-and-elses?

Sming/Services/WebHelpers/escape.cpp
mikee47: With embedded stuff I tend to use if because the resulting code is more predictable. In this case I probably did it because it makes the code a little more readable (less verbose). I doubt it will affect code performance one way or the other here.

Sming/Services/WebHelpers/escape.cpp
slaff: Ok, then please put back the switch statements.

Issue #1173, opened 2017-07-06
List of objects that can be moved to flash


gbl08ma:
I have noticed that static data RAM usage, mainly rodata, has been increasing a lot in more recent Sming versions.

This is a list of constant objects that, in principle, can be annotated to never leave the flash:

- [ ] [**http_parser** `tokens`](https://github.com/nodejs/http-parser/blob/335850f6b868d3411968cbf5a4d59fe619dee36f/http_parser.c#L189) (256 bytes)
- [ ] [**http_parser** `unhex`](https://github.com/nodejs/http-parser/blob/335850f6b868d3411968cbf5a4d59fe619dee36f/http_parser.c#L224) (256 bytes)
- [ ] [**http_parser** `normal_url_char`](https://github.com/nodejs/http-parser/blob/335850f6b868d3411968cbf5a4d59fe619dee36f/http_parser.c#L243) (32 bytes)
- [ ] [**http_parser** `http_strerror_tab`](https://github.com/nodejs/http-parser/blob/335850f6b868d3411968cbf5a4d59fe619dee36f/http_parser.c#L465) (264 bytes) - alternatively, can we get rid of this entirely or make it a compile-time option? Most people probably don't need these human-friendly errors on a constrained platform as the ESP8266.

Note that for some of those, there might be alignment issues when reading from flash. I'm not sure, but I think there are problems reading from non-4-byte-aligned addresses in flash. In that case, moving arrays of 8 and 16 bit types to flash might be problematic. They could be converted to arrays of 32 bit elements which are then casted to the original types, but that will waste some flash space (four times the space they take in RAM). There are more advanced solutions (e.g. always reading 4 bytes at a time, then using bitwise operations to retrieve the correct byte), but this is slower and requires too much changes to third-party code.

I'll update this list if I find more large constants that can eventually be moved to the flash.

slaff:
Fixed with the merge of PR #1459.

PR #1459, opened 2018-10-05
Move HTTP strings into flash and provide suitable functions

mikee47:
http-parser - patch update
* place tokens[] into flash memory and use get_token() function instead - doesn't rely on mforce32 compiler support to save RAM
* Remove http_errno_name() - replaced with httpGetErrnoName() defined in HttpCommon.cpp. Note the replacement returns a String, not a const char*

HttpCommon
* Add httpGetErrnoName(), httpGetErrnoDescription() and httpGetStatusText() functions - saves RAM by storing text strings in flash

HttpServerConnection
* getStatus() method removed, superseded by httpGetStatusText() function. The _message_ parameter is now a const String&, instead of const char*. As the default value is nullptr, if () statement works as expected.

FlashString.h
* Add FSTR_TABLE macro

Samples/Basic_Progmem
* Add demo code for use of FSTR_TABLE

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpCommon.cpp
+++ b/Sming/SmingCore/Network/Http/HttpCommon.cpp
@@ -1,0 +1,69 @@
+/****
+ * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
+ * Created 2015 by Skurydin Alexey
+ * http://github.com/SmingHub/Sming
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ * @author: 2018 - Mikee47 <mike@sillyhouse.net>
+ *
+ * 	httpGetErrnoName() and httpGetStatusText() functions added
+ *
+ ****/
+
+#include "HttpCommon.h"
+
+// Define flash strings and lookup table for HTTP error names
+#define XX(_n, _s) static DEFINE_FSTR(hpename_##_n, "HPE_" #_n);
+HTTP_ERRNO_MAP(XX)
+#undef XX
+
+static FSTR_TABLE(hpeNames) = {
+#define XX(_n, _s) FSTR_PTR(hpename_##_n),
+	HTTP_ERRNO_MAP(XX)
+#undef XX
+};
+
+String httpGetErrnoName(enum http_errno err)
+{
+	if(err > HPE_UNKNOWN)
+		return F("HPE_#") + String(err);
+
+	return *hpeNames[err];
+}
+
+// Define flash strings and lookup table for HTTP error descriptions
+#define XX(_n, _s) static DEFINE_FSTR(hpedesc_##_n, _s);
+HTTP_ERRNO_MAP(XX)
+#undef XX
+
+static FSTR_TABLE(hpeDescriptions) = {
+#define XX(_n, _s) FSTR_PTR(hpedesc_##_n),
+	HTTP_ERRNO_MAP(XX)
+#undef XX
+};
+
+String httpGetErrnoDescription(enum http_errno err)
+{
+	if(err > HPE_UNKNOWN)
+		return F("HPE_#") + String(err);
+
+	return *hpeDescriptions[err];
+}
+
+// Define flash strings for HTTP status codes
+#define XX(_num, _name, _string) static DEFINE_FSTR(hpsText_##_num, #_string);
+HTTP_STATUS_MAP(XX)
+#undef XX
+
+String httpGetStatusText(enum http_status code)
+{
+	switch(code) {
+#define XX(_num, _name, _string)                                                                                       \
+	case _num:                                                                                                         \
+		return hpsText_##_num;
+		HTTP_STATUS_MAP(XX)
+#undef XX
+	default:
+		return F("<unknown_") + String(code) + '>';
+	}
+}
--- a/Sming/SmingCore/Network/Http/HttpCommon.cpp
+++ b/Sming/SmingCore/Network/Http/HttpCommon.cpp
@@ -6,7 +6,7 @@
  *
  * @author: 2018 - Mikee47 <mike@sillyhouse.net>
  *
- * 	httpGetErrnoName() and httpGetStatusText() functions added
+ * 	httpGetErrorName(), httpGetErrorDescription() and httpGetStatusText() functions added
  *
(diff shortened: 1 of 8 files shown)


mikee47:
Test code:

```
#include "Network/Http/HttpCommon.h"

void testHttpCommon()
{
	for (int i = 0; i < 100; ++i) {
		auto err = static_cast<http_errno>(i);
		debug_i("httpError(%d) = \"%s\", \"%s\"", i, httpGetErrnoName(err).c_str(),
			httpGetErrnoDescription(err).c_str());
	}

	for (int i = 100; i < 550; ++i) {
		debug_i("http_status(%d) = \"%s\"", i, httpGetStatusText(i).c_str());
	}

}
```

samples/Basic_ProgMem/app/TestProgmem.cpp
```diff
@@ -131,6 +131,23 @@ void testFSTR(Print& out)
...
+
+	static DEFINE_FSTR(fstr1, "Test string #1");
+	static DEFINE_FSTR(fstr2, "Test string #2");
+
+	static FSTR_TABLE(table) = {
+			FSTR_PTR(fstr1),
+			FSTR_PTR(fstr2),
+	};
+
+	// Table entries may be accessed directly as they are word-aligned
```
slaff: Make sure to run `make cs` and commit again.

```
!!! samples/Basic_ProgMem/app/TestProgmem.cpp not compliant to coding style, here is the fix:
--- samples/Basic_ProgMem/app/TestProgmem.cpp	2018-10-05 08:17:28.813803129 +0000
+++ /dev/fd/63	2018-10-05 08:17:34.313872466 +0000
@@ -131,15 +131,14 @@
 	TEST(demoFSTR1 == String(demoFSTR2))
 #undef TEST
 
-
 	// FSTR table
 
 	static DEFINE_FSTR(fstr1, "Test string #1");
 	static DEFINE_FSTR(fstr2, "Test string #2");
 
 	static FSTR_TABLE(table) = {
-			FSTR_PTR(fstr1),
-			FSTR_PTR(fstr2),
+		FSTR_PTR(fstr1),
+		FSTR_PTR(fstr2),
 	};
 
 	// Table entries may be accessed directly as they are word-aligned
```

Sming/SmingCore/Network/Http/HttpServerConnection.cpp
```diff
@@ -505,34 +505,20 @@ void HttpServerConnection::onError(err_t err)
 	TcpClient::onError(err);
 }
 
-const char* HttpServerConnection::getStatus(enum http_status code)
```
slaff: Don't remove `HttpServerConnection::getStatus` because you will break compatibility. Leave it and use`httpGetStatusText` internally.

Sming/SmingCore/Network/Http/HttpCommon.cpp
```diff
@@ -0,0 +1,69 @@
...
+HTTP_ERRNO_MAP(XX)
+#undef XX
+
+static FSTR_TABLE(hpeNames) = {
+#define XX(_n, _s) FSTR_PTR(hpename_##_n),
+	HTTP_ERRNO_MAP(XX)
+#undef XX
+};
+
+String httpGetErrnoName(enum http_errno err)
```
slaff: Mandatory: Rename this to httpGetErrorName because that is what this function does.

slaff: Needs further discussion: One other thing that is coming to my mind is that maybe we should think about removing the `http` prefix from those functions and replace it with a C++ namespace.

Sming/SmingCore/Network/Http/HttpServerConnection.cpp
mikee47: Tricky one that, since getStatus() returns const char*. All we could do there is leave the current implementation as-is and deprecate it. Plus, it should be a static method.

Sming/SmingCore/Network/Http/HttpCommon.cpp
mikee47: I've a similar treatment for WebConstants.h, which lives in the ContentType namespace.
That also involves a change from const char* to String return value, which raises the same issue of compatibility.

samples/Basic_ProgMem/app/TestProgmem.cpp
mikee47: Weird. The travis run on my repo didn't pick this up!

Sming/SmingCore/Network/Http/HttpCommon.cpp
mikee47: How about ```namespace HttpText``` ?

samples/Basic_ProgMem/app/TestProgmem.cpp
mikee47: I set travis up on my github account so it could pre-check the commit before submitting it. Same file, here's the result:
```The command "if [ "$SDK_VERSION" == "2.0.0" ]; then ../.travis/tools/clang/format-pr.sh; fi" exited with 0.```

Not sure how to go about debugging that - it shouldn't need any additional settings to get it working?

File: Sming/SmingCore/Network/Http/HttpServerConnection.cpp (hunk 2 above)
Comment (mikee47) [reply Sming/SmingCore/Network/Http/HttpServerConnection.cpp]: I'll leave as-is then.

File: Sming/SmingCore/Network/Http/HttpCommon.cpp (hunk 3 above)
Comment (mikee47) [reply Sming/SmingCore/Network/Http/HttpCommon.cpp]: OK

File: samples/Basic_ProgMem/app/TestProgmem.cpp (hunk 1 above)
Comment (slaff) [reply samples/Basic_ProgMem/app/TestProgmem.cpp]: Just run locally the following commands

```
cd %SMING_HOME%
make cs # or in the worst cases open the file with very basic text editor and fix the indentation.
git add -u 
git commit -m "Fixed coding style issues"
git push
```

mikee47: Yep, I need to remember to run ```make cs``` ! Thanks.

mikee47:
Closes #1173

PR #1460, opened 2018-10-05
Move ContentType text into flash and revise functions

mikee47:
* fromFileExtension() and fromFullFileName() return String instead of const char*
* Code moved into new file, WebConstants.cpp
* Bugfix: File extension matching is now done on entire extension, not just the start; this caused a mis-interpretation of 'json' as 'js'.
* MIME_TYPE_MAP performs a 1:1 mapping between file extension and MIME type. For HTML files the extension has been changed to .html as this appears to be the generally preferred format. The alternative, .htm, is checked in the code when mapping to MIME type.
* toString() returns an invalid String so the caller can use if() to check for validity. The function previously returned "<unknown>", which is a rather strange string to be embedding in an HTML header.

### DIFF
--- a/Sming/SmingCore/Network/WebConstants.cpp
+++ b/Sming/SmingCore/Network/WebConstants.cpp
@@ -26,7 +26,7 @@ static FSTR_TABLE(mimeStrings) = {
 MIME_TYPE_MAP(XX)
 #undef XX
 
-static FSTR_TABLE(extStrings) = {
+static FSTR_TABLE(extensionStrings) = {
 #define XX(_name, _ext, _mime) FSTR_PTR(extstr_##_name),
 	MIME_TYPE_MAP(XX)
 #undef XX
@@ -38,8 +38,8 @@ String fromFileExtension(const char* extension)
 	if(strcasecmp(extension, _F("htm")) == 0)
 		return mimestr_HTML;
 
-	for(unsigned i = 0; i < ARRAY_SIZE(extStrings); ++i) {
-		if(*extStrings[i] == extension)
+	for(unsigned i = 0; i < ARRAY_SIZE(extensionStrings); ++i) {
+		if(*extensionStrings[i] == extension)
 			return *mimeStrings[i];
 	}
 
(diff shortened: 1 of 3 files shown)


slaff:
@mikee47 What is your non-mforce-l32 compiler saying about the free heap improvements?

mikee47:
Going by the travis builds, and using the HttpServer_WebSockets sample, free RAM goes from 49239 to 49507 for this PR.

mikee47:
For the past few PRs we have free RAM / IRAM as:
Fix/String_concat 44768 / 7934
feature/optimise-hashmap 44763 / 7934
feature/uri-escaping 44495 / 7934
feature/http-strings 49239 / 7934
feature/improve-WebConstants 49507 / 7934

The uri-escaping result is weird; goes down then goes backup up again afterwards!

mikee47:
I discovere

slaff:
> ... PROGMEM_L32 attribute on tokens[] (line 195), but the next PR replaced it with PROGMEM.

Yes, I saw that one but did not bother you because the next PR was already suggesting proper fix.

Sming/SmingCore/Network/WebConstants.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+	return *mimeStrings[m];
+}
+
+String fromFullFileName(const char* fileName)
+{
+	if(!fileName)
+		return nullptr;
+
+	const char* pExt = strrchr(fileName, '.');
```
slaff: Just rename the variable to `extension`. Try to use more descriptive names. I guess `p` is a prefix for "pointer" but try to avoid it.

Sming/SmingCore/Network/WebConstants.cpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+String fromFileExtension(const char* extension)
+{
+	// We accept 'htm' or 'html', but the latter is preferred
+	if(strcasecmp(extension, _F("htm")) == 0)
+		return mimestr_HTML;
+
+	for(unsigned i = 0; i < ARRAY_SIZE(extStrings); ++i) {
+		if(*extStrings[i] == extension)
```
slaff: More descriptive name, please.

PR #1461, opened 2018-10-06
Encapsulate encoding and decoding of HTTP header fields

mikee47:
* Move HttpHeaders from structures.h into a separate module, and expand
* Replace common HTTP header field names with enumerated type - eliminates errors due to mis-typing, simplifies code
* Replace instances of NULL with nullptr as appropriate in any updated files
* Wrap constant string data with _F() or F() as appropriate

Methods no longer used:
* HttpResponse::hasHeader()
* HttpRequest::getHeader()


mikee47:
I've held-off making any changes to Samples (HttpClient, HttpClient_Instapush, HttpServer_ConfigNetwork, MeteoControl, but can add those if you want.

Sming/SmingCore/Data/HttpHeaders.cpp
```diff
@@ -0,0 +1,51 @@
...
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ * @author: 2018 - Mikee47 <mike@sillyhouse.net>
+ *
+ ****/
+
+#include "HttpHeaders.h"
+
+// Define field name strings and a lookup table
+#define XX(_tag, _str, _comment) static DEFINE_FSTR(hhfnStr_##_tag, _str);
```
slaff: Can you replace the `hhfnStr_` prefix with `HTTP_HEADER_` ?

Sming/SmingCore/Data/HttpHeaders.h
```diff
@@ -0,0 +1,147 @@
...
+/*
+ * Common HTTP header field names. Enumerating these simplifies matching
+ * and generating headers. The strings themselves are stored in flash to
+ * save RAM.
+ *
+ * According to RFC 2616: 4.2, field names are case-insensitive.
+ *
+ */
+#define HTTP_HEADER_FIELDNAME_MAP(XX)                                                                                  \
+	XX(AccessControlAllowOrigin, "Access-Control-Allow-Origin", "")                                                    \
```
slaff: Replace the first element to be upper-case words separated with underscore. This way it will be quite natural to set the header to `HTTP_HEADER_AUTHORIZATION` or `HTTP_HEADER_ACCESS_CONTROL_ALLOW_ORIGIN`.

slaff:
> Clearly using the enum for lookups is a lot faster so I may revisit this at some point.

We should not forbid or make it very complicated to set custom headers. If there is an elegant way to use enum as lookup and also allow custom headers I will be keen to see it (Can be done also as a separate PR ).

slaff:
> HttpClient, HttpClient_Instapush, HttpServer_ConfigNetwork, MeteoControl, but can add those if you want.

As long as the changes are related to the main goal of that PR then go ahead. Because the samples are the first thing that new developers copy and modify we should give them a very good start by applying the best practises in using Sming and C++.

Sming/SmingCore/Data/HttpHeaders.cpp
```diff
@@ -0,0 +1,51 @@
+/****
```
slaff: A better location of the HttpHeaders.* files will be under `Sming/SmingCore/Network/Http/`. Please, move them there.

PR #1466, opened 2018-10-08
Remove leading underscore from class variables

mikee47:


Sming/SmingCore/Data/CStringArray.h
```diff
@@ -80,16 +80,16 @@ class CStringArray : protected String
...
 	}
 
 	unsigned count() const
 	{
-		return count_;
+		return stringCount;
 	}
 
 private:
-	unsigned count_ = 0;
```
slaff: @mikee47 Are there any other places in the code where you succeeded to "smuggle" underscores in the variable names?

mikee47: Wiring/FlashString.h. I've updated the coding style page as suggested to clarify.

PR #1472, opened 2018-10-09
Move EspDigitalPins instantiation out of ESP8266EX.h

mikee47:
Fixes issue where each module using EspDigitalPins gets their own copy - currently Digital.cpp and HardwarePWM.cpp.

Effect can be demonstrated by adding `pullup(2);` to Basic_HwPWM sample, then running `objdump -x out\build\app.out | grep EspDigital`:

```
3ffe87a4 l     O .rodata        000000c0 _ZL14EspDigitalPins
3ffe8864 l     O .rodata        000000c0 _ZL14EspDigitalPins
```


slaff:
> Does making this const cause problems with mforce32 compiler?

I don't think so but have to check on real device to be sure. 

> add DRAM_ATTR... although this isn't yet defined. .iram.text? .iram.literal?

There is IRAM_ATTR that will instruct the compiler to put the code in .iram.text. What is the DRAM_ATTR for?

mikee47:
It's part of the ESP32 IDF, basically same job as IRAM_ATTR but for data. Here it might be a good idea to implement even if it only tags the data as 'might be accessed from ISR'. Specifically this would apply to global/static `const` data only.

slaff:
> I may be over-complicating this... just thinking aloud.

Out of the box the data will go to the .dram section. If you look at the [memory map](https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map)  that is already the case for
```
3ffe87a4 l     O .rodata        000000c0 _ZL14EspDigitalPins
```

Unless you instruct the compiler to move it to flash, as is the case with the flash strings in the recent changes. In that case the developer should be aware of the fact that inside an interrupt handler that data might not be available because all other interrupts may be blocked meaning also reading from the flash will not work.

PR #1463, opened 2018-10-06
Improve HttpHeaders using HashMap(enum, String)

mikee47:
Revision of HttpHeaders to use enum instead of String for key parameter.

* Custom field names handled using CStringArray
* Add assert() statements to HashMap::operator[] methods to check index bounds

PR #1473, opened 2018-10-10
Add method which should have been part of #1461 and #1463

mikee47:
* Comparison accomplished using stack, avoids allocating temporary String variable on heap


slaff:
@mikee47 Can you test this PR on real device? And let me know if it works for you.

mikee47:
@slaff Yes, already tested and working.

slaff:
> Are you having issues?

Thanks for checking. I don't have a device right now but will double check later today.

PR #1475, opened 2018-10-10
Implement global task queue

mikee47:
* Move out of HardwareSerial into SystemClass, making queue available for applications and other framework code.
* SDK has two priority queues, this is a simple implementation of one of them. If the situation arises where some tasks need to be elevated then a second task queue could be added, but this would consume more RAM. The API can be extended with a second parameter to support this.

### DIFF
--- a/Sming/SmingCore/Interrupts.cpp
+++ b/Sming/SmingCore/Interrupts.cpp
@@ -5,13 +5,20 @@
  * All files of the Sming Core are provided under the LGPL v3 license.
  ****/
 
-#include "../SmingCore/Interrupts.h"
-#include "../SmingCore/Digital.h"
-#include "../Wiring/WiringFrameworkIncludes.h"
+#include "Interrupts.h"
+#include "Digital.h"
+#include "WiringFrameworkIncludes.h"
+#include "System.h"
 
-InterruptCallback _gpioInterruptsList[16] = {0};
-Delegate<void()> _delegateFunctionList[16];
-bool _gpioInterruptsInitialied = false;
+static InterruptCallback _gpioInterruptsList[16] = {0};
+static InterruptDelegate _delegateFunctionList[16];
+static bool _gpioInterruptsInitialied = false;
+
+/** @brief  Interrupt handler
+ *  @param  intr_mask Interrupt mask
+ *  @param  arg pointer to array of arguments
+ */
+static void interruptHandler(uint32 intr_mask, void* arg);
 
 void attachInterrupt(uint8_t pin, InterruptCallback callback, uint8_t mode)
 {
@@ -38,7 +45,7 @@ void attachInterrupt(uint8_t pin, Delegate<void()> delegateFunction, GPIO_INT_TY
 {
 	if(pin >= 16)
 		return; // WTF o_O
-	_gpioInterruptsList[pin] = NULL;
+	_gpioInterruptsList[pin] = nullptr;
 	_delegateFunctionList[pin] = delegateFunction;
 	attachInterruptHandler(pin, mode);
 }
@@ -48,7 +55,7 @@ void attachInterruptHandler(uint8_t pin, GPIO_INT_TYPE mode)
 	ETS_GPIO_INTR_DISABLE();
 
 	if(!_gpioInterruptsInitialied) {
-		ETS_GPIO_INTR_ATTACH((ets_isr_t)interruptHandler, NULL); // Register interrupt handler
+		ETS_GPIO_INTR_ATTACH((ets_isr_t)interruptHandler, nullptr); // Register interrupt handler
 		_gpioInterruptsInitialied = true;
 	}
 
@@ -61,7 +68,7 @@ void attachInterruptHandler(uint8_t pin, GPIO_INT_TYPE mode)
 
 void detachInterrupt(uint8_t pin)
 {
-	_gpioInterruptsList[pin] = NULL;
+	_gpioInterruptsList[pin] = nullptr;
 	_delegateFunctionList[pin] = nullptr;
 	attachInterruptHandler(pin, GPIO_PIN_INTR_DISABLE);
 }
@@ -120,18 +127,27 @@ static void IRAM_ATTR interruptHandler(uint32 intr_mask, void* arg)
 	do {
 		gpio_status = GPIO_REG_READ(GPIO_STATUS_ADDRESS);
 		processed = false;
-		for(uint8 i = 0; i < ESP_MAX_INTERRUPTS; i++, gpio_status << 1) {
-			if((gpio_status & BIT(i)) && (_gpioInterruptsList[i] || _delegateFunctionList[i])) {
-				//clear interrupt status
-				GPIO_REG_WRITE(GPIO_STATUS_W1TC_ADDRESS, gpio_status & BIT(i));
-
-				if(_gpioInterruptsList[i])
-					_gpioInterruptsList[i]();
-				else if(_delegateFunctionList[i])
-					_delegateFunctionList[i]();
-
-				processed = true;
+		for(uint8 i = 0; i < ESP_MAX_INTERRUPTS; i++) {
+			if(!bitRead(gpio_status, i)) {
+				continue;
 			}
+
+			// clear interrupt status
+			GPIO_REG_WRITE(GPIO_STATUS_W1TC_ADDRESS, gpio_status & _BV(i));
+
+			if(_gpioInterruptsList[i]) {
+				_gpioInterruptsList[i]();
+			} else if(_delegateFunctionList[i]) {
+				System.queueCallback(
+					[](uint32_t interruptNumber) {
+						auto& delegate = _delegateFunctionList[interruptNumber];
+						if(delegate)
+							delegate();
+					},
+					i);
+			}
+
+			processed = true;
 		}
 	} while(processed);
 }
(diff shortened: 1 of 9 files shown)


mikee47:
I've built and run the Basic_Serial sample and all appears well.

Sming/SmingCore/Platform/System.cpp
```diff
@@ -6,14 +6,21 @@
...
 SystemClass System;
 
-SystemClass::SystemClass()
+os_event_t SystemClass::taskQueue[TASK_QUEUE_LENGTH];
+
+/** @brief OS calls this function which invokes user-defined callback
+ *  @note callback function pointer is placed in event->sig, with parameter
+ *  in event->par.
+ */
+static void __taskHandler(os_event_t* event)
```
slaff: Remove the underscores.

Sming/SmingCore/Platform/System.h
```diff
@@ -109,14 +123,25 @@ class SystemClass
      */
 	void onReady(ISystemReadyHandler* readyHandler);
 
+	/**
+	 * @brief Queue a deferred callback.
+	 * @param callback The function to be called
+	 * @param param Parameter passed to the callback
+	*/
+	static __forceinline bool IRAM_ATTR deferCallback(task_callback_t callback, os_param_t param = 0)
```
slaff: Can you rename the method to `queueCallback` because that it is what it does?

Sming/SmingCore/Platform/System.h
```diff
@@ -12,12 +12,24 @@
...
+#ifndef TASK_QUEUE_LENGTH
+#define TASK_QUEUE_LENGTH 10
+#endif
+
+/** @brief Task callback function type
+ * 	@ingroup event_handlers
+ * 	@note Callback code does not need to be in IRAM
+ *  @todo Integrate delegation into callbacks
+ */
+typedef void (*task_callback_t)(os_param_t param);
```
slaff: Can you replace the C-style callback with C++ std::function?

slaff:
> I've built and run the Basic_Serial sample and all appears well.

If you want try it with Basic_rBoot too.

Sming/SmingCore/Platform/System.cpp
```diff
@@ -30,11 +32,20 @@ void SystemClass::initialize()
 	state = eSS_Intializing;
 
 	// Initialise the global task queue
-	system_os_task(__taskHandler, USER_TASK_PRIO_1, taskQueue, TASK_QUEUE_LENGTH);
+	system_os_task(taskHandler, USER_TASK_PRIO_1, taskQueue, TASK_QUEUE_LENGTH);
 
 	system_init_done_cb(staticReadyHandler);
 }
 
+bool SystemClass::queueCallback(TaskCallback callback, os_param_t param)
```
slaff: Maybe you can add priority - a third optional parameter to queueCallback with default value USER_TASK_PRIO_1 ?

mikee47: If I do that, I'd need to check the parameter and fail it. I could create the second task queue on-the-fly, but should also include some way of checking we're not in interrupt context. Alternatively I could add another method to SystemClass to create the elevated task queue. Or I could include the code to create the second task queue using `#define`. It's getting a bit complicated, which is why I'm thinking just leave it until the need arises? It wouldn't affect any existing code...

slaff: > which is why I'm thinking just leave it until the need arises?

Ok, that's fine.

mikee47:
@slaff Regarding the size of the task queue, I just kept it at 10 entries from the HardwareSerial usage, but I've never actually bothered checking how full the queue gets. I've an app which uses it extensively, so I'll add a counter to track the maximum number of requests, give it a thrashing and see what happens...

slaff:
> so I'll add a counter to track the maximum number of requests, give it a thrashing and see what happens...

I will wait for the results.

mikee47:
> 
> > so I'll add a counter to track the maximum number of requests, give it a thrashing and see what happens...
> 
> I will wait for the results.

I'd not been able to get the maximum task count above 2. Clearly the tasks are getting invoked immediately my function returns control to the SDK, which is as expected. To increase that number I have to deliberately queue multiple items in a single function. I'd be suspicious of code which does that since you'd roll these up into a single task anyway, so I think the limit of 10 is fine.

mikee47:
@slaff My implementation using std:function uses heap, because we need to take a copy of it. I agree std:function is more flexible but it gets called from interrupt context so heap allocation v. bad...

mikee47:
I'll update Samples for this PR. Basic_Interrupt is a good candidate.

slaff:
> My preference would be to go back to function pointer and use lambda where necessary to wrap method calls, unless you have any ideas?

Sounds reasonable. Ok, let's do it this way. Plus make sure to add an example how to call an object method, if needed, using lambda.

Sming/SmingCore/Interrupts.cpp
```diff
@@ -5,13 +5,51 @@
...
 
-#include "../SmingCore/Interrupts.h"
-#include "../SmingCore/Digital.h"
-#include "../Wiring/WiringFrameworkIncludes.h"
+#include "Interrupts.h"
+#include "Digital.h"
+#include "WiringFrameworkIncludes.h"
+#include "System.h"
+
+static InterruptCallback _gpioInterruptsList[16] = {0};
```
slaff: Remove the underscores, please.

Sming/SmingCore/Interrupts.cpp
```diff
@@ -5,13 +5,51 @@
...
+static InterruptDelegate _delegateFunctionList[16];
+static bool _gpioInterruptsInitialied = false;
```
slaff: Underscores.

Sming/SmingCore/Platform/System.cpp
```diff
@@ -11,15 +11,26 @@
...
  *  @note callback function pointer is placed in event->sig, with parameter
  *  in event->par.
  */
-static void taskHandler(os_event_t* event)
+void SystemClass::taskHandler(os_event_t* event)
 {
 	auto callback = reinterpret_cast<TaskCallback*>(event->sig);
 	if(callback) {
+		/*
+		 * @todo how to guarantee atomic operation?
```
slaff: Atomic operations are possible in ESP8266. You can take a look here: 
https://github.com/wizard97/Embedded_RingBuf_CPP/blob/9e091e9632bc3eba513f422ee7cb17895c565a1a/RingBufHelpers.h#L24

 I remember also using [GCCs __cleanup](http://echorand.me/site/notes/articles/c_cleanup/cleanup_attribute_c.html) but have to scratch my head a bit harder figure out which was the project...

slaff: https://github.com/wizard97/SimplyAtomic/blob/master/esp8266.h

samples/Gesture_APDS-9960/app/application.cpp
```diff
@@ -38,11 +38,11 @@ void IRAM_ATTR handleGesture()
 	}
 }
 
-void IRAM_ATTR interruptRoutine()
```
slaff: Without the IRAM_ATTR the function will be stored in Flash. If that function is called during interrupt is there  a guarantee that it would be always possible to load the function machine code from flash?

mikee47: With the modifications all `InterruptDelegate` callbacks are handled via the task queue, in the same context as any other regular code, so there's no special handling required.

slaff: >  I've attempted to explain this in the Basic_Interrupts sample.

Thanks, that is exactly what I was looking for :)

Sming/SmingCore/Interrupts.cpp
```diff
@@ -5,13 +5,51 @@
...
+static void IRAM_ATTR interruptHandler(uint32 intr_mask, void* arg)
+{
+	boolean processed;
+	uint32 gpio_status;
 
-InterruptCallback _gpioInterruptsList[16] = {0};
-Delegate<void()> _delegateFunctionList[16];
-bool _gpioInterruptsInitialied = false;
+	do {
+		gpio_status = GPIO_REG_READ(GPIO_STATUS_ADDRESS);
```
slaff: Coding style: `gpio_status` => `gpioStatus`.

Sming/SmingCore/Platform/System.cpp
mikee47: Thanks. It's a bit heavy! I was hoping there was an intrinsic to do a simple decrement rather than have to disable interrupts... I'm inclined to just leave it as-is, it's only for debugging and system info so not critical...

Sming/SmingCore/Platform/System.cpp
```diff
@@ -22,7 +22,11 @@ void SystemClass::taskHandler(os_event_t* event)
 {
 	auto callback = reinterpret_cast<TaskCallback>(event->sig);
 	if(callback) {
+		// If we get interrupt during adjustment of the counter, do it again
```
slaff: make cs

```
!!! Sming/SmingCore/Platform/System.cpp not compliant to coding style, here is the fix:
--- Sming/SmingCore/Platform/System.cpp	2018-10-11 17:53:14.397090476 +0000
+++ /dev/fd/63	2018-10-11 17:53:19.857040330 +0000
@@ -25,7 +25,7 @@
 		// If we get interrupt during adjustment of the counter, do it again
 		uint8_t oldCount = taskCount;
 		--taskCount;
-		if (taskCount != oldCount - 1)
+		if(taskCount != oldCount - 1)
 			--taskCount;
 		callback(event->par);
 	}
```

PR #1481, opened 2018-10-11
Change HardwareTimer to use NMI (Non-Maskable Interrupts)

mikee47:
Prevents other interrupts causing timing jitter.

* Move trivial code into header to improve performance
* Fix #include guard naming
* Change return type for stop() to void - meaningless
* Change SYSTEM_ERROR() to emit using debug_e so it uses flash instead of RAM for format strings
* Remove default parameter values - no point

Samples/Basic_Debug

* Change to use HardwareTimer and add a note about using SimpleTimer and Timer

### DIFF
--- a/Sming/SmingCore/HardwareTimer.cpp
+++ b/Sming/SmingCore/HardwareTimer.cpp
@@ -64,23 +64,27 @@ typedef enum {
 	NMI_SOURCE = 1,
 } FRC1_TIMER_SOURCE_TYPE;
 
-static void IRAM_ATTR hw_timer_isr_cb(void* arg)
+// Used by ISR
+static HardwareTimer* isrTimer;
+
+static void IRAM_ATTR hw_timer_isr_cb()
 {
-	if(arg == null)
-		return;
-	HardwareTimer* ptimer = (HardwareTimer*)arg;
-	ptimer->call();
+	if(isrTimer) {
+		isrTimer->call();
+	}
 }
 
 HardwareTimer::HardwareTimer()
 {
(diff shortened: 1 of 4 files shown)


slaff:
The code looks good but did you have the chance to test it on real-life example where slight inaccuracy can be noticed easily?

mikee47:
> The code looks good but did you have the chance to test it on real-life example where slight inaccuracy can be noticed easily?

Yes! I made this change a couple of months ago to improve the reliability of the RFSwitch library (rewrite of RCSwitch). It's deployed in a client's house to automate smart light switches, which use 433MHz receivers, plus other connected items (hence the IO Control library to provide a common interface to it all). 

The control packet for such an RF device is only 24 bits, but must be repeated a number of times to deal with random interference, warm-up timing variations, receiver settling, etc. With a standard priority interrupt (the current version of HardwareTimer) it only takes a serial interrupt to kill the packet. To get reliable results I had to repeat the packets around 80 times.

With the new version, the HardwareTimer interrupt will occur even if in the middle of servicing another interrupt, so the timing accuracy is much better. Because of this the system operates reliably with only 45 repeats.

It's a pity the ESP8266 doesn't have any proper Output Compare circuitry as that would guarantee accurate results regardless of interrupt activity. Ditto for input capture. But of course it makes up for all that by sheer speed!

slaff:
> With the new version, the HardwareTimer interrupt will occur even if in the middle of servicing another interrupt, so the timing accuracy is much better. 

Ok, sounds. Let's see what the other developers can confirm that.

PR #1487, opened 2018-10-13
Split DataSourceStream contents into separate files

mikee47:
Test only. For a reason I cannot fathom, the travis build in my fork fails the Basic_SSL sample with missing sntp_xx functions at link stage, but only for SDK 2.0. I've been unable to reproduce the problem on my dev system. So I need to see if the build still fails on the main repo (I've seen querks like this before).

This is only the initial commit for sorting the streams out. No code changes have been made, just rearranging content and tidying #includes.


mikee47:
Yes, failed here too.

PR #1490, opened 2018-10-15
Rewrite flashmem driver to cope with misaligned RAM buffers

mikee47:
spiffs_check just hangs without this. No exception raised.
Required to support random read/write. Note SPIFFS uses block accesses in normal operation so this problem wouldn't normally present.


Sming/system/flashmem.c
```diff
@@ -1,102 +1,205 @@
...
+/*
+ * To ensure memory alignment a temporary buffer is used by flashmem_read and flashmem_write functions.
+ *
+ * The buffer must be an integer multiple of INTERNAL_FLASH_WRITE_UNIT_SIZE.
+ */
+#define FLASH_BUFFERS 32
+
+/** @brief determines if the given value is aligned to a word (4-byte) boundary */
+#undef IS_ALIGNED
+#define IS_ALIGNED(_x) (((uint32_t)(_x)&0x00000003) == 0)
```
slaff: Rename _x to x

Sming/system/flashmem.c
```diff
@@ -1,102 +1,205 @@
...
-uint32_t flashmem_read( void *to, uint32_t fromaddr, uint32_t size )
+// For release builds, bypass internal flash read routines and go direct to SDK
+#ifdef DEBUG_BUILD
+
+#define FLASHMEM_READ flashmem_read_internal
+#define FLASHMEM_WRITE flashmem_write_internal
+
+#else
+
+static inline uint32_t _flashmem_write_internal(const void* from, uint32_t toaddr, uint32_t size)
```
slaff: `_flashmem_write_internal` -> `flashmem_write_internal` .

Sming/system/flashmem.c
```diff
@@ -1,102 +1,205 @@
...
-    for( i = 0; size && ( i < rest ); i ++, size --, pto ++ )
-      *pto = tmpdata[ i ];
-  }
-  return ssize;
+	SpiFlashOpResult r = spi_flash_write(toaddr, (uint32*)from, size);
+	return (SPI_FLASH_RESULT_OK == r) ? size : 0;
 }
 
-bool flashmem_erase_sector( uint32_t sector_id )
+static inline uint32_t _flashmem_read_internal(void* to, uint32_t fromaddr, uint32_t size)
```
slaff: `_flashmem_read_internal` -> `flashmem_read_internal`. Make appropriate fixes If there are also other places in this PR where the name of the function ot the variables start or end with underscore.

Sming/system/flashmem.c
```diff
@@ -1,102 +1,205 @@
...
+		for(i = 0; i < rest; ++i)
+			*pto++ = tmpdata[i];
+	}
+
+	return size;
+}
+
+bool flashmem_erase_sector(uint32_t sector_id)
+{
+	WDT_FEED();
```
slaff: Is `WDT_FEED()` really needed? I am just curious if you have experienced any issues with flash secor erasing causing the software watchdog to kick in?

slaff: This question applies also to the other new lines where WDT_FEED() is called.

mikee47: I haven't had any issues, but then this was always in the code as `WRITE_PERI_REG(0x60000914, 0x73);`. I had to dig around to find out what that actually did, turned out there was a useful definition for it.

Sming/system/flashmem.c
mikee47: OK, but we'll have to leave `_flash_code_end` though

Sming/system/flashmem.c
mikee47: I think the main risk from watchdog timeouts is with `spiffs_format`. Erase/write operations for normal file access would normally be much shorter, unlikely enough to trip a watchdog.

Sming/system/flashmem.c
slaff: > OK, but we'll have to leave _flash_code_end though

Yes, that's correct.

Sming/system/flashmem.c
mikee47: My previous statement about 3ms, ignore please, it's wrong! I spent a little more time on this today and put timing checks on `fileWrite`. Writes take an average of about 10us per byte, reads no more than 1us. Unlikely to write more than 10K in one operation (RAM limitations) so, with overheads, maybe 150ms for a fileWrite, 15ms for a fileRead.

Given these figures, the WDT seems un-necessary. If we wanted to remove them we'd need to take a closer look at `spiffs_format` and `spiffs_check`. Everything else is done with buffering and callbacks so should be fine. For now, probably easier to leave them in.

mikee47:
> @mikee47 Is there any performance penalty coming with that PR?

I've timed various fileRead() and fileWrite() operations using multiple read/writes of the same file, and averaged the values. There is no observable difference between them.

Here's a simple test you can try to illustrate why the revisions are necessary. Try reading some data from the start of flash memory:

```
	  char buffer[32];
	  auto result = flashmem_read(&buffer[3], 0, 16);
	  debug_i("flashmem_read() returned %u", result);
```

Because of the misaligned memory buffer the old driver just hangs.

slaff:
> Because of the misaligned memory buffer the old driver just hangs.

Perfect, thanks. I will try to test it myself next week and if there is no performance drawback will merge the PR.

PR #1491, opened 2018-10-16
Added code that allows programmatically to manage the phy data.

slaff:
The initial physical data is used initialize an ESP8266 chip and controls its behaviour. Sometimes you may want to change that data for a certain ROM on the device without changing it for all ROMs on the device. 
This PR introduces a new feature that allows programmatically to control and change the initial physical data.

That option is not enabled by default but can be enabled with ENABLE_CUSTOM_PHY=1 directive.

Closes #985.

### DIFF
--- a/Sming/system/esp8266_phy.c
+++ b/Sming/system/esp8266_phy.c
@@ -22,14 +22,14 @@
 
 #include <user_config.h>
 
-#ifndef F_CRYSTAL
-#define F_CRYSTAL 40000000
+#ifndef ICACHE_RAM_ATTR
+#define ICACHE_RAM_ATTR __attribute__((section(".iram.text")))
 #endif
 
-static const uint8_t phyInitData[128] =
+static const uint8_t ICACHE_FLASH_ATTR phyInitData[128] =
 {
     [0] = 5,  // Reserved, do not change
-    [1] = 8,  // Reserved, do not change
+    [1] = 0,  // Reserved, do not change
     [2] = 4,  // Reserved, do not change
     [3] = 2,  // Reserved, do not change
     [4] = 5,  // Reserved, do not change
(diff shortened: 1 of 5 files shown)


slaff:
@tius2000 Any comments or feedback from your side?

Sming/system/esp8266_phy.c
```diff
@@ -0,0 +1,268 @@
...
+
+int __attribute__((weak)) get_adc_mode()
+{
+	return 33;
+}
+
+int ICACHE_FLASH_ATTR __attribute__((weak)) custom_register_chipv6_phy(uint8_t* initData)
+{
+	if (initData != NULL) {
+	  memcpy(initData, phyInitData, sizeof(phyInitData));
```
mikee47: Shouldn't this be memcpy_P?

slaff: I don't think that is needed but will test it later just to be sure.

tius2000:
@slaff : I do not understand the details, but I think that is great idea!

piperpilot:
I will give this a +1 as I didn't realize that the phy data was different between SDKs.  I am migrating to 3.0 to fix some stability issues with wifi.  So far I haven't seen any major problems with OTA updating to a version of my app on the 3.0 SDK, but I can also say it hasn't really fixed some of the issues I'm seeing.  I would say updating the PHY with SDK moving forward would be somewhat required for anyone that has used SMING for a period of time.

mikee47:
@slaff I've been trying to get analogue reading working so tried patching PHY to see if that helped (it doesn't, probably broken hardware). Anyway, I've got this ported onto current develop but not sure best way to submit it.

slaff:
The problem with this PR is that it works only with the latest SDK. If we want to use it for 1.5.4 or 2.0.0 it will not work. There are different tricks to intercept reading or writing to special area on the flash memory but unfortunately I was not able to find a solution that works on all supported SDKs. 

We can eventually say that this will work only with SDK >= 3.0.0.

> Anyway, I've got this ported onto current develop but not sure best way to submit it

Submit your code as a new PR and I will close this one.

mikee47:
OK, thanks.

mikee47:
You might find this funny. I had signal connected to D0 instead of A0 on my nodeMCU. Shoot me now :-(

mikee47:
@slaff With #2171 changing the PHY could be done by adding a build variable with the partition name for the PHY initialisation data (default would be `phy_init`) so it gets picked up in `user_pre_init`. Would that be sufficient to achieve the goals for this PR?

(NB. This still leaves the issue of how to construct the PHY data but with a HEX editor and the [Official configuration guide](https://www.espressif.com/sites/default/files/documentation/esp8266_phy_init_bin_parameter_configuration_guide_en.pdf) not too difficult. If there's a need could be handled using a python script.)

mikee47:
Closing in favour of #2830

PR #1492, opened 2018-10-16
Add makeHexString() function

mikee47:
* Simplifies and eliminates duplicate code
* More efficient than using sprintf, etc.


Sming/Wiring/HexString.cpp
```diff
@@ -0,0 +1,29 @@
+/****
```
slaff: Can you move the HexString.* files to SmingCore/Data/ ? I would prefer to keep in the Wiring directory mostly classes coming from and/or improving [Wiring](http://wiring.org.co/).

PR #1493, opened 2018-10-16
Upgrade serial port driver

mikee47:
Revised driver does not (should not) break backward compatibility; default behaviour is unchanged.

* Tidy #includes. A side effect of this is that some Arduino libraries (ITEADLIB_Arduino_Nextion) use sprintf so stdio.h has been added to ArduinoCompat.h.
* Rewrite driver to incorporate transmit buffering in a similar manner to receive buffering. This change allows serial protocols such as RS485 to be implemented without requiring hardware to switch transfer direction.
* Add transmit buffer empty callback to HardwareSerial - users may implement alternative buffering schemes, e.g. direct stream to serial port.
* Driver can operate in low-memory mode with no additional buffers, just running on 127-byte hardware FIFOs, or with dynamically resizeable buffers
* Interrupt handler is shared between both serial ports, so deals with that issue
* Add option flags to allow user to customise behaviour. Currently this defines whether uart_write waits on buffer full or returns immediately with short count. This allows a large transmit buffer to be used with minimal impact on system performance whilst capturing debug information.
* Move some code from HardwareSerial (e.g. findChar) into uart driver
* Update Basic_Serial sample application


slaff:
@mikee47 I like this PR a lot ! Did you have the time to test it extensively on a real device?

mikee47:
I haven't tested the CommandDelegate stuff, by the way.

mikee47:
> I haven't tested the CommandDelegate stuff, by the way.

CommandProcessing_Debug works fine :-)

Sming/SmingCore/HardwareSerial.cpp
```diff
@@ -162,129 +98,101 @@ void HardwareSerial::systemDebugOutput(bool enabled)
...
-	return uart_get_baudrate(uart);
-}
-
-HardwareSerial::operator bool() const
-{
-	return uart != nullptr;
-}
-
-size_t HardwareSerial::indexOf(char c)
-{
```
mikee47: I want to fix the return type of this method, but that requires changing it in `Stream` as well. These are the only two classes affected in Sming, but it's potentially a BC. What say you?

slaff: > return type of this method (should be int) ... What say you?

Good catch! Please make the changes and add in the commit message information about the BC. The impact should be relatively low.

mikee47:
I've discovered an odd behaviour with my modbus stuff, it's started hanging on serial receive callback. I'll post back when it's sorted, may need a change to this PR.

mikee47:
> I've discovered an odd behaviour with my modbus stuff, it's started hanging on serial receive callback. I'll post back when it's sorted, may need a change to this PR.

Sorted by latest commit. I've rebased to latest develop.

samples/Basic_Serial/files/Readme.md
```diff
@@ -0,0 +1,322 @@
+# Sming
```
slaff: Maybe it would be better to change the makefile of the app to copy the Readme file, if not present, from the Sming root directory and store it under `files`.

samples/Basic_Serial/include/SerialReadingDelegateDemo.h
```diff
@@ -1,54 +1,28 @@
...
-				charReceived++;
-				Serial.print(cur);
-			}
-			Serial.println();
-		}
+		this->callback = callback;
 	}
 
 private:
+	HardwareSerial* serial;
```
slaff: Can you set the initial value to nullptr?

PR #1496, opened 2018-10-18
General changes to Stream classes

mikee47:
* Move trivial code into header
* Use member variable initialisers instead of setting defaults in constructor
* Change int to unsigned/size_t as appropriate
* Change appropriate methods to const (except where there's a risk of breaking other code)
* Amend comments to clarify operation
* Mark inherited methods with virtual (if not already)
* Add ReadWriteStream::write(uint8_t) implementation to call write(const char*, size_t) method by default

### DIFF
--- a/Sming/SmingCore/Data/Stream/EndlessMemoryStream.h
+++ b/Sming/SmingCore/Data/Stream/EndlessMemoryStream.h
@@ -25,21 +25,22 @@
 class EndlessMemoryStream : public ReadWriteStream
 {
 public:
-	virtual ~EndlessMemoryStream();
+	virtual ~EndlessMemoryStream()
+	{
+		delete stream;
+	}
 
-	//Use base class documentation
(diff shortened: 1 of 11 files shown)


Sming/SmingCore/Data/Stream/EndlessMemoryStream.h
```diff
@@ -25,33 +25,40 @@
...
-
 	/** @brief  Write chars to stream
 	 *  @param  buffer Pointer to buffer to write to the stream
 	 *  @param  size Quantity of chars to write
 	 *  @retval size_t Quantity of chars written to stream
 	 */
 	virtual size_t write(const uint8_t* buffer, size_t size);
 
-	virtual bool isFinished();
+	/** @todo is this behaviour consistent with DataSourceStream ?
```
slaff: Remove the todo comment and leave the return value as it is for now.

PR #1498, opened 2018-10-21
HttpRequest improvements and move HttpParams into separate module

mikee47:
* Rename HttpRequest::stream to bodyStream
* Move trivial code into headers
* Simplify getHeader() and getPostParameter() methods by calling `HashMap[] const` method, thus ensuring missing values aren't added. Also more efficient as doesn't require two lookups (one for contains(), another for operator[]).
* Simplify and optimise HttpRequest::getBody() by reading content directly into pre-allocated String

WebHelpers/escape
* bugfix: encoding of space as '+' added
* uri_escape(const char*, int) rewritten without using heap
* add uri_unescape_inplace(char*, int)
* bugfix: output buffer must account for nul terminator in uri_unescape_inplace(String&)

structures.h
* Move HttpParams definition into separate module

UrlencodedOutputStream
* Remove (duplicate) definition of HttpParams
* Move output code into HttpParams::printTo() method

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpRequest.h
+++ b/Sming/SmingCore/Network/Http/HttpRequest.h
@@ -169,7 +169,7 @@ protected:
 	RequestBodyDelegate requestBodyDelegate;
 	RequestCompletedDelegate requestCompletedDelegate;
 
-	ReadWriteStream* stream = nullptr;
+	ReadWriteStream* bodyStream = nullptr;
 	ReadWriteStream* responseStream = nullptr;
 
 #ifdef ENABLE_HTTP_REQUEST_AUTH
--- a/Sming/SmingCore/Network/Http/HttpRequest.h
+++ b/Sming/SmingCore/Network/Http/HttpRequest.h
@@ -21,6 +21,7 @@
 #include "Data/Stream/DataSourceStream.h"
 #include "Data/Stream/MultipartStream.h"
 #include "Network/Http/HttpHeaders.h"
+#include "HttpParams.h"
 
 class HttpClient;
 class HttpServerConnection;
--- a/Sming/SmingCore/Network/Http/HttpRequest.h
+++ b/Sming/SmingCore/Network/Http/HttpRequest.h
@@ -30,6 +30,9 @@ typedef Delegate<int(HttpConnection& client, HttpResponse& response)> RequestHea
 typedef Delegate<int(HttpConnection& client, const char* at, size_t length)> RequestBodyDelegate;
 typedef Delegate<int(HttpConnection& client, bool successful)> RequestCompletedDelegate;
 
+/*
+ * Encapsulates an incoming or outgoing request
+ */
 class HttpRequest
 {
 	friend class HttpClient;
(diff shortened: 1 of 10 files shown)


Sming/SmingCore/Network/Http/HttpRequest.h
```diff
@@ -30,82 +31,157 @@ typedef Delegate<int(HttpConnection& client, HttpResponse& response)> RequestHea
...
+
 	HttpRequest* setBody(uint8_t* rawData, size_t length);
 
 	/**
 	 * @brief Instead of storing the response body we can set a stream that will take care to process it
 	 * @param ReadWriteStream *stream
 	 *
 	 * @retval HttpRequest*
+	 *
+	 * @note The response to this message will be stored in the user-provided stream.
```
slaff: `The response to this message`. Did you mean `The response to this request` ?

Sming/SmingCore/Network/Http/HttpRequest.h
```diff
@@ -30,82 +31,157 @@ typedef Delegate<int(HttpConnection& client, HttpResponse& response)> RequestHea
...
 
-	HttpRequest* setHeaders(const HttpHeaders& headers);
+	HttpRequest* setHeader(const String& name, const String& value)
+	{
+		headers[name] = value;
+		return this;
+	}
 
-	HttpRequest* setHeader(const String& name, const String& value);
+	/** @deprecated this replaces existing post parameter set, which is different to
```
slaff: Can you add here also a very small example of the new recommended way? That target audience is beginners.

Sming/SmingCore/Network/Http/HttpRequest.h
mikee47: Yes! Well spotted.

Sming/SmingCore/Network/Http/HttpRequest.h
mikee47: setPostParameters() isn't actually used anywhere in Sming or its samples. Would it be safe to delete?

mikee47: If we're keeping it, I'd probably amend the comment to something like:
```
To set all post parameters, use request.postParams = params
To copy additional post parameters, use request.postParams.setMultiple(params)
```

slaff:
@mikee47 Is it tested on real device with real app ?

mikee47:
> @mikee47 Is it tested on real device with real app ?

Yes, the escaping changes are tested and verified. This PR is mainly refactoring, the substantive changes are yet to come.
I've attempted to get the `Basic_WebSkeletonApp` sample working as it exercises `HttpResult::getBody()` but it fails to work as-is. I've pushed a few changes to make it work.

Sming/SmingCore/Network/Http/HttpServerConnection.cpp
```diff
@@ -280,6 +280,8 @@ int HttpServerConnection::staticOnBody(http_parser* parser, const char* at, size
 	}
 
 	// TODO: ...
+	connection->request.setBody((const uint8_t*)&at, length);
```
slaff: `connection->request.setBody((const uint8_t*)&at, length);`

That is something that we don't do. Because it is the easiest way to crash remotely the http server and the device. Just send a request with very long body. What we do is use body parsers and decide in there how to operate over the incoming response body.

That said, please, remove that new line.

mikee47:
@slaff I've reverted everything to do with the `Basic_WebSkeletonApp`, bad example for this PR.

mikee47:
> Is it tested on real device with real app ?

slaff:
> Tested with HttpServer_ConfigNetwork and works fine. Code is migrated from my dev. branch which it's tested on first. We're moving into trickier territory now, lots of dependencies.

I guess that means that the PR is ready to be test by me, right?

mikee47:
> I guess that means that the PR is ready to be test by me, right?
Yes please, I've nothing else to add for this PR.

slaff:
@mikee47 That PR broke my HTTPS client connections... My wild guess is that either some data is sent  before the handshake is over... or I will have to update my code which means that we should add more information how to migrate to the new code.

Sming/SmingCore/Network/Http/HttpRequest.h
```diff
@@ -117,15 +193,38 @@ class HttpRequest
...
+
+	/** @brief Clear buffers and reset to default state in preparation for another request */
 	void reset();
 
 #ifdef ENABLE_SSL
-	HttpRequest* setSslOptions(uint32_t sslOptions);
-	uint32_t getSslOptions();
+	HttpRequest* setSslOptions(uint32_t sslOptions)
+	{
+		sslOptions = sslOptions;
```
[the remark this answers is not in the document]
slaff: Yep, that was the problem. Can you put a PR with the small fix? Sorry for not noticing this earlier. There should be a way to detect such problems automatically.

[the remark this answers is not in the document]
mikee47: I made the error moving the code into the header - yes, it would be useful to be able to track those sorts of changes in diff.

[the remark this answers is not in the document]
mikee47: @slaff Thank you!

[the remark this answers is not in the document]
slaff: The issue is fixed with #1504. No action is needed from your side.

> yes, it would be useful to be able to track those sorts of changes in diff.

I guess my IDE, cpplint or codacy should be able to detect the issue. Will look tomorrow how to enable that in codacy by default.

PR #1500, opened 2018-10-22
 Refactor TemplateFileStream class using TemplateStream

mikee47:
DataSourceStream
* [BC] Amend StreamType enumeration eSST_TemplateFile -> eSST_Template
* [BC] Make getStreamType() method const
* [BC] make id() method const
* Add virtual getName() method to provide generic way of getting filename from TemplateStream wrapper
* Add virtual isValid() method

TemplateFileStream
* Refactor as TemplateStream class to allow any other stream type to provide source template data
* Replace int with unsigned or size_t as appropriate
* Minor code improvements

HttpRequest
* setFile() method can use any ReadWriteStream-derived class, doesn't have to be FileStream.


slaff:
@mikee47 rebase and push -f to update the code. 
```
cd %SMING_HOME%
git checkout develop
git pull
git checkout feature/TemplateStream
git merge develop
# if the merge errored then
# git mergetool 
git rebase develop 
# if rebase failed
# git mergetool 
````

mikee47:
> @mikee47 Did you have the time to test the PR with the HttpServer_ConfigNetwork sample?

Not sure. I'll do that test and report back. I'd also like to make this PR a bit more interesting so will add something to samples to illustrate using TemplateStream with flash strings instead of files.

mikee47:
Re-tested with `HttpServer_ConfigNetwork` (lovely sample!) yes, I remember already doing this so works fine.

PR #1502, opened 2018-10-23
Streaming support for flash data

mikee47:
* Add FlashMemoryStream class
* Add FlashTemplateStream class
* Add IMPORT_FSTR macro to bind a file into firmware image as FlashString object
* Demonstrate usage using HttpServer_ConfigNetwork sample application. Note behaviour is unchanged, but settings.html is statically linked rather than using SPIFFS.


mikee47:
Absolutely.

PR #1503, opened 2018-10-23
Rename SimpleConcurrentQueue to ObjectQueue and simplified code.

mikee47:
Rename SimpleConcurrentQueue to ObjectQueue and simplify


slaff:
@mikee47 Did you test that code?

PR #1505, opened 2018-10-23
Enable more compiler checks

mikee47:
Modify Makefile to enable ALL compiler warnings, then selectively disable so we can see exactly which warnings are being suppressed.

Note: This PR disables warnings-as-errors because it won't (quite) build at the moment.

Core plus ArduinoJson build with only these warnings disabled:

```
	comment
	strict-aliasing
	unused-function
	unused-variable
```

Libraries require these ones disabled also

```
	reorder
	parentheses
	unused-but-set-variable
	maybe-uninitialized *
	return-type *
	sign-compare +
	array-bounds *
```

   \* masks serious issues
   \+ frail code, likely to break

Bugfixes:
* WebsocketFrame.cpp line 152 - error in expression logic (<< used instead of <<=)
* CommandExecutor::executorReceive() - missing return statement

Changes:
* Change Vector::contains() method to use indexOf() - duplicate code
* Change AtClient retries parameter to unsigned
* Use flash strings in gdbstub.c
* Remove SimpleQueue class from AtClient, replace with basic FIFO<>
* Modify esp-open-lwip patch to include additional missing function prototypes, also use _F() on flash strings to conserve RAM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,4 @@ GPATH
 # Microsoft Visual SourceSafe files
 vssver2.scc
 
-/.metadata
+.metadata
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ GPATH
 
 # Microsoft Visual SourceSafe files
 vssver2.scc
+
+/.metadata
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -267,9 +267,12 @@ endif
 MFORCE32 := $(shell $(CC) --help=target | grep mforce-l32)
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
-CFLAGS += -Wall -Wno-comment \
+CFLAGS += -Wall -Wundef -Wpointer-arith -Wno-comment \
 			-Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections \
          	-D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL) -DENABLE_CMD_EXECUTOR=$(ENABLE_CMD_EXECUTOR) -DESP8266=1 -DSMING_INCLUDED=1 
+ifneq ($(STRICT),1)
+CFLAGS += -Werror -Wno-sign-compare -Wno-parentheses -Wno-unused-variable -Wno-unused-but-set-variable -Wno-strict-aliasing -Wno-return-type -Wno-maybe-uninitialized
+endif
 ifeq ($(SMING_RELEASE),1)
 	# See: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html
 	#      for full list of optimization options
@@ -291,6 +294,9 @@ endif
 CFLAGS += -DCUST_FILE_BASE=$$* -DDEBUG_VERBOSE_LEVEL=$(DEBUG_VERBOSE_LEVEL) -DDEBUG_PRINT_FILENAME_AND_LINE=$(DEBUG_PRINT_FILENAME_AND_LINE)
 
 CXXFLAGS	= $(CFLAGS) -fno-rtti -fno-exceptions -std=c++11 -felide-constructors
+ifneq ($(STRICT),1)
+CXXFLAGS += -Wno-reorder
+endif
 
 # linker flags used to generate the main object file
 LDFLAGS		= -nostdlib -u call_user_start -Wl,-static -Wl,--gc-sections -Wl,-wrap,system_restart_local 
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -140,7 +140,7 @@ COM_SPEED_SERIAL  ?= $(COM_SPEED)
 # By default `debugf` does not print file name and line number. If you want this enabled set the directive below to 1
 DEBUG_PRINT_FILENAME_AND_LINE ?= 0
 
-# Defaut debug verbose level is INFO, where DEBUG=3 INFO=2 WARNING=1 ERROR=0 
+# Default debug verbose level is INFO, where DEBUG=3 INFO=2 WARNING=1 ERROR=0 
 DEBUG_VERBOSE_LEVEL ?= 2
 
 # Disable CommandExecutor functionality if not used and save some ROM and RAM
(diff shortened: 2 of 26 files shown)


mikee47:
@slaff I may be a little premature with this one but it does highlight the areas I think need to be sorted before 3.7.x

.gitignore
```diff
@@ -17,3 +17,5 @@ GPATH
 
 # Microsoft Visual SourceSafe files
 vssver2.scc
+
+/.metadata
```
slaff: You can safely remove the leading slash. Ex: `/.metadata` -> `.metadata`.

Sming/Wiring/WVector.h
```diff
@@ -119,14 +119,14 @@ Vector<Element>::Vector(unsigned int initialCapacity, unsigned int capacityIncre
   _capacity = initialCapacity;
   _data = new Element*[ _capacity ];
   _increment = capacityIncrement;
-  if (_data == NULL) _capacity = _increment = 0;
-};
+  if (_data == nullptr) _capacity = _increment = 0;
+}
```
slaff: Can you leave the closing semicolon? This way the files in Wiring will be consistent (see also WHashMap for example).  Or was that change due to a warning reported from the compiler?

mikee47: It's only for method implementations, not the class definition - WHashMap does everything inline so unless I'm missing something it's entirely consistent. I removed them because autoformat makes it look very strange (although we're not doing that, at least for now), no compiler warnings.

slaff: > It's only for method implementations, not the class definition

You are absolutely right.  Now comes the question how to solve the quite-so-often error of creating a (template) class definition without the trailing semicolon :). There are two major ways to solve that
- we remove all semicolons from template method definitions (which is correct C++). And spend time educating every newcomer that made the mistake. 
- we put semicolons everywhere on both template classes and template methods. This way, hopefully, there will be less explaining to do.

I would prefer the first solution. But that should be a part of a separate PR where also the other template definitions are revised and consistent. If you agree - please put back the semicons here and create a new PR where they are removed from the template method definitions. 

What do you say?

mikee47: No probs, I'll put the semicolons back!

mikee47: It's pretty simple rule: no semicolons after function (or method) definition :-)

slaff: > We love those semicolons!

But not for long. The new PR will fix this, right :)

mikee47: Natürlich. I've had a quick look through and as far as I can tell WVector.h is the only offender. Do you know of any others?

Sming/third-party/.patches/esp-open-lwip.patch
```diff
@@ -343,3 +385,165 @@ index af6e360..71b2178 100644
...
+ #include "lwip/mem.h"
+ #include "lwip/tcp_impl.h"
+ #include "lwip/udp.h"
++#include "lwip/igmp.h"
+ 
+ #include "lwip/app/espconn_udp.h"
+ 
++
++/*
++ * wifi_get_ip_info()
```
slaff: Can you put a bit more information in the comment? Or just remove it if it does not contain useful information.

Sming/Wiring/WVector.h
slaff: > Do you know of any others?

I ran quick reg-ex search over the code and only WVector.h showed up. Which is good.

slaff:
> I may be a little premature with this one but it does highlight the areas I think need to be sorted before 3.7.x. I have PRs to fix most of the remaining problems

Can you add a directive that switches to the higher level of error reporting? For example STRICT=1. Strict by default should be disabled. What this will bring is:
* not scare new, returning or non-core Sming/C++ developers
* but at the same time allow us, the core developers, to see the issues and take care to gradually fix them (the travis and appveyor build should have STRICT=1 by default).

Once all major and minor issues are resolved we can set the STRICT directive to be `on` by default.

Sming/Makefile
```diff
@@ -267,12 +267,13 @@ endif
 MFORCE32 := $(shell $(CC) --help=target | grep mforce-l32)
 
 # compiler flags using during compilation of source files. Add '-pg' for debugging
-CFLAGS += -Wpointer-arith -Wundef -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -finline-functions -fdata-sections -ffunction-sections \
-         -D__ets__ -DICACHE_FLASH -DARDUINO=106 -DCOM_SPEED_SERIAL=$(COM_SPEED_SERIAL) -DENABLE_CMD_EXECUTOR=$(ENABLE_CMD_EXECUTOR) -DESP8266=1 -DSMING_INCLUDED=1 
+CFLAGS += -Wall -Wno-comment \
```
slaff: > I may be a little premature with this one but it does highlight the areas I think need to be sorted before 3.7.x. I have PRs to fix most of the remaining problems

Can you add a directive that switches to the higher level of error reporting? For example STRICT=1. Strict by default should be disabled. What this will bring is:
* not scare new, returning or non-core Sming/C++ developers
* but at the same time allow us, the core developers, to see the issues and take care to gradually fix them (the travis and appveyor build should have STRICT=1 by default).

Once all major and minor issues are resolved we can set the STRICT directive to be `on` by default.

mikee47: Good idea. Thinking about `-Werror`, I'll put that back in for non-STRICT but perhaps leave it out for `STRICT` might be more helpful for development. What do you think?

mikee47: I've also deliberately left `Makefile-rboot.mk` and `Makefile-project.mk` out of all  this - something for future consideration perhaps...

Sming/appinit/user_main.cpp
```diff
@@ -119,11 +119,11 @@ extern "C" void ICACHE_FLASH_ATTR  __attribute__((weak)) user_pre_init(void)
 
 	enum flash_size_map sizeMap = system_get_flash_size_map();
 	if(!system_partition_table_regist(partitions, ARRAY_SIZE(partitions), sizeMap)) {
-		os_printf(_F("system_partition_table_regist: failed\n"));
-		os_printf(_F("size_map = %u\n"), sizeMap);
+		os_printf("system_partition_table_regist: failed\n");
```
slaff: Just out of interested: Is the optimized os_printf(X) doing the same or similar to os_printf(_F(X))?

mikee47: Yes. When I looked at other instances of os_printf() (in LWIP especially) I realised it would be simpler redefining `os_printf()` instead, but someone else had already thought of it! Format strings are emitted to `.irom.text` (ICACHE_RODATA_ATTR) rather than `.irom0.text` but end up in the same place.

PR #1508, opened 2018-10-24
Revise SplitString

mikee47:
* Doesn't need to be a friend of String, but #included inside WString.h for convenience
* Use unsigned instead of int where appropriate
* Minor improvements


Sming/Wiring/SplitString.cpp
```diff
@@ -1,94 +1,62 @@
...
-    if (chars[i] == delim) splitCount++;
-  }
-  if (splitCount == 0)
-  {
-    splits.addElement(atol(what.buffer));
-    return 1;
-  }
-
-  int pieceCount = splitCount + 1;
+/* $Id: WMemory.cpp 1156 2011-06-07 04:01:16Z bhagman $
```
slaff: `WMemory.cpp` -> `SplitString.cpp`.

PR #1537, opened 2018-11-26
Added testing OTA server

### LINKED ISSUE #1536: Feature request: Make target that builds and launches web server for rboot OTA
It would be advantageous to be able to launch a web server to serve rBoot OTA images. I have tweaked my makefile thus:

```
build_ota: all
	cd $(FW_BASE) && python -m SimpleHTTPServer 8080
```

slaff:
Closes #1536.

### DIFF
--- a/Sming/Makefile-rboot.mk
+++ b/Sming/Makefile-rboot.mk
@@ -8,6 +8,8 @@
 
 ### Defaults ###
 
+SERVER_OTA_PORT ?= 9999
+
 # rBoot options, overwrite them in the projects Makefile-user.mk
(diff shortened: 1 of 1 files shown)


riban-bw:
How about using a variable for the port number?

> $OTA_PORT ?= 8080

(I chose 8080 because it happened to be a port that I had already opened in the firewall on my dev machine but others may want a different value.)
It would be useful to hear from others as to whether they feel this is of sufficiently wide benefit or whether it is so niche that it may be better placed in individual project makefiles (which is where I put it when I first tested it).

PR #1540, opened 2018-12-02
Add toHTTPDate to DateTime class with supporting changes to CStringArray

riban-bw:
Corrects ISO8601 year to use 4 characters.
Add toHTTPDate function to DateTime class.
Calculate stringCount for CStringArray if not set, e.g. if object instantiated using assignment. Note: stringCount now mutable.
Add missing documentation for CStringArray class.
Use uppercase Y,M,D for date formats in documentation.

### DIFF
--- a/Sming/Services/DateTime/DateTime.cpp
+++ b/Sming/Services/DateTime/DateTime.cpp
@@ -155,7 +155,7 @@ String DateTime::toISO8601()
 
 String DateTime::toHTTPDate()
 {
-    char buf[30];
+	char buf[30];
 	m_snprintf(buf, sizeof(buf), _F("%s, %02d %s %04d %02d:%02d:%02d GMT"), CStringArray(flashDayNames)[DayofWeek], Day, CStringArray(flashMonthNames)[Month], Year, Hour, Minute, Second);
 	return String(buf);
 }
--- a/Sming/Services/DateTime/DateTime.cpp
+++ b/Sming/Services/DateTime/DateTime.cpp
@@ -4,7 +4,7 @@
 
   This library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
-  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. 
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
 */
 
 #include "DateTime.h"
@@ -17,6 +17,7 @@
  * Used to parse HTTP date strings - see parseHttpDate()
  */
 static DEFINE_FSTR(flashMonthNames, "Jan\0Feb\0Mar\0Apr\0May\0Jun\0Jul\0Aug\0Sep\0Oct\0Nov\0Dec");
+static DEFINE_FSTR(flashDayNames, "Sun\0Mon\0Tue\0Wed\0Thu\0Fri\0Sat");
 
 /** @brief Get the number of days in a month, taking leap years into account
  *  @param month 0=jan
@@ -148,7 +149,14 @@ String DateTime::toFullDateTimeString()
 String DateTime::toISO8601()
 {
 	char buf[32];
-	m_snprintf(buf, sizeof(buf), _F("%02d-%02d-%02dT%02d:%02d:%02dZ"), Year, Month + 1, Day, Hour, Minute, Second);
+	m_snprintf(buf, sizeof(buf), _F("%04d-%02d-%02dT%02d:%02d:%02dZ"), Year, Month + 1, Day, Hour, Minute, Second);
+	return String(buf);
+}
+
+String DateTime::toHTTPDate()
+{
+    char buf[30];
+	m_snprintf(buf, sizeof(buf), _F("%s, %02d %s %04d %02d:%02d:%02d GMT"), CStringArray(flashDayNames)[DayofWeek], Day, CStringArray(flashMonthNames)[Month], Year, Hour, Minute, Second);
 	return String(buf);
 }
 
--- a/Sming/Services/DateTime/DateTime.h
+++ b/Sming/Services/DateTime/DateTime.h
@@ -76,20 +76,20 @@ typedef enum {
 class DateTime
 {
 public:
-    /** @brief  Instantiate an uninitialised date and time object
-     */
+	/** @brief  Instantiate an uninitialised date and time object
+	 */
 	DateTime()
 	{
 	}
 
-    /** @brief  Instantiate a date and time object
-     *  @param  time Unix time to assign to object
-     */
+	/** @brief  Instantiate a date and time object
+	 *  @param  time Unix time to assign to object
+	 */
 	DateTime(time_t time);
 
-    /** @brief  Get current Unix time
-     *  @retval time_t Quantity of seconds since 00:00:00 1970-01-01
-    */
+	/** @brief  Get current Unix time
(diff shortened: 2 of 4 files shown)


riban-bw:
@slaff I avoided changing the coding style of the existing code (which was inconsistent and probably written before Sming project had agreed a style guide) so as to reduce the difficulty in identifying the functional changes to the library, i.e. ease validation and acceptance. Running
> make cs

results in lots of
> : CommandLine Error: Option 'disable-symbolication' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options

This may be because clang-tidy is not available for cygwin. I will try to tidy up the source code to match Sming coding style. I'll run it through clang-tidy on a Linux machine then update the PR.

Sming/Services/DateTime/DateTime.cpp
```diff
@@ -148,7 +149,14 @@ String DateTime::toFullDateTimeString()
...
 {
 	char buf[32];
-	m_snprintf(buf, sizeof(buf), _F("%02d-%02d-%02dT%02d:%02d:%02dZ"), Year, Month + 1, Day, Hour, Minute, Second);
+	m_snprintf(buf, sizeof(buf), _F("%04d-%02d-%02dT%02d:%02d:%02dZ"), Year, Month + 1, Day, Hour, Minute, Second);
+	return String(buf);
+}
+
+String DateTime::toHTTPDate()
+{
+    char buf[30];
```
slaff: Fix the indentation please. Should be as the line below.

Sming/SmingCore/Data/CStringArray.cpp
```diff
@@ -1,55 +1,67 @@
...
+			if(index == 0)
+				return s;
+			offset += strlen(s) + 1;
+		}
+	}
+
+	return nullptr;
+}
+
+unsigned CStringArray::count() const
```
slaff: Can you explain here why is that change needed?

Sming/Services/DateTime/DateTime.cpp
```diff
@@ -4,7 +4,7 @@
```
slaff: For consistency the following methods in the DateTime class should be renamed:

* parseHttpDate(const String& httpDate) -> fromHttpDate(const String& httpDate)
* convertFromUnixTime(time_t timep,..) -> fromFromUnixTime(time_t timep,..) 
* convertToUnixTime(int8_t sec, ..)  -> toUnixTime(int8_t sec, ..)

In addition if the DateTime class should deal with different date formatting then the following method could be a very good addition

`format(const String& format)`

Example:
```
DateTime d;
String httpDate = d.format("D, d M Y H:i:s GMT");
```

Sming/SmingCore/Data/CStringArray.h
```diff
@@ -1,98 +1,122 @@
...
+		stringCount = 0;
+	}
+
+	/** @brief  Get quantity of strings in array
+	*   @retval unsigned Quantity of strings
+	*/
+	unsigned count() const;
+
+private:
+	mutable unsigned stringCount = 0;
```
slaff: Why is this private variable set as mutable? Use-cases, please.

Sming/SmingCore/Data/CStringArray.h
```diff
@@ -1,98 +1,122 @@
...
+	*   @retval bool True if string exists in array
+	*   @note   Search is not case-sensitive
+	*/
+	bool contains(const String& str) const
+	{
+		return indexOf(str) >= 0;
+	}
+
+	/** @brief Get string at the given position
+	 * 	@param index 0-based index of string to obtain
```
slaff: Can you align `@param` with `@brief` in all places where you have added the documentation ?

Sming/SmingCore/Data/CStringArray.cpp
riban-bw: When an instance of CStringArray is created using an assignment, e.g. CStringArray cs("first\0second"); the quantity of elements is not set. This might normally be set within the constructor but I was unable to figure out how to do that in this case so instead I ensured count() returned the correct quantity of elements in the array by deducing whether it had already been calculated and if not then one-time calculating.

Sming/SmingCore/Data/CStringArray.h
riban-bw: count() is a const function so cannot change the value of the private variable stringCount unless stringCount is mutable. See explanation in other comment about why count() needs to set stringCount.

Sming/Services/DateTime/DateTime.cpp
riban-bw: Will do - IDE did its own thing. Will edit in text editor until I can whip IDE into line.

Sming/Services/DateTime/DateTime.cpp
slaff: > Do you want me to rationalise this?

Yes, but that should be done as a separate PR (after this one gets merged). The following refactoring changes should be done:
- Move DateTime from Services to SmingCore.
- Rename the old methods according to the rules above.
- In the DateTime.h header added the old methods that should redirect to the new ones, plus mark them as deprecated. Example:

```C++
class DateTime {
public: 
    /** @brief  Parse a HTTP full date and set time and date
     *  @param  httpDate HTTP full date in RFC 1123 format, e.g. Sun, 06 Nov 1994 08:49:37 GMT
     *  @note   Also supports obsolete RFC 850 date format, e.g. Sunday, 06-Nov-94 08:49:37 GMT where 2 digit year represents range 1970-2069
     *  @note   GMT suffix is optional and is always assumed / ignored
     *
     *   @deprecated Use `fromHttpDate` instead
     */
     bool parseHttpDate(const String& httpDate)   __attribute__ ((deprecated))
     {
            return fromHttpDate(httpDate);
     }
}
```

slaff:
@mikee47 When you have a bit of time can you tell me what is your opinion about the suggested changes in CStringArray ?

Sming/SmingCore/Data/CStringArray.cpp
```diff
@@ -1,55 +1,67 @@
...
+{
+	if(stringCount == 0 && length() > 0) {
+		//If array is created by assignment (e.g. CStringsArray csa = "Hello\0World";) then stringCount is not set so set it here
+		for(unsigned offset = 0; offset < length() + 1; offset++) {
```
mikee47: why not just 'offset <= length()' ?

mikee47: @slaff I like the changes, makes the class more useful.

slaff: @riban-bw Can you make the suggested change ?

`for(unsigned offset = 0; offset < length() + 1; offset++) {`
to be
`for(unsigned offset = 0; offset <= length(); offset++) {`

riban-bw: Doh! Done!

Sming/Services/DateTime/DateTime.h
```diff
@@ -137,11 +137,20 @@ class DateTime
...
+     *  @retval String Date and time in format DD.MM.YYYY hh:mm:ss
      */
 	String toFullDateTimeString();
+
+   /** @brief  Get human readable date and time
+     *  @retval String Date and time in format YYYY-MM-DDThh:mm:ssZ
+     */
 	String toISO8601();
 
+   /** @brief  Get human readable date and time
```
slaff: @riban-bw Open that file with very simple text editor and then find and fix all the places where the indentation is wrong. As above, for example.

Issue #1545, opened 2018-12-07
Refactoring of DateTime


slaff:
The following refactoring changes should be done:
- Move DateTime from Services to SmingCore.
- Rename the old methods according to the rules above.
- In the DateTime.h header added the old methods that should redirect to the new ones, plus mark them as deprecated. 
Example:

```C++
class DateTime {
public: 
    /** @brief  Parse a HTTP full date and set time and date
     *  @param  httpDate HTTP full date in RFC 1123 format, e.g. Sun, 06 Nov 1994 08:49:37 GMT
     *  @retval bool True on success
     *  @note   Also supports obsolete RFC 850 date format, e.g. Sunday, 06-Nov-94 08:49:37 GMT where 2 digit year represents range 1970-2069
     *  @note   GMT suffix is optional and is always assumed / ignored
     *
     *   @deprecated Use `fromHttpDate` instead
     */
     bool parseHttpDate(const String& httpDate)   __attribute__ ((deprecated))
     {
            return fromHttpDate(httpDate);
     }
}
```

riban-bw:
As discussed, you can allocate this to me if you want. I will pick it up over the next couple of weeks.

riban-bw:
Related to #1540.

riban-bw:
@slaff suggested adding:

format(const String& format), e.g.
```
DateTime d;
String httpDate = d.format("D, d M Y H:i:s GMT");
```
I wonder whether something more like strftime would be better, e.g.
```
DateTime d;
String httpDate = d.format("%a, %d %b %Y %H:%M:%S GMT");
```
If so, we may want to limit the implementation or add pre-processor conditional compilation blocks to avoid bloat.

I have done the refactoring (PR imminent) and will work up a strftime style _format_ function for review.

slaff:
> I wonder whether something more like strftime would be better, e.g.

Better not for the moment.

PR #1549, opened 2018-12-11
Fix/refactor date time

riban-bw:
In response to issue #1545 this PR refactors DateTime to rationalise method names, deprecating old (inconsistent) names.
Moved from Services to SmingCore.
Also adds format method which returns a string based on formatting, providing strftime formatting of string. (Some elements not implemented but mostly there.)
Provides a sample project demonstrating DateTime string output features.
Fixes bug with toHttpDate not setting day of week.
Changed convenience string methods to use new format method.
Not (yet) implemented any locale - all currently in English with existing variation in date formats, e.g. US date format with dot separator used by some European countries.


slaff:
@riban-bw Don't forget to run `make cs` and commit the coding style changes.

Sming/SmingCore/DateTime/DateTime.h
```diff
@@ -112,14 +112,26 @@ class DateTime
...
 	 */
-	bool parseHttpDate(const String& httpDate);
+	bool fromHttpDate(const String& httpDate);
+
+	/** @brief  Parse a HTTP full date and set time and date
+	 *  @param  httpDate HTTP full date in RFC 1123 format, e.g. Sun, 06 Nov 1994 08:49:37 GMT
+	 *  @retval bool True on success
+	 *  @note   Also supports obsolete RFC 850 date format, e.g. Sunday, 06-Nov-94 08:49:37 GMT where 2 digit year represents range 1970-2069
+	 *  @note   GMT suffix is optional and is always assumed / ignored
+	 *  @deprecated Use 'fromHttpDat' instead
```
slaff: `fromHttpDat` change to `fromHttpDate`.

Sming/SmingCore/DateTime/DateTime.cpp
```diff
@@ -0,0 +1,416 @@
+/*
```
slaff: Move `Sming/SmingCore/DateTime/DateTime.*` to `Sming/SmingCore/DateTime.*`. We don't need additional sub-directory for just one class.

Sming/SmingCore/SmingCore.h
```diff
@@ -55,7 +55,8 @@
 #include "Data/Stream/FileStream.h"
```
slaff: You not' only change the content of the file but also its permissions. `Sming/SmingCore/SmingCore.h 100644 → 100755`. The file permission change has to be reverted.

Sming/SmingCore/DateTime/DateTime.h
riban-bw: Fixed in working copy. Will push when coding style checked and committed.

Sming/SmingCore/DateTime/DateTime.cpp
riban-bw: Fixed in working copy. Will push when coding style checked and committed. I am currently unable to remove the directory with git. Will investigate further.

Sming/SmingCore/SmingCore.h
riban-bw: Not sure what happened there. Cygwin seems to do odd things at times. Fixed in working copy. Will push when coding style checked and committed.

Sming/SmingCore/DateTime/DateTime.cpp
slaff: Try these commands:
```
git mv Sming/SmingCore/DateTime/DateTime.h Sming/SmingCore/DateTime.h
git mv Sming/SmingCore/DateTime/DateTime.cpp Sming/SmingCore/DateTime.cpp
git rm -r Sming/SmingCore/DateTime/
```

riban-bw:
> @riban-bw Don't forget to run `make cs` and commit the coding style changes.

I have downgraded Doxygen and Clang in Cygwin to one that works so I can now run clang-format. Hopefully this should improve my ability to submit changes within Sming coding style guidelines. I may even be able to beat my IDE into submission to produce the required style to begin with!

Sming/SmingCore/DateTime/DateTime.h
riban-bw: Fixed in last push.

Sming/SmingCore/DateTime/DateTime.cpp
riban-bw: I tried rm -r but git doesn't think the directory exists so maybe something I did removed it. It looks like it is now deleted.

Sming/SmingCore/SmingCore.h
riban-bw: Fixed in last push.

Sming/SmingCore/Network/Http/HttpConnection.cpp
```diff
@@ -89,7 +89,7 @@ DateTime HttpConnection::getLastModifiedDate()
 {
```
slaff: `Sming/SmingCore/Network/Http/HttpConnection.cpp 100644 → 100755`

Fix the permissions here.

Sming/SmingCore/Network/Http/HttpConnectionBase.cpp
```diff
@@ -44,7 +44,7 @@ void HttpConnectionBase::init(http_parser_type type)
 #endif
 		parserSettings.on_url = staticOnPath;
 
-		// Data callbacks: on_url, (common) on_header_field, on_header_value, on_body;
+// Data callbacks: on_url, (common) on_header_field, on_header_value, on_body;
```
slaff: That change is not needed. Take it out of the PR.

Sming/SmingCore/Network/MqttClient.h
```diff
@@ -309,7 +309,7 @@ class MqttClient : protected TcpClient
 
 	// client flags
 	uint8_t flags = 0;
-	/* 7 8 6 5 4 3 2 1 0
+/* 7 8 6 5 4 3 2 1 0
```
slaff: That change is not needed. Take it out of the PR.

Sming/SmingCore/Network/NtpClient.h
```diff
@@ -18,7 +18,7 @@
 #include "UdpConnection.h"
 #include "Platform/System.h"
 #include "Timer.h"
-#include "../Services/DateTime/DateTime.h"
+#include <DateTime.h>
```
slaff: Use `#include "DateTime.h"` instead of `#include <DateTime.h>`.

Sming/SmingCore/SPI.cpp
```diff
@@ -119,8 +119,9 @@ uint32 SPIClass::transfer32(uint32 data, uint8 bits)
 	regvalue |= SPI_USR_MOSI | SPI_DOUTDIN | SPI_CK_I_EDGE;
 	WRITE_PERI_REG(SPI_USER(SPI_NO), regvalue);
 
-	WRITE_PERI_REG(SPI_USER1(SPI_NO), (((bits - 1) & SPI_USR_MOSI_BITLEN) << SPI_USR_MOSI_BITLEN_S) |
-										  (((bits - 1) & SPI_USR_MISO_BITLEN) << SPI_USR_MISO_BITLEN_S));
+	WRITE_PERI_REG(SPI_USER1(SPI_NO),
```
slaff: Those changes is not needed. `Sming/SmingCore/SPI.cpp` should not be changed in this PR.

Sming/SmingCore/SystemClock.h
```diff
@@ -13,7 +13,7 @@
 #ifndef SMINGCORE_SYSTEMCLOCK_H_
 #define SMINGCORE_SYSTEMCLOCK_H_
 
-#include "../Services/DateTime/DateTime.h"
+#include <DateTime.h>
```
slaff: Use #include "DateTime.h" instead of #include <DateTime.h>.

Sming/SmingCore/Wire.cpp
```diff
@@ -271,7 +271,7 @@ void TwoWire::onRequest(void (*function)(void))
 	//user_onRequest = function;
 }
 
-	// Preinstantiate Objects //////////////////////////////////////////////////////
+// Preinstantiate Objects //////////////////////////////////////////////////////
```
slaff: Wire.cpp should not be modified in this PR.

samples/Basic_DateTime/.project
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>Basic_Blink</name>
```
slaff: `<name>Basic_Blink</name>`: Change the name to Basic_DateTime.

samples/Basic_DateTime/app/application.cpp
```diff
@@ -0,0 +1,119 @@
+/**	Basic_DateTime sample application
+	Author: Brian Walton (brian@riban.co.uk)
+	Date: 2018-12-10
+	Provides serial interface, accepting Unix timestamp
+	Prints each type of DateTime::format option
+*/
+#include <user_config.h>
+#include <SmingCore/SmingCore.h>
+
+time_t g_timestamp = 0;
```
slaff: Use camelCase. For this variable and the one below.

samples/Basic_Delegates/app/application.cpp
```diff
@@ -53,8 +53,8 @@ class Task
 		taskTimer
```
slaff: Don't modify this file in the current PR.

samples/Basic_ProgMem/app/TestProgmem.cpp
```diff
@@ -123,7 +123,7 @@ void testFSTR(Print& out)
 	}
 	out.println("}");
 
-	// Test equality operators
+// Test equality operators
```
slaff: Don't modify this file in the current PR.

samples/SDCard/app/application.cpp
```diff
@@ -106,7 +106,7 @@ void stat_file(char* fname)
 }
 
 FRESULT ls(const char* path /* Start node to be scanned (also used as work area) */
-)
+		   )
```
slaff: Don't modify this file in the current PR.

riban-bw:
I sometimes hate git - grrrrrr! The hard part should be writing the code not committing it to version control. Hopefully it looks in better shape now.
I have added one of the missing format parameters (%U week number) which leaves two others (both permutations of week number). I also added locale which may be useful in other modules. I have added what looks right to me for USA, UK, Germany, France and Australia which are the countries / languages that jump to mind where contributors mostly reside. Please check if my assumptions are correct (blame Google if they are wrong) and offer any missing locale. (There are obviously lots missing but it feels like unproportional effort to implement everywhere in the world on this first commit!)

Sming/SmingCore/Network/Http/HttpConnectionBase.cpp
riban-bw: Fixed

samples/Basic_DateTime/app/application.cpp
riban-bw: Fixed

samples/Basic_Delegates/app/application.cpp
riban-bw: Fixed

samples/Basic_ProgMem/app/TestProgmem.cpp
riban-bw: Fixed

samples/SDCard/app/application.cpp
riban-bw: Fixed

riban-bw:
> But please DON'T use percentage Something (%S) as format. Use "S". Please, see the format parameters for PHP for example and follow them http://php.net/manual/en/function.date.php#format .

Sming is a C++ framework. I think the strftime format may be familiar to C++ programmers. It is also more flexible than the PHP implementation - normal text may be included within the format string, e.g. "Time: %X". (I prefer Sming to Arduino because Sming implements printf which is an invaluable feature.) I understand that someone who is familiar with PHP may prefer the PHP format, someone who is familiar with Python may want the Python format, etc. but I recommend implementing the % format for all those C++ programmers who are already familiar with it. We could implement a similar function with different formatting if you want (under a separate PR). This could be designed to avoid overheads due to the way that only used functions are added to the firmware.

Regarding locale - this is implemented mostly with those horrid pre-compiler directives which ensure only the required bits are put into the code, e.g. `#define LOCALE LOCALE_DE_DE` should only include the German locale elements where they are actually used by the firmware. I don't think that Locale.h adds any bloat to the library as it simply defines **only** the strings that are already used in DateTime and if a user does not call _format_ in their source code then it won't be added to the firmware.

I agree that we need to constrain firmware size with particular attention to each type of memory and do try to code with an _embedded_ ethos, though may not create optimal code on a first attempt. (That is what peer review is for - right?) On the point of bloat, I was considering adding a feature request to measure libsming size after each commit and compare with previous / develop head, so that we can monitor the change in size that each commit has which may help to keep a handle on bloat.

slaff:
> On the point of bloat, I was considering adding a feature request to measure libsming size after each commit and compare with previous / develop head, so that we can monitor the change in size that each commit has which may help to keep a handle on bloat.

Do you know a free web service where we can send the data and it will generated the needed graphs for us? It is good to see how data size, free heap size and code size vary.

slaff:
> Sming is a C++ framework. I think the strftime format may be familiar to C++ programmers.

I agree with you. Let it be `http://www.cplusplus.com/reference/ctime/strftime/`

One more point - make sure that the code passes CI tests. At the moment it seem to fail under Travis due to coding style issues and under Windows due to 
```c++
'setlocale' is not a member of 'std'
     char* __old = std::setlocale(LC_NUMERIC, 0);
                   ^
```

riban-bw:
I can see that the coding style issue relates to indentation I added to Locale.h to make things easier to understand and edit. I will fix that (and also fix some errors with German I was able to identify when chatting to a German colleague). The Windows issue may be more tricky. I looks like something peculiar to Chocolatey which I cannot install on my development machine (the only Windows machine I have access to). This error does not occur with cygwin or Ubuntu. I suspect that the use of Unicode characters in some of the languages may be triggering espressif libraries to kick in locale elements which fail for Windows. This may be a dormant bug in their software. I will look at the errors from appveyor and try to diagnose but this may prove challenging. If anyone with a Chocolatey install is available to help diagnose and test this it would be useful.

slaff:
> I suspect that the use of Unicode characters in some of the languages may be triggering xtensa libraries to kick in locale elements which fail for Windows

I guess the issue is more trivial. On Windows the files names are case insensitive. Which means that your Locale.h will be loaded when locale.h. Try renaming it to something like CoreLocale.h and see if that works.

riban-bw:
Thanks slaff, I will try that. (I configured my Windows machine to be case insensitive to allow me to build xtensa. Maybe I should change it back to pick up stuff like this.) I was about to commit these changes but realised I had started to implement the last couple of formats (variations on week number) but not yet tested so need to complete that before I commit. Hopefully sort in next few days. (I didn't have to change the German - I had got it right first time! Native German speakers feel free to contradict me.)

slaff:
> German - I had got it right first time

The months and the days of the week are correct. If there is something more I can check it too.

riban-bw:
The API documentation has a table showing which parameters are influenced by locale. I guess I could change the locale of a test machine and compare Sming output with a native test app. (I think this can be done with an environmental variable in Linux so I could probably create a test for this for the supported languages.) I have been doing similar with cygwin but its implementation of strftime seems simplistic and does not include all the options so firing up a Linux machine may be a better test environment. The only other locale element I set for Germany is the short date where Germany uses the dot separator.

slaff:
> samples/SDCard/app/application.cpp 100644 → 100755

That unneeded modification is still in the PR. Remove it using:

```
cd $SMING_HOME
git checkout cc7af2263e41e24c12c047878c95f13489d5c81c
```

After that the older version of the file will be ready to commit to this PR.

riban-bw:
Thanks for the tip @slaff. I used 
```
git diff cc7af2263e41e24c12c047878c95f13489d5c81c|grep -A 1 -B 1 "old mode"
```
to list all the changed permission files and found a few more so corrected those. Knowing the ID of the version to compare to was handy. I have just seen in github where to find that so maybe I can be more efficient with git in future! Life was so much easier with subversion :-).

slaff:
@riban-bw Thanks for the PR. It took a bit longer but hopefully you are now better prepared `git` user :).

PR #1554, opened 2018-12-15
Fixes #1553 coding style errors

### LINKED ISSUE #1553: Several coding style errors
Running `make cs` updates several files. These should be corrected and committed to avoid interfering with other PRs.

riban-bw:


slaff:
Please, use clang-format version 5.0.0. I don't know which is your version but it definitely destroys the formatting instead of improving it.

PR #1558, opened 2018-12-18
Create separate libraries with name suffix to indicate build option, …

riban-bw:
…e.g. libsmingssld.a for debug library with SSL support.

Use different build locations to avoid need to clean and rebuild everything every time.

Allow creation of libsming, libsmingssl, libsmingd and libsmingssld. Suffix 'd' indicates ENABLE_GDB=1. Suffix 'ssl' indicates ENABLE_SSL=1. This allows simultaneous existence of normal and debug version of SSL and non-SSL versions of the library. Note: common.ld needed to be adjusted to ensure the correct map. SSL and non-SSL (and gdb) builds use the same value so use of wildcard '*' allowed all to be handled by single line. All relevant makefiles use the new naming convention.

An advantage of this PR is to allow debug libraries to be available without time consuming recompiling or awkward file renaming.

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -29,9 +29,6 @@
 # MacOS / Linux:
 # COM_PORT = /dev/tty.usbserial
 
-# Com port speed
-COM_SPEED = 115200
-
 # Git command
 GIT ?= git
 
@@ -159,16 +156,21 @@ USER_LIBDIR = compiler/lib
 
 LIBSMING = libsming
 ifeq ($(ENABLE_SSL),1)
-	LIBSMING= libsmingssl
+	LIBSMING := $(LIBSMING)ssl
+	BUILD_BASE := $(BUILD_BASE)ssl
 endif
 
+ifeq ($(ENABLE_GDB),1)
+	LIBSMING := $(LIBSMING)d
+	BUILD_BASE := $(BUILD_BASE)d
+endif
 
 # name for the target project
 TARGET		= app
 
 CUSTOM_TARGETS ?=
 
-ARDUINO_LIBRARIES = $(shell git submodule status Libraries | cut -c2- | cut -f2 -d ' ' | sed -r 's/Libraries\/([^/]*)/Libraries\/\1\/library.properties/g' )
+ARDUINO_LIBRARIES =
 
 # which modules (subdirectories) of the project to include in compiling
 MODULES     = system system/helpers Wiring appinit \
--- a/Sming/Makefile-project.mk
+++ b/Sming/Makefile-project.mk
@@ -200,8 +200,12 @@ THIRD_PARTY_DIR = $(SMING_HOME)/third-party
 SMING_FEATURES = none
 LIBSMING = sming
 ifeq ($(ENABLE_SSL),1)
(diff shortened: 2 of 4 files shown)


slaff:
It would be better to check if the following is possible:
- compile separately gdb (which is better to be called stack_dumper)
- compile separately esp-gdbstub

And at link time, based on a directive, decide which one to include so that the linker can added to the final static binary.

Sming/Makefile
```diff
@@ -159,16 +156,21 @@ USER_LIBDIR = compiler/lib
...
+	BUILD_BASE := $(BUILD_BASE)d
+endif
 
 # name for the target project
 TARGET		= app
 
 CUSTOM_TARGETS ?=
 
-ARDUINO_LIBRARIES = $(shell git submodule status Libraries | cut -c2- | cut -f2 -d ' ' | sed -r 's/Libraries\/([^/]*)/Libraries\/\1\/library.properties/g' )
+ARDUINO_LIBRARIES =
```
slaff: That change is plain wrong. Please, remove it from here. Did you use the `master` branch and tried to merge changes from it to the `develop` branch?

riban-bw: Hmmm... I took a copy of the modified files from the messy fork then recreated a clean fork and copied the files back. I should have done a more granular compare of the changed files which would have shown this up. The change must have happened in the mess that was the last fork! I will sort this out in next commit.

riban-bw:
I am worried that having a separate debug library may not provide sufficient debug access in the main libsming but I will check whether this is possible. It would be a much neater solution.

slaff:
There will be one `libsming` library and the additions will be added at compile time. As for example SSL/TLS support.

riban-bw:
That's a shame. I find it awkward to keep track of which libs have what features enabled.

slaff:
> That's a shame. I find it awkward to keep track of which libs have what features enabled.

The end user should know only which option he needs to enable. As is the case at the moment. During link time the build system will take care to use the correct libs. The advantage would be shorter compile times and there will be no need for recompilation of the same library just to provide additional feature. As for example TLS.

riban-bw:
I am confused...

Are we aiming for a single Sming library with all features enabled and those features may be subsequently enabled / disabled at app compile / link time?

Are we aiming for current behaviour where the library needs to be compiled with the features the app requires (which may mean recompiling library for different apps)?

Are we aiming for a combination of these where the library may be compiled with as many or few features then at app compile time, linking may occur to any of the available features?

I like the last option. Previously we have had to compile different libraries for SSL / non-SSL because non-SSL apps would not link to the SSL lib.

mikee47:
@riban-bw I think this is definitely the time to make sure the issues you've raised get properly addressed since some major restructuring has been done, with more changes in the pipeline.

We're having some discussion about adding an alternative/improved build system using CMake over at #1684, so I'll try to address these issues there.

Regarding SSL, have a look at #1713.

riban-bw:
Thanks both! I can see the ideas for the cmake build process may resolve these concerns so will happily leave this issue closed and follow the related issues.

PR #1563, opened 2018-12-26
Add HttpHeaders::setMultiple() method

### LINKED ISSUE #1562: Custom headers are not sent correctly
Setting custom header results in invalid HTTP request.
```
	HttpHeaders headers;
	headers[F("X-Powered-By")] = _F("Sming");				  // Use text for non-standard field names
	getRequest->setHeaders(headers);
```

image

As can be seen from the snapshot above the header name is missing.

mikee47:
Using inherited HashMap version doesn't work. Fixes #1562

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpHeaders.h
+++ b/Sming/SmingCore/Network/Http/HttpHeaders.h
@@ -169,6 +169,15 @@ public:
 		remove(fromString(name));
 	}
 
+	void setMultiple(const HttpHeaders& headers)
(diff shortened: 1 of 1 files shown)


mikee47:
I'm just testing `class HttpHeaders : private HashMap<HttpHeaderFieldName, String>` rather than `public`, which would have flagged this issue up earlier.

slaff:
@mikee47 Thanks a lot for the fast reaction!

PR #1565, opened 2019-01-02
Fix compilation when SMING_RELEASE=1

etmmahi:
Fix for compilation of crash_handler when SMING_RELEASE=1

### DIFF
--- a/Sming/system/crash_handler.c
+++ b/Sming/system/crash_handler.c
@@ -31,9 +31,7 @@ extern int ets_printf(const char*, ...);
 void uart_write_char_d(char c);
 static void uart0_write_char_d(char c);
 static void uart1_write_char_d(char c);
-#ifndef SMING_RELEASE
-static void print_stack(uint32_t start, uint32_t end);
-#endif
+static void print_stack(uint32_t start, uint32_t end) SMING_UNUSED;
 //static void print_pcs(uint32_t start, uint32_t end);
 
 extern void __custom_crash_callback( struct rst_info * rst_info, uint32_t stack, uint32_t stack_end ) {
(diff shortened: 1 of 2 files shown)


slaff:
How to reproduce the issue? What is the error log showing? I have an application using the crash handler and it is compiling just fine with SMING_RELEASE=1, but I guess I am missing something.

etmmahi:
Hi,
the issue arises when I try to compile the development branch with:

make SMING_RELEASE=1

trace of failing part of compilation:
```
make[1]: Entering directory '/home/martijn/espdev/Sming/Sming'
CC system/crash_handler.c
system/crash_handler.c:98:13: error: 'print_stack' defined but not used [-Werror=unused-function]
 static void print_stack(uint32_t start, uint32_t end) {
             ^
cc1: all warnings being treated as errors
Makefile:596: recipe for target 'out/build/system/crash_handler.o' failed
make[1]: *** [out/build/system/crash_handler.o] Error 1
make[1]: Leaving directory '/home/martijn/espdev/Sming/Sming'
```


Due to the STRICT option added in the Makefile,  warnings are now treated as errors and the compilation stops. See pull request for solution.

Best

slaff:
Your solution introduces a problem in the cases where the function is actually used. There should be a better solution.  For example you can do the following:

Add that definition in `Sming/system/include/esp_systemapi.h`
```c
/*
* Use this definition in the cases where a function or a variable is meant to be possibly unused. GCC will not produce a warning for it. 
*/
#define SMING_UNUSED  __attribute__ ((unused))
```

and then in `Sming/system/crash_handler.c` add the following change

```c
static void print_stack(uint32_t start, uint32_t end) SMING_UNUSED  {
...
```

etmmahi:
Implemented the suggested solution. Code builds clean with SMING_RELEASE=1.

PR #1571, opened 2019-01-08
General improvements to UdpConnection class

mikee47:
bugfix: Potential memory leak in UdpConnection::staticOnReceive
Simplify #includes
Remove redundant code
Move trivial method implementations into header
Provide initialisers for all member variables
Replace NULL with nullptr
Consistent use of braces

### DIFF
--- a/Sming/SmingCore/Network/UdpConnection.cpp
+++ b/Sming/SmingCore/Network/UdpConnection.cpp
@@ -8,13 +8,19 @@
 #include "UdpConnection.h"
 #include "WString.h"
 
-void UdpConnection::initialize(udp_pcb* pcb)
+bool UdpConnection::initialize(udp_pcb* pcb)
 {
 	if(pcb == nullptr) {
 		pcb = udp_new();
+		if(pcb == nullptr) {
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/Network/UdpConnection.cpp
```diff
@@ -68,37 +56,21 @@ bool UdpConnection::connect(IPAddress ip, uint16_t port)
 void UdpConnection::send(const char* data, int length)
```
slaff: I would prefer to have the return type of `UdpConnection::send` changed from `void` to `bool`.  `false` when the allocation did not work and `true` otherwise.  
Return type change should be applied also to the `UdpConnection::sendString` , `UdpConnection::sendTo` and `UdpConnection::sendStringTo` methods.

mikee47: Agreed. As `send()` & `sendTo()` are `virtual` it could break existing code, but the compiler would warn so easy to fix. I've also added `bool` return value for `initialize()`.

PR #1572, opened 2019-01-08
Dns server improvements

mikee47:
Remove leading underscores from member variable names
Provide initialisers for all member variables
Store `ttl` member variable in host order, not network 
Use ip_addr for resolvedIP
Tidy #includes
Move trivial code into header
downcaseAndRemoveWwwPrefix is a static method
Use new/delete instead of malloc/free

### DIFF
--- a/Sming/SmingCore/Network/DNSServer.cpp
+++ b/Sming/SmingCore/Network/DNSServer.cpp
@@ -14,17 +14,14 @@
 #include "DNSServer.h"
 #include "UdpConnection.h"
 #include "NetUtils.h"
-#include "../Wiring/WString.h"
+#include "WString.h"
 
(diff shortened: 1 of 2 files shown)


mikee47:
Tested using DNSCaptivePortal sample, works fine on my phone, debug logs look fine.

Sming/SmingCore/Network/DNSServer.cpp
```diff
@@ -14,79 +14,56 @@
...
-
-void DNSServer::setErrorReplyCode(const DNSReplyCode& replyCode)
-{
-	_errorReplyCode = replyCode;
-}
-
-void DNSServer::setTTL(const uint32_t& ttl)
-{
-	_ttl = htonl(ttl);
+	free(buffer);
```
slaff: I guess this should be replaced with `delete[] buffer`.

PR #1575, opened 2019-01-19
HardwarePWM: more fine grained control

verybadsoldier:
* add control functions based on channel numbers
* added option to not updated PWM when setting duty (for bulk channel update)
* added function to actually update PWM (for bulk channel update)

Please consider to add this. It is meant to be able to use channel numbers instead of pin numbers and to do bulk updates of multiple channels.


slaff:
@verybadsoldier We are enforcing coding standard for files under `samples` and `SmingCore`.  Please make sure to install clang-format and then run `make cs`. For more information read the `Coding Style Rules` section in the [release notes for v.3.6.1](https://github.com/SmingHub/Sming/releases/tag/3.6.1).

```
!!! Sming/SmingCore/HardwarePWM.cpp not compliant to coding style, here is the fix:
--- Sming/SmingCore/HardwarePWM.cpp	2019-01-19 11:19:48.659791451 +0000
+++ /dev/fd/63	2019-01-19 11:19:52.999996779 +0000
@@ -81,7 +81,7 @@
  */
 uint32 HardwarePWM::getDutyChan(uint8 chan)
 {
-	if (chan == PWM_BAD_CHANNEL)
+	if(chan == PWM_BAD_CHANNEL)
 		return 0;
 	else
 		return pwm_get_duty(chan);
@@ -109,7 +109,7 @@
 		return false;
 	} else if(duty <= maxduty) {
 		pwm_set_duty(duty, chan);
-		if (update)
+		if(update)
 			updatePWM();
 		return true;
 	} else {
!!! Sming/SmingCore/HardwarePWM.h not compliant to coding style, here is the fix:
--- Sming/SmingCore/HardwarePWM.h	2019-01-19 11:19:48.659791451 +0000
+++ /dev/fd/63	2019-01-19 11:19:53.007997154 +0000
@@ -51,13 +51,13 @@
      */
 	bool analogWrite(uint8 pin, uint32 duty);
 
-    /** @brief  Set PWM duty cycle for a channel
+	/** @brief  Set PWM duty cycle for a channel
      *  @param  channel Channel to set
      *  @param  duty Value of duty cycle to set channel to
      *  @param  update Update PWM output
      */
-	bool setDutyChan(uint8 chan, uint32 duty, bool update=true);
+	bool setDutyChan(uint8 chan, uint32 duty, bool update = true);
 
 	/** @brief  Set PWM duty cycle
      *  @param  pin GPIO to set
@@ -65,7 +65,7 @@
      *  @param  update Update PWM output
      */
-	bool setDuty(uint8 pin, uint32 duty, bool update=true);
+	bool setDuty(uint8 pin, uint32 duty, bool update = true);
 
 	/** @brief  Get PWM duty cycle
 	 *  @param  chan Channel to get duty cycle for
```

slaff:
@mikee47 When you have time can you review this PR?

verybadsoldier:
@slaff 
Ok sorry, updated.

slaff:
@verybadsoldier What is the impact of the changes to users using current HardwarePWM implementation. Are there any side effects or incompatibilities that they should be aware of?

verybadsoldier:
The patch is meant to be fully backwards compatible and only introduces additional functions to directly address PWM channels (`setDutyChan` & `getDutyChan`).

Sming/SmingCore/HardwarePWM.h
```diff
@@ -87,6 +102,10 @@ class HardwarePWM
      */
 	uint32 getMaxDuty();
 
+	/** @brief  This function is used to actually update the PWM.
+	 */
+	void updatePWM();
```
mikee47: Does this need to be a public method?

Sming/SmingCore/HardwarePWM.cpp
```diff
@@ -75,39 +75,61 @@ bool HardwarePWM::analogWrite(uint8 pin, uint32 duty)
 	return setDuty(pin, duty);
 }
 
+/* Function Name: getDutyChan
+ * Description: This function is used to get the duty cycle number for a given channel
+ * Parameters: chan -Esp8266 channel number
+ */
+uint32 HardwarePWM::getDutyChan(uint8 chan)
+{
+	if(chan == PWM_BAD_CHANNEL)
```
mikee47: For consistency, we should be enclosing all conditional statements in braces { }

Sming/SmingCore/HardwarePWM.cpp
```diff
@@ -75,39 +75,61 @@ bool HardwarePWM::analogWrite(uint8 pin, uint32 duty)
...
-bool HardwarePWM::setDuty(uint8 pin, uint32 duty)
+bool HardwarePWM::setDutyChan(uint8 chan, uint32 duty, bool update)
 {
-	uint8 chan = getChannel(pin);
 	if(chan == PWM_BAD_CHANNEL) {
 		return false;
 	} else if(duty <= maxduty) {
 		pwm_set_duty(duty, chan);
-		pwm_start();
+		if(update)
```
mikee47: Braces

mikee47:
> 
> 
> @mikee47 When you have time can you review this PR?

@slaff Looks good, I like the idea of accessing by channel. Also, the changes don't affect the external API just enhance it.

Sming/SmingCore/HardwarePWM.h
verybadsoldier: Yes, it is needed in my opinion because after changing, let's say, duty of 5 channels (via `setDutyChan`) you need a way to activate the new duty values. Just setting `update` to `True` on the last channel you modified is not enough since it might get skipped (and therefore not updating PWM) if the new duty value is equal to the current one.

Sming/SmingCore/HardwarePWM.cpp
verybadsoldier: Updated the braces.

Sming/SmingCore/HardwarePWM.h
mikee47: OK, that's good. Maybe put that into the comment for `setDutyChan` ?

slaff: @verybadsoldier Please, update the comment of  `setDutyChan` and specify that updatePWM() must be called in order to activate the new duty value. 

Another thing from my side - please, update the name of the `HardwarePWM::updatePWM` method to `HardwarePWM::update` and update the documentation and comments.

Once these two changes are done I think we are good to merge the PR.

verybadsoldier: `updatePWM` does neccessarily has to be called after calling `setDutyChan`/`setDuty`. Please notice that `setDutyChan`/`setDuty` will auto-update the PWM by default. This is controlled by the (new) parameter `update` which defaults to `true`.

```
bool setDutyChan(uint8 chan, uint32 duty, bool update = true);
```
```
bool setDuty(uint8 pin, uint32 duty, bool update = true);
```

Calling `updatePWM` manually is only needed after calling one of those two set-Functions with `update=false` (and didn't call another set with `update=true` later).

verybadsoldier: Sorry, added *NOT* in the first line in the last comment... pretty important...

slaff: @verybadsoldier Ok, your explanation sounds good. Can you add it as a comment and change the name of the method to `update`? Once you push these changes and the CI system has checked it we will merge it so that we can release it this week with v3.7.1.

verybadsoldier:
Sorry, I didn't get to make the changes you requested yet. I planned to do it today or tomorrow. Will post a follow-up PR with them tho!

slaff:
@verybadsoldier Ok, I will delay the release until then.

PR #1584, opened 2019-01-28
Revert to default MAC string format for `getMAC` method...

### LINKED ISSUE #1583: {StationClass,AccessPointClass}::getMAC - undocumented change in behaviour
https://github.com/SmingHub/Sming/commit/483f3bfb34bea8cf1247c73be016455673c0e42d changed the behaviour of the getMAC methods on `AccessPointClass` and `StationClass` - they now return a MAC address as `00:11:22:aa:bb:cc`, while previously they returned it as `001122aabbcc`. This is a sudden API change on the `master` branch, seemingly undocumented.

In my opinion, the behaviour change should be rolled back, as it breaks applications that rely on a fixed format of the MAC for identification. I have discovered this due to an application breaking in this exact way.

Preferably, an alternative/extended method would return a MAC object that can be rendered as a string in different formats...

mikee47:
...in `AccessPointClass` and `StationClass`

Fixes #1583, so by default reverts to previous behaviour, but supports alternative formats.

### DIFF
--- a/Sming/SmingCore/Platform/AccessPoint.h
+++ b/Sming/SmingCore/Platform/AccessPoint.h
@@ -70,9 +70,10 @@ public:
 	bool setIP(IPAddress address);
 
 	/** @brief  Get WiFi AP MAC address
+	 *  @param optional separator between bytes (e.g. ':')
      *  @retval String WiFi AP MAC address
      */
-	String getMAC();
+	String getMAC(char sep = '\0');
 
 	/** @brief  Get WiFi AP network mask
      *  @retval IPAddress WiFi AP network mask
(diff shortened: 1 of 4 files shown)


Sming/SmingCore/Platform/AccessPoint.h
```diff
@@ -70,9 +70,10 @@ class AccessPointClass : protected ISystemReadyHandler
 	bool setIP(IPAddress address);
 
 	/** @brief  Get WiFi AP MAC address
+	 *  @param optional separator between bytes (e.g. ':')
      *  @retval String WiFi AP MAC address
      */
-	String getMAC();
+	String getMAC(char sep = '\0');
```
slaff: > as 00:11:22:aa:bb:cc, while previously they returned it as 001122aabbcc

@mikee47 Shouldn't the delimiter be an empty string `''` , instead of `'\0'` ?

slaff: Thanks.

PR #1586, opened 2019-01-28
 Do not flash blankfs.bin if SPIFFS is disabled 

fooker:


slaff:
@fooker I like the PR but it should be done against the `develop` branch. I will close it and you should open a new PR with the right settings. If needed take a look at the [Contributing Guide](https://github.com/SmingHub/Sming/blob/develop/CONTRIBUTING.md)

PR #1589, opened 2019-01-30
Switched to clang-format v6.

slaff:


mikee47:
For info, I've done `make cs` with `clang-format` 7.01 (current) and output is consistent with 6.01.

PR #1593, opened 2019-02-03
StreamTransformer improvements

mikee47:
Add virtual `transform` method to `StreamTransformer` class, optimise `QuotedPrintableOutputStream` and `ChunkedStream`.

transformCallback retained, but no longer required by framework; could deprecate and make `transform` a pure virtual method?

### DIFF
--- a/Sming/SmingCore/Data/Stream/Base64OutputStream.cpp
+++ b/Sming/SmingCore/Data/Stream/Base64OutputStream.cpp
@@ -13,19 +13,15 @@
 // Produce line breaks in output encodings
 const unsigned CHARS_PER_LINE = 72;
 
-Base64OutputStream::Base64OutputStream(IDataSourceStream* stream, size_t resultSize /* = 512 */)
+Base64OutputStream::Base64OutputStream(IDataSourceStream* stream, size_t resultSize)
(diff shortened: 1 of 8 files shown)


Sming/SmingCore/Data/Stream/Base64OutputStream.cpp
```diff
@@ -13,19 +13,15 @@
...
 const unsigned CHARS_PER_LINE = 72;
 
-Base64OutputStream::Base64OutputStream(IDataSourceStream* stream, size_t resultSize /* = 512 */)
+Base64OutputStream::Base64OutputStream(IDataSourceStream* stream, size_t resultSize)
 	: StreamTransformer(stream, nullptr, resultSize, (resultSize / 4))
-
 {
 	base64_init_encodestate(&state, CHARS_PER_LINE);
-
-	transformCallback = std::bind(&Base64OutputStream::encode, this, std::placeholders::_1, std::placeholders::_2,
```
slaff: > Note: transformCallback retained, but no longer required by framework; could deprecate and make `transform` a pure virtual method

Can you add that information as comments in the code? Ex: 
```
/**
*  ...
* @deprecated
*/ 
StreamTransformerCallback transformCallback  __attribute__ ((deprecated))  = nullptr;
```

PR #1595, opened 2019-02-05
TCP improvements

mikee47:
Consistent use of braces
Tidy #includes
NULL -> nullptr
Move trivial code into header
Some methods are const
Ensure ALL member variables have initializers; remove corresponding redundant assignments from constructors
Used unsigned types where appropriate

Remove `TcpServer` as friend of `TcpConnection`, use accessor methods instead

`TcpConnection`: Move callback handler code out of static methods

### DIFF
--- a/Sming/SmingCore/Network/MqttClient.h
+++ b/Sming/SmingCore/Network/MqttClient.h
@@ -9,15 +9,15 @@
 #define _SMING_CORE_NETWORK_MqttClient_H_
 
 #include "TcpClient.h"
-#include "../Network/URL.h"
-#include "../../Wiring/WString.h"
-#include "../../Wiring/WHashMap.h"
+#include "URL.h"
+#include "WString.h"
+#include "WHashMap.h"
 #include "Data/ObjectQueue.h"
 #include "Mqtt/MqttPayloadParser.h"
 #include "../mqtt-codec/src/message.h"
 #include "../mqtt-codec/src/serialiser.h"
 #include "../mqtt-codec/src/parser.h"
-#include <functional>
+//#include <functional>
 
 /** @defgroup   mqttclient MQTT client
  *  @brief      Provides MQTT client
@@ -67,13 +67,16 @@ public:
 	 * Sets keep-alive time. That information is sent during connection to the server
 	 * @param uint16_t seconds
 	 */
-	void setKeepAlive(uint16_t seconds); //send to broker
+	void setKeepAlive(uint16_t seconds) //send to broker
+	{
+		keepAlive = seconds;
+	}
 
 	/**
 	 * Sets the interval in which to ping the remote server if there was no activity
-	 * @param int seconds
+	 * @param unsigned int seconds
 	 */
-	void setPingRepeatTime(int seconds);
+	void setPingRepeatTime(unsigned seconds);
 
 	/**
 	 * Sets last will and testament
@@ -95,7 +98,10 @@ public:
 	bool subscribe(const String& topic);
 	bool unsubscribe(const String& topic);
 
-	void setEventHandler(mqtt_type_t type, MqttDelegate handler);
+	void setEventHandler(mqtt_type_t type, MqttDelegate handler)
+	{
+		eventHandler[type] = handler;
+	}
 
 	/**
 	 * Sets or clears a payload parser (for PUBLISH messages from the server to us)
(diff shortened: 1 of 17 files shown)


Sming/SmingCore/Network/MqttClient.h
```diff
@@ -9,15 +9,15 @@
...
+#include "URL.h"
+#include "WString.h"
+#include "WHashMap.h"
 #include "Data/ObjectQueue.h"
 #include "Mqtt/MqttPayloadParser.h"
 #include "../mqtt-codec/src/message.h"
 #include "../mqtt-codec/src/serialiser.h"
 #include "../mqtt-codec/src/parser.h"
-#include <functional>
+//#include <functional>
```
slaff: Remove the commented line.

Sming/SmingCore/Network/MqttClient.h
```diff
@@ -316,9 +322,9 @@ class MqttClient : protected TcpClient
 
 #ifndef MQTT_NO_COMPAT
 	// @deprecated
-	MqttMessageDeliveredCallback onDelivery = 0;
+	MqttMessageDeliveredCallback onDelivery = nullptr;
```
slaff: Since these are the old-style C++ delegates I think that using nullptr instead of 0 will not work as expected. I seem to recall that we already had such an issue but I am not 100% certain.

Sming/SmingCore/Network/MqttClient.h
```diff
@@ -316,9 +322,9 @@ class MqttClient : protected TcpClient
...
 	// @deprecated
-	MqttStringSubscriptionCallback subscriptionCallback = 0;
+	MqttStringSubscriptionCallback subscriptionCallback = nullptr;
```
slaff: Using nullptr instead of 0 for old-style C++ delegates might not work as expected. Has to be checked.

Sming/SmingCore/Network/TcpConnection.h
```diff
@@ -228,6 +268,16 @@ class TcpConnection
 	virtual err_t onSslConnected(SSL* ssl);
 #endif
 
+	// These do most of the work - called from static methods
+	err_t tcpOnConnected(err_t err);
```
slaff: Can you rename the `tcpOn*` to `interalOn*`? Otherwise having staticOnConnect, tcpOnConnect, onConnect will cause additional confusion as to which method is the right one to use.

While `internalOn*` might be a bit better than `tcpOn*` and not quite happy with it so I am also open to suggestions from your side.

Sming/SmingCore/Network/MqttClient.h
mikee47: Should  match `Delegate(FunctionDeclaration m)` constructor as `FunctionDeclaration` is a pointer type `ReturnType (*FunctionDeclaration)(ParamsList...)`. Can't see how else it could be interpreted?

Sming/SmingCore/Network/TcpConnection.h
mikee47: We could adopt the LWIP convention e.g. `tcp_connected` instead of `tcpOnConnect` to indicate it's all happening at a lower level. Could do the same with the static methods, e.g. `static_tcp_connected`...

mikee47: Just noticed, we already have an `internalTcpConnect`!

mikee47: This will come up again handling HTTP and Websocket parser callbacks. These are simpler though as the static methods are all virtually identical (just a null pointer check then call the method). Both parsers use a function table, so I adopted the same names as those to help make things clearer, e.g. `HttpConnection::on_message_begin`.

slaff: > Just noticed, we already have an `internalTcpConnect`!

Nice catch :)

For consistency I can suggest the following method renames:
```
internalTcpConnect -> internalConnect
tcpOn* -> internalOn* (Ex: tcpOnConnected -> internalOnConnected )
```

Can you do that change? And don't forget to remove the commented line 20 in Sming/SmingCore/Network/MqttClient.h.

slaff:
@mikee47 Is the PR  tested from your side and ready for merging or you plan to add more changes?

mikee47:
> @mikee47 Is the PR tested from your side and ready for merging or you plan to add more changes?

That's it for this PR.

PR #1601, opened 2019-02-07
Http connection base improvements

mikee47:
General tidy up, moving `parserSettings` into flash saves 40 bytes.
Builds and runs fine.

### DIFF
--- a/Sming/SmingCore/Network/Http/HttpConnectionBase.cpp
+++ b/Sming/SmingCore/Network/Http/HttpConnectionBase.cpp
@@ -12,39 +12,37 @@
 
 #include "HttpConnectionBase.h"
 
-bool HttpConnectionBase::parserSettingsInitialized = false;
-http_parser_settings HttpConnectionBase::parserSettings;
+/** @brief http_parser function table
+ *  @note stored in flash memory; as it is word-aligned it can be accessed directly
+ *  Notification callbacks: on_message_begin, on_headers_complete, on_message_complete
+ *  Data callbacks: on_url, (common) on_header_field, on_header_value, on_body
+ */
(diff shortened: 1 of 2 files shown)


slaff:
@mikee47 A new worrying message started to show up. It started to appear since the latest PRs (This one included). 

I am not sure which PR is causing it exactly but the message is:
```
86494066 Warning: HttpRequest streams are not copied..
E:M 0
```

The 'E:M 0' comes from the SDK and is usually showing up to specify that there is an error with allocating the desired size of memory. For example `E:M 1024` means that there was Error:Memory not able to locate 1024 in one piece. 

The code continues to work without issues but it might mean that somewhere we try to allocate memory segment with size 0, which is not quite useful and may pinpoint to a new issue in the code. 

Can you take a look at it when you have some time?

mikee47:
So the `Warning: HttpRequest streams are not copied` was introduced on 25/6/18, in the `HttpRequest(const HttpRequest&)` constructor. As far as I can see, the only place it gets is called by the `clone()` method, which in turn is only used in `rBootHttpUpdate::start()` - that's been there since forever.

As for the The `E:M 0` message I suspect it's almost certainly unrelated to the `HttpRequest` warning. I _think_ the SDK outputs it's debug messages via `uart_debug_putc` in `system\uart.cpp`; if so could modify that function to buffer the previous few characters and dump a stack trace when it detects `E:M` in the stream... GDB is another option but haven't used that with Sming yet.

For confirmation of all this, I use port 1 for debug output with a putty session running all the time; everything gets logged to file. The logs date back to August, so include all the output from PR tests. The `HttpRequest` warning doesn't appear anywhere, which supports it only being invoked via `rBootHttpUpdate` (which I've never used). The `E:M 0` does appear but that was specifically with code I was developing and the cause since fixed.

Issue #1520, opened 2018-11-13
problem with OTA in Basic_rBoot


petrynchyn:
I have problem with OTA in sample Basic_rBoot, please check someone this example in yourself.

```
make flash
Killing Terminal to free /dev/ttyUSB0
/opt/Sming/Sming/Makefile-rboot.mk:614: recipe for target 'flash' failed
make: [flash] Вбито (ігнорується)
/opt/esp-open-sdk/esptool/esptool.py -p /dev/ttyUSB0 -b 115200 write_flash -ff 40m -fm qio -fs 32m 0x00000 out/firmware/rboot.bin 0x02000 out/firmware/rom0.bin 0x100000 out/firmware/spiff_rom.bin
WARNING: Flash size arguments in megabits like '32m' are deprecated.
Please use the equivalent size '4MB'.
Megabit arguments may be removed in a future release.
esptool.py v2.2-dev
Connecting....
Detecting chip type... ESP8266
Chip is ESP8266EX
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Compressed 2032 bytes to 1429...
Wrote 2032 bytes (1429 compressed) at 0x00000000 in 0.1 seconds (effective 122.6 kbit/s)...
Hash of data verified.
Compressed 319168 bytes to 220389...
Wrote 319168 bytes (220389 compressed) at 0x00002000 in 19.5 seconds (effective 131.1 kbit/s)...
Hash of data verified.
Compressed 65536 bytes to 191...
Wrote 65536 bytes (191 compressed) at 0x00100000 in 0.0 seconds (effective 22285.3 kbit/s)...
Hash of data verified.

Leaving...
Hard resetting...
python -m serial.tools.miniterm /dev/ttyUSB0 115200 
--- Miniterm on /dev/ttyUSB0  115200,8,N,1 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---
��␃�␛�g�s��o|�␌l␌lll ␃��{�l��: 079069 trying to mount spiffs at 0x00100000, length 65536
79378 fs.start: size:64 Kb, offset:0x100000

81006 mount res: 0

81624 Station configuration is: MKVL

Currently running rom 0.
Type 'help' and press enter for instructions.

mode : sta(60:01:94:2b:ae:0b)
add if0
scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
[...]

PR #1604, opened 2019-02-08
Fix OTA. Disable interrupts during flash erase/write

etmmahi:
Update the patch file for rboot.
This disables interrupts during flash erase/write operations during OTA updates

### DIFF
--- a/Sming/third-party/.patches/rboot.patch
+++ b/Sming/third-party/.patches/rboot.patch
@@ -38,14 +38,14 @@ index ca234de..0c830d8 100644
  	CFLAGS += $(addprefix -I,$(RBOOT_EXTRA_INCDIR))
  endif
 diff --git a/appcode/rboot-api.c b/appcode/rboot-api.c
-index 523560f..dada359 100644
+index 523560f..110718d 100644
 --- a/appcode/rboot-api.c
 +++ b/appcode/rboot-api.c
 @@ -28,6 +28,8 @@
  extern "C" {
  #endif
  
-+#define FLASH_INT_MASK ((1 << ETS_SDIO_INUM) | (1 << ETS_GPIO_INUM) | (1 << ETS_UART_INUM) | (1 << ETS_FRC_TIMER1_INUM))
++#define FLASH_INT_MASK ((1 << ETS_GPIO_INUM) | (1 << ETS_UART_INUM) | (1 << ETS_FRC_TIMER1_INUM))
 +
  #if defined(BOOT_CONFIG_CHKSUM) || defined(BOOT_RTC_ENABLED)
  // calculate checksum for block of data
--- a/Sming/third-party/.patches/rboot.patch
+++ b/Sming/third-party/.patches/rboot.patch
@@ -1,16 +1,16 @@
 diff --git a/Makefile b/Makefile
-index ca234de..5f1c72b 100644
+index ca234de..0c830d8 100644
 --- a/Makefile
 +++ b/Makefile
 @@ -22,10 +22,16 @@ else
  Q := @
  endif
-
+ 
 -CFLAGS    = -Os -O3 -Wpointer-arith -Wundef -Werror -Wl,-EL -fno-inline-functions -nostdlib -mlongcalls -mtext-section-literals  -D__ets__ -DICACHE_FLASH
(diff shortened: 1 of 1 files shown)


etmmahi:
See issue #1520

slaff:
@petrynchyn @Ph3r0x Can you check if this PR is fixing the problem that you are having with OTA?

slaff:
@etmmahi In SDK 1.5 `ETS_SDIO_INUM` does not seem to be defined? Can you update the code set the right definition, if not already defined?

```
/home/travis/build/SmingHub/Sming/Sming/third-party/rboot/appcode/rboot-api.c:71:15: note: in expansion of macro 'FLASH_INT_MASK'
  ets_isr_mask(FLASH_INT_MASK);
               ^
/home/travis/build/SmingHub/Sming/Sming/third-party/rboot/appcode/rboot-api.c: In function 'rboot_write_flash':
/home/travis/build/SmingHub/Sming/Sming/third-party/rboot/appcode/rboot-api.c:31:31: error: 'ETS_SDIO_INUM' undeclared (first use in this function)
 #define FLASH_INT_MASK ((1 << ETS_SDIO_INUM) | (1 << ETS_GPIO_INUM) | (1 << ETS
                               ^
/home/travis/build/SmingHub/Sming/Sming/third-party/rboot/appcode/rboot-api.c:161:17: note: in expansion of macro 'FLASH_INT_MASK'
    ets_isr_mask(FLASH_INT_MASK);
                 ^
make[1]: *** [out/build//home/travis/build/SmingHub/Sming/Sming/third-party/rboot/appcode/rboot-api.o] Error 1
make[1]: Leaving directory `/home/travis/build/SmingHub/Sming/samples/Accelerometer_MMA7455'
make: *** [Accelerometer_MMA7455] Error 2
The command "make samples" exited with 2.
```

slaff:
So the code changes are workaround for the following problem:  
There is, or are, interrupt routines that are not in IRAM. It would be better to find these routines and fix them. Can you paste here the watchdog reset stack dump? Maybe in it stays the name of the function that is causing the issue.

etmmahi:
Correct. This exactly the problem.  By playing with the flags in the interrupt  it is easy to pinpoint the problem. If I leave out the UART bit, the OTA fails! So somewhere in the UART interrupt chain which is executed in the interrupt, there is a function NOT in IRAM. Just find it :-;

andr2000:
Hm, in the rboot example we have serial callback... which is in flash... and is called from the interrupt handler... Can it be why we fail?

etmmahi:
Just did some more experiments. It seems the 'printing' just before the OTA update is causing problems. There is still serial output + interrupts going on when the OTA code is activated. This means there is something in the serial interrupt chain not in IRAM. When we cannot solve this, the workaround with disable interrupts is a good fix,

andr2000:
@etmmahi the strings we print reside in flash and uart_write has the following comment:
`		// If TX buffer not in use or it's empty then write directly to hardware FIFO`
Can it be that we are trying to touch flash while we shouldn't?

slaff:
> If I leave out the UART bit, the OTA fails! So somewhere in the UART interrupt chain ...

@mikee47 When you have some time can you check what can be the issue with the modified UART? There most probably is an IRQ function that is not in IRAM.

mikee47:
@slaff I traced through the interrupt path again last night, and all code is either IRAM_ATTR or __forceinline. I have a suspicion the problem is more subtle or we'd have seen this sooner. I'll think on this!

etmmahi:
Fix #1610 tested, it supersedes this pull request. Closing this one

Ph3r0x:
@slaff sorry for my late response
i've updated Sming with:
make dist-clean
make DEBUG_VERBOSE_LEVEL=3 ENABLE_SSL=1

and i am still getting resettet after a few seconds
my fix for that is the following:
init_method{
...
	// Setup the WIFI connection
	WifiStation.enable(true);

	WifiEvents.onStationGotIP(connectOk);
}

void connectOk(IPAddress ip, IPAddress mask, IPAddress gateway){
	Serial.setCallback(serialCallBack);
}

i dont know but is make dist-clean and make again enough to get the latest develop commit?

mikee47:
> is make dist-clean and make again enough to get the latest develop commit?
@Ph3r0x `git checkout develop`, then `git pull` first

PR #1606, opened 2019-02-08
Add null/zero-length checks in `SSLKeyCertPair::assign()`

mikee47:
Fixes `E:M 0` messages as reported in #1601.
Did some exploration with `HttpClient` sample which reproduces this issue, now fixed.


slaff:
@mikee47 Thanks a lot!

Issue #1523, opened 2018-11-15
HttpServer_WebSocket sample and broadcast before server.shutdown()


Evgsurf:
Hello.

I build and run HttpServer_WebSocket sample.
I changed index.html code and add text field and button to send data to server
This code added before output div
`<div><input id='sendtext'><input type="button" id="disconnect" onclick="doSend(document.getElementById('sendtext').value)" value="Send"/></div>
`
So, when I open serveral session (different tabs in browser) server sends messages to each session by broadcast command
```
	String message = "New friend arrived! Total: " + String(totalActiveSockets);
	socket.broadcast(message.c_str(), message.length());
```

`RESPONSE: New friend arrived! Total: 2`

But when I type **shutdown** message/command, only active session get message 
`The server is shutting down...`

by code 
```
		String message = "The server is shutting down...";
		socket.broadcast(message.c_str(), message.length());

```
Other sessions do not get this broadcasted message.

mikee47:
Sounds probable that the HTTP server is getting shut down before all messages can be sent. Try commenting out line 58 `server.shutdown();` to see if that fixes it. It would be better anyway to call `server.shutdown()` using a timer to delay by a few seconds.

mikee47:
Try replacing `server.shutdown();` with this:
```
	  auto timer = new SimpleTimer;
	  timer->setCallback([](void* timer) {
		delete static_cast<SimpleTimer*>(timer);
		server.shutdown();
	  }, timer);
	  timer->startMs(3000);
```

Evgsurf:
Of couse it does work correctly!
But hardcode 3 secs is not very good by me.

I guess, it will better to user another way

  ```
auto timer = new SimpleTimer;
	  timer->setCallback([](void* timer) {
                if (server. READY()) { // I am not sure what is function is neccessary to use
		      delete static_cast<SimpleTimer*>(timer);
		      server.shutdown();
                }
	  }, timer);
	  timer->startMs(100);
```

Issue #1275, opened 2017-11-02
Samples rBoot Fatal on 4M flash


lrodorigo:
Hi. I'm getting the Fatal exception 28 when trying the rBoot sample on an Ai-Thinker ESP-12-F (no SSL download, I don't know if can be related to https://github.com/SmingHub/Sming/issues/1167).
I noticed that the firmware seems to be flashed, but the rom switch is not working.
The next reboot always happens from rom 0.
I didn't copy any linked file  into the sample folder, because I am using  the example as-is on a 4M flash.

I also get two distinct fatal exceptions, here's the debug output and the two stack traces:

**Debug Output:**
```
staticOnMessageComplete: Execution queue: 0, http://xxxxxx/ver0.bin

Firmware download finished!
 - item: 0, addr: 202000, len: 287360 bytes
In callback...
Firmware updated, rebooting to rom 1...
state: 5 -> 0 (0)
rm 0
TCP connection error: -8
del if0
usl
Firmware updated.


***** Fatal exception 28
pc=0x401063f7 sp=0x3ffff9f0 excvaddr=0x0000007c
ps=0x00000030 sar=0x00000010 vpri=0x40100de4
r00: 0x401063f2=1074815986 r01: 0x3ffff9f0=1073740272 r02: 0x00000000=         0
r03: 0x00000000=         0 r04: 0x60000000=1610612736 r05: 0x0000007e=       126
r06: 0x6000001c=1610612764 r07: 0x00000004=         4 r08: 0x3fff05b0=1073677744
r09: 0x00000130=       304 r10: 0x00000000=         0 r11: 0x3ffef790=1073674128
r12: 0x3fff1110=1073680656 r13: 0x00000000=         0 r14: 0x3fff11b0=1073680816
r15: 0x3ffffaf0=1073740528

```
**first stack trace:**
```
0x401063f7: HttpConnection::staticOnMessageComplete(http_parser*) at /opt/Sming/Sming/SmingCore/Network/Http/HttpConnection.cpp:125
0x40100de4: pvPortRealloc at ??:?
0x401063f2: Delegate<int (HttpConnection&, bool)>::operator()(HttpConnection&, bool) const at /opt/Sming/Sming/SmingCore/Network/Http/HttpConnection.cpp:125
 (inlined by) HttpConnection::staticOnMessageComplete(http_parser*) at /opt/Sming/Sming/SmingCore/Network/Http/HttpConnection.cpp:209
[...]

PR #1611, opened 2019-02-09
Add optional `deferMillis` parameter to `System::restart()`

mikee47:
A system restart may be requested from within a deeply-nested callback routine, such as on network transfer completion. This happens in the `Basic_rBoot` sample application, which I have found occasionally crashes at the end of an update cycle; this is probably more likely in debug mode where there is more going on.

In such scenarios it is best to defer the actual restart request so that the callback can unwind and complete fully. Therefore, this modification will _always_ defer a restart request, using either the task queue (if deferMillis == 0) or a system timer if a longer delay is required by the application.

### DIFF
--- a/Sming/SmingCore/Platform/System.cpp
+++ b/Sming/SmingCore/Platform/System.cpp
@@ -6,6 +6,7 @@
  ****/
 
 #include "System.h"
+#include "SimpleTimer.h"
 
 SystemClass System;
 
@@ -80,6 +81,22 @@ void SystemClass::onReady(ISystemReadyHandler* readyHandler)
 				handler->onSystemReady();
 			},
 			reinterpret_cast<uint32_t>(readyHandler));
+	}
+}
+
+void SystemClass::restart(unsigned deferMillis)
+{
+	if(deferMillis == 0) {
+		queueCallback([](uint32_t) { system_restart(); });
+	} else {
+		auto timer = new SimpleTimer;
+		timer->setCallback(
+			[](void* timer) {
+				delete static_cast<SimpleTimer*>(timer);
+				system_restart();
+			},
+			timer);
+		timer->startMs(deferMillis);
 	}
 }
 
--- a/Sming/SmingCore/Platform/System.h
+++ b/Sming/SmingCore/Platform/System.h
@@ -107,12 +107,13 @@ public:
 		return state == eSS_Ready;
 	}
 
-	/** @brief  Restart system
+	/** @brief Request a restart of the system
+	 *  @param deferMillis defer restart request by a number of milliseconds
+	 *  @note A delay is often required to allow network callback code to complete correctly.
+	 *  The restart is always deferred, either using the task queue (if deferMillis == 0)
+	 *  or using a timer. This method always returns immediately.
      */
-	void restart()
-	{
-		system_restart();
-	}
+	void restart(unsigned deferMillis = 0);
 
 	/** @brief  Set the CPU frequency
      *  @param  freq Frequency to set CPU


mikee47:
See also #1523. There are other framework calls which would benefit from this approach, e.g.#1598

mikee47:
I found with testing rBoot sample that no delay is actually required, deferring the call is sufficient. I could simplify this PR... thoughts?

mikee47:
Probably fix for #1275

slaff:
> I could simplify this PR... thoughts?

Can you elaborate on that?

mikee47:
> > I could simplify this PR... thoughts?
> 
> Can you elaborate on that?

Well, only the question of whether it's _necessary_ or useful to have the `deferMillis` parameter at all - could just use the task queue if that is all that is required to fix the problem. Can always leave it up to the user application to use timers, etc. if required.

slaff:
What bothers me about this PR is that it fixes an issue that we don't quite well understand. Is the SDK causing the segfault, is the UART code causing it or it is something else? 

I am having a suspicion that we are proposing a wrong fix. The following is true for ESP8266 and NodeMCU in particular: 
If you flash the device and call software restart having the UART lines up it will  cause a WDT reset. Similar to the following:

> Then I reasoned that as I was using GPIO as an output, maybe the hardware wasn't reseting it to an input on a soft reset, so the reset would depend upon the state of GPIO#0.
> 
> So, I set GPIO#0 high before doing the restart and it works!

 
From here https://github.com/esp8266/Arduino/issues/1722

What I am thinking is that just  setting GIO0 to high will be enough. Can you try this?

mikee47:
I suspect that's a different, unrelated issue. I'll use the `basic_rBoot` sample as a specific example. Here's what happens with a straight call to `system_restart()`:

```
26229500 staticOnMessageComplete: Execution queue: 0, http://192.168.1.68:80/rom0.bin
26230111 
Firmware download finished!
26235083  - item: 0, addr: 2000, len: 307872 bytes
In callback...
Firmware updated, rebooting to rom 0...
Restarting...

state: 5 -> 0 (0)
rm 0
pm close 7
26300490 TCP connection error: -8
del if0
usl
26301301 Firmware updated.

***** Fatal exception 28
```

Now with the deferred restart:

```
11839537 staticOnMessageComplete: Execution queue: 0, http://192.168.1.68:80/rom0.bin
11840147 
Firmware download finished!
11845128  - item: 0, addr: 2000, len: 307760 bytes
In callback...
Firmware updated
Swapping from rom 1 to rom 0.
Restarting...

11902755 Firmware updated.
11903240 TCP connected
11903366 HttpConnection::onReadyToSendData: waitingQueue.count: 0
11903545 Nothing in the waiting queue
11908313 TCP onReadyToSendData: 0
11913276 TCP received: 813 bytes
11917263 HttpConnection::onReadyToSendData: waitingQueue.count: 0
11926034 Nothing in the waiting queue
11931114 TCP onReadyToSendData: 1
state: 5 -> 0 (0)
rm 0
pm close 7
11940916 TCP connection error: -8
del if0
usl
*P*VETT*uU+UZUTTQTQQjP*uZTUQTj*QEuP+TQTQQBUuZT(EQREj*QREQQujUKZVVUU(TTUUEJPQZUJZEEQQu],QuVUPQota1 not set
ota2 not set
>> 0x3JQRQrf cal sector: 1019
freq trace enable 0
rf[112] : 00
98375 spiffs disabled
```

My immediate impression is that the rug is getting pulled out from under the network stack and leading to the crash. Doubtless there is work to be done to specifically address _why_ the `Exception 28` occurs - I will take a look at that in due course - but this PR simply aims to make the transition a bit cleaner and improve reliability of the operation.

slaff:
> My immediate impression is that the rug is getting pulled out from under the network stack and leading to the crash.

I would have expected that the Espressif guys have already taken care of this, but without the source code I cannot tell. Will try asking @me-no-dev and @igrr if they can shed some light.

igrr:
I haven't looked at the SDK code in a while, but I think calling system_restart from an LwIP callback is not a great idea, as that call will tear down the network interface. Deferring that using a task queue seems to be the correct approach.

slaff:
> could just use the task queue if that is all that is required to fix the problem.

## I. If this is the only case where restart segfaults 

@mikee47 With the information from @igrr I would suggest the following - remove the new code from System::restart and add a new method called `restart` in the rBoot client  that has the task queue code. The method needs to have a description explaining why it is here and why it should be called instead of System::restart. And finally the sample(s) should be updated accordingly. 

## II. If there are more places where restart segfaults

Then we could think about adding System::restart to a queue. But this might hide the real cause of the problem.

What do you think?

mikee47:
The SDK API docs. say of `system_restart`: 'The ESP8266 will not restart immediately. Please do not call other functions after calling this API.'. That would seem to imply that other system functions are affected, not just the network APIs.

I think using the 'does it segfault?' approach is perhaps risky here - it may only do it sometimes! The risk of masking problem code could be mitigated by disabling the behaviour during testing - perhaps another compiler flag.

For user applications, we should provide an API which makes life as easy as possible for user applications. That means _always_ deferring the call, and providing the option to include an additional delay so their own stuff can finish up if necessary.

So I'm voting in favour of the general case which this PR addresses.

slaff:
> So I'm voting in favour of the general case which this PR addresses.

I had a second thought on this and have to agree with you. This will bring stability without sacrificing performance or resource usage so it's fine.

PR #1612, opened 2019-02-10
Fix/mqtt fifo memleak

### LINKED ISSUE #1576: Frequent mqtt publish eats heap
Hello.

I have a problem that is probably related to String.

If the code is executed each second, the heap gets exhausted in few minutes. About a kB each few tens of seconds.

If the timer is increased to two seconds, then the heap stays at almost constant value (currently at 38856 for few minutes already).

I think that I saw somewhere through the issues a similar problem, but can't recall the topic.

Here is the code.
PS: I tried to pass the strings by reference, but this doesn't change anything.
PS2: I tried to track down if some memory is allocated within the mqtt class and couldn't resolve how this code releases the memory. However it should work properly since my other projects work fine.

[MqttClient.cpp](https://github.com/SmingHub/Sming/blob/develop/Sming/SmingCore/Network/MqttClient.cpp)
```c++
#define COPY_STRING(TO, FROM)                                                                                          \
	{                                                                                                                  \
		TO.length = FROM.length();                                                                                     \
		TO.data = (uint8_t*)malloc(FROM.length());                                                                     \
		if(!TO.data) {                                                                                                 \
			debug_e("Not enough memory");                                                                              \
			return false;                                                                                              \
		}                                                                                                              \
		memcpy(TO.data, FROM.c_str(), FROM.length());                                                                  \
	}
```

```c++
struct electricData {
	float V;
	float I;
	float P;
	uint32_t status;
	uint32_t conf;
};
```

```c++
void publishElectricity() {
	electricData elData = measureElectricity();

	String topic;
	//topic.reserve(1);

	String data;
	//data.reserve(5);

	topic = String( "P" );
	data = String( elData.P , 5 );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "V" );
	data = String( elData.V , 4 );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "I" );
	data = String( elData.I , 4 );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "kWh" );
	data = String( kWh , 4 );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "status" );
	data = String( elData.status );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "conf" );
	data = String( elData.conf );
	mqttClient.pubMsgRef( topic , data );

	topic = String( "heap" );
	data = String( system_get_free_heap_size() );
	mqttClient.pubMsgRef( topic , data );

/*
	mqttClient.pubMsg( String("P") , String( elData.P , 5 ) );
	mqttClient.pubMsg( String("V") , String( elData.V , 4 ) );
	mqttClient.pubMsg( String("I") , String( elData.I , 4 ) );
	mqttClient.pubMsg( String("kWh") , String( kWh , 4 ) );
	mqttClient.pubMsg( String("status") , String( elData.status ) );
	mqttClient.pubMsg( String("conf") , String( elData.conf ) );
	mqttClient.pubMsg( String("heap") , String( system_get_free_heap_size() ) );
	*/
}
```

```c++
void mqttHelper::pubMsgRef( String& topic , const String& data ) {
	if(mqtt.getConnectionState() == eTCS_Connected) {
		topic = MqttConfig.DeviceTopic + topic;
		mqtt.publish( topic , data);
	}
	//auto reconnect?
}
```

At the end - heap is still at 38856...

etmmahi:
When adding too many messages to the MQTT requestqueue they get dropped when the queue is full. In this case, the allocated memory is not released. This fix first checks if there is room in  the queue.

May also fix #1576

### DIFF
--- a/Sming/SmingCore/Network/MqttClient.cpp
+++ b/Sming/SmingCore/Network/MqttClient.cpp
@@ -238,6 +238,10 @@ bool MqttClient::connect(const URL& url, const String& clientName, uint32_t sslO
 
 bool MqttClient::publish(const String& topic, const String& content, uint8_t flags)
 {
+	if (requestQueue.full()) {
+		return false;
+	}
+
 	mqtt_message_t* message = (mqtt_message_t*)malloc(sizeof(mqtt_message_t));
 	mqtt_message_init(message);
 	message->common.type = MQTT_TYPE_PUBLISH;
@@ -256,6 +260,10 @@ bool MqttClient::publish(const String& topic, IDataSourceStream* stream, uint8_t
 {
 	if(!stream || stream->available() < 1) {
 		debug_e("Sending empty stream or stream with unknown size is not supported");
+		return false;
+	}
+
+	if (requestQueue.full()) {
 		return false;
 	}
 
@@ -278,6 +286,10 @@ bool MqttClient::subscribe(const String& topic)
 {
 	debug_d("subscription '%s' registered", topic.c_str());
 
+	if (requestQueue.full()) {
+		return false;
+	}
+
 	mqtt_message_t* message = (mqtt_message_t*)malloc(sizeof(mqtt_message_t));
 	mqtt_message_init(message);
 	message->common.type = MQTT_TYPE_SUBSCRIBE;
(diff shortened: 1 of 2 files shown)


slaff:
@etmmahi Thanks for the fix!

PR #1614, opened 2019-02-10
Expand `SSLSessionId` structure to manage allocated memory.

mikee47:


Sming/SmingCore/Network/TcpConnection.h
```diff
@@ -156,10 +156,55 @@ struct SSLKeyCertPair {
 	}
 };
 
-typedef struct {
+/** @brief Manages buffer to store SSL Session ID
+ */
+struct SSLSessionId {
```
slaff: The TcpConnection.h is starting to have too much code that is not strictly Tcp related but SSL related. Can you put SSLSessionId, SslFingerprintType, SSLFingerprints  and  SSLKeyCertPair into a separate header file?

mikee47:
It occurred to me doing all this that all three structures could be more easily implemented using `String`s...

Sming/SmingCore/Network/TcpConnection.h
slaff: > Can you put SSLSessionId, SslFingerprintType, SSLFingerprints and SSLKeyCertPair into a separate header file?

Sorry, I was not clear enough. I've meant one header file and not header file per structure. Can you put them back to one header file OR create a separate header file in which all new SSL related header files are included?

mikee47: I'll be elaborating on `SSLFingerprints` so thought it best to split them functionally - it's like the 'fingerprint API' header. Yes, can certainly use a common header - is it worth having an `SSL` sub-folder perhaps?

slaff:
>  is it worth having an SSL sub-folder perhaps?

Sounds good to me. In it you can put also the SSL Validator stuff. 

> It occurred to me doing all this that all three structures could be more easily implemented using `String`

That would be great. As long as it simplifies the code.

mikee47:
There are a few of naming inconsistencies:

`SslValidatorCallback` -> `SSLValidatorCallback`
`SslFingerprintType` -> `SSLFingerprintType`
`SslValidator.h` -> `SSLValidator.h`
`SslValidator.cpp` -> `SSLValidator.cpp`

Do you want those changed, or left as-is?

slaff:
> Do you want those changed, or left as-is?

Yes, please. Make them consistent. Since we don't use all caps for abbreviations, for example for TCP* we use Tcp*, I would suggest that we follow the same rule also for SSL. That said SSL* should be Ssl*. 
At the moment there are expectations from the rule, like DNSServer instead of DnsServer but that should not be part of this PR.

slaff:
@mikee47 Mike, before I merge this PR can you provide a list of incompatible changes? 
For example `keyPassword` is no longer public and in order to access it one needs to use getKeyPassword().

mikee47:
@slaff None of these changes should break users' applications as neither `SSLSessionId` nor `SSLKeyCertPair` are used in any user APIs. `SSLFingerprintType` is used as an API parameter so it's possible some user apps have it as well. Shall I add a deprecated version of that to the PR?

Possible breaking changes:

struct SSLKeyCertPair -> class SslKeyCertPair
  certificate -> getCertificate()
  certificateLength -> getCertificateLength()
  key -> getKey()
  keyLength -> getKeyLength()
SSLFingerprintType -> SslFingerprintType

Deprecated so old version still works with compiler warning:

SSLFingerprints -> SslFingerprints
SSLKeyCertPair -> SslKeyCertPair
SSLSessionId -> SslSessionId

PR #1616, opened 2019-02-10
Fix/ssl fingerprints

mikee47:
Improve memory management of SSL fingerprint data by extending SSLFingerprints structure and moving into separate header file. Create `SSLValidatorList` class for `TCPClient` and move relevant code into methods.

Uses may now use methods of `SSLFingerprints` to provide certificate data. I've updated the `Basic_Ssl` and `HttpClient` sample applications accorginly.


mikee47:
This (should) just be a re-factoring of existing code without changing the validation mechanism. However, it feels to me like a very complex way to crack what appears to be a fairly simple problem.

The user's application passes their fingerprints to HttpRequest, which then passes them into TcpClient. Validation eventually happens in `TcpClient::onSslConnected`. This is essentiaally a 'one-time fingerprint cache' but we could bypass all of this by just using a callback to the user's application instead.

Am I missing anything?

PR #1618, opened 2019-02-12
Improve memory management of SSL fingerprint data

mikee47:
Uses may now use methods of `SslLFingerprints` to provide certificate data. I've updated the `Basic_Ssl` and `HttpClient` sample applications accordingly.

Add methods to SslFingerprints structure to manage allocation/de-allocation of fingerprint data.
Create `SslValidatorList` class for `TcpClient` and move relevant code into methods.
Take care to set `SslFingerprints` members to `nullptr` when ownership relinquished - previous code `TcpClient::pinCertificate()`.
bugfix: In `TcpClient::onSslConnected`, the first matching validator causes checking to stop, which means the validator dataj for any subsequent fingerprints never gets released. This is fixed in the new `SslValidatorList` implementation.
Revise `Basic_Ssl` sample to use different web-page with fingerprint that doesn't change.
Also change `HardwareSerial` to base on `ReadWriteStream` instead of `Stream`, allows incoming data to be streamed direct.


mikee47:
This is more-or-less just be a re-factoring of existing code without changing the validation mechanism. However, it feels to me like an overly complex way to crack what should be a fairly simple problem.

The user's application passes their fingerprints to HttpRequest, which keeps them as an `SslFingerprints` structure. They are then passed into TcpClient, which pulls the values out into a hashmap. Validation eventually happens in `TcpClient::onSslConnected`. This is essentially a 'one-time fingerprint cache'; the mechanism could be more simply implemented wrapping everything in the SslValdation class, then passing that around.

Alternatively, we could bypass all of this by just using a callback (to the user's application instead...

Thoughts / comments?

Sming/SmingCore/HardwareSerial.h
```diff
@@ -239,11 +244,21 @@ class HardwareSerial : public Stream
 	 *  @note Although this shares the same name as the method in IDataSourceStream,
 	 *  behaviour is different because in effect the 'seek' position is changed by this call.
 	 */
-	virtual uint16_t readMemoryBlock(char* buf, size_t max_len)
+	virtual uint16_t readMemoryBlock(char* buf, int max_len)
```
mikee47: Note that 'override' would have caught this :-)

mikee47:
@slaff As I've started using `__attribute__((deprecated))` I can prepare a PR for all the remaining items which only have `@deprecated` comment.

Another similar general change was use of `override`.

Both of these I think shoukd be introduced sooner rather than later...

Sming/SmingCore/Network/Ssl/SslFingerprints.h
```diff
@@ -14,15 +14,107 @@ enum SslFingerprintType {
...
+	bool setSha256_P(const uint8_t* cert, unsigned length)
+	{
+		// Word-aligned and sized buffers don't need special handling
+		return setSha256(cert, length);
+	}
+
+	/** @brief Moves values out of source */
+	SslFingerprints& operator=(SslFingerprints& source)
+	{
+		delete[] certSha1;
```
mikee47: Add self check

Sming/SmingCore/Network/Ssl/SslFingerprints.h
```diff
@@ -14,15 +14,107 @@ enum SslFingerprintType {
...
+		pkSha256 = source.pkSha256;
+		source.pkSha256 = nullptr;
+
+		return *this;
+	}
+
+	/** @brief Make copy of values from source */
+	SslFingerprints& operator=(const SslFingerprints& source)
+	{
+		setSha1(source.certSha1, SHA1_SIZE);
```
mikee47: Add self check

Sming/SmingCore/Network/Ssl/SslValidator.h
```diff
@@ -16,20 +16,40 @@
...
 
 typedef std::function<bool(SSL* ssl, void* data)> SslValidatorCallback;
 
 bool sslValidateCertificateSha1(SSL* ssl, void* data);
 
 bool sslValidatePublicKeySha256(SSL* ssl, void* data);
 
-#endif /* ENABLE_SSL */
+struct SSLValidator {
+	SslValidatorCallback callback;
```
mikee47: Could ditch callbacks and use `SslFingerprintType` with `switch`, can then include length check

Sming/SmingCore/Network/Ssl/SslValidator.h
```diff
@@ -16,20 +16,40 @@
...
+	void* data; ///< Fingerprint
+};
+
+class SSLValidatorList : private Vector<SSLValidator>
```
mikee47: Any advantage to using a separate list for validation? Could this be wrapped into `SslFingerprints`?

slaff:
> @slaff As I've started using `__attribute__((deprecated))` I can prepare a PR for all the remaining items which only have `@deprecated` comment.
> Another similar general change was use of `override`.

Yes, please. Make sure also to modify the https://github.com/SmingHub/Sming/wiki/Coding-Style-Rules wiki page to reflect those changes.

samples/HttpClient/app/application.cpp
```diff
@@ -98,24 +98,28 @@ void setSslFingerprints(HttpRequest* request)
...
+	fingerprints.setSha256_P(publicKeyFingerprint, sizeof(publicKeyFingerprint));
 
 	/*
 	 * The line below shows how to trust only a certificate that matches the SHA1 fingerprint.
 	 * When google changes their certificate the SHA1 fingerprint should not match any longer.
 	 */
-	//	request->->pinCertificate(sha1Fingerprint, eSFT_CertSha1)
+	fingerprints.setSha1_P(sha1Fingerprint, sizeof(sha1Fingerprint));
+
+	// request->pinCertificate(fingerprints);
```
slaff: Maybe I am missing something but where are the fingerprints attached to the request?

Sming/SmingCore/Network/TcpClient.h
```diff
@@ -149,20 +152,18 @@ class TcpClient : public TcpConnection
...
 
 	/**
 	 * @brief	Requires(pins) the remote SSL certificate to match certain fingerprints
-	 *
 	 * @note	The data inside the fingerprints parameter is passed by reference
-	 *
 	 * @param	fingerprints - passes the certificate fingerprints by reference.
 	 *
-	 * @retval	bool  true of success, false or failure
+	 * @retval bool  true of success, false or failure
```
slaff: Can you change this also to 'bool true on success, false on failure' ?

Sming/SmingCore/Network/Ssl/SslValidator.h
slaff: Fingerprints are just one small part of the things that you can check in a certificate and decide if it is valid or not. The callback is there to provide more flexibility

Sming/SmingCore/Network/Ssl/SslFingerprints.h
```diff
@@ -14,15 +14,107 @@ enum SslFingerprintType {
...
+	{
+		return setValue(pkSha256, SHA256_SIZE, cert, length);
+	}
+
+	/** @brief Make copy of SHA1 certificate from data stored in flash
+	 *  @param cert
+	 */
+	bool setSha1_P(const uint8_t* cert, unsigned length)
+	{
+		// Word-aligned and sized buffers don't need special handling
```
slaff: Is the word-aligning planned to be added here? If not then this method is identical to the non _P versions and my question is why is that needed?

Sming/SmingCore/Data/Stream/DataSourceStream.h
```diff
@@ -19,6 +19,7 @@
 enum StreamType {
 	eSST_Invalid,	///< Stream content not valid
 	eSST_Memory,	 ///< Memory data stream
+	eSST_Serial,	 ///< Serial port
```
slaff: Hmm... Do we need those specific types or more general ones would be better?

Sming/SmingCore/Network/Ssl/SslFingerprints.h
mikee47: It's a matter of semantics; it also means we could add an optimisation internally which avoids copying the fingerprint data but instead takes a reference to it.

samples/HttpClient/app/application.cpp
mikee47: In the commented-out line - so behaviour isn't changed. Can un-comment if you prefer, or re-comment `fingerprints.setXXX` lines?

slaff: Uncomment and change the comments text if needed.

Sming/SmingCore/Network/Ssl/SslValidator.h
mikee47: Can you give me some examples? I'm just trying to understand this fully in case I'm over-simplifying.

samples/HttpClient/app/application.cpp
mikee47: OK, I'll also update the fingerprints (they've changed)

Sming/SmingCore/Data/Stream/DataSourceStream.h
mikee47: I do wonder if we need `StreamType` at all? In the framework it's barely used:

`IDataSourceStream::isValid()`

Inherited classes can just override this method directly, rather than returning `eSST_Invalid` from `getStreamType()`

`HttpRequest::getBody()` and `HttpResponse::getBody()`

Both of these fail if the stream type isn't `eSST_Memory`, but I'm not sure that check is even necessary.


So we could fix these points and deprecate the `getStreamType()` method completely. We change `IDataSourceStream::getStreamType()` to return `eSST_Unknown` and can then omit it from `HardwareSerial` completely.

slaff: > So we could fix these points and deprecate the getStreamType() method completely. We change IDataSourceStream::getStreamType() to return eSST_Unknown and can then omit it from HardwareSerial completely.

That sounds good to me. Just make sure that it is in a separate PR.

Sming/SmingCore/Network/Ssl/SslFingerprints.h
slaff: > but I'd also change to a `class` and make the certificate data private, which would break existing user's code...

Let's leave this change for now and schedule it for the next version.

Sming/SmingCore/Network/Ssl/SslValidator.h
slaff: > Can you give me some examples? I'm just trying to understand this fully in case I'm over-simplifying.

Let me see if I can do this. 

A developer may want to have the fingerprint checks and in addition have a validator that checks some properties of the certificate. 
In the callback the developer will get access to the SSL certificate. Such a certificate contains multiple data objects (https://www.cryptologie.net/article/262/what-are-x509-certificates-rfc-asn1-der/) and for example there can be SSL validation check that makes sure that the length of the public key used in the certificate is greater than 1024 in order to reject weak keys.

mikee47: Thanks, that helps.

Sming/SmingCore/Network/Ssl/SslFingerprints.h
```diff
@@ -5,24 +5,138 @@
...
+
+private:
+	/** @brief Internal method to set a fingerprint
+	 *  @param value Reference to fingerprint value in this structure
+	 *  @param length Required length for value
+	 *  @param newValue
+	 *  @param newLength
+	 *  @retval bool true on success, false on invalid length or memory allocation failure
+	 */
+	bool setValue(uint8_t*& value, unsigned length, const uint8_t* newValue, unsigned newLength)
```
mikee47: Note: If user provides fingerprint of wrong length (or memory allocation fails) the validator list could end up being empty in which case it will succeed. Preferred behaviour would be a fail.

Sming/SmingCore/Network/Ssl/SslValidator.h
```diff
@@ -16,20 +16,64 @@
...
+public:
+	~SslValidatorList()
+	{
+		// Make sure memory gets released
+		validate(nullptr);
+	}
+
+	bool add(SslValidatorCallback callback, void* data)
+	{
+		return Vector::add(SslValidator{callback, data});
```
mikee47: Annoyingly, `Vector::add` returns true without actually checking if it succeeded. `WVector.h` needs an overhaul! Do you want it fixed in this PR, or a separate one?

slaff: > `WVector.h` needs an overhaul! Do you want it fixed in this PR, or a separate one?

A seperate one, please.

mikee47:
OK, I've tested this last commit with various dodgy keys and behaves predictably. Users should be able to throw any old crap in there and it'll play nice. I've also taken the liberty of adding `flashmem_address_valid` which may come in useful for other code.

mikee47:
OK, I've tested this last commit with various dodgy keys and behaves predictably. Users should be able to throw any old crap in there and it'll play nice.

PR #1623, opened 2019-02-14
Fix remaining `STRICT` compiler warnings

mikee47:
Core code (excluding Libraries) should build entirely without warnings or errors.

### DIFF
--- a/Sming/SmingCore/DateTime.cpp
+++ b/Sming/SmingCore/DateTime.cpp
@@ -379,7 +379,7 @@ String DateTime::format(String sFormat)
 void DateTime::calcDayOfYear()
 {
 	DayofYear = 0;
-	for(int i = 0; i < Month; ++i) {
+	for(unsigned int i = 0; i < Month; ++i) {
 		switch(i) {
 		case 8:  //Sep
 		case 3:  //Apr
--- a/Sming/SmingCore/DateTime.cpp
+++ b/Sming/SmingCore/DateTime.cpp
@@ -379,7 +379,7 @@ String DateTime::format(String sFormat)
 void DateTime::calcDayOfYear()
 {
 	DayofYear = 0;
-	for(unsigned int i = 0; i < Month; ++i) {
+	for(int i = 0; i < Month; ++i) {
 		switch(i) {
 		case 8:  //Sep
 		case 3:  //Apr
--- a/Sming/SmingCore/Network/TcpServer.cpp
+++ b/Sming/SmingCore/Network/TcpServer.cpp
@@ -129,7 +129,8 @@ err_t TcpServer::onAccept(tcp_pcb* clientTcp, err_t err)
 	connections.add(client);
 	debug_d("Opening connection. Total connections: %d", connections.count());
 
-	onClient((TcpClient*)client);
+	// @todo Dangerous up-cast - needs looking at
+	onClient(reinterpret_cast<TcpClient*>(client));
 
 	return ERR_OK;
 }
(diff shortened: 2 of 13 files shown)


Sming/SmingCore/DateTime.cpp
```diff
@@ -379,7 +379,7 @@ String DateTime::format(String sFormat)
 void DateTime::calcDayOfYear()
 {
 	DayofYear = 0;
-	for(unsigned int i = 0; i < Month; ++i) {
+	for(int i = 0; i < Month; ++i) {
```
slaff: Just for my understanding: Why not use `unsigned` here?

Sming/SmingCore/Network/TcpServer.cpp
```diff
@@ -129,7 +129,8 @@ err_t TcpServer::onAccept(tcp_pcb* clientTcp, err_t err)
 	connections.add(client);
 	debug_d("Opening connection. Total connections: %d", connections.count());
 
-	onClient((TcpClient*)client);
+	// @todo Dangerous up-cast - needs looking at
```
slaff: I guess you will take a look at this before the PR is declared ready for merging?

Sming/SmingCore/DateTime.cpp
mikee47: `Month` is signed (`int8_t`) as are all the member variables and methods. I can re-write the `DateTime` class using unsigned values, which is rather more extensive and would be a separate PR.
This seemed like an easier solution unless there's actually a problem with `DateTime`. What'cha wanna do?

Sming/SmingCore/Network/TcpServer.cpp
mikee47: Here be demons. We have `virtual createClient()` returning a `TcpConnection*` which, in most cases, is actually a `TcpClient*`, except for `FTPServerConnection` which creates an `FTPServerConnection` inherited directly from `TcpConnection`. Logically, I'd expect `createClient` to `TcpClient*'. If that's the case it's a significant change to the framework and FTPServer needs looking at - there's a lot of extra stuff in `TcpClient` vs. `TcpConnection`.

If you have any thoughts on this, most welcome!

This'll need a separate PR, so I'll remove the comment.

Sming/SmingCore/DateTime.cpp
slaff: > This seemed like an easier solution unless there's actually a problem with `DateTime`

Then it is ok.

Sming/SmingCore/Network/TcpServer.cpp
slaff: > This'll need a separate PR, so I'll remove the comment and put up with the 'C-style typecast' warning for now.

Perfect. Thanks!

slaff: Just remove the comment so that I can merge the PR.

Sming/SmingCore/DateTime.cpp
mikee47: > unless there's actually a problem with `DateTime`

Had a read through and found a few things. Will submit separate PR for this.

Sming/SmingCore/Network/TcpServer.cpp
slaff: @mikee47 Is the PR ready for merging?

mikee47: @slaff Yes, all done.

Issue #1626, opened 2019-02-16
git submodules errors during first make after git clone


andr2000:
After I have cloned the repo at
`* develop 3a7fbb2d0b87 Fix `fromHttpDate()`, parameters from signed to unsigned and String optimisation. (#1625)`
I set up the environment and tried to make, but I see there are errors during git submodules fetch
[git_log.txt](https://github.com/SmingHub/Sming/files/2872063/git_log.txt)

mikee47:
You'll need to `make dist-clean` first.
Don't worry about the `Error 128 (ignored)` - that is because their is no patch file for those.

andr2000:
Well, this doesn't really help:
```
$:/tmp/Sming/Sming$ make dist-clean 
make[1]: Entering directory '/tmp/Sming/Sming/third-party/esp-open-lwip'
make[1]: Makefile.open: No such file or directory
make[1]: *** No rule to make target 'Makefile.open'.  Stop.
make[1]: Leaving directory '/tmp/Sming/Sming/third-party/esp-open-lwip'
Makefile:483: recipe for target 'clean' failed
make: [clean] Error 2 (ignored)
make[1]: Entering directory '/tmp/Sming/samples/Accelerometer_MMA7455'
```
and then
```
$:/tmp/Sming/Sming$ make
Fetching Arduino Library Libraries/Adafruit_BME280_Library/ ...
Submodule 'Sming/Libraries/Adafruit_BME280_Library' (https://github.com/adafruit/Adafruit_BME280_Library.git) registered for path 'Libraries/Adafruit_BME280_Library'
Cloning into '/tmp/Sming/Sming/Libraries/Adafruit_BME280_Library'...
Submodule path 'Libraries/Adafruit_BME280_Library': checked out '321186220e1e04080bb70fa45aae6e6a5820899f'
error: unrecognized input
Makefile:533: recipe for target 'Libraries/Adafruit_BME280_Library/library.properties' failed
make: [Libraries/Adafruit_BME280_Library/library.properties] Error 128 (ignored)
```

andr2000:
The reason I am worried about those "ignored" errors is that I build Sming with Yocto [1], [2] and those make the build fail. This doesn't happen with master branch though... and because of rboot fixes I decided to temporarily switch to develop branch.
[1] https://www.yoctoproject.org/
[2] https://github.com/andr2000/meta-prod-smarthome/blob/master/recipes-esp8266/sming/sming.bb

[...]

PR #1627, opened 2019-02-16
Do not try to apply non-existent patches (#1626)

andr2000:
Current build system assumes that all Arduino libraries
have relevant patches and tries to apply those without
checking if patches really exist. Fix this by checking
if patch file exists before applying it.

Signed-off-by: Oleksandr Andrushchenko <andr2000@gmail.com>


mikee47:
Works well building under Windows, thanks.

andr2000:
Great as I am only able to check under Linux

PR #1631, opened 2019-02-19
Ran code through linter, fixed issues arising

mikee47:
Add missing `override`s
Add missing SMING_DEPRECATED
Constructor variables assignments out of code
Some missing member variable intialisation
Remove comments from optional parameters
Replace `boolean` with `bool`
Fix issue with changing default `minHeapSize` in `HttpServer`

### DIFF
--- a/Sming/SmingCore/Delegate.h
+++ b/Sming/SmingCore/Delegate.h
@@ -45,8 +45,9 @@ public:
 	__forceinline void decrease()
 	{
 		references--;
-		if(references == 0)
+		if(references == 0) {
 			delete this;
+		}
 	}
 
 private:
@@ -78,7 +79,7 @@ public:
      *  @param  args The delegate method parameters
      *  @retval ReturnType The return value from the invoked method
      */
-	ReturnType invoke(ParamsList... args)
+	ReturnType invoke(ParamsList... args) override
 	{
 		return (mClass->*mMethod)(args...);
 	}
@@ -105,7 +106,7 @@ public:
      *  @param  args The delegate function parameters
      *  @retval ReturnType The return value from the invoked function
      */
-	ReturnType invoke(ParamsList... args)
+	ReturnType invoke(ParamsList... args) override
 	{
 		return (mMethod)(args...);
 	}
@@ -131,7 +132,6 @@ public:
     */
 	__forceinline Delegate()
 	{
-		impl = nullptr;
 	}
 
 	// Class method
@@ -141,10 +141,9 @@ public:
 	 */
 	template <class ClassType> __forceinline Delegate(MethodDeclaration<ClassType> m, ClassType* c)
 	{
-		if(m != NULL)
+		if(m != nullptr) {
 			impl = new MethodCaller<MethodDeclaration<ClassType>>(c, m);
-		else
-			impl = nullptr;
+		}
 	}
 
 	// Function
@@ -153,16 +152,16 @@ public:
 	 */
 	__forceinline Delegate(FunctionDeclaration m)
 	{
-		if(m != NULL)
+		if(m != nullptr) {
 			impl = new FunctionCaller<FunctionDeclaration, ReturnType, ParamsList...>(m);
-		else
-			impl = nullptr;
+		}
 	}
 
 	__forceinline ~Delegate()
 	{
-		if(impl != nullptr)
+		if(impl != nullptr) {
 			impl->decrease();
+		}
 	}
 
 	/** @brief  Invoke a delegate
@@ -208,8 +207,9 @@ public:
 	Delegate& operator=(Delegate&& that) // move assignment
 	{
 		if(this != &that) {
-			if(impl)
+			if(impl != nullptr) {
 				impl->decrease();
+			}
 
 			impl = that.impl;
 			that.impl = nullptr;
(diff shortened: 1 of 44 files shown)


Sming/SmingCore/Network/HttpServer.h
```diff
@@ -89,11 +97,6 @@ class HttpServer : public TcpServer
 protected:
 	TcpConnection* createClient(tcp_pcb* clientTcp) override;
 
-protected:
-#ifdef ENABLE_SSL
-	int minHeapSize = 16384;
```
slaff: @mikee47 What are the advantages of setting the `minHeapSize` value for every constructor instead of the old way as initial value for the protected variable `minHeapSize` ? What is the issue that this change is addressing?

samples/ScreenTFT_ILI9340-ILI9341/app/BMPDraw.cpp
```diff
@@ -49,8 +49,8 @@ void bmpDraw(Adafruit_ILI9341 tft, String fileName, uint8_t x, uint8_t y)
 	uint32_t rowSize;					// Not always = bmpWidth; may have padding
 	uint8_t sdbuffer[3 * BUFFPIXEL];	// pixel buffer (R+G+B per pixel)
 	uint8_t buffidx = sizeof(sdbuffer); // Current position in sdbuffer
-	boolean goodBmp = false;			// Set to true on valid header parse
-	boolean flip = true;				// BMP is stored bottom-to-top
+	bool goodBmp = false;				// Set to true on valid header parse
```
slaff: The newly added training spaces are probably not needed here.

samples/ScreenTFT_ST7735/app/BMPDraw.cpp
```diff
@@ -49,8 +49,8 @@ void bmpDraw(Adafruit_ST7735 tft, String fileName, uint8_t x, uint8_t y)
 	uint32_t rowSize;					// Not always = bmpWidth; may have padding
 	uint8_t sdbuffer[3 * BUFFPIXEL];	// pixel buffer (R+G+B per pixel)
 	uint8_t buffidx = sizeof(sdbuffer); // Current position in sdbuffer
-	boolean goodBmp = false;			// Set to true on valid header parse
-	boolean flip = true;				// BMP is stored bottom-to-top
+	bool goodBmp = false;				// Set to true on valid header parse
```
slaff: The newly added training spaces are probably not needed here.

Sming/SmingCore/Delegate.h
```diff
@@ -229,11 +229,13 @@ template <class ReturnType, class... ParamsList> class Delegate<ReturnType(Param
 	void copy(const Delegate& other)
 	{
 		if(impl != other.impl) {
-			if(impl)
+			if(impl != nullptr) {
```
slaff: Any possible side effects that you can think of from the changes in Delegate.h ?

Sming/SmingCore/Network/HttpServer.h
mikee47: This declaration creates a new `int minHeapSize` which hides the one in `TcpServer`. A simpler, probably better, way would be to modify the statement in `TcpServer.h` instead of putting it here. That would change behaviour, but would it be an issue?

Sming/SmingCore/Delegate.h
mikee47: No side effects. Because `impl` has a static initialiser we can be certain that all the constructors get a default `nullptr` value for this, so setting it to `nullptr` again isn't necessary.

Sming/SmingCore/Network/HttpServer.h
slaff: Ok, I see that this change https://github.com/SmingHub/Sming/commit/fa422059039a901f9cb23786e60b2cffbfac0fb9 changed `int minHeapSize` to `size_t minHeapSize` in TcpServer thus making the protected int minHeapSize in HttpServer to be a completely new one.

> A simpler, probably better, way would be to modify the statement in `TcpServer.h` instead of putting it here.

Can you elaborate a bit?

slaff:
@mikee47 Can you take a look at the issues reported from Codacy?

image

Sming/SmingCore/Network/HttpServer.h
mikee47: In `TcpServer.h`:

```
protected:
#ifdef ENABLE_SSL
	size_t minHeapSize = 16384;
#else
	size_t minHeapSize = 3000;
#endif
```

samples/ScreenTFT_ILI9340-ILI9341/app/BMPDraw.cpp
mikee47: Checked, there  aren't any - it's tabs added by clang-format

slaff: Ok, then it is fine.

Sming/SmingCore/Network/HttpServer.h
slaff:
```
protected:
#ifdef ENABLE_SSL
	size_t minHeapSize = 16384;
#else
	size_t minHeapSize = 3000;
#endif
```

Ok, understood. Let's have that change then.

Issue #1637, opened 2019-02-20
Web server RE initialization


icodk:
Is (how) it possible to restructure the paths a web server is responding to later in the life time of the application ?
in old version of sming (about 2.0) I just added a path.clear() function that cleared the path map just before I populated it with a subset of the original pages (saving space and access to some functionality not needed in the long run). 
I could do it again of course but it is not version portable. Could be nice to have it build in.
The simplest way would probably be to expose the resource tree as public.

mikee47:
> The simplest way would probably be to expose the resource tree as public.

That's a good idea. At present, it's unsafe because `ResourceTree` inherits from `HashMap` which doesn't manage objects very well - it has the potential to leak like a sieve and objects have to be manually destroyed. I'm currently working up a PR to rectify this situation, which I think will make your suggestion practical and simplify things considerably.

icodk:
Looking forward...
Thanks

mikee47:
@icodk You can now use `HttpServer::paths.clear()`

slaff:
> Is (how) it possible to restructure the paths a web server is responding to later in the life time of the application ?

@icodk That should be possible with the latest version of the code from the `develop` branch. Give it a try and write back if you need some help.

icodk:
@mikee47 & @slaff 
That's great Thanks

PR #1638, opened 2019-02-21
Introduce ObjectMap

mikee47:
Add `ObjectMap` class
Move `ResourceTree` class into separate file, rename as `HttpResourceTree`
Rename `HttpServer::resourceTree` as `paths`, make public
Deprecate `HttpServer::addPath` and `HttpServer::setDefault` methods

Addresses #1637

### DIFF
--- a/Sming/SmingCore/Data/ObjectMap.h
+++ b/Sming/SmingCore/Data/ObjectMap.h
@@ -103,6 +103,16 @@ public:
 			return map.remove(key);
 		}
 
+		/**
+		 * @brief Get the value for a given key and remove it from the map, without destroying it
+		 * @retval V*
+		 * @note The returned object must be freed by the caller when no longer required
+		 */
+		V* extract()
+		{
+			return map.extract(key);
+		}
+
 	private:
 		ObjectMap<K, V>& map;
 		K key;
@@ -267,6 +277,24 @@ public:
 			removeAt(index);
 			return true;
 		}
+	}
+
+	/**
+	 * @brief Get the value for a given key and remove it from the map, without destroying it
+	 * @param key
+	 * @retval V*
+	 * @note The returned object must be freed by the caller when no longer required
+	 */
+	V* extract(const K& key)
+	{
+		V* value = nullptr;
+		int i = indexOf(key);
+		if(i >= 0) {
+			value = entries[i].value;
+			entries[i].value = nullptr;
+			entries.remove(i);
+		}
+		return value;
 	}
 
 	/**
--- a/Sming/SmingCore/Data/ObjectMap.h
+++ b/Sming/SmingCore/Data/ObjectMap.h
@@ -1,0 +1,249 @@
+/****
+ * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
+ * Created 2015 by Skurydin Alexey
+ * http://github.com/SmingHub/Sming
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ * ObjectMap.h
+ *
+ * @author: 31 Jul 2018 - Mikee47 <mike@sillyhouse.net>
+ *
+ * Implementation of a HashMap for owned objects, i.e. anything created with new().
+ * Once added to the map the object is destroyed when no longer required.
+ *
+ */
+
+#ifndef _SMING_CORE_DATA_OBJECT_MAP_H_
+#define _SMING_CORE_DATA_OBJECT_MAP_H_
+
+#include "WVector.h"
+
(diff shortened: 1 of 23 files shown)


Sming/SmingCore/Network/HttpClient.cpp
```diff
@@ -13,25 +13,21 @@
...
 
-	int i = httpConnectionPool.indexOf(cacheKey);
-	if(i >= 0) {
+	if(connection != nullptr) {
 		// Check existing connection
-		connection = httpConnectionPool.valueAt(i);
 		if(connection->getConnectionState() > eTCS_Connecting && !connection->isActive()) {
 			debug_d("Removing stale connection: State: %d, Active: %d", connection->getConnectionState(),
 					connection->isActive());
 			delete connection;
```
mikee47: I wonder, is it necessary to delete the connection? Could we just reset it, perhaps?

Sming/SmingCore/Network/HttpServer.h
```diff
@@ -66,32 +59,51 @@ class HttpServer : public TcpServer
...
+	 * @param callback -The callback that will handle this path
 	 */
 	void addPath(String path, const HttpPathDelegate& callback);
+
 	void addPath(const String& path, const HttpResourceDelegate& onRequestComplete);
-	void addPath(const String& path, HttpResource* resource);
 
-	void setDefaultHandler(const HttpPathDelegate& callback);
-	void setDefaultResource(HttpResource* resource);
+	void addPath(const String& path, HttpResource* resource)
```
mikee47: Maybe deprecate some of these, use resourceTree directly?

slaff: That is a good idea. Deprecate it now and the removal will be after two or three major releases. The Readme.md has to be updated too.

Sming/SmingCore/Data/ObjectMap.h
```diff
@@ -0,0 +1,249 @@
...
+	 *			ObjectMap<String, MyType> map;
+	 *			MyType* object1 = new MyType();
+	 *			MyType* object2 = new MyType();
+	 *			map.set("key1", object1);
+	 *			map.set("key1", object2); // object1 is freed automatically
+	 *			// As soon as `map` goes out of scope, object2 is released
+	 *		}
+	 *
+	 */
+	void set(const K& key, V* value)
```
slaff: That `set` will cause confusion. We have two ways to add things to the map that behave diffently:

```
ObjectMap<String, MyType> map;
map["key1"] = object1;
// and 
map.set("key1, object1);
```


The best would have been to make `map["key1"] = object1;` behave as `set`.  But probably it would be better to remove completely `V& operator[](const K& key)` since I cannot think of a proper way to implement the above.

mikee47: > The best would have been to make `map["key1"] = object1;` behave as `set`

Done :-)

Sming/SmingCore/Network/HttpServer.h
mikee47: I've renamed `resourceTree` to `paths` as it seems to make more sense to me when in use. I left the internal references (in `HttpServerConnection`) alone.

Sming/SmingCore/Network/Http/HttpServerConnection.cpp
```diff
@@ -124,11 +123,12 @@ int HttpServerConnection::onHeadersComplete(const HttpHeaders& headers)
...
 		types.add(majorType);
-		types.add("*");
+		types.add(String('*'));
 
 		for(unsigned i = 0; i < types.count(); i++) {
-			if(bodyParsers->contains(types.at(i))) {
-				bodyParser = (*bodyParsers)[types.at(i)];
+			const String& type = types[i];
+			if(bodyParsers->contains(type)) {
+				bodyParser = (*bodyParsers)[type];
```
mikee47: bodyParsers->get(type) would be nicer

Sming/SmingCore/Data/ObjectMap.h
```diff
@@ -0,0 +1,271 @@
...
+	 */
+	const V* operator[](const K& key) const
+	{
+		return find(key);
+	}
+
+	/** @brief Access map entry by reference
+	 *  @param key
+	 *  @retval Value Guarded access to mapped value corresponding to given key
+	 *  @note If the given key does not exist in the map then it will be created and a null value entry returned.
```
mikee47: To avoid this (usually) un-desirable side-effect we could take advantage of our `Value` functor and use it to create an entry in the map only when a value is actually assigned.

PR #1639, opened 2019-02-23
Fix/HttpConnection memory leak

mikee47:
`waitingQueue` is never destroyed.

* Change `waitingQueue` from un-owned `RequestQueue*` to `RequestQueue`
* Free request if queue is full in `send(HttpRequest*)`


mikee47:
@slaff It seems apparent now that we don't need both HttpConnection and HttpClientConnection, so I'll merge them. But which name? HttpClientConnection is more logical, can retain HttpConnection as an alias and deprecate it.

slaff:
> Update: I've worked through this but should wait until `ObjectMap` is merged then I'll rebase with additional changes.

The ObjectMap is merged. You can rebase now.

slaff:
@mikee47 Is this PR tested and ready to roll ?

mikee47:
> Is this PR tested and ready to roll ?
@slaff I've tested on `HttpServer_Websockets` sample and is OK. I'm re-testing on `HttpClient` sample - there's something odd happening so want to get to the bottom of that first.

The refactoring means we should be able to use `HttpConnection` (almost) interchangeably for either client or server - there are also potentially further simplifications that could be made to delegates as these would mostly be identical for both client and server.

If you've any misgivings I'm happy to roll stuff back, take another look, etc.

mikee47:
> I'm re-testing on `HttpClient` sample - there's something odd happening

Found it, fix in PR #1641

slaff:
> Found it, fix in PR #1641 - I'll rebase this PR again after merging

Ready to rebase.

> The refactoring means we should be able to use `HttpConnection` (almost) interchangeably for either client or server - there are also potentially further simplifications that could be made to delegates as these would mostly be identical for both client and server.

I will be more than happy to simplify things and decrease the code that needs to be maintained.

PR #1647, opened 2019-03-13
Add `WifiSniffer` class and sample application

### LINKED ISSUE #1643: promiscuous mode in Sming
Promiscuous mode works fine in Arduino/esp8266 with both  sdk 1.5.3 [https://www.esp8266.com/viewtopic.php?f=29&t=12127 ](https://www.esp8266.com/viewtopic.php?f=29&t=12127)  and 3.0.0 but I could not make it work with Sming (tried only with 1.5.3). Everything compiles and uploads but the call back function that should be called upon reception of wifi activity is never called. Could it be an issue with  C/C++ calling calling convention ?

mikee47:
Uses promiscuous mode to provide generic mechanism for Wifi sniffing.

Closes #1643.

### DIFF
--- a/Sming/SmingCore/Platform/WifiSniffer.cpp
+++ b/Sming/SmingCore/Platform/WifiSniffer.cpp
@@ -1,0 +1,174 @@
+/****
+ * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
+ * Created 2015 by Skurydin Alexey
+ * http://github.com/anakod/Sming
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ * WifiSniffer.cpp
+ *
+ * Original code by Ray Burnette http://www.hackster.io/rayburne/projects
+ *
+ * Adapted for use with Sming March 2019 mikee47 <mike@sillyhouse.net>
+ *
+ ****/
+
+#include "WifiSniffer.h"
+#include "esp_wifi_sniffer.h"
+
+NewBeaconCallback WifiSniffer::newBeaconCallback;
+NewClientCallback WifiSniffer::newClientCallback;
+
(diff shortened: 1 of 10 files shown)


Sming/SmingCore/Platform/WifiSniffer.cpp
```diff
@@ -0,0 +1,174 @@
...
+	} else {
+		bi.err = -3;
+	}
+
+	bi.capa[0] = frame[34];
+	bi.capa[1] = frame[35];
+	memcpy(bi.bssid, frame + 10, ETH_MAC_LEN);
+}
+
+void WifiSniffer::promisc_cb(uint8_t* buf, uint16_t len)
```
slaff: Please, rename `WifiSniffer::promisc_cb` to `WifiSniffer::parseData`.

Sming/SmingCore/Platform/WifiSniffer.h
```diff
@@ -0,0 +1,112 @@
...
+	}
+};
+
+typedef std::function<void(const BeaconInfo& beacon)> NewBeaconCallback;
+typedef std::function<void(const ClientInfo& client)> NewClientCallback;
+
+class WifiSniffer
+{
+public:
+	/** @brief Initialise the sniffer
```
slaff: Please, add a note here that the `begin` method cannot be called directly in the `init()` and should be called when the system is ready. Example:
```C++
void ready()
{
	sniffer.begin(onNewBeacon, onNewClient);
}

 void init()
{
// ... some code
 	System.onReady(ready);
}
```

slaff:
@mikee47 When you have time can you take a look at the requested changes? They are not much and IMHO quite easy to implement.

PR #1653, opened 2019-03-26
Fixed the SPIFFS size calculation.

slaff:
Closes #1313.

### DIFF
--- a/Sming/Services/SpifFS/spiffs_sming.c
+++ b/Sming/Services/SpifFS/spiffs_sming.c
@@ -56,7 +56,13 @@ bool spiffs_format_internal(spiffs_config *cfg)
 {
   if (cfg->phys_addr == 0)
   {
-	SYSTEM_ERROR("Can't format file system, wrong address");
+	SYSTEM_ERROR("Can't format file system, wrong address given.");
+	return false;
+  }
+
+  if (cfg->phys_size == 0)
+  {
+	SYSTEM_ERROR("Can't format file system, wrong size given.");
 	return false;
   }
 
@@ -67,14 +73,14 @@ bool spiffs_format_internal(spiffs_config *cfg)
   sect_last = flashmem_get_sector_of_address(sect_last);
   debugf("sect_first: %x, sect_last: %x\n", sect_first, sect_last);
   ETS_INTR_LOCK();
-  int total = sect_last - sect_first;
+  int total = sect_last - sect_first + 1;
   int cur = 0;
   int last = -1;
   while( sect_first <= sect_last )
   {
 	if(flashmem_erase_sector( sect_first++ ))
 	{
-		int percent = cur++ * 100 / total;
+		int percent = ++cur * 100 / total;
 		if (percent > last)
 			debugf("%d%%", percent);
 		last = percent;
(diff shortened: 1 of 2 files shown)


slaff:
@jonnykl Please, check this PR and tell me if it is working for you.

Sming/Services/SpifFS/spiffs_sming.c
```diff
@@ -67,14 +73,14 @@ bool spiffs_format_internal(spiffs_config *cfg)
   sect_last = flashmem_get_sector_of_address(sect_last);
   debugf("sect_first: %x, sect_last: %x\n", sect_first, sect_last);
   ETS_INTR_LOCK();
-  int total = sect_last - sect_first;
+  int total = sect_last - sect_first + 1;
```
jonnykl: This is not correct (see [PR 1313](https://github.com/SmingHub/Sming/pull/1313#issuecomment-476843308))

jonnykl:
Except the change I commented a few minutes ago everything is correct but does not solve the problem. You prevent the developer from using illegal values which is good but anyways the function `spiffs_format_internal()` erases the sectors for SpifFS and one sector after the end of SpifFS. E.g. if `SPIFF_SIZE` is set to the maximum allowed value, the sector where `esp_init_data_default.bin` is flashed will be erased.

Sming/Services/SpifFS/spiffs_sming.c
jonnykl: As you can see ([here](https://github.com/SmingHub/Sming/pull/1313/commits/43a9a4f719fefba4921b573b7c69633243605b4d)) I also did the same change but in conjunction with the change in line 67 (which is the fix of the bug).

jonnykl:
The fix works now if `cfg->phys_addr` and `cfg->phys_size` is a multiple of the physical block size.

Pro (for this fix):
- almost all of the space requested is used ("almost" because SpifFS is organized in logical blocks, e.g. if logical block size would be 5 and the request space is 13, then only 2 logical blocks / 10 bytes would be used)

Con:
- data which does not belong do SpifFS may be deleted

This fix is OK, but this behavior should be documented (bytes from `cfg->phys_addr + cfg->phys_size` (inclusive) to the end of the physical block will be erased when `spiffs_format_internal()` is called).

Pro (my fix):
- no other data which does not belong to SpifFS is deleted

Con:
- less space than requested is used (if requested space is not of multiple of physical block size)

I think both versions of the fix are OK. They only differ if the developer specifies a offset/size that the FS does not fit exactly in physical blocks.

Sming/appspecific/rboot/overrides.c
```diff
@@ -18,15 +18,15 @@ spiffs_config spiffs_get_storage_config()
 #error "Define either RBOOT_SPIFFS_0 or RBOOT_SPIFFS_1"
 #endif
 
-  maxAllowedEndAddress = INTERNAL_FLASH_SIZE - 1;
-  requestedEndAddress =  cfg.phys_addr + SPIFF_SIZE - 1;
-  if(requestedEndAddress > maxAllowedEndAddress) {
+  cfg.phys_addr &= 0xFFFFF000; // get the start address of the sector
```
jonnykl: I think the following code would be better:
```
cfg.phys_addr -= cfg.phys_addr % INTERNAL_FLASH_SECTOR_SIZE;
```

PR #1655, opened 2019-03-28
GDB and exception handling improvements

mikee47:
Makefile changes:
* Remove esp-gdbstub as third-party module
* Remove all GDB stuff from `Makefile` - gdbstub is compiled into application, not framework
* gdbstub can be used with release or debug builds.
* Add support in all makefiles for assembly code (both .s and .S files)
* Remove all COM port settings from `Makefile` - application code does all this now
* Bring `Makefile-project.mk`inline with `Makefile-rboot.mk` (compiler warnings, etc.)
* Add new 'gdb' phony target to make it easier to start debugger with correct port, baud rate and other settings

HardwareSerial
* Fix so that end() and systemDebugOutput() don't screw up initialisation of other serial port(s)
* Put callback handling code into static methods - saves IRAM and simplifies code
* Revise interrupt callback to prevent task queue flooding, e.g. on receive overflow
* Add `getStatus()` method to report serial errors

HardwareTimer
* Make HardwareTimer mode selectable (maskable or NMI)

user_main.cpp
* Don't change serial port baud rates - do that in gdb_init() instead which is compiled into application

system/crash_handler.c
* Remove stack dumping code and all serial port access - moved into gdb_hooks.cpp
* Include text for exception code in report

system/SerialBuffer.h
* Move non-trivial code into .cpp module
* Add `getReadData()` and `skipRead()` methods for efficient block reading

esp_systemapi.h
* Remove uart prototypes - driver must be used instead
* Ensure wdt functions are available

uart.h, uart.cpp
* Add UART2 and provide generic notification callback handling
* Add UART_OPT_CALLBACK_RAW option
* Save some IRAM by moving uart_get_uart, uart_set_callback, uart_peek_last_char, uart_rx_available and moving uart_detach into flash
* Use inline functions in preference to macros
* Default debug port should be undefined, not UART0
* Use bit manipulation macros to simplify/clarify code
* Disable interrupts in `uart_tx_free()` and `uart_rx_available()` to ensure result is accurate
* Fix incorrect RX FIFO size reading, causes problems in FIFO overflow situation
* Add `uart_get_status()` to support reporting of errors and Rx BREAK condition
* Add mode parameter to uart_flush() so tx/rx can be flushed independently, e.g. if a receive overrun occurs.
* Add `uart_set_break()` function

gdbstub, exception handling and stack dumping

* appspecific/gdb/ code is always compiled into application, contains system exception handler and stack dumper
* gdb/ code is only compiled when ENABLE_GDB=1
* gdbstub revised for readability and checking
* uart handling code in separate module (gdbuart.cpp)
* Add build option to support both patched and unpatched GDB versions
* Extend GDBSTUB_CTRLC_BREAK to prevent accidental breaking
* Add gdb_enable() function
* Include options for stack dumping, with defaults to provide consistent behaviour (i.e. OFF in release builds, ON in debug builds)
* Add debugging code to stub which can be optionally enabled to see realtime behaviour
* Implement register read/write (p & P) and binary write (X) commands
* Provide API for GDB system functions

LiveDebug sample

* Revise to highlight possible debugging issues with hardware timer, and illustrate other timer types which are 'debug safe'
* Add serial read callback to demonstrate UART2 function
* Implement simple interactive debug console using GDB syscall support
* Add error checking/reporting to serial receive callback

Basic_Serial sample
* Add error checking/reporting to serial receive callback


slaff:
@riban-bw Mike is updating and improving the GDB debugging code. Can you test this PR and give us your feedback?

Sming/appspecific/gdb/gdbstub-entry.S
```diff
@@ -0,0 +1,314 @@
...
+	wsr		a0, LITBASE
+	l32i	a0, a2, GDBSR_sar
+	wsr		a0, SAR
+	l32i	a15, a2, GDBSR_A(15)
+	l32i	a14, a2, GDBSR_A(14)
+	l32i	a13, a2, GDBSR_A(13)
+	l32i	a12, a2, GDBSR_A(12)
+	l32i	a11, a2, GDBSR_A(11)
+	l32i	a10, a2, GDBSR_A(10)
+	l32i	a9, a2, GDBSR_A(8)
```
mikee47: Should be GDBSR_A(9)

Sming/gdb/gdbstub.cpp
```diff
@@ -0,0 +1,898 @@
...
+		}
+		if(c == '$') {
+			// Wut, restart packet?
+			debug_e("Unexpected '$' received");
+			checksum = 0;
+			cmdLen = 0;
+			continue;
+		}
+		if(c == '}') { // escape the next char
+			c = gdbReceiveChar() ^ 0x20;
```
mikee47: Wrong: Checksum must be updated

slaff: > I'll add in next commit.

Thanks a lot Mike!

riban-bw:
I will try to review this week (whilst I am on a business trip to Inverness, Scotland - nice!). I have it running but need to remind myself how to use gdb and look at how Mike has implemented the sample. Hope to respond this week.

mikee47:
The last commits include additions to the LiveDebug sample so it interacts with GDB during a debug session. If you like the way it's going, I could take a look at revising the serial receive handler to process input in a similar way, so the program works the same whether using a regular terminal or GDB.

slaff:
> If you like the way it's going, I could take a look at revising the serial receive handler to process input in a similar way, so the program works the **same whether using a regular terminal or GDB.**

Yep, that will be best.

slaff:
I am trying to achieve something very simple but it still does not work for me. I am setting a software breakpoint for the `init` function

```
(gdb) br init
Breakpoint 1 at 0x40105da4: file app/application.cpp, line 399.
```

Then I am trying to run the code until the breakpoint
```
(gdb) c
Continuing.
```
But that ends up shortly after that with segfault

```
Program received signal SIGSEGV, Segmentation fault.
0x4000dfa9 in ?? ()
```

image

Any idea why is that happening?

mikee47:
> But that ends up shortly after that with segfault
@slaff I get the same! It's because of an optimisation to the `writeMemoryBlock()` function I made, which uses `memcpy` for RAM addresses. Clearly that doesn't work for IRAM; If you take that bit out it works again. I'll look at it in more detail later on, but like just use memcpy for DRAM and slow writes for everything else.

Sming/gdb/gdbstub.cpp
```diff
@@ -0,0 +1,941 @@
...
+static ERRNO_T ATTR_GDBEXTERNFN writeMemoryBlock(uint32_t addr, const void* data, unsigned length)
+{
+	// write memory from gdb
+	if(!isValidWriteAddr(addr, length)) {
+		return EFAULT;
+	}
+
+	// If writing to RAM, use memcpy for efficiency
+	if(isRamAddr(addr)) {
+		memcpy(reinterpret_cast<void*>(addr), data, length);
```
mikee47: This fails for IRAM addresses, which must be word aligned.

slaff:
@mikee47 I like the improvements and the debugging seems to be quite stable. 
What I am missing is:
-  Information about the additional system calls. Why are they? How can they be used? What can they  help you with during debugging?
- Very simple code example where the user enters keystrokes and those keystrokes are outputted. To demonstrate the fact that also interactive code can be debugged and the results can be shown in the output.

mikee47:
> @mikee47 I like the improvements and the debugging seems to be quite stable.

@slaff Good to hear!

> What I am missing is:
> 
>     * Information about the additional system calls. Why are they? How can they be used? What can they  help you with during debugging?

Excellent questions! I'm inclined to respond by saying "read the GDB manual" as it is pretty comprehensive. However, it's a lot to go through and contains a lot of stuff which isn't relevant to us. What is needed is a manual specifically focused on remote serial debugging. It sounds like the sort of thing which should already exist - if not, perhaps add a Wiki page?

For now, we definitely need some documentation in `gdb_syscall.h`. Probably also some more comments in LiveDebug sample, which I'll try to simplify perhaps by putting all the syscall stuff in a separate module.

>     * Very simple code example where the user enters keystrokes and those keystrokes are outputted. To demonstrate the fact that also interactive code can be debugged and the results can be shown in the output.

Do you mean like the `Basic_Serial` sample?

slaff:
> Do you mean like the Basic_Serial sample?

Simpler than that :) A simple echoer that reads the incoming data from the input and outputs it back to the output in reversed order. Example: "echo -> ohce".
The function should output a prompt to hint the users to type:
```
(type something) >  
```

And outputs the output on a new line with reversed characters.

 Plus the function that handles the keystrokes should have the GDB_IRAM_ATTR attribute so that users can easily put a software breakpoint in there and see how it works.

> It sounds like the sort of thing which should already exist - if not, perhaps add a Wiki page?

Wiki page would be great.

riban-bw:
I concur. We want a sample app that provides some minimal normal functionality and demonstrates gdb in action and we need some simple docs / instructions on how to get started with Sming + gdb. I have yet to get this PR to do anything useful. Ctrl + c breaks out of code but I haven't actually seen the code or the debugger do anything. Let's make it easy to see gdb in action 😀.

slaff:
> Let's make it easy to see gdb in action 

I have this in my TODO list. Once the PR is merged I will take care to write a Step-By-Step HowTo and probably make a video to so that everyone who is willing to read or watch should be able to understand how to use debugger in Sming :)

riban-bw:
For me to do any useful testing I need some idea how the sample is supposed to behave. If someone tells me that within the next few hours then I can test this eBay. After that I am unlikely to have much free time for a while. Cheers

mikee47:
@riban-bw The LiveDebug sample is just that - a sample - so the testing should really be whether the GDB integration itself is sufficient. To that end it should be straightforward to compile any application (make ENABLE_GDB=1 flash) and to then be greeted with the GDB prompt.
I'll add some documentation for the syscall stuff but that is very much an 'add on' for GDB which suggest new ways that GDB can be helpful in developing applications. Applications do not need to support it at all.

riban-bw:
Thanks @mikee47 that does help but what helped the most was to realise that my test device needed specific parameters to work, e.g. SPI_SIZE=4M. I now have this working and will play a bit. I have seen some issues but need to familiarise more to understand expected behaviour better so that I can identify and report such issues.

Some fixes for the code:
Line 320: Rresume... typo
Add "time" option to the help output.

slaff:
> Very simple code example where the user enters keystrokes and those keystrokes are outputted. To demonstrate the fact that also interactive code can be debugged and the results can be shown in the output.

@mikee47 I can take care of that. The documentation can also come later. You have already done an amazing job and brought the ESP-GDBStub to a new level :).
That said is there anything more as code changes that need to be added and tested as part of this PR before I merge it?

mikee47:
@slaff Thanks for your support, much appreciated. There are some things which need addressing though. Your thoughts would be welcome.

# Outstanding issues

* Reported by @riban-bw
Line 320: Rresume... typo
Add "time" option to the help output.
Add to gdbcmds: "echo Type 'continue' to run application" to give user clue how to start.

* LiveDebug sample

Add the 'hang' command to force an infinite loop.
Maybe some other types of exception?

* gdb_syscall documentation

* Add Data/Streams/GdbHostFileStream class

To wrap gdb_syscall_* file I/O and make them easier to use. A typical use for this is to allow applications to create debugging log files on the host.

* General code structure and file naming

Please check you are happy with the layout, and naming of code, etc:

```
# Main header added to SmingCore.h so functions available to applications
System/
	gdb_hooks.h

# 'Always build' code, includes stack dumping and exception handlers
appspecific/gdb/
	gdb_hooks.cpp		 
	gdbstub-entry.S

# Code modules only built with ENABLE_GDB=1, plus header files required by gdb_hooks.cpp
gdb/
```


* System restart hooking

I added a 'hang' command to LiveDebug to check what happens with WDT. We get a message displayed but GDB does not get invoked. I attempted to add a 'break' at that point but it doesn't work - requires further investigation as this is clearly a necessary function otherwise we can't effectively debug those 'mysterious restarts'.

* Host I/O

Ref. gdbstub.cpp, line 652, I started looking at the 'Host I/O' capabilities of GDB. These clearly separate into two parts: File I/O and Flash writing.

The flash commands are there to support the 'load' command, intended to program images remotely. If you think that would be useful to have, let me know and I'll take a look, otherwise I propose just removing the code from the stub.

I've got a working implementation of the File I/O part, so you can use the 'remote get', 'remote put' and 'remote delete' commands from GDB. This calls the appropriate file* functions, though of course interrupts are disabled whilst this happens. One issue here is that if breaking into the application using Ctrl+C, we cannot be sure it's in a safe place to do this; what if we break into the middle of another fileWrite, for example?

* Ctrl+C handling

This breaks into the debugger using directly inside the UART interrupt routine. In most situations, it would be better to have the break occur inside regular code by queuing a task callback wrapping the gdbstub_do_break() call.

There are some side-effects though, and it does complicate operation slightly. For example, if the application code is 'stuck' somewhere then we'd need to press Ctrl+C again to 'force' a break.

* Initial baud rate setting

I removed this from framework, which means initial debug output (including from rBoot) appears at default 74880 baud.
Instead, we need a call to a hook function (defined in gdb_hooks.cpp) inserted at the earliest possible opportunity in the boot process to set the baudrate.
The question here is: where is the best place to put this hook? It could go back into `user_init()` (and perhaps also `user_rf_pre_init()`) but that is too late for rBoot.
Maybe the call could go into rBoot itself?

riban-bw:
I think you have done a brilliant job of implementing a really useful implementation. I suggest resolving the most urgent / highest impact issues and merging then creating issues and PR for outstanding issues. It would be good to get this in the next stable release.

slaff:
> Stack dumping and serial ports
The previous code dumps directly to both serial ports without any option  ... Instead it uses the m_print* functions which can be redirected anywhere. **Is this change OK**?

Yes, absolutely.

> Availability of GDB for Windows 
Needs to be integrated into Choco and probably also manual build process.

That should be tackled in a separate PR in the choco-build scripts. Just give me a link from where I can get the xtensa debugger for Windows and I will try to do the rest.

> Initial baud rate setting
I removed this from framework, which means initial debug output (including from rBoot) appears at default 74880 baud.
Instead, we need a call to a hook function (defined in gdb_hooks.cpp) inserted at the earliest possible opportunity in the boot process to set the baudrate.
The question here is: where is the best place to put this hook? It could go back into user_init() (and perhaps also user_rf_pre_init()) but that is too late for rBoot.

- `user_rf_pre_init`  is a bad idea because it is not called at all in some SDK versions. 
- `user_init` is ok. Even if we change the bootloader it should still work. 
- rBoot bootloader is also fine. The extra code will increase the code size of the bootloader. As long as that increase is negligible (for embedded bootloader) I am ok with this.

> Ctrl+C handling
This breaks into the debugger using directly inside the UART interrupt routine. In most situations, **it would be better to have the break occur inside regular code** by queuing a task callback wrapping the gdbstub_do_break() call.

I like the idea but **will it be possible to implement this till end of next week**?

>  Host I/O
> If you think that would be useful to have, let me know and I'll take a look, otherwise I propose just removing the code from the stub.

That code will definitely raise questions how to use it.  My suggestion is to leave the code as it is and improve it in the next release. The improvement should be a simple and useful example to the LiveDebug code.

> System restart hooking
I added a 'hang' command to LiveDebug to check what happens with WDT. We get a message displayed but GDB does not get invoked. I attempted to add a 'break' at that point but it doesn't work - requires further investigation as this is clearly a necessary function otherwise we can't effectively debug those 'mysterious restarts'.

Nice to have but can wait for the next release.

> Reported by @riban-bw
Line 320: Rresume... typo
Add "time" option to the help output.
Add to gdbcmds: "echo Type 'continue' to run application" to give user clue how to start.

**That should be fixed for this release.**

> LiveDebug sample
Add the 'hang' command to force an infinite loop.
Maybe some other types of exception?

Should wait for the next release(s). It is better to have a small set of issues that can be debugged with GDB. 

> CR/LF
Stack dumping code needs to use \r\n instead of just \n.

**That should be fixed for this release.**

> gdb_syscall documentation
> Add Data/Streams/GdbHostFileStream class
To wrap gdb_syscall_* file I/O and make them easier to use. A typical use for this is to allow applications to create debugging log files on the host.

That can wait for the next release so that it is also used in the GDB Host I/O example code.

> General code structure and file naming

Looks good for now. That can be modified in future releases if needed.

mikee47:
> 
> 
> > Stack dumping and serial ports
> > The previous code dumps directly to both serial ports without any option  ... Instead it uses the m_print* functions which can be redirected anywhere. **Is this change OK**?
> 
> Yes, absolutely.
> 
> > Availability of GDB for Windows
> > Needs to be integrated into Choco and probably also manual build process.
> 
> That should be tackled in a separate PR in the choco-build scripts. Just give me a link from where I can get the xtensa debugger for Windows and I will try to do the rest.
> 
> > Initial baud rate setting
> > I removed this from framework, which means initial debug output (including from rBoot) appears at default 74880 baud.
> > Instead, we need a call to a hook function (defined in gdb_hooks.cpp) inserted at the earliest possible opportunity in the boot process to set the baudrate.
> > The question here is: where is the best place to put this hook? It could go back into user_init() (and perhaps also user_rf_pre_init()) but that is too late for rBoot.
> 
>     * `user_rf_pre_init`  is a bad idea because it is not called at all in some SDK versions.
> 
>     * `user_init` is ok. Even if we change the bootloader it should still work.
> 
>     * rBoot bootloader is also fine. The extra code will increase the code size of the bootloader. As long as that increase is negligible (for embedded bootloader) I am ok with this.
> 
> 
> > Ctrl+C handling
> > This breaks into the debugger using directly inside the UART interrupt routine. In most situations, **it would be better to have the break occur inside regular code** by queuing a task callback wrapping the gdbstub_do_break() call.
> 
> I like the idea but **will it be possible to implement this till end of next week**?
> 
> > Host I/O
> > If you think that would be useful to have, let me know and I'll take a look, otherwise I propose just removing the code from the stub.
> 
> That code will definitely raise questions how to use it. My suggestion is to leave the code as it is and improve it in the next release. The improvement should be a simple and useful example to the LiveDebug code.
> 
> > System restart hooking
> > I added a 'hang' command to LiveDebug to check what happens with WDT. We get a message displayed but GDB does not get invoked. I attempted to add a 'break' at that point but it doesn't work - requires further investigation as this is clearly a necessary function otherwise we can't effectively debug those 'mysterious restarts'.
> 
> Nice to have but can wait for the next release.
> 
> > Reported by @riban-bw
> > Line 320: Rresume... typo
> > Add "time" option to the help output.
> > Add to gdbcmds: "echo Type 'continue' to run application" to give user clue how to start.
> 
> **That should be fixed for this release.**
> 
> > LiveDebug sample
> > Add the 'hang' command to force an infinite loop.
> > Maybe some other types of exception?
> 
> Should wait for the next release(s). It is better to have a small set of issues that can be debugged with GDB.
> 
> > CR/LF
> > Stack dumping code needs to use \r\n instead of just \n.
> 
> **That should be fixed for this release.**
> 
> > gdb_syscall documentation
> > Add Data/Streams/GdbHostFileStream class
> > To wrap gdb_syscall_* file I/O and make them easier to use. A typical use for this is to allow applications to create debugging log files on the host.
> 
> That can wait for the next release so that it is also used in the GDB Host I/O example code.
> 
> > General code structure and file naming
> 
> Looks good for now. That can be modified in future releases if needed.

mikee47:
@slaff @riban-bw Thank you both for your patience. I've tested the latest changes on Windows/Linux and reasonably happy with its behaviour. I've checked it using the LiveDebug sample in all three modes (GDB, Terminal and non-GDB) and behaves as expected.

For the `Initial baud rate setting` I've made the change at the `user_init()` stage as it's simplest, but early rBoot and ROM messages are still output at 74880 as at present.
For `Ctrl+C handling` and `Host I/O` I've implemented the propsed changes, but if you prefer to revert any of this that's fine.

I do have a further point about the behaviour of `GDBSTUB_BREAK_ON_EXCEPTION`, which may benefit from being redefined as with I did with the `GDBSTUB_CTRLC_BREAK` setting:
0: Disable exception trapping under GDB completely;
1: Enable exception reporting only when debugger is attached;
2: Enable exception reporting at any time (the existing behaviour, consistent with previous gdbstub implementations).
Ditto for `GDBSTUB_BREAK_ON_RESTART`. My reasoning here is that compiling with a setting of '1' will allow the application to run normally provided GDB isn't attached. So, if an exception occurs or the system crashes it will restart automatically. Contrast this with setting '2', where it will wait indefinitely for GDB. Perhaps setting '1' should be the default.

mikee47:
> resolving the most urgent / highest impact issues

I've been thinking about this one, and definitely the most critical part of this PR are that it doesn't break anything - and hopefully improves behaviour - when GDB is _not_ enabled. The main culprit would be the uart driver here, of course, along with `HardwareSerial`, with the addition of error reporting and overflow checking should be more robust whilst keeping 100% backward compatibility. If you've any apps. which stress those in particular worth checking them for sure.

slaff:
@mikee47 Is the PR ready for merging?

mikee47:
@slaff Yes, all done.

slaff:
> @slaff Yes, all done.

Thanks a lot!

Sming/gdb/gdbuart.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+	case UART_NOTIFY_BEFORE_READ:
+		break;
+	}
+}
+#endif
+
+static void IRAM_ATTR doCtrlBreak()
+{
+	if(break_requests != 0) {
```
slaff: One side-effect that I am experiencing in Eclipse, and probably other debuggers where you cannot send twice Ctrl-C, is that it looks like the debugging session has paused in the block below.

mikee47: > One side-effect that I am experiencing in Eclipse, and probably other debuggers where you cannot send twice Ctrl-C, is that it looks like the debugging session has **paused in the block below.**

@slaff Sorry, I don't understand - is this something you'd like me to look at before 3.8.0?

slaff:
>> .. cannot send twice Ctrl-C ..

> Sorry, I don't understand - is this something you'd like me to look at before 3.8.0?

Don't do anything related to this for now. Let me first try to get used to it myself and I will open a new issue if needed.

mikee47:
>Availability of GDB for Windows
>Needs to be integrated into Choco and probably also manual build process.
>>That should be tackled in a separate PR in the choco-build scripts. Just give me a link from where I can get the xtensa debugger for Windows and I will try to do the rest.

@slaff You can find pre-built version of this at [SysProgs](http://gnutoolchains.com/esp8266/). Download and run the executable installer, then copy the `C:\SysGCC\esp8266\opt\xtensa-lx106-elf\bin\xtensa-lx106-elf-gdb.exe` to a suitable location. This is the Espressif 'patched' version.

mikee47:
@slaff All the Arduino toolchains are listed [here](https://github.com/esp8266/Arduino/blob/master/package/package_esp8266com_index.template.json). Their current toolchains use the unpatched version of GDB. Aside from the size of the register sets, I've noticed that they sometimes produce different stack trace results. Not sure why though, and probably never will without details of the Espressif GDB patch, which I believe is closed source.

slaff:
@mikee47 I am still experiencing issues with the debugger. For example a very simple session like the one below ends up in flames

```
1. New LiveDebugg application is flashed on the NodeMCU device.
2. make gdb is executed.
---
For help, type "help".
Type "apropos word" to search for commands related to "word".
0xffffffff in ?? ()
overlapping memory region
overlapping memory region
overlapping memory region
overlapping memory region
overlapping memory region
overlapping memory region
add symbol table from file "out/build/bootrom.elf" at
	.text_addr = 0x40000000

Welcome to SMING!
Type 'c' (continue) to run application

A program is being debugged already.  Kill it? (y or n) y
Remote debugging using /dev/ttyUSB0
gdbstub_init () at /home/slavey/dev/esp8266.dev.box/dev/Sming/Sming//gdb/gdbstub.cpp:908
908			gdbstub_do_break();
(gdb) 
(gdb) c
Continuing.
LiveDebug sample
Explore some capabilities of the GDB debugger.

[OS] mode : sta(18:fe:34:f3:74:e7)
[OS] add if0
252348026 GdbAttach(1)
252391206 open log 0
(Attached) break
255528829 gdb_read_console() returned 6
Calling gdb_do_break()

Program received signal SIGTRAP, Trace/breakpoint trap.
gdb_do_break () at /home/slavey/dev/esp8266.dev.box/dev/Sming/Sming//gdb/gdbstub.cpp:937
937		gdbstub_do_break();
(gdb) br blink
Breakpoint 1 at 0x40105d4c: file app/application.cpp, line 53.
(gdb) c
Continuing.

Breakpoint 1, blink () at app/application.cpp:53
53	{
(gdb) 
Continuing.

Breakpoint 1, blink () at app/application.cpp:53
53	{
(gdb) 
Continuing.
(Attached) 


***** Fatal exception 20 (INSTR_PROHIBITED)
pc=0xffffffff sp=0x3ffffec0 excvaddr=0x00000000
ps=0x00000030 sar=0x0000001e vpri=0x402072cc
r00: 0xffffffff=        -1 r01: 0x3ffffec0=1073741504 r02: 0x00000000=         0 
r03: 0x00000001=         1 r04: 0x00000000=         0 r05: 0x3ffeebf0=1073671152 
r06: 0x40224020=1075986464 r07: 0x3ffe8ac0=1073646272 r08: 0x00000008=         8 
r09: 0x00000000=         0 r10: 0x3ffe8f64=1073647460 r11: 0x3ffe8a9c=1073646236 
r12: 0x3fffc230=1073726000 r13: 0x401062fc=1074815740 r14: 0x3fffc200=1073725952 
r15: 0x00000022=        34 

Program received signal SIGSEGV, Segmentation fault.
0xffffffff in ?? ()
```

PR #1657, opened 2019-04-01
Change fileSetContent() to return a result

kmihaylov:
Instead of void, fileSetContent() now returns a result of the operation (fileOpen or fileWrite) and in case of error, the error code (available in third-party/spiffs/src/spiffs.h) will be returned.

### DIFF
--- a/Sming/SmingCore/FileSystem.cpp
+++ b/Sming/SmingCore/FileSystem.cpp
@@ -115,17 +115,17 @@ void fileClearLastError(file_t fd)
 	SPIFFS_clearerr(&_filesystemStorageHandle);
 }
 
-signed int fileSetContent(const String& fileName, const String& content)
+int fileSetContent(const String& fileName, const String& content)
 {
 	return fileSetContent(fileName, content.c_str());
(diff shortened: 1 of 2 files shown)


Sming/SmingCore/FileSystem.cpp
```diff
@@ -115,16 +115,22 @@ void fileClearLastError(file_t fd)
 	SPIFFS_clearerr(&_filesystemStorageHandle);
 }
 
-void fileSetContent(const String& fileName, const String& content)
+signed int fileSetContent(const String& fileName, const String& content)
```
slaff: @kmihaylov Just use `int` instead of `signed int`. In C/C++ `int` represents signed integers. Apply that also to the other changes coming from you.

slaff:
@kmihaylov Make sure to run `make cs` to auto-fix the coding style once you are done with your changes. Take a look at [Coding-Style-Rules](https://github.com/SmingHub/Sming/wiki/Coding-Style-Rules) for details. We use clang-format version 6.

PR #1659, opened 2019-04-06
makeHexString() incorrectly appends nul character to result

### LINKED ISSUE #1658: Incompatibility between String class and Mosquitto broker
After a freshly built & flashed ESP8266 node failed to connect to a Mosquitto MQTT broker with nothing but a brief `'socket error on client <unknown>'`, I had a look at why this (previously working) code failed to work, with the MQTTClient_Hello sample working.

In the end I traced it down to a null-byte in the client ID that gets sent to the Mosquitto broker. When a null byte is added to the client ID string, Mosquitto's connect routine pushes it through its `mosquitto_validate_utf8()` routine in `utf8_mosq.c`, which returns an error when it finds a null byte, which consequently aborts the connection attempt.

As to why the sample code worked, that is because it provided a const string to the String constructor of the first parameter, which invokes the `String::copy()` routine, which sets the length of the String instance to be the number of characters, minus the null byte that gets added as well.

However, since I used the MAC address of the ESP8266 as the client ID (a practice I have used with Sming since early 2017), it gets a String instance from `StationClass::getMAC()`, which in turn invokes `makeHexString()` in Data/HexString.cpp.

It is this function which sets the String length to:

`if(!result.setLength(charsPerByte * length + 1))`

This means the characters plus the null byte as the length of the String. Consequently, when the MQTT client then copies this client ID from the String instance, it copies it without null byte in the case of the const character string, and with a null byte in the case of the MAC address String.

I hacked around this issue for now by having `makeHexString()` set the String length to just the characters (removing the + 1), but this is obviously far from ideal.

Ideally I guess String would be like std::string, with an internally managed buffer and an API that prevents such direct manipulations. Right now I'm not sure what the right course of action here is, beyond maybe slotting in that quick hack for now.

mikee47:
Fixes #1658

### DIFF
--- a/Sming/SmingCore/Data/HexString.cpp
+++ b/Sming/SmingCore/Data/HexString.cpp
@@ -12,20 +12,28 @@
 
 String makeHexString(const uint8_t* data, unsigned length, char separator)
 {
+	if(data == nullptr || length == 0) {
+		return nullptr;
+	}
+
 	String result;
 
-	unsigned charsPerByte = separator ? 3 : 2;
-	if(!result.setLength(charsPerByte * length + 1))
+	unsigned stringLength = length * 2; // 2 hex chars per byte
+	if(separator != '\0') {
+		stringLength += length - 1;
+	}
+	if(!result.setLength(stringLength)) {
 		return nullptr;
+	}
 
 	char* p = result.begin();
 	for(unsigned i = 0; i < length; ++i) {
-		if(i && separator)
+		if(i != 0 && separator != '\0') {
 			*p++ = separator;
+		}
 		*p++ = hexchar(data[i] >> 4);
 		*p++ = hexchar(data[i] & 0x0F);
 	}
-	*p = '\0';
 
 	return result;
 }


mikee47:
Test code:

```
void testMakeHexString()
{
	uint8 hwaddr[6];
	wifi_get_macaddr(STATION_IF, hwaddr);

	m_printHex("MAC", hwaddr, 6);

	auto test = [](const uint8_t* data, unsigned length, char separator = '\0') {
		String s = makeHexString(data, length, separator);
		m_printHex(nullptr, s.c_str(), s.length(), -1, 24);
	};

	test(nullptr, 6);
	test(hwaddr, 0);
	test(hwaddr, 6);
	test(hwaddr, 6, ':');
	test(hwaddr, 7);
	test(hwaddr, 7, ':');
	test(hwaddr, 1, ':');
	test(hwaddr, 0, ':');
}

```

Output:

```
MAC: dc 4f 22 1c 26 66                                .O".&f
64 63 34 66 32 32 31 63 32 36 36 36                                      dc4f221c2666
64 63 3a 34 66 3a 32 32 3a 31 63 3a 32 36 3a 36 36                       dc:4f:22:1c:26:66
64 63 34 66 32 32 31 63 32 36 36 36 30 30                                dc4f221c266600
64 63 3a 34 66 3a 32 32 3a 31 63 3a 32 36 3a 36 36 3a 30 30              dc:4f:22:1c:26:66:00
64 63                                                                    dc
```

PR #1613, opened 2019-02-10
Hook system debug messages

mikee47:
To help track down "E:M" errors output by system functions, #define UART_DEBUG_SNOOP and rebuild. Any lines starting with "E:M" will then receive a full stack dump.

PR #1662, opened 2019-04-10
Minor gdb fixes

mikee47:
* `HardwareSerial::updateUartCallback()` requires a return value (warning in STRICT mode)
* Bad debug statement in `GdbFileStream::open()` - only causes an issue if building gdbstub in debug mode
* Make sure any buffered debug output appears before exception message in GDB console


mikee47:
@slaff Anything else you want to add to this?

slaff:
> Anything else you want to add to this?

If you have time maybe you can rebase #1613 on the latest code from develop. But it is not urgent and can wait also for the next release.

PR #1665, opened 2019-04-12
Fix 'corrupt stack?' warning in GDB backtrace

mikee47:
One cause is missing symbols, so tell GDB about ESP8266 ROM symbols defined in `bootrom.elf` file.

See https://github.com/jcmvbkbc/esp-elf-rom for how the `bootrom.elf` file was created.


mikee47:
This issue was raised in #1613.

Sming/Makefile-rboot.mk
```diff
@@ -540,7 +542,11 @@ endef
 
 .PHONY: all checkdirs spiff_update spiff_clean clean kill_term terminal gdb
 
-all: $(USER_LIBDIR)/lib$(LIBSMING).a checkdirs $(LIBMAIN_DST) $(RBOOT_BIN) $(RBOOT_ROM_0) $(RBOOT_ROM_1) $(SPIFF_BIN_OUT) $(FW_FILE_1) $(FW_FILE_2) 
+all: $(USER_LIBDIR)/lib$(LIBSMING).a checkdirs $(LIBMAIN_DST) $(RBOOT_BIN) $(RBOOT_ROM_0) $(RBOOT_ROM_1) $(SPIFF_BIN_OUT) $(FW_FILE_1) $(FW_FILE_2) $(BUILD_BASE)/$(BOOTROM_ELF)
+
+# File contains boot rom symbols required by GDB, put it somewhere easy to find
+$(BUILD_BASE)/$(BOOTROM_ELF):
+	cp $(SMING_HOME)/gdb/$(BOOTROM_ELF) $(BUILD_BASE)
```
slaff: I guess `bootrom.elf` should be included in this PR.

mikee47:
A couple of notes if this issue is revisisted in the future.

* This PR always copies the bootrom.elf file regardless of build mode. A better approach would be to add a `GDB_SYMBOLS` target to makefile, defined only when ENABLE_GDB=1.

* I was unable to find a way to directly load the `$(SMING_HOME)/gdb/bootrom.elf` file. Adding the directory to GDB's data or debug directory didn't work. Also, in Windows GDB doesn't recognise drive separators. Copying the file to out/build means we can use a relative reference.

* An alternative to loading symbols separately is to merge the symbols into the `app_0.out` file.

slaff:
Do you want to make the changes this weekend? Is yes - go ahead and give it a try.

mikee47:
Commit in #1671 copies symbol files only when ENABLE_GDB=1.

PR #1666, opened 2019-04-12
Add support in m_printf() for flash strings

mikee47:
* Move `__PRETTY_FUNCTION__` and `__FUNCTION__` definitions into flash.
* Add `isFlashPtr` macro to `FakePgmSpace.h` - this was implemented as an inline function in `SslFingerprints.cpp`
* Revise `m_vsnprintf`handling of `%s` to detect and handle flash strings correctly

The changes also allow `m_printf`, `debug_*`, etc. to use flash strings directly without having to load them into RAM first, however the main point of this PR is to allow the `__PRETTY_FUNCTION__` and `__FUNCTION__` to be used without consuming RAM.

Example:

```
void MyClass::testMethod()
{
  debug_i("Entering %s", __PRETTY_FUNCTION__);
  debug_i("%s\r\n", PSTR("This is a flash string"));
}
```

### DIFF
--- a/Sming/SmingCore/Network/Ssl/SslFingerprints.cpp
+++ b/Sming/SmingCore/Network/Ssl/SslFingerprints.cpp
@@ -12,13 +12,6 @@
 
 #include "SslFingerprints.h"
 #include <user_config.h>
-#include "flashmem.h"
-
-static inline bool isFlashPtr(const uint8_t* ptr)
-{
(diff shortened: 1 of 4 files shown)


slaff:
The windows build fails: https://ci.appveyor.com/project/slaff/sming-sb483/builds/23802455

PR #1663, opened 2019-04-10
Prevent unwanted format of SPIFFS

kmihaylov:
It was found that spiffs_format_internal initiates a format after SPIFFS has written enough data to reach the top memory address (typical wear-leveling procedure).

The proposed code tries to handle in a better way the mounting and initialization of the filesystem.

Discussion #1660

Issue #1660, opened 2019-04-07
SPIFFS crash on api_spiffs_erase (example provided)


kmihaylov:
Hello. Finally I found when my FS is crashing. It it some time after data has been written to it, that's why it was a little bit hard to figure it out when exactly the bug appears.

I found that while there is no need for api_spiffs_erase, the consistency of the FS is OK, the device can be rebooted w/o problems (data loss, broken FS).

On the first appearance of "api_spiffs_erase", however, the FS is immediately broken.

I made a simple program where one can write some data to the FS.
Issue command ```save 10``` where the integer is how many times the fileSetContent will be called. The sample text is in the cpp file and can be changes. I set it to 10 bytes for convenience (idk how SPIFFS reserves bytes for the filename).

You have to try with some random repetition values, 100 is good to begin with. When you figure out where "api_spiffs_erase" is shown you can decrease the argument of ```save```to this position -1.

I got "api_spiffs_erase" on the 63rd write (630 bytes of data excluding what bytes were reserved for file name, spiffs, etc).

If I try to repeat the file save 62 times, the FS works OK, the device can be safely rebooted w/o data loss.
After the reboot only one ```save``` operation is needed to reach "api_spiffs_erase" and the FS gets immediately broken. On the next reboot it is formatted  by the firmware.

Example code is provided here: https://github.com/kmihaylov/Sming_fsTest.git

Tested with commit  5300c0b (current develop)

```
/opt/esp-open-sdk/esptool/esptool.py -p /dev/ttyUSB0 -b 115200 write_flash -ff 40m -fm qio -fs 32m 0x00000 out/firmware/rboot.bin 0x02000 out/firmware/rom0.bin 0x100000 out/firmware/spiff_rom.bin
WARNING: Flash size arguments in megabits like '32m' are deprecated.
Please use the equivalent size '4MB'.
Megabit arguments may be removed in a future release.
esptool.py v2.6-beta1
Serial port /dev/ttyUSB0
Connecting....
Detecting chip type... ESP8266
Chip is ESP8266EX
Features: WiFi
MAC: bc:dd:c2:91:cf:b1
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Compressed 2032 bytes to 1429...
Wrote 2032 bytes (1429 compressed) at 0x00000000 in 0.1 seconds (effective 123.9 kbit/s)...
Hash of data verified.
Compressed 275280 bytes to 196722...
[...]

PR #1667, opened 2019-04-12
Disable SPIFFS init in spiffs_mount_internal()

kmihaylov:
Dead simple solution to the problems in #1660 and childish solution in #1663.

I'm considering these scenarios:
1. Flashing a firmware almost always requires something like a ```flashinit```.
2. If flash wasn't erased, probably there is some reason not to have been erased, probably there is already existing spiffs.
3. If it is an example or other working application that requires flash (contents), probably it already flashes its own image.

If it is not any of the cases above, the current code checks the first 4 bytes of the beginning of the partition, if they're 0xFF (explicitly UINT32_MAX), which is the expected state for erased flash, then it does erase the flash again, which doesn't make sense. (This is assumption).

If there was some non spiffs data and the flash wasn't erased (like my try to mount random-data) a user should notice non-functioning fs and will probably take care (most probably this will be the developer).

If it is working application that has been tested and it was working for some time, if some unexpected data appears in the spiffs address space, then there is some worse problem - software bug or hardware failure.

---------

What do we get now?
1. If the flash was (most probably) erased, we erase it again in order to achieve:
2. Granted data loss once the partition was filled (due to the wear-leveling technique).

This is very quick workaround before better method for mount/erase/check is implemented.

### DIFF
--- a/Sming/Components/spiffs/spiffs_sming.c
+++ b/Sming/Components/spiffs/spiffs_sming.c
@@ -107,18 +107,6 @@ static void spiffs_mount_internal(spiffs_config *cfg)
   cfg->hal_write_f = api_spiffs_write;
   cfg->hal_erase_f = api_spiffs_erase;
   
-  uint32_t dat;
-  bool writeFirst = false;
-  flashmem_read(&dat, cfg->phys_addr, 4);
-  //debugf("%X", dat);
-
-  if (dat == UINT32_MAX)
-  {
-	  debugf("First init file system");
-	  spiffs_format_internal(cfg);
-	  writeFirst = true;
-  }
-
   int res = SPIFFS_mount(&_filesystemStorageHandle,
     cfg,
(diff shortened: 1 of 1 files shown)


kmihaylov:
May I know your opinion on this PR?

mikee47:
Simplifying the initialisation is a good thing. Ideally `spiffs_mount()`, `spiffs_mount_internal()` and `spiffs_mount_manual()` would return an error code. However, that would be inconsistent with `spiffs_format()` and `spiffs_format_manual()` which return `bool`. Either way it would be for the application to check the return value and act accordingly, choosing to check/format as required.

dpechman:
is this issue fix or have another treatment in Sming 4.1.0 ? I have just migrated a project from v3.8.1 and would like to have sure its fixed...

kmihaylov:
@dpechman #2034 was merged.

PR #1669, opened 2019-04-13
Wrong address for `isFlashPtr` - 0x40200000 not 0x4020000

mikee47:


mikee47:
At some point we need to sort out all these addresses and put them in a single file - like 'memory-map.h' or similar.

slaff:
> At some point we need to sort out all these addresses and put them in a single file - like 'memory-map.h' or similar.

Where are the other places where we need the addresses from the memory map?

mikee47:
> > At some point we need to sort out all these addresses and put them in a single file - like 'memory-map.h' or similar.
> 
> Where are the other places where we need the addresses from the memory map?

# ESP8266 Memory Map

I would prefer to see all the memory base definitions in a single header file, so we can see at a glance how the memory is organised.

https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map

At present those base addresses are scattered around various places - see list below.

The starting point for this should probably be `xtensa/config/core-isa.h` as that defines the core architecture on which the ESP8266 is built. For example, `XCHAL_INSTROM0_VADDR` is the flash ROM start address.

The ESP32 IDF does this -

https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/include/soc/soc.h

Hardware modules are defined in separate files; in contrast we have `peri.h` which lumps most of them together.

The IDF also provides struct versions, such as:

https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/include/soc/uart_struct.h

I've used this approach in my SPI driver rewrite and it makes the code a lot easier to understand.


## system/flashmem.h

	32
	#define INTERNAL_FLASH_START_ADDRESS 0x40200000

## gdb/gdbstub.cpp

	139
	static bool ATTR_GDBEXTERNFN isValidWriteAddr(uint32_t addr)
	{
	  return (addr >= 0x3ff00000 && addr < 0x40000000) || (addr >= 0x40100000 && addr < 0x40140000) ||
	       (addr >= 0x60000000 && addr < 0x60002000);
	}

	153
	static bool ATTR_GDBEXTERNFN isValidReadAddr(uint32_t addr)
	{
	  // @todo Consider whether it's necessary to restrict read addresses at all
	  return addr >= 0x20000000 && addr < 0x60000000;
	}


## system/include/espinc/peri.h

	31
	#define ESP8266_REG(addr) *((volatile uint32_t *)(0x60000000+(addr)))
	#define ESP8266_DREG(addr) *((volatile uint32_t *)(0x3FF00000+(addr)))

Throughout this file register blocks are defined in absolute terms, rather than grouped by function. For example:

	194
	//UART 0 Registers
	#define U0F   ESP8266_REG(0x000) //UART FIFO
	#define U0IR  ESP8266_REG(0x004) //INT_RAW
	#define U0IS  ESP8266_REG(0x008) //INT_STATUS
	#define U0IE  ESP8266_REG(0x00c) //INT_ENABLE
	... etc

	//UART 1 Registers
	#define U1F   ESP8266_REG(0xF00) //UART FIFO
	#define U1IR  ESP8266_REG(0xF04) //INT_RAW
	#define U1IS  ESP8266_REG(0xF08) //INT_STATUS
	#define U1IE  ESP8266_REG(0xF0c) //INT_ENABLE

I would be in favour of removing these - the uart driver uses the alternate definitions at lines 228 onwards.


## system/include/espinc/uart_register.h

Not used by the framework - remove this file?

	10
	#define REG_UART_BASE( i )  (0x60000000+(i)*0xf00)


## gdb/gdb_hooks.cpp

	144
	debug_print_stack(reg.a[1], 0x3fffffb0);

	
	
## crash_handler.c

	62
	uint32_t stack_end = 0x3fffffb0;
	// it's actually 0x3ffffff0, but the stuff below ets_run
	// is likely not really relevant to the crash


## Platform/System.cpp

	106
	void SystemClass::setCpuFrequency(CpuFrequency freq)
	{
		if(freq == eCF_160MHz)
			REG_SET_BIT(0x3ff00014, BIT(0));
		else
			REG_CLR_BIT(0x3ff00014, BIT(0));
	
		ets_update_cpu_frequency(freq);
	}
	

	
	In the Non-OS SDK some of these are defined in eagle_soc.h:

	//Peripheral device base address define{{
	#define PERIPHS_DPORT_BASEADDR              0x3ff00000
	#define PERIPHS_GPIO_BASEADDR               0x60000300
	#define PERIPHS_TIMER_BASEDDR               0x60000600
	#define PERIPHS_RTC_BASEADDR                0x60000700
	#define PERIPHS_IO_MUX						0x60000800
	//}}

Not complete by any means (uart base is missing, for instance)

PR #1672, opened 2019-04-14
Refactoring/buffer classes

slaff:


mikee47:
@slaff Other than the fact these classes all have `buffer` in their names, any other reason for this change? In particular, `SerialBuffer` is not a framework class, but an essential part of the uart driver.

PR #1677, opened 2019-04-29
Updated ArduinoJson to version 6

slaff:


mikee47:
Thanks for this Slaff. There's a lot to like about the new version, not least support for msgpack.

To stress things a bit I ported my `HttpServer_IFS` sample application, which uses Json extensively.

The move to `DynamicJsonDocument` with its predefined allocation size is potentially a tricky thing to deal with for serialisation as more care is required to select an appropriate limit. In particular, serialising file listings via Json required a buffer size of 4K to get it working. Something for the future is to find a way to get true Json serialisation streaming; using the file listing as an example, we'd ideally just serialise one file record at a time via callback so potentially large datasets could be serialised with low memory footprint.

I patched ArduinoJson V5 so `FlashString`s could be used as arguments, e.g.

```
DEFINE_FSTR(varname1, "varname1");
...
json[varname1] = "value of var1";

```

This is accomplished even more easily with V6 by writing a StringAdapter for `FlashString`. We don't need to patch ArduinoJson for this, instead we could provide our own `ArduinoJson.h` - I put it in `SmingCore`:

```
#ifndef SMINGCORE_ARDUINOJSON_H_
#define SMINGCORE_ARDUINOJSON_H_

#include "Libraries/ArduinoJson/ArduinoJson.h"
#include "Data/FlashStringRefAdapter.hpp"

#endif /* SMINGCORE_ARDUINOJSON_H_ */
```

We therefore only need to add `#include <ArduinoJson.h>` to our code to get a customisable version of the library. Here's `FlashStringRefAdapter.hpp` -


```
// ArduinoJson - arduinojson.org
// Copyright Benoit Blanchon 2014-2019
// MIT License
//
// Sming FlashString adapter mikee47 April 2019 <mike@sillyhouse.net>

#pragma once

namespace ARDUINOJSON_NAMESPACE
{
class FlashStringRefAdapter
{
public:
	FlashStringRefAdapter(const FlashString& str) : _str(str)
	{
	}

	bool equals(const char* expected) const
	{
		return _str.isEqual(expected);
	}

	bool isNull() const
	{
		return false;
	}

	char* save(MemoryPool* pool) const
	{
		size_t n = _str.size();
		char* dup = pool->allocFrozenString(n);
		if(dup)
			memcpy_P(dup, _str.data(), n);
		return dup;
	}

	const char* data() const
	{
		return nullptr;
	}

	size_t size() const
	{
		return _str.length();
	}

	bool isStatic() const
	{
		return false;
	}

private:
	const FlashString& _str;
};

inline FlashStringRefAdapter adaptString(const FlashString& str)
{
	return FlashStringRefAdapter(str);
}

template <> struct IsString<FlashString> : true_type {
};

} // namespace ARDUINOJSON_NAMESPACE
```

mikee47:
In V5, `str` gets invalidated, but in V6 gets set to "Null".

```
String str = json["non-existent key"];
```

Sming/Services/Yeelight/YeelightBulb.cpp
```diff
@@ -157,32 +156,31 @@ bool YeelightBulb::onResponse(TcpClient& client, char* data, int size)
...
+					JsonArray result = doc["result"];
+					String resp = result[0];
 					parsePower(resp);
 				}
 			}
-			if (root.containsKey("method") && root.containsKey("params"))
+			if (doc.containsKey("method") && doc.containsKey("params"))
 			{
-				String method = root["method"].asString();
+				String method = doc["method"];
```
mikee47: should be `doc["method"].as<const char*>()`

samples/HttpServer_ConfigNetwork/include/AppSettings.h
```diff
@@ -24,34 +24,33 @@ struct ApplicationSettingsStorage {
...
+				dhcp = network["dhcp"];
 
-			dhcp = network["dhcp"];
-
-			ip = network["ip"].asString();
-			netmask = network["netmask"].asString();
-			gateway = network["gateway"].asString();
+				ip = network["ip"].as<String>();
+				netmask = network["netmask"].as<String>();
+				gateway = network["gateway"].as<String>();
```
mikee47: `ssid`, `password`, `ip`, `netmask` and `gateway` should probably all use `as<const char*>()`

mikee47:
For efficient reading of values I try to avoid calling twice if possible, i.e. calling `containsKey(KEY)` first then another call to get the value. I've worked up this template function to assist. Could add it to `Core/ArduinoJson.h` perhaps.

```
/**
 * @brief Helper function to copy a Json data value, provided it's valid
 * @param data Value returned from JsonObject[key], JsonDocument[key] or JsonVariant[key]
 * @param value Where to write value, unchanged if data is null
 * @retval bool true if value was read, false if not
 * @note Example:
 *
 *		JsonDocument doc;
 *		...
 *		String value1;
 *		if(!jsonGetValue(doc["key1"], value1)) {
 *			value1 = F("default value");
 *		}
 *
 *		// Alternative for String values
 *		String value1 = doc["key1"].as<const char*>() ?: F("default value");
 */
template <typename TObject, typename TStringRef, typename TValue>
bool jsonGetValue(const ARDUINOJSON_NAMESPACE::MemberProxy<TObject&, TStringRef&>& data, TValue& value)
{
	if(data.isNull()) {
		return false;
	} else {
		value = data;
		return true;
	}
}
```

Sming/Services/Yeelight/YeelightBulb.cpp
slaff: As far as I understood from ["Automatic conversion"](https://arduinojson.org/v6/doc/upgrade/) that should happen automatically and there is not extra need to cast it to string.

slaff:
> Could add it to Core/ArduinoJson.h perhaps.

Very good idea! If you want create PR(s) adding to the current PR. And I will merge the updates. One way to to this would be:
- git clone https://github.com/slaff/Sming.git 
- git checkout update/arduino-json-v6
- git branch improvements/arduino-json-v6
- //  add your improvements.
- // commit locally your improvements
- // once ready push to remote
- git remote add my https://github.com/mikee47/Sming.git
- git push -u my improvements/arduino-json-v6
// if needed add more improvements.
// Once ready create a PR in slaff/Sming
// I will merge them and do the rest.

Sming/Services/Yeelight/YeelightBulb.cpp
mikee47: https://github.com/bblanchon/ArduinoJson/issues/420

mikee47:
```
void testAsString()
{
	StaticJsonDocument<256> doc;
	doc["exists"] = "abc";
	String s1 = doc["method"];
	debug_i("S1 = %d chars '%s'", s1.length(), s1.c_str());
	String s2 = doc["method"].as<const char*>();
	debug_i("S2 = %d chars '%s'", s2.length(), s2.c_str());
	String s3 = (const char*)doc["method"];
	debug_i("S3 = %d chars '%s'", s3.length(), s3.c_str());
}
```
Output:
```
4642436 S1 = 4 chars 'null'
4642528 S2 = 0 chars ''
4642608 S3 = 0 chars ''
```

mikee47:
OK. Do you intend to use `ArduinoJson` as a submodule, or keep it as-is ?

Sming/Services/Yeelight/YeelightBulb.cpp
bblanchon: The most portable and explicit way is to use [JsonVariant::operator|](https://arduinojson.org/v6/api/jsonvariant/or/) to specify the default value:

```c++
String value = doc["key"] | "";
```

See a comparative demo here:
https://wandbox.org/permlink/oxMsD1f3qYFq9jC0

mikee47: That's a much cleaner solution. Thank you.

slaff:
> OK. Do you intend to use ArduinoJson as a submodule, or keep it as-is ?

I will move it to a submodule.

slaff:
@mikee47 When you have time can you review the latest changes (https://github.com/SmingHub/Sming/pull/1677/commits/184fefcda10f0e4fc3a3f42af74d09a2619d9bfc) ?

Sming/Services/Yeelight/YeelightBulb.cpp
```diff
@@ -157,33 +156,29 @@ bool YeelightBulb::onResponse(TcpClient& client, char* data, int size)
...
-				if (id == propsId)
-				{
-					auto &result = root["result"].asArray();
-					String resp = result[0].asString();
+				JsonArray result = doc["result"];
+				String resp = result[0] | "";
+				if(resp.length()) {
 					parsePower(resp);
 				}
 			}
```
mikee47: I'd be tempted to rewrite like this:

```
String resp = result[0].as<const char*>;
if(resp) {
	parsePower(resp);
}
```

Because we're testing for the existence of a value, not so much providing a default.

Sming/Services/Yeelight/YeelightBulb.cpp
```diff
@@ -157,33 +156,29 @@ bool YeelightBulb::onResponse(TcpClient& client, char* data, int size)
...
-			if (root.containsKey("method") && root.containsKey("params"))
+
+			String method = doc["method"] | "";
```
mikee47: Again, I'd just use `String method = doc["method"].as<const char*>;`

Sming/Core/Data/Stream/JsonObjectStream.h
```diff
@@ -24,7 +24,7 @@ class JsonObjectStream : public MemoryDataStream
 public:
 	/** @brief  Create a JSON object stream
     */
-	JsonObjectStream() : rootNode(buffer.createObject())
+	JsonObjectStream() : doc(1024)
```
mikee47: Given that buffer size is fixed, I'd add a parameter here `JsonObjectStream(size_t capacity = 1024) : doc(capacity)`

Sming/Core/Data/Stream/JsonObjectStream.h
```diff
@@ -51,12 +51,11 @@ class JsonObjectStream : public MemoryDataStream
 	 */
 	int available() override
 	{
-		return rootNode.success() ? rootNode.measureLength() : 0;
+		return (doc.isNull() ? 0 : measureJson(doc));
```
mikee47: I was wondering whether the null test was necessary, so I tested `measureJson()` on an empty document - returns 4! ("null"). Which leads to another question of what constitutes a valid 'empty' JSON document. Technically, it should probably be `{}` though there is some disagreement on that... I wonder what a browser would accept?

Sming/Services/Yeelight/YeelightBulb.cpp
```diff
@@ -157,33 +156,29 @@ bool YeelightBulb::onResponse(TcpClient& client, char* data, int size)
...
+				for (JsonObject::iterator it=result.begin(); it!=result.end(); ++it)
 				{
-					auto &result = root["params"].asObject();
-					for (JsonObject::iterator it=result.begin(); it!=result.end(); ++it)
-					{
-						if (strcmp(it->key, "power") == 0)
-							parsePower(it->value);
-					}
+					if (strcmp(it->key().c_str(), "power") == 0)
+						parsePower(it->value());
```
mikee47: I wonder why we're bothering with the iterator here, as we 're only after one value...

Sming/Wiring/FakePgmSpace.h
```diff
@@ -117,6 +117,7 @@ extern "C"
 	char *strcpy_P(char * dest, const char * src_P);
 	char *strncpy_P(char * dest, const char * src_P, size_t size);
 	int strcmp_P(const char *str1, const char *str2_P);
+	int strncmp_P(const char *str1, const char *str2_P, const size_t size);
 	int strcasecmp_P(const char* str1, const char* str2_P);
```
mikee47: Should also add this lower down for #ifndef ICACHE_FLASH

```
#define strncmp_P(str1, str2_P, size) strncmp(str1, str2_P, size)
```

mikee47:
`strncmp_P` still missing for non-ICACHE_FLASH...

slaff:
> strncmp_P still missing for non-ICACHE_FLASH...

Very good eyes. Thanks for noticing this. The change is now part of this PR ( https://github.com/SmingHub/Sming/pull/1677/files#diff-d545a6368c80287a194853b644bb931aR180 ).  If you want make a last check before merging.

mikee47:
Looks good. I was wondering about adding `Json::saveToStream` etc. for consistency?

slaff:
> Looks good. I was wondering about adding Json::saveToStream etc. for consistency?

Yep, go for it.

mikee47:
I've opened a PR on ArduinoJson to add support for Stream pointers - if that gets accepted we can simplify this PR.

Sming/Wiring/FakePgmSpace.cpp
```diff
@@ -55,6 +55,18 @@ int strcmp_P(const char *str1, const char *str2_P)
 	return *(unsigned char *)str1 < (unsigned char)pgm_read_byte(str2_P) ? -1 : 1;
 }
 
+
+int strncmp_P(const char *str1, const char *str2_P, const size_t size)
+{
+	for (int i=0; *str1 == pgm_read_byte(str2_P); str1++, str2_P++, i++) {
```
mikee47: for(unsigned i

mikee47:
> I've opened a PR on ArduinoJson to add support for Stream pointers in serialisation/de-serialisation - if that gets accepted we can simplify this PR.

Nope, doesn't like [Stream pointers](https://github.com/bblanchon/ArduinoJson/pull/978) so wrapper functions will have to do.

PR #1680, opened 2019-05-03
Fix FileStream write caching

### LINKED ISSUE #1679: SPIFFS write caching not working with FileStream
Excessive flash writing is occurring when using a `FileStream` for writing to a file in small chunks (e.g. serialising JSON data directly to a file stream). The behaviour is as if SPIFFS write caching isn't working at all.

Investigation shows that `FileStream::write` calls `fileSeek` before a `fileWrite` to ensure the data is written to the end of a file. The problem is that `SPIFFS_lseek` flushes the write cache, every time.

Removing the call to `fileSeek` fixes the problem and performance is restored. However, that's probably too simple as a proper fix as we'd potentially lose position tracking. Maybe arrange things so that `fileSeek` is only called if the file position has been changed, e.g. by another seek or read operation.

mikee47:
Add position check in `FileStream::write` before calling `fileSeek` to avoid flushing SPIFFS write cache.

Fixes #1679

### DIFF
--- a/Sming/Core/Data/Stream/FileStream.cpp
+++ b/Sming/Core/Data/Stream/FileStream.cpp
@@ -91,9 +91,7 @@ size_t FileStream::write(const uint8_t* buffer, size_t size)
 	int written = fileWrite(handle, buffer, size);
 	if(check(written)) {
 		pos += size_t(written);
-		if(pos > this->size) {
-			this->size = pos;
-		}
+		this->size = pos;
 	}
 
 	return written;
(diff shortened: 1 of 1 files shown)


slaff:
@mikee47 Is tested and ready to merge?

mikee47:
@slaff Yes, I've tested this fix and happy with it for now, but generally I think we need to take another look at streams to ensure they all behave consistently, probably document behaviour a little better. Some of the streams are 'forward-only' (e.g. StreamTransformer-based ones) whilst others support proper seeking (`FileStream` does, `MemoryDataStream` should but doesn't, etc.). Opening a FileStream for read/write definitely needs clarifying - expected behaviour would be for `write` to happen at the current position, but it doesn't.

PR #1687, opened 2019-05-08
Fix multipart stream boundary value calculation

mikee47:
* May contain invalid characters (probably NULs) because `sizeof(pool)` is 64, but actual size is 63.
* Clarify use of `PSTR_ARRAY` and getting correct size by re-defining flash data using `PSTR_` prefix.


mikee47:
Needs backporting to LTS. Sorry.

PR #1695, opened 2019-05-21
Revise system headers / enforce c11

mikee47:
Build `C` files within modules to c11 standard (previously defaulted to `gnu90`)

Standardise data types, e.g. use `uint32_t` not `uint32`
* Arduino libraries use a mixture so still need to maintain definitions for those

Tidy and simplify `c_types.h`
* Rename `espinc/c_types_compatible.h` as `c_types.h`
* Revise to use `stdint` and `stdbool` definitions
* Move attribute definitions and macros into `esp_attr.h`
* Put common attribute definitions into `sming_attr.h`

`user_config.h` shouldn't contain system definitions
* Move system definitions from `user_config.h` into `esp_systemapi.h` (exclude LWIP)
* Don't #include `user_config.h` from `esp_systemapi.h`
* Move missing memory API definitions out of `esp_systemapi.h` and into `mem.h`

Simplify Library/sample code using <esp_systemapi.h> instead of assortment of low-level headers

Add `WEAK_ATTR` macro to allow selective weak symbol support
* Not a C/C++ standard feature and not consistently applied on all platforms, so avoid where possible.


Sming/Arch/Esp8266/Components/axtls-8266/axtls-8266.patch
```diff
@@ -21,8 +21,7 @@ diff -Nuar a/replacements/libc.c b/replacements/libc.c
 + Modified 03 April 2015 by Markus Sattler
 + */
 +
-+#include <stdint.h>
-+#include "espinc/c_types_compatible.h"
++#include <c_types.h>
 +#include <stdarg.h>
 +
 +extern int ets_putc(int);
```
slaff: In this PR  `make dist-clean` did not clean the axtls directory and the new patches was not applied cleanly. Any idea why is that happening?

mikee47: Try `make dist-clean ENABLE_SSL=1`

mikee47: We have these optional custom libraries which get built:

	pwm_open
	mainmm
	lwip_full
	lwip_open
	lwip2
	axtls
	mqttc

Behaviour is that the library only gets cleaned if the relevant flags are defined. It would be better of course if they all got cleaned regardless. I can look at this in another PR.

slaff: >  It would be better of course if they all got cleaned regardless. I can look at this in another PR.

Absolutely. It will also match the current default behaviour.

PR #1711, opened 2019-05-31
Move the generated flash.bin to out/firmware.

slaff:


mikee47:
I deliberately didn't do that because `make clean` and it's gone. It represents the flash state of a virtual device (and we may want several) so should be persistent.

slaff:
> I deliberately didn't do that because make clean and it's gone. It represents the flash state of a virtual device (and we may want several) so should be persistent.

Hm... having this file in the root folder of the app is not something that I really like. Also flash.bin has to be ignored by git.

mikee47:
In actual use the default location should be probably somewhere entirely separate, like `~.sming/host/` or `~sming/data/`. There are other things I considered, like adding versioning/backup so if something messes the flash up we can compare with the previous state. Also using a filename based on the name of the project. In the end I decided to keep things simple and let the user decide how they want to roll.

Maybe make `FLASH_BIN` easy to change, possibly rename to `SMING_FLASH_BIN`? 

```
diff --git a/Sming/Arch/Host/flash.mk b/Sming/Arch/Host/flash.mk
index a5a1a546..e8e009a2 100644
--- a/Sming/Arch/Host/flash.mk
+++ b/Sming/Arch/Host/flash.mk
@@ -27 +27 @@ DD			:= dd
-FLASH_BIN	:= $(FW_BASE)/flash.bin
+SMING_FLASH_BIN	?= $(FW_BASE)/flash.bin
```

mikee47:
Wonder if there are other related emulator files which might be required? If so, maybe keep `FLASH_BIN` internal and add a `SMING_HOST_DIR` as the default location.

slaff:
> Maybe make FLASH_BIN easy to change, possibly rename to SMING_FLASH_BIN? (And update docs.)

Sounds good to me. I have updated the PR.

Sming/Arch/Host/flash.mk
```diff
@@ -1,41 +1,41 @@
...
+endif
+
+CFLAGS		+= -DSPIFF_SIZE=$(SPIFF_SIZE)
+
+
+# Use DD to update $(FW_BASE)/flash.bin
+
+DD			:= dd
+
+SMING_FLASH_BIN	?= $(FW_BASE)/flash.bin
```
mikee47: Need to tag add `CONFIG_VARS += SMING_FLASH_BIN` so `make list-config` shows the value.

Sming/Arch/Host/flash.mk
```diff
@@ -1,41 +1,41 @@
...
-endef
-
-#
-ERASE_FLASH	:= $(DD) if=/dev/zero ibs=1 count=$(SPI_SIZE) | tr "\000" "\377" > $(FLASH_BIN) 
+#
+# Flash parameters
+#
+
+# SPI_SIZE: 512K, 256K, 1M, 2M, 4M
+SPI_SIZE				?= 512K
```
mikee47: Can we add `CONFIG_VARS += SPI_SPEED SPI_MODE SPI_SIZE SPIFF_SIZE` ?

mikee47:
.. and app.mk (line 86) needs updating

slaff:
All recommended changes are added to this PR. Other recommendations or shall I merge the PR?

mikee47:
Nothing else to suggest, though now I'm not so sure about whether `SMING_FLASH_BIN` or `FLASH_BIN` is more appropriate! (sorry) There are lots of other config vars which don't prefix `SMING_` so maybe shouldn't start now... I'll let you decide!

slaff:
> I'll let you decide! 

Here we go: SMING_ prefix should be used only for important settings. Thus FLASH_BIN is the better name. I will change this, push PR and merge it.

PR #1719, opened 2019-06-07
Fix/gcc8 compiler issues 1 and Host makefile extraneous character

frankdownunder:
I use G++ 8.3.0 which is fussy, it found some dodgy code in RCSwitch.
I also found an extraneous $ sign in the makefile that build the Host architecture

### DIFF
--- a/Sming/Arch/Host/sming.mk
+++ b/Sming/Arch/Host/sming.mk
@@ -62,7 +62,7 @@ $(call UserLibPath,lwip): $(LWIP_BUILD_DIR)/Makefile
 
 $(LWIP_BUILD_DIR)/Makefile: $(LWIP_BASE)/lwip/.submodule | $(LWIP_BUILD_DIR)
 	$(Q) cd $(LWIP_BUILD_DIR); \
-	$(CMAKE) -DUSER_LIBDIR="$(SMING_HOME)/$(USER_LIBDIR)" $(CMAKE_OPTIONS) $(SMING_HOME$)/$(LWIP_BASE)/$(UNAME)
+	$(CMAKE) -DUSER_LIBDIR="$(SMING_HOME)/$(USER_LIBDIR)" $(CMAKE_OPTIONS) $(SMING_HOME)/$(LWIP_BASE)/$(UNAME)
 
 $(LWIP_BUILD_DIR):
 	mkdir -p $@
--- a/Sming/Libraries/RCSwitch/RCSwitch.cpp
+++ b/Sming/Libraries/RCSwitch/RCSwitch.cpp
@@ -285,7 +285,8 @@ char* RCSwitch::getCodeWordB(int nAddressCode, int nChannelCode, boolean bStatus
    
    const char* code[5] = { "FFFF", "0FFF", "F0FF", "FF0F", "FFF0" };
    if (nAddressCode < 1 || nAddressCode > 4 || nChannelCode < 1 || nChannelCode > 4) {
-    return nullptr;
+     sReturn[0] = '\0';
+     return sReturn;
    }
    for (int i = 0; i<4; i++) {
      sReturn[nReturnPos++] = code[nAddressCode][i];
@@ -348,7 +349,8 @@ char* RCSwitch::getCodeWordC(char sFamily, int nGroup, int nDevice, boolean bSta
   int nReturnPos = 0;
   
   if ( (byte)sFamily < 97 || (byte)sFamily > 112 || nGroup < 1 || nGroup > 4 || nDevice < 1 || nDevice > 4) {
-    return nullptr;
+     sReturn[0] = '\0';
+     return sReturn;
   }
   
   const char* sDeviceGroupCode =  dec2binWcharfill(  (nDevice-1) + (nGroup-1)*4, 4, '0'  );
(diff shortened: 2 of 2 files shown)


Sming/Arch/Host/sming.mk
```diff
@@ -62,7 +62,7 @@ $(call UserLibPath,lwip): $(LWIP_BUILD_DIR)/Makefile
 
 $(LWIP_BUILD_DIR)/Makefile: $(LWIP_BASE)/lwip/.submodule | $(LWIP_BUILD_DIR)
 	$(Q) cd $(LWIP_BUILD_DIR); \
-	$(CMAKE) -DUSER_LIBDIR="$(SMING_HOME)/$(USER_LIBDIR)" $(CMAKE_OPTIONS) $(SMING_HOME$)/$(LWIP_BASE)/$(UNAME)
+	$(CMAKE) -DUSER_LIBDIR="$(SMING_HOME)/$(USER_LIBDIR)" $(CMAKE_OPTIONS) $(SMING_HOME)/$(LWIP_BASE)/$(UNAME)
```
slaff: Nice catch!

Sming/Libraries/RCSwitch/RCSwitch.cpp
```diff
@@ -285,7 +285,7 @@ char* RCSwitch::getCodeWordB(int nAddressCode, int nChannelCode, boolean bStatus
    
    const char* code[5] = { "FFFF", "0FFF", "F0FF", "FF0F", "FFF0" };
    if (nAddressCode < 1 || nAddressCode > 4 || nChannelCode < 1 || nChannelCode > 4) {
-    return '\0';
+    return nullptr;
```
slaff: @mikee47 What do you think about this change?

mikee47: Code will behave as it did before, which is probably wrong! It's used by `sendTriState()` which will then just segfault if it gets a NULL value. It should therefore be:

```
   if (nAddressCode < 1 || nAddressCode > 4 || nChannelCode < 1 || nChannelCode > 4) {
     sReturn[0] = '\0';
     return sReturn;
   }
```

Sming/Libraries/RCSwitch/RCSwitch.cpp
```diff
@@ -348,7 +348,7 @@ char* RCSwitch::getCodeWordC(char sFamily, int nGroup, int nDevice, boolean bSta
   int nReturnPos = 0;
   
   if ( (byte)sFamily < 97 || (byte)sFamily > 112 || nGroup < 1 || nGroup > 4 || nDevice < 1 || nDevice > 4) {
-    return '\0';
+    return nullptr;
```
mikee47: As for `getCodeWordB()`

Sming/Libraries/RCSwitch/RCSwitch.cpp
```diff
@@ -418,7 +418,7 @@ char* RCSwitch::getCodeWordD(char sGroup, int nDevice, boolean bStatus){
         case 'D':
             sGroupCode = dec2binWcharfill(1, 4, 'F'); break;
         default:
-            return '\0';
+            return nullptr;
```
mikee47: As for `getCodeWordB()`

Sming/Libraries/RCSwitch/RCSwitch.cpp
```diff
@@ -437,7 +437,7 @@ char* RCSwitch::getCodeWordD(char sGroup, int nDevice, boolean bStatus){
         case 3:
             sDevice = dec2binWcharfill(1, 3, 'F'); break;
         default:
-            return '\0';
+            return nullptr;
```
mikee47: As for `getCodeWordB()`

mikee47:
`RCSwitch ` appears to have had some updates since this version, but curiously this bug hasn't been fixed. Perhaps it should be re-implemented as a patched submodule?

slaff:
> RCSwitch appears to have had some updates since this version,

The latest update that I can find is from Nov 22, 2017. Is https://github.com/sui77/rc-switch the repo you are talking about or there is newer one?

mikee47:
That's the one mentioned in the README.md, so I guess that's it. According to GIT the code hasn't been updated since it was added 1/5/2016.

mikee47:
There are quite a few libraries which have been copied in rather than added as submodules, likely they're out of date too. I haven't done anything about this myself because it would require testing the changes, although we could just patch it and wait for the bug reports to come in...

slaff:
> we could just patch it and wait for the bug reports to come in..

As long as the patch(es) fix issues in the RCSwitch code it would be best to send them upstream to the original project. Once they are merged I will take care to put RCSwitch as a submodule and provide 
a patch with our own changes.

frankdownunder:
How long do you think it will take to get a proper fix? In the mean time, it doesnt compile for me!
Can we add it to the list of EXCLUDE_LIBRARIES or remove it or something? 
Or, since I havent changed the behaviour, implement my change as a temporary measure?

slaff:
> How long do you think it will take to get a proper fix

That is the proper fix as suggested by Mike. 
```
if (nAddressCode < 1 || nAddressCode > 4 || nChannelCode < 1 || nChannelCode > 4) {
     sReturn[0] = '\0';
     return sReturn;
   }
```

@frankdownunder please, apply the fix it in this PR and I will take care to submit it as PR  to upstream.

PR #1721, opened 2019-06-12
Add additional checks to `Vector` class methods

mikee47:
https://github.com/SmingHub/Sming/pull/1618#discussion_r256467818

* Change `ensureCapacity` to adjust new capacity by `increment`, thus simplifying code which uses it
* Add `bool` return value to:
	`addElement()` - so that `add()` return value isn't meaningless
	`ensureCapacity()`
	`setSize()`
	`insertElementAt()`
	`setElementAt()`

### DIFF
--- a/Sming/Wiring/WVector.h
+++ b/Sming/Wiring/WVector.h
@@ -48,25 +48,24 @@ class Vector : public Countable<Element>
      void copyInto(Element* array) const;
      bool add(const Element& obj)
     {
-      addElement(obj);
-      return true;
+      return addElement(obj);
     }
-     void addElement(const Element& obj);
-     void addElement(Element* objp);
+     bool addElement(const Element& obj);
+     bool addElement(Element* objp);
      void clear()
     {
       removeAllElements();
     }
-     void ensureCapacity(unsigned int minCapacity);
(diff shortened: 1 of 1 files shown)


Sming/Wiring/WVector.h
```diff
@@ -266,55 +265,49 @@ unsigned int Vector<Element>::size() const
...
 
 template <class Element>
-void Vector<Element>::addElement(const Element &obj)
+bool Vector<Element>::addElement(const Element &obj)
 {
-  if (_size == _capacity)
-    ensureCapacity(_capacity + _increment);
-  if (_size < _capacity)
-    _data[ _size++ ] = new Element(obj);
+  if (!ensureCapacity(_size + 1)) return false;
```
slaff: Can you please at least put the `return` on the next line with a bit of offset ? Best would be to use curly braces. 
Apply those changes to the other places where you have a return on the same line.

mikee47: It would be simplest to just fix all the coding style issues in this file, i.e. just apply the Sming CS and fix all the style issues. The Arduino Esp8266 doesn't use this class (uses std:;vector instead) so it's really a Sming-only file ATM.

PR #1722, opened 2019-06-12
Revise `BMPDraw.cpp` as a template function and add to Adafruit_GFX library

mikee47:
* Allows use with a wide range of TFT displays without requiring a virtual base class
* Eliminate code duplication


mikee47:
This is just code refactoring, but haven't yet tested on hardware.

Sming/Libraries/Adafruit_GFX/BMPDraw.h
```diff
@@ -0,0 +1,193 @@
...
+ ***************************************************/
+/*
+ * Revised as a template class so it can be used with any supported Adafruit display library,
+ * such as Adafruit_ILI9341, Adafruit_ST7735, etc.
+ * June 2019 mikee47 <mikee@sillyhouse.net>
+ */
+
+#pragma once
+
+#include <SmingCore.h>
```
slaff: (Nice-to-have) If we use this include then the code won't be easily reusable by the Arduino community. Would it be possible to replace it with <Arduino.h> and other standard Arduino includes?

mikee47: Good point, I've replaced with Arduino.h which is sufficient to compile it. However, it does use the Sming file API so isn't portable from that perspective. There are various solutions to that but my motivation for including this PR now was the issue of maintaining two (almost) identical source code files.

PR #1723, opened 2019-06-16
Fix `return-type` and `maybe-uninitialized` compiler warnings

mikee47:


Sming/Arch/Esp8266/Components/gdbstub/gdbhostio.cpp
```diff
@@ -59,6 +59,7 @@ void ATTR_GDBEXTERNFN gdbHandleHostIo(char* commandBuffer, unsigned cmdLen)
 		unsigned flags = GdbPacket::readHexValue(data);
 		++data; // Skip ,
 		unsigned mode = GdbPacket::readHexValue(data);
+		(void)mode; // unused
```
slaff: It would be better to have a global SMING_UNUSED(X) macro. 

```
#define SMING_UNUSED(X) (void)X
```

So that the code above would look like:
```
SMING_UNUSED(mode); 
```

That change should be applied also to `Sming/Arch/Host/Components/hostlib/sockets.cpp` line 134. 
This way later on we call easily discover unused variables and update/improve the code if needed.

mikee47: I've re-coded to avoid the situation, should have done that in the first place!

PR #1726, opened 2019-06-21
Use explicit signed types in base64_decode

### LINKED ISSUE #1725: base64 decoding problem padding bytes
I have a problem regarding the base64 API. This base64 encoded string:
`YWRtaW46dGVzdA==`

gets decoded by this code:
`userPass = base64_decode(userPass)`

to this
`admin:test␏�`

So there are one or two extra bytes appended (depending on encoding).

When I remove the padding bytes manually before decoding:
`YWRtaW46dGVzdA`
then I get
`admin:test`
(which is what I want)

Is it me failing to understand base64 or is this a bug?

mikee47:
Code breaks if compiler doesn't default to signed chars, therefore explicit types must be used. In this case, the `int8_t` type is more appropriate than char. Fixes #1725.

### DIFF
--- a/Sming/Components/libb64/cdecode.c
+++ b/Sming/Components/libb64/cdecode.c
@@ -6,14 +6,16 @@ For details, see http://sourceforge.net/projects/libb64
 */
 
 #include "cdecode.h"
+#include <stdint.h>
 
-int base64_decode_value(char value_in)
+int8_t base64_decode_value(char value_in)
 {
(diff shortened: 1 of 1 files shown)


slaff:
@verybadsoldier Please, check this PR if it fixes the base64 problem that you are experiencing.

verybadsoldier:
Issue is fixed for me aftet picking this commit. Thanks alot.

PR #1727, opened 2019-06-22
Fixes ibraries/MFRC522/MFRC522.cpp:1460:4: error: 'invertedError' may…

kmihaylov:
… be used uninitialized in this function [-Werror=maybe-uninitialized]

The library should be probably updated though...

### DIFF
--- a/Sming/Libraries/MFRC522/MFRC522.cpp
+++ b/Sming/Libraries/MFRC522/MFRC522.cpp
@@ -1371,6 +1371,7 @@ void MFRC522::PICC_DumpMifareClassicSectorToSerial(Uid *uid,			///< Pointer to U
 	byte buffer[18];
 	byte blockAddr;
 	isSectorTrailer = true;
+	invertedError = false; // Avoid "unused variable" warning.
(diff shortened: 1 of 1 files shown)


Sming/Libraries/MFRC522/MFRC522.cpp
```diff
@@ -1371,6 +1371,7 @@ void MFRC522::PICC_DumpMifareClassicSectorToSerial(Uid *uid,			///< Pointer to U
 	byte buffer[18];
 	byte blockAddr;
 	isSectorTrailer = true;
+	invertedError = false; // Avoid "unused variable" warning.
```
slaff: If this variable is not used then the proper fix will be to remove line 1351 completely.

kmihaylov: The variable is used. I checked the original repo.

https://github.com/miguelbalboa/rfid/blob/72312cfe32718bcf7e4cd78210fac57afd50ebdc/src/MFRC522.cpp#L1534

PR #1730, opened 2019-06-23
Wemos D1 mini has pin 16, so change test to allow 16 as a valid pin.

slaff:


Sming/SmingCore/Digital.cpp
```diff
@@ -20,7 +20,7 @@ const uint8_t esp8266_gpioToFn[16] = {0x34, 0x18, 0x38, 0x14, 0x3C, 0x40, 0x1C,
 
 void pinMode(uint16_t pin, uint8_t mode)
 {
-	if(pin < 16) {
+	if(pin <= 16) {
```
mikee47: `pin == 16` is handled below separately, no fix required

slaff: Well spotted! Thanks, I will revert the commit shortly.

PR #1732, opened 2019-06-26
Added interface for std::function to Delegate

verybadsoldier:
Make it possible to pass `std::function` to `Delegate`.

### DIFF
--- a/Sming/Core/Delegate.h
+++ b/Sming/Core/Delegate.h
@@ -114,6 +114,32 @@ private:
 	MethodDeclaration mMethod;
 };
 
+/** @brief  Delegate std::function caller class
+*/
+template <class ReturnType, typename... ParamsList>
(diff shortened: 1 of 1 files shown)


slaff:
@verybadsoldier Please, fix the coding style. 
```
cd $SMING_HOME
make cs
```

and add the changes to this PR.

verybadsoldier:
Sorry, fixed.

Unsure how to handle nicely this ambiguity appearing in the samples betwen `std::function` and function pointer when using a lambda expression. An explicit casting might work but its looking quite ugly.

verybadsoldier:
Ok, I am giving up on this since I cannot find a way to solve this ambiguity problem regarding lambdas in a way that is not super ugly in the API. Sorry, for this waste of time.

Issue #1337, opened 2018-03-02
Make Delegators Compatible with Lambdas and std::function


zigarrre:
## Problem

Currently one can only create `Delegate` instances from raw function pointers or from a pointer to a member function and a pointer to a corresponding instance.

This means that `Delegate` does not work with modern C++ features such as lambda functions or `std::function`. (non-capturing lambdas already work as they can be casted to function pointers but captures are a integral part of lambdas that makes them much more useful)

This would be a minimum code example of using lambdas with delegates wich currently does NOT work:

```c++
Timer t;
int foo = 123;
t.initializeMs(1000,
    [foo] {
        // do something with foo
    }).start();
```

This trivial example might not fully show the usefulness of captures but it becomes obvious with more complex code. For example you can also capture `this` which means you could write callbacks using lambdas where you can access other members from within the callback as if it where part of the enclosing class. So without a doubt this would be an extremely useful feature to have.

## Solution

The `Delegate` class should accept ether a `std::function` or - probably even better - use a template to accept anything that is callable.

I don't fully understand all the stuff that is going on in `Delegate.h` so I don't feel comfortable making those changes myself in fear of unknown consequences.

slaff:
It would be probably better to replace completely Delegate with std::function as the latter is coming directly since C++ 11, as you correctly noted above. And [std::function](http://en.cppreference.com/w/cpp/utility/functional/function) should cover all the use cases that we have with the Delegates in Sming.

frankdownunder:
I had a quick try of std::function, with code like:
`typedef std::function<void(bool)> PumpDisabledDelegate;`
` class Pump `
`{ `
`    public: `
`        Pump(PumpDisabledDelegate pumpDisabledCallback); `
but got linker errors:
out/build/app_app.a(pump.o): In function std::function<void (bool)>::operator()(bool) const:
pump.cpp:390: undefined reference to std::__throw_bad_function_call()

Is there something I need to do to dismiss the exceptions?

slaff:
> Is there something I need to do to dismiss the exceptions?

Due to size considerations we are using microc / microgcc and that is most probably causing the problem. You can try to add the following code:

```
namespace std {void __throw_bad_function_call() { while(1); }; }
```

and check again.

frankdownunder:
Thanks Slaff that did the trick!

Sming currently provides Delegates to handle callbacks to member functions, eg
```
[...]

PR #1733, opened 2019-06-28
added interface for std::function to interrupt callbacks

verybadsoldier:
Ok, this is my next try in adding support for `std::function` to interrupt callbacks. This time it's less generic since I didn't touch `Delegate` directly. It is adding a dedicated type `InterruprDelegateStdFunction`. It follows the same pattern as `TimerDelegateStdFunction` so it is consistent in my opinion.

### DIFF
--- a/Sming/Arch/Esp8266/Core/Interrupts.cpp
+++ b/Sming/Arch/Esp8266/Core/Interrupts.cpp
@@ -15,6 +15,7 @@
 
 static InterruptCallback gpioInterruptsList[16] = {0};
 static InterruptDelegate delegateFunctionList[16];
+static InterruptDelegateStdFunction stdFunctionList[16];
 static bool gpioInterruptsInitialied = false;
 
 /** @brief  Interrupt handler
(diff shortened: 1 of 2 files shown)


mikee47:
Further to https://github.com/SmingHub/Sming/issues/1337#issuecomment-374584133, I'm wonder if we might just replace the delegate support here with std:;function? Perhaps there's a way to modify delegates as well so they can be used in place of std:;function, something like an implicit cast?

slaff:
> I'm wonder if we might just replace the delegate support here with std:;function? 

That is exactly what the original plan was.  So your PR #1734 fits perfectly.

hreintke:
Beware that std::function itself is not in IRAM -> not interrupt safe.
We have a lot going on the latest week in on Arduino/ESP8266 you can check those discussions.

Also it is not possible to get lamda's in iram -> needs documenting

verybadsoldier:
As far as I know the `std::function` does not become the actual interrupt function but just gets stored internally. The actually interrupt function (which is in IRAM) then just uses `System.queueCallback` to call the stored `std::function`.

mikee47:
> Also it is not possible to get lamda's in iram -> needs documenting

It appears that lambdas simply get put in the same section as the function they're created in, so if that function is in IRAM then the lambda should be as well. Pity __attrbibute__ doesn't work on them though, be very useful.

slaff:
@verybadsoldier @mikee47 @hreintke I will close this PR in favour of #1734 .

PR #1734, opened 2019-06-29
Revise `Delegate` to use `std::function`

mikee47:
Trying to use both `std::function` and `Delegate` causes issues, and they seem to do the same job anyway.

Further to #1337, #1732 and #1733, this PR is a rewrite of the `Delegate` class based on `std::function`.

__Improvements__

* Full compatibility with `std::function`, `std::bind` and lambdas
* `std::bind` is not required for standard callbacks, so existing code does not require modifying
* When invoking class methods, Deprecate uses lambda functions instead of `std::bind`
* No heap usage

__Call types__

* `Basic_Delegates` updated with speed testing to show performance of various call methods with a single parameter. Indicative results on an Esp8266:

	Method | CPU cycles per call | current Delegate
	--- | --- | ---
	Callback | 23 | 23
	Delegate (function) | 45 | 46
	Delegate (lambda) | 43 | -
	Delegate (method) | 51 | 51
	Delegate (bind) | 51 | -

* Note that `std::bind` _may_ also use the heap, and there is anecdotal evidence that type inferences can sometimes go wrong. It's generally safer to use lambdas.
* `std::bind` is only required to bind callbacks to functions or class methods where parameter lists differ.

__Notes__

* Reference counting is not supported any more
* Clearing a Delegate must be done using `nullptr`, not `0` or `NULL`
* `stdc++` library has been added for Esp8266 builds as it may be required for placeholders when building with optimisations disabled

__Deprecations__

* `TimerDelegateStdFunction` not required, just use `TimerDelegate` instead
* Rename `commandFunctionDelegate` to `CommandFunctionDelegate`
* Use `HardwareSerial::onDataReceived` instead of `setCallback`

__Additional changes__

* Replace `std::function` types with `Delegate`
* Replace `std::bind` with Delegate, simpler and uses lambdas
* Remove redundant declarations and includes

### DIFF
--- a/Sming/Arch/Esp8266/Components/gdbstub/gdbuart.cpp
+++ b/Sming/Arch/Esp8266/Components/gdbstub/gdbuart.cpp
@@ -406,8 +406,6 @@ bool ATTR_GDBINIT gdb_uart_init()
 	auto uart1 = uart_init(UART1, SERIAL_BAUD_RATE, UART_8N1, UART_TX_ONLY, 1, 0, 0);
 	if(uart1 != nullptr) {
 		uart_set_debug(UART1);
-
-		using namespace std::placeholders;
 		m_setPuts(std::bind(&uart_write, uart1, _1, _2));
 	}
 
--- a/Sming/Arch/Esp8266/app.mk
+++ b/Sming/Arch/Esp8266/app.mk
@@ -8,6 +8,7 @@
 LIBS += \
 	microc \
 	microgcc \
+	stdc++ \
 	hal \
 	phy \
 	pp \
--- a/Sming/Core/AtClient.cpp
+++ b/Sming/Core/AtClient.cpp
@@ -134,7 +134,7 @@ void AtClient::sendDirect(AtCommand command)
 	stream->print(command.text);
 	debugf("Sent: timeout: %d, current %d ms, name: %s", currentCommand.timeout, millis(),
 		   command.text.substring(0, 20).c_str());
-	commandTimer.initializeMs(currentCommand.timeout, std::bind(&AtClient::ticker, this)).startOnce();
+	commandTimer.initializeMs(currentCommand.timeout, TimerDelegate(&AtClient::ticker, this)).startOnce();
 }
 
 // Low Level Queue Functions
--- a/Sming/Core/AtClient.cpp
+++ b/Sming/Core/AtClient.cpp
@@ -20,7 +20,7 @@
 
 AtClient::AtClient(HardwareSerial* stream) : stream(stream)
 {
-	this->stream->setCallback(StreamDataReceivedDelegate(&AtClient::processor, this));
+	this->stream->onDataReceived(StreamDataReceivedDelegate(&AtClient::processor, this));
 }
 
 void AtClient::processor(Stream& source, char arrivedChar, uint16_t availableCharsCount)
--- a/Sming/Core/AtClient.h
+++ b/Sming/Core/AtClient.h
@@ -17,7 +17,6 @@
 
 #include "HardwareSerial.h"
 #include "FILO.h"
-#include "Delegate.h"
 #include "Timer.h"
 
 #define AT_REPLY_OK "OK"
--- a/Sming/Core/AtClient.h
+++ b/Sming/Core/AtClient.h
@@ -34,13 +34,13 @@ typedef Delegate<bool(AtClient& atClient, String& reply)> AtCompleteCallback;
 //     finished successfully processing the command
 
 typedef struct {
-	String text;					   ///< the actual AT command
-	String response2;				   ///< alternative successful response
-	unsigned timeout;				   ///< timeout in milliseconds
-	unsigned retries;				   ///< number of retries before giving up
-	bool breakOnError = true;		   ///< stop executing next command if that one has failed
-	AtReceiveCallback onReceive = 0;   ///< if set you can process manually all incoming data in a callback
-	AtCompleteCallback onComplete = 0; ///< if set then you can process the complete response manually
+	String text;				   ///< the actual AT command
+	String response2;			   ///< alternative successful response
+	unsigned timeout;			   ///< timeout in milliseconds
+	unsigned retries;			   ///< number of retries before giving up
+	bool breakOnError = true;	  ///< stop executing next command if that one has failed
+	AtReceiveCallback onReceive;   ///< if set you can process manually all incoming data in a callback
+	AtCompleteCallback onComplete; ///< if set then you can process the complete response manually
 } AtCommand;
 
 typedef enum { eAtOK = 0, eAtRunning, eAtError } AtState;
--- a/Sming/Core/Data/Stream/MultipartStream.h
+++ b/Sming/Core/Data/Stream/MultipartStream.h
@@ -13,7 +13,6 @@
 #pragma once
 
 #include "MultiStream.h"
-#include "Delegate.h"
 #include "Network/Http/HttpHeaders.h"
 
 /**
--- a/Sming/Core/Data/StreamTransformer.h
+++ b/Sming/Core/Data/StreamTransformer.h
@@ -25,8 +25,7 @@
  * @brief Callback specification for the stream transformers
  * @note See StreamTransformer::transform() method for details
  */
-typedef std::function<size_t(const uint8_t* in, size_t inLength, uint8_t* out, size_t outLength)>
-	StreamTransformerCallback;
+typedef Delegate<size_t(const uint8_t* in, size_t inLength, uint8_t* out, size_t outLength)> StreamTransformerCallback;
 
 class StreamTransformer : public IDataSourceStream
 {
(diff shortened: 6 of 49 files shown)


verybadsoldier:
I cannot test it against my application code as I would have to port to new API first but I am sure I am fine with just `std::function`. And I basically think it is a good step forward.

mikee47:
What Sming version are you using?

verybadsoldier:
I am on 3.8.0. When I tried building against this PR branch I got warnings regarding porting to API 4.0.0.

mikee47:
It would be helpful to see the warnings, perhaps work through them here and we can use it to improve the migration guide.

verybadsoldier:
Output is like this:
```
+ Libraries/RGBWWLed/RGBWWLedColor.cpp
In file included from Libraries/RGBWWLed/RGBWWTypes.h:3:0,
                 from Libraries/RGBWWLed/RGBWWLed.h:40,
                 from Libraries/RGBWWLed/RGBWWLedColor.cpp:8:
Libraries/RGBWWLed/../../SmingCore/SmingCore.h:2:0: error: ignoring #pragma warning  [-Werror=unknown-pragmas]
 #pragma warning "Please update to #include <SmingCore.h> - see https://github.com/SmingHub/Sming/wiki/Migrating-from-Sming-v3.8.x-to-v4"
 ^
In file included from Libraries/RGBWWLed/RGBWWconst.h:11:0,
                 from Libraries/RGBWWLed/RGBWWTypes.h:4,
                 from Libraries/RGBWWLed/RGBWWLed.h:40,
                 from Libraries/RGBWWLed/RGBWWLedColor.cpp:8:
Libraries/RGBWWLed/../../SmingCore/SmingCore.h:2:0: error: ignoring #pragma warning  [-Werror=unknown-pragmas]
 #pragma warning "Please update to #include <SmingCore.h> - see https://github.com/SmingHub/Sming/wiki/Migrating-from-Sming-v3.8.x-to-v4"
 ^
In file included from Libraries/RGBWWLed/RGBWWconst.h:16:0,
                 from Libraries/RGBWWLed/RGBWWTypes.h:4,
                 from Libraries/RGBWWLed/RGBWWLed.h:40,
                 from Libraries/RGBWWLed/RGBWWLedColor.cpp:8:
Libraries/RGBWWLed/../../SmingCore/SmingCore.h:2:0: error: ignoring #pragma warning  [-Werror=unknown-pragmas]
 #pragma warning "Please update to #include <SmingCore.h> - see https://github.com/SmingHub/Sming/wiki/Migrating-from-Sming-v3.8.x-to-v4"
 ^
In file included from Libraries/RGBWWLed/RGBWWLedColor.h:10:0,
                 from Libraries/RGBWWLed/RGBWWLed.h:43,
                 from Libraries/RGBWWLed/RGBWWLedColor.cpp:8:
Libraries/RGBWWLed/../../SmingCore/SmingCore.h:2:0: error: ignoring #pragma warning  [-Werror=unknown-pragmas]
 #pragma warning "Please update to #include <SmingCore.h> - see https://github.com/SmingHub/Sming/wiki/Migrating-from-Sming-v3.8.x-to-v4"
 ^
```
Looks clear to me but I want to first get rid of two hacks I did for my 3.8.0 version (includes this std::function topic) before going forward further.

mikee47:
@verybadsoldier I've ported your `esp_rgbww_firmware` onto latest Sming develop (plus some upcoming changes). Haven't had a change to actually run it yet but it's a start!

mikee47:
Possible thorn for this PR - https://github.com/verybadsoldier/esp_rgbww_firmware/pull/1#issuecomment-507046790

slaff:
@mikee47 how is this PR looking? Something to add here from your side?

mikee47:
Re. issue https://github.com/verybadsoldier/esp_rgbww_firmware/pull/1#issuecomment-507046790, the 'undefined placeholder` problem. This is the sort of error the linker throws up:

```
r:/sandboxes/esp_rgbww_firmware/out/Esp8266/debug/lib/clib-App-d368dbc221bcd658e6ecbdb532e2b09e.a(webserver.o): In function `ApplicationWebserver::onIndex(HttpRequest&, HttpResponse&)':
r:\sming-v4\sming\core\data/arduinojson.h:86: undefined reference to `std::placeholders::_2'
r:\sming-v4\sming\core\data/arduinojson.h:86: undefined reference to `std::placeholders::_1'
```

This is only evident on the Windows Esp8266 compiler, linux is OK.

The problem manifests when compiling with `-Og` or `-O0` so suspect there's a piece of RTL code missing from somewhere. And yes, if I add `stdc++` to LIBS it links fine.

mikee47:
Second point. The existing `Delegate` is reference-counted. Or, more precisely, the 'caller' object which the `Delegate` wrapper class contains gets copied onto the heap and ends up being destroyed automatically when all the Delegates referencing it are destroyed.

The advantage of reference counting is we can treat the Delegate almost like a function pointer without doing anything special, like this (from `System.cpp`):

```
void SystemClass::onReady(SystemReadyDelegate readyHandler)
{
	if(readyHandler) {
		auto handler = new SystemReadyDelegate(readyHandler);
		queueCallback(
			[](uint32_t param) {
				SystemClass::state = eSS_Ready;
				auto handler = reinterpret_cast<SystemReadyDelegate*>(param);
				(*handler)();
				delete handler;
			},
			reinterpret_cast<uint32_t>(handler));
	}
}
```

It does make the overhead of such calls clearer. Note that when I wrote this I should have used
 `readyHandler->increase()` and `readyHandler->decrease()`, but because I didn't the code works exactly the same using `std::function`.

It also seems that `bind` can do heap allocation - https://stackoverflow.com/questions/46163607/avoid-memory-allocation-with-stdfunction-and-member-function. Because this all happens at run-time, and is probably compiler dependent, it'll need some investigation to find out what the downsides are going to be.

mikee47:
@slaff I kind of threw this PR together as a 'what if', so is a bit more drastic than the approach in https://github.com/SmingHub/Sming/issues/1337#issuecomment-374951893. I'm OK with it though!

Porting `esp_rgbww_firmware` was a useful exercise to establish how much work migration would require. As far as delegates were concerned it's fairly straightforward. (ArduinoJson definitely took more work, maybe we could add V5 back in as an option to ease things a bit?)

This PR eliminates Delegate completely, and removes `TimerDelegateStdFunction` instead of deprecating it. I've not renamed anything, so still using the term `Delegate`, but it actually works better for me than, say, `StdFunction` as it ably describes _what_ it does, rather than _how_.

One thing I considered was retaining `Delegate` as an alias for `std::function`, perhaps retaining `Delegate.h` for that which would also include <functional> and <placeholders>. If we're going to keep the term 'delegate' then that might work better. I could also revert a bunch of changes which replaced `Delegate` with `std:;function`. What do you think?

I'll need to edit the top comment to summarise impact before merging.

mikee47:
I'm also wondering if there's a way we can avoid the need to explicitly use `std::bind`, so we could keep doing this: 

`ntpcp = new NtpClient("pool.ntp.org", 30, NtpTimeResultDelegate(&ntpClientDemo::ntpResult, this));`

instead of having to do this:
`ntpcp = new NtpClient("pool.ntp.org", 30, std::bind(&ntpClientDemo::ntpResult, this, _1, _2));`

slaff:
> One thing I considered was retaining Delegate as an alias for std::function, perhaps retaining Delegate.h for that which would also include and . If we're going to keep the term 'delegate' then that might work better. I could also revert a bunch of changes which replaced Delegate with std:;function. What do you think?

Yes, why not.

slaff:
@mikee47  Can you help me understand better the table below?  
- current Delegate ?= the old Delegate mechanism ?
- callback ?= calling anonymous functions/lambdas ?

	Method | time (ns) per call | current Delegate
	--- | --- | ---
	Callback | 275 | -
	Delegate (function) | 663 | 538
	Delegate (method) | 738 | 600
	Delegate (bind) | 738 | -


If I have understood the results it seems that the old Delegate mechanism is faster and the only real disadvantage is that it does not support anonymous functions. Would it be better to add support for lambdas in the old Delegates instead of replacing it completely with the new Delegate suggested in this PR? I am looking for pro and contra arguments :)

mikee47:
I've rewritten the testing code to show figures in CPU cycles, and add a lambda test case. I also had a suspicion the optimiser was messing things up, that should now be sorted.

Method | CPU cycles per call | current Delegate
--- | --- | ---
Callback | 23 | 23
Delegate (function) | 45 | 46
Delegate (lambda) | 43 | -
Delegate (method) | 51 | 51
Delegate (bind) | 51 | -

To test on a different build just need to comment-out the lambda and std::bind tests.

Hopefully this is now a no-brainer!

I'll also just point out that the old Delegate puts the callback function wrapper on the heap, so for every registered callback there's a RAM cost (and another possible source of heap fragmentation). Using lambdas means that the code is created in the calling context (usually flash, but if the caller is in IRAM then the lambda will be too).

In addition to captured lambdas, the old Delegate doesn't support std::bind either. Because the new Delegate is based on std::function we can guarantee compatibility with all the standard uses.

I think the key point here is that if speed is important, use a function callback, but for flexibility use a Delegate.

slaff:
> I think the key point here is that if speed is important, use a function callback, but for flexibility use a Delegate.

Ok, sounds good to me. I will test it tonight and schedule it for merging till the end of the week. Unless you suggest something else :)

mikee47:
Task queue should support Delegates, but that's a separate PR.

slaff:
@mikee47 Last words before merging this PR ? :)

mikee47:
Nope, had enough of this one. Next!

PR #1736, opened 2019-07-01
Improve Host `make run` behaviour

mikee47:
With 0126e4e29a073593559fb579f045617f7bbb9fc8 a better default for`SMING_TARGET_OPTIONS` is just `--flashfile=$(FLASH_BIN) --flashsize=$(SPI_SIZE)`.

Removing `--uart=0 --uart=1` defaults to console output for serial, but no input possible.

Removing `--pause=5` runs the code without delay. In practice `--pause` (wait for keypress) is more useful. The 5s delay was a compromise so that integration tests could run unattended, no longer required.

Where serial I/O is required for a project, manually running telnet, etc. on each run is annoying. This PR now automatically launches the appropriate telnet sessions in new terminal windows and passes the correct options to the emulator. The project defines which ports are required by setting `ENABLE_HOST_UARTID` in their `component.mk` file.

Tested in Windows and Ubuntu.

`SMING_TARGET_OPTIONS` is no longer published, instead use other variables which are managed by their relevant Components (flash, uart, wifi, etc.)

Given the complexity `make flash` no longer runs the emulator, however this can be done using `make flash run`.


mikee47:
I'm wondering if we need the `telnet` target at all now?

slaff:
> I'm wondering if we need the telnet target at all now?

I guess in the cases where a developer needs both to use/see uart 0 and uart 1. As for example in the Basic_Serial sample.

slaff:
I have tried this PR and as far as I can see now there is a difference between `make run` and `make flash` related to serial output.  Is that intentional?

If I run `make run` it starts the application and opens two new terminal windows. But if I run `make flash` it automatically starts sending uart0 to the same terminal and there is no uart1 output(or server listening for telnet connection).

mikee47:
That is a good point. I figured that `make flash` should also run the emulator, but if we follow that logic then `flashapp` should do the same. However, I have no idea what a `flashapp` target might actually do for Host, so I left it out and instead added `run`.

I think it might be less confusing if `make flash` ONLY updates the virtual flash image. If we want to run it as well just do `make flash run`.

Also, the `telnet` target isn't very helpful. It's just as easy (and more flexible) to type `telnet 10000`, etc.

mikee47:
OK, I've updated `make gdb` so it starts up uart servers as for `make run` but then pauses so user can startup telnet, etc. if required.

mikee47:
Haven't considered all this in terms of debugging via eclipse. Could maybe create a command file dynamically... Any thoughts?

Sming/Arch/Host/Components/esp_wifi/component.mk
```diff
@@ -1,3 +1,8 @@
+# Options to add for configuring host network behaviour
+CACHE_VARS				+= HOST_NETWORK_OPTIONS
+HOST_NETWORK_OPTIONS	?=
+SMING_TARGET_OPTIONS	+= $(HOST_NETWORK_OPTIONS)
```
slaff: Maybe you can rename `SMING_TARGET_OPTIONS` to `CLI_TARGET_OPTIONS`. Let's use the prefix `SMING_` for important things like SMING_HOME and SMING_ARCH.

slaff:

mikee47:
Yes, all done.

PR #1738, opened 2019-07-02
CI: Updated coding style check.

slaff:


mikee47:
@slaff Thanks. Doesn't pick up the `HardwarePWM.cpp` from previous commit; is there any reason we can't just check every file? i.e. get rid of commit filter?

slaff:
>  Doesn't pick up the HardwarePWM.cpp

Yes, there are two issues that I have seen: 
- if you force-commit then travis does not pass correct sha1 values and the script ends up in flames with
```
fatal: ambiguous argument '4fbd54bb1f2e...698c1dc1a076': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

> i.e. get rid of commit filter?

It would be probably better to get rid of the script and use `make cs` and check the diff. If there is a diff content then we print it and exit with 1. This way we need to keep the folders that have to be checked only in one place.

mikee47:
Nice solution!

PR #1743, opened 2019-07-04
Add malloc_count Component to track heap usage

mikee47:
Currently, `system_get_free_heap_size()` doesn't return anything meaningful. This PR fixes that using a Component based on `malloc_count` https://panthema.net/2013/malloc_count/.

I've also added a couple of functions to enable logging. To use it, add `#include <malloc_count.h>` and call `MallocCount::enableLogging()` and `MallocCount::setLogThreshold()` as required.

The Component also works for the Esp8266 (standard or custom heaps). Enable by adding `COMPONENT_DEPENDS += malloc_count` to the project's `component.mk` file.

The Component is enabled by default for the Host emulator, but can be disabled using `ENABLE_MALLOC_COUNT=0`.

### DIFF
--- a/Sming/Arch/Host/Components/heap/component.mk
+++ b/Sming/Arch/Host/Components/heap/component.mk
@@ -1,0 +1,9 @@
+# We require malloc_count to keep track of heap usage for system_get_free_heap_size()
+# If using valgrind, for example, disabling hooking provides a cleaner trace
+COMPONENT_VARS			:= ENABLE_MALLOC_COUNT
+ENABLE_MALLOC_COUNT		?= 1
+
+ifeq ($(ENABLE_MALLOC_COUNT),1)
+GLOBAL_CFLAGS			+= -DENABLE_MALLOC_COUNT
+COMPONENT_DEPENDS		:= malloc_count
+endif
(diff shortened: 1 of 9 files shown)


mikee47:
It would probably be useful to add an option to disable this if using valgrind, `ENABLE_MALLOC_COUNT`, 
but I think should be enabled by default.

As the API is only available for Host builds it needs to be conditionally compiled (using `#ifdef ARCH_HOST`). An alternative is to make this a generic Component so it can be used on any architecture, with a common API. However, the ESP32 (and Esp8266 RTOS) use Espressif's 'capability heap' and includes tracing support. As a separate issue then, perhaps we should take a look at that and see if it offers any improvements over the existing heap allocation options?

slaff:
> It would probably be useful to add an option to disable this if using valgrind, ENABLE_MALLOC_COUNT,

Yes, please.

slaff:
The Windows build is failing. That is probably due to the concurrency and the way windows handles files. I saw "Error 5" in the error output which lead me to this post: https://stackoverflow.com/questions/14201870/cmake-fails-to-compile-simple-test-error-5-access-denied.

mikee47:
I've logged in via RDP to find out what's going on here. There's some mingw startup code which calls `free` on memory which wasn't allocated via malloc_count, which leads to a sentinel check failure. The code then goes on to call __real_free() on an invalid location (ptr - 16) hence access denied. I'll do some more testing and fix this.

slaff:

mikee47:
@slaff Yep, all done thanks.

PR #1745, opened 2019-07-06
Fix custom heap

mikee47:
Should call umm_malloc, etc. not malloc().
Update to latest version of umm_malloc, and add support for poison checking.


slaff:
@mikee47 Did you try this PR on a real device?

mikee47:
Yes, works on NodeMCU. I've never got the custom heap to work, until I found the reason today. It's odd, though, it must have worked at some point but can't find when that was; candidate for LTS I guess?

PR #1752, opened 2019-07-07
Add back support for ArduinoJson version 5

mikee47:
Whilst ArduinoJson version 6 is recommended for new projects, existing projects can continue to use version 5, or a combination if absolutely necessary.

ArduinoJson has been removed as a dependency from Sming, so projects must now specify which (if any) is required:

* Set required version in project's `component.mk` file, e.g. `ARDUINO_LIBRARIES := ArduinoJson5`
* If required, add `#include <JsonObjectStream.h>` to code, otherwise add `#include <ArduinoJson.h>`.

Using both versions in a project is supported, but will increase the code size so is not recommended:

* Instead of using `ARDUINO_LIBRARIES`, set `COMPONENT_DEPENDS := ArduinoJson5 ArduinoJson6`. Because ArduinoJson5 is listed first, that will be the default. To use version 6 use `#include <JsonObjectStream6.h>`, or `#include <ArduinoJson6.h>` as appropriate.

`Yeelight` has been moved out of Services and into a separate Component. Note that Components must always explicitly specify which version to use since they have no knowledge of which is default.

If you need to determine which version is in use, check `ARDUINOJSON_VERSION_MAJOR`, e.g:

	#if ARDUINOJSON_VERSION_MAJOR == 5
		...
	#endif

Additional changes:

* Add `Json::getValueChanged()` helper function to version 6
* Update both libraries to latest versions
* Update `HostTests` to include LONGLONG checks for both versions
* Port `Basic_WebSkeletonApp_LTS` but retain ArduinoJson5, add note explaining the steps taken to convert
* Fix samples and tests as required
* Remove deprecated method `HttpResponse::sendJsonObject`


slaff:
I am very happy that the current Sming version can support this. But supporting two major versions of the same component is not a good idea. It would be better to fix the integration with ArduinoJson 6, and, as suggested in this PR, make ArduinoJson an optional component and not a mandatory one for the framework.

mikee47:
>
> ...supporting two major versions of the same component is not a good idea.

In general, I would agree, and for new projects there is little argument that V6 should be the standard.

However, there is nothing wrong with ArduinoJson V5 - it's reliable and works very well. For larger projects, it's not a trivial exercise moving to V6, in fact the user may deliberately choose to remain with V5 and hence be discouraged from continuing with Sming.

The [esp_rgbww_firmware](https://github.com/verybadsoldier/esp_rgbww_firmware.git) contains a lot of Json code, and moving that to V6 took a fair bit of work. If I hadn't needed to do that, it would have been a much simpler exercise and could have had the project up and running much quicker. For most users I suspect they'd have just not bothered.

So, in order to make it as easy as possible to get existing projects started with Sming 3.9+, I believe support for V5 is essential. Some users may even prefer to use it for very good reasons, but that's there decision and one of the things I really like about Sming / Arduino is the flexibility to pick and choose the bits we need to get a job done.

samples/Basic_WebSkeletonApp_LTS/app/application.cpp
```diff
@@ -0,0 +1,59 @@
...
+	// Attach Wifi events handlers
+	WifiEvents.onStationDisconnect(STADisconnect);
+	WifiEvents.onStationGotIP(STAGotIP);
+
+	startWebServer();
+
+	counterTimer.initializeMs(1000, counter_loop).start();
+}
+
+void counter_loop()
```
slaff: Change this to `counterLoop` instead of `counter_loop`

samples/Basic_WebSkeletonApp_LTS/app/webserver.cpp
```diff
@@ -0,0 +1,106 @@
...
+			}
+		}
+		saveConfig(ActiveConfig);
+	} else {
+		response.setCache(86400, true); // It's important to use cache for better performance.
+		response.sendFile("config.html");
+	}
+}
+
+void onConfiguration_json(HttpRequest& request, HttpResponse& response)
```
slaff: Change this to `onConfigurationJson`.

samples/Basic_WebSkeletonApp_LTS/app/webserver.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	if(file[0] == '.')
+		response.code = HTTP_STATUS_FORBIDDEN;
+	else {
+		response.setCache(86400, true); // It's important to use cache for better performance.
+		response.sendFile(file);
+	}
+}
+
+void onAJAXGetState(HttpRequest& request, HttpResponse& response)
```
slaff: Change this to `onAjaxGetState`.

slaff:
@mikee47 Is the PR ready for merging?

mikee47:
@slaff Yes thanks.

mikee47:
I've just updated the Wiki with a note on this.

PR #1768, opened 2019-07-16
Fix line-ending problems

mikee47:
When developing on Windows, files get committed with CRLF line endings instead of the Unix LF standard. This causes problems tracking changes. Instead of retrospectively fixing this using dos2unix, this PR simply sets `text=auto` in .gitattributes to ensure line endings are normalised.

* Remove `dos2unix` make target, no longer necessary


.gitattributes
```diff
@@ -1 +0,0 @@
-* -text
```
slaff: Please, restore that file. While we can enforce line endings for our code that is not the case for third-party code. Without that setting we have to inform every windows user (https://github.com/SmingHub/Sming/wiki/Windows-Quickstart#configure-git-for-windows-environment) how to use git and how to set the line endings in order for the patching to work there.

mikee47: Can you suggest an alternative solution to the problem (other than not using Windows!) ?

mikee47: In order to track changes through commits properly, every single commit needs to be consistently checked in with autocrlf behaviour. To fix a bunch of commits I used this:

```
git filter-branch --tree-filter "git add --renormalize ." develop..HEAD
```

But that doesn't work with the current setting in `.gitattributes`, so I had to rebase to get rid of that first before running the filter.

Could we perhaps amend the filter so it specifies which files it should/shouldn't apply to?

slaff: Our code should use Linux line endings. That can be enforced with running dos2unix. 
When a contributor has added new code and his IDE has changed the line endings he should run dos2unix on his local branch, commit  the dos2unix change and only then rebase. 

> To fix a bunch of commits I used this:

Since I am squash-merging the PRs it does not really matter if the dos2unix was the first commit in the PR, the middle one or the last one.

mikee47: >
> While we can enforce line endings for our code that is not the case for third-party code.

Wouldn't third-party code be in a submodule anyway? What else would qualify as 'third-party' code and might require different line endings?

>
> Without that setting we have to inform every windows user (https://github.com/SmingHub/Sming/wiki/Windows-Quickstart#configure-git-for-windows-environment) how to use git and how to set the line endings in order for the patching to work there.

Patches are applied using `--ignore-whitespace --whitespace=nowarn` - is that not sufficient? I'm trying to understand the problem which .gitattributes (#1421) is supposed to fix?

>
> Since I am squash-merging the PRs it does not really matter if the dos2unix was the first commit in the PR, the middle one or the last one.

No, but with complex PR's (like the docs update) if there aren't consistent line endings through every commit it makes tracking changes and rebasing almost impossible.

What I'm trying to pin down is how to stop this problem occurring in the first place, without having to fix it afterwards and complicate the dev. process un-necessarily.

slaff: > but with complex PR's (like the docs update) if there aren't consistent line endings 

Ok,  If that makes life easier let's remove .gitattributes.

slaff:
@mikee47 Do you have anything else to add to this PR before merging?

mikee47:
@slaff If GIT is now sorting out line endings we can probably get rid of the `dos2unix` target ;-)

mikee47:
@slaff All done. I've run every file in the repo. against dos2unix and nothing further crops up.

slaff:
> If GIT is now sorting out line endings we can probably get rid of the dos2unix target ;-)

Different contributor might have configured line-endings that are different than the default one. Thus we should use the .gitattribute file and specify what are the settings for this repository. 

I would suggest to re-add the .gitattribute file but change the content to contain only:
```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto
```

More on this https://help.github.com/en/articles/configuring-git-to-handle-line-endings

@mikee47 Do you think this will work also for you?

mikee47:
@slaff Perfect! Confirmed via https://git-scm.com/docs/gitattributes this will achieve what we need.

slaff:
> Confirmed via ...

Wonderful! Last words before merging?

mikee47:
Amen.

PR #1770, opened 2019-07-19
Add whole-file submodule patching, and move logic into `build.mk`

mikee47:
Where files need to be added or replaced in a submodule, using a .patch file is clumsy. Putting them into a separate sub-directory allows easier change tracking, editing. This also provides a means to provide a component.mk for a submodule without having to first pull in the entire submodule. This was worked around using an additional `components.mk` file, but this is a better solution and more easily documented.

* Remove `all-submodules` target from project.mk, instead use original `submodules` target in main Makefile
* Add COMPONENT_DEPENDS as necessary to ensure dependencies are fully documented
* .patches directories may contain sub-directories containing files which will be copied into target component
* No need for the 'components.mk' fix, instead provide regular component.mk files
* Whole-file patches applied during component dependency parsing, before submodule is actually pulled in. Reapplied afterwards.


slaff:
Looks good to me. Anything to add before merging?

mikee47:
Good to go thanks.

PR #1795, opened 2019-08-03
Refactoring of the rBoot Http Client and output stream.

slaff:
- Renamed rBootItemOutputStream to rBootOutputStream
- Decoupled Item and Output stream
- Added optional maximum allowed length to the output stream
- Simplified the HttpServer_FirmwareUpload example
- Moved the rBootOutputStream to Data/Stream
- RbootOutputStream can be extended to allow Signed OTA. See:
https://github.com/SmingHub/Sming/blob/develop/docs/source/experimental/signed-ota.rst


samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include <Data/Stream/LimitedMemoryStream.h>
 #include <MultipartParser/HttpMultipartResource.h>
-#include <Network/rBootHttpUpdate.h>
+#include <Data/Stream/RbootOutputStream.h>
 
 HttpServer server;
 String lastModified;
```
mikee47: `response.sendTemplate` -> `sendNamedStream`

Issue #1570, opened 2019-01-06
Recommended SDK? esp-open-sdk with SDK 2.0.0?


verybadsoldier:
I have a basic question:
Which is the recommended SDK you would propose for production use?

In current `esp-open-sdk` they default to SDK `2.1.0` but I get compile errors with that:
```
vbs@ubuntu:~/Projects/esp_rgbww_firmware/Sming/Sming$ make
make[1]: Entering directory '/home/vbs/Projects/esp_rgbww_firmware/Sming/Sming'
make[2]: Entering directory '/home/vbs/Projects/esp_rgbww_firmware/Sming/Sming/third-party/esp-open-lwip'
/home/vbs/Projects/esp-open-sdk/xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc -DLWIP_OPEN_SRC -DPBUF_RSV_FOR_WLAN -DEBUF_LWIP -DICACHE_FLASH -Os -Iinclude -I/home/vbs/Projects/esp-open-sdk/sdk/include -Wl,-EL -mlongcalls -mtext-section-literals -I../../system/include -I../../Wiring   -c -o lwip/core/def.o lwip/core/def.c
/home/vbs/Projects/esp-open-sdk/xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc -DLWIP_OPEN_SRC -DPBUF_RSV_FOR_WLAN -DEBUF_LWIP -DICACHE_FLASH -Os -Iinclude -I/home/vbs/Projects/esp-open-sdk/sdk/include -Wl,-EL -mlongcalls -mtext-section-literals -I../../system/include -I../../Wiring   -c -o lwip/core/dhcp.o lwip/core/dhcp.c
In file included from include/lwip/stats.h:37:0,
                 from lwip/core/dhcp.c:73:
lwip/core/dhcp.c: In function 'dhcp_start':
lwip/core/dhcp.c:676:52: error: too many arguments to function 'pvPortMalloc'
     dhcp = (struct dhcp *)mem_malloc(sizeof(struct dhcp));
                                                    ^
include/lwip/mem.h:58:36: note: in definition of macro 'mem_malloc'
 #define mem_malloc(s) pvPortMalloc(s, "", __LINE__, false)
```

It compiles fine when setting SDK to `2.0.0`. So I conclude `2.1.0` is not supported.

As support for `3.0` is experimental currently I guess the recommended way is to use current `esp-open-sdk` and set SDK version back to `2.0.0`?

[...]

PR #1796, opened 2019-08-03
Misc updates

mikee47:
* In host flash routines, don't assert if range check fails, just print a warning and fail the call
* Normalise COMPONENT_SEARCH_DIRS so path can contain '..', etc.
* StreamTransformer - nullptr check and implement `getName()`
* Update _BV macro so it can be used with enumerated values
* Update NtpClient so nullptr (or invalid String) as reqServer parameter to use default - Fixes problem with section mismatch passing flash String value as default
* Update Espressif SDK and ArduinoJson6 to latest versions
* HttpResponse: add null pointer check and set content type for compressed files

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -62,6 +62,7 @@ if [ "$TRAVIS_BUILD_STAGE_NAME" == "Test" ]; then
 	# Build the documentation
 	mv $SMING_PROJECTS_DIR/samples ..
 	mv $SMING_PROJECTS_DIR/tests ..
+	unset SMING_PROJECTS_DIR
 	make docs V=1
 else
 	$MAKE_PARALLEL samples
--- a/Sming/Arch/Host/Components/spi_flash/flashmem.cpp
+++ b/Sming/Arch/Host/Components/spi_flash/flashmem.cpp
@@ -28,7 +28,12 @@ static const char* flashFileName = "flash.bin";
 #define SPI_FLASH_SEC_SIZE 4096
 
 #define CHECK_ALIGNMENT(_x) assert(((uint32_t)(_x)&0x00000003) == 0)
-#define CHECK_RANGE(_addr, _size) assert((_addr) + (_size) <= flashFileSize);
+
+#define CHECK_RANGE(_addr, _size)                                                                                      \
+	if((_addr) + (_size) > flashFileSize) {                                                                            \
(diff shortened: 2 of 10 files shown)


slaff:
> Update Espressif SDK and ArduinoJson6 to latest versions

Did you test the latest SDK on a real device?

mikee47:
Yes, haven't had any issues with updated SDK. However, testing on the nodeMCU (4MB memory) with smaller values of `SPI_SIZE` has issues. There's the fix to startup code, but I the values in esptool component.mk look very suspect.

mikee47:
See also https://github.com/SmingHub/Sming/issues/1570#issuecomment-466719355

mikee47:
Have added support for 8/16MB flash, commands run OK but obviously won't boot and don't have the hardware to test it.

Using Basic_Blink, all sizes work fine, except for SDK 3 which fails on 256K or 512K. This is what I get for 512K:

```
tail 8
chksum 0xb1
load 0x3ffe8000, len 600, room 0
tail 8
chksum 0xe6
csum 0xe6

rBoot v1.4.2 - richardaburton@gmail.com
Flash Size:   4 Mbit
Flash Mode:   DIO
Flash Speed:  40 MHz
rBoot Option: Big flash

Booting rom 0 at 2000, load addr 32530.
emapg
map 0 err
system param partition error
ota1 not set
ota2 not set
system_partition_table_regist: failed
size_map = 0
partition[0]: 1, 0x00000000, 0x00001000
partition[1]: 100, 0x00001000, 0x0000100
partition[2]: 101, 0x00002000, 0x0007900
partition[3]: 4, 0x0007b000, 0x00001000
partition[4]: 5, 0x0007c000, 0x00001000
partition[5]: 6, 0x0007d000, 0x00003000
```

mikee47:
The examples provided with SDK3 all list maps 0 & 1 (256/512K) as unsupported...

slaff:
> The examples provided with SDK3 all list maps 0 & 1 (256/512K) as unsupported...

Ok.. then we should add a note in this PR and in the documentation that devices with flash memory size 256 and 512K are not supported by SDK v3.0.0 or greater. And for those cases SDK 1.5.4 or SDK 2.0.0 is recommended.

mikee47:
I've moved out partition table changes to separate PR.

slaff:
I am planning to merge this PR. Is it ready or you would like to add something more?

mikee47:
All done thanks.

Issue #1781, opened 2019-07-26
Proposal: (Signed) OTA firmware updates using HTTP POST file upload


aemseemann:
This proposal is intended to evaluate interest in an alternative OTA update mechanism for Sming and discuss potential realization aspects.

As far as I'm aware, Sming currently only provides the OTA update method demonstrated in samples/Basic_rBoot.
In this approach, the ESP downloads the firmware image from a server provided by the developer. The update is initiated using a secondary channel, e. g. by a command received from the serial console. While the second channel can be considered an additional security measure, it is sometimes inconvenient, especially if you do not have access to the serial console, e. g. for mains-powered devices like Sonoff switches.
 
As an alternative, I propose an update mechanism where the new firmware is pushed to the device via an HTTP POST file upload request. This is the type of request generated by a web form containing an `<input type="file"...>` control and thus would allow to easily integrate a convenient firmware upgrade page to an existing web interface. While this is probably the preferred method employed by end users, the same interface can be used during development by emulating the upload request from a Makefile target (e. g. using `curl -F ...`) to enable a completely wireless development workflow.

I implemented an initial draft [here](https://github.com/aemseemann/Sming/tree/feature/push-ota-signed)  (branch feature/push-ota-signed => [samples/Signed_OTA](https://github.com/aemseemann/Sming/tree/feature/push-ota-signed/samples/Signed_OTA)) and would really like to known your opinion in general and if it is worth pursuing an integration into Sming somehow.

**Some technical details:**

[...]

PR #1797, opened 2019-08-03
Integrate libsodium v1.0.18 as a component.

aemseemann:
This PR adds the [sodium](https://github.com/jedisct1/libsodium) library of cryptographic functions to Sming as discussed in https://github.com/SmingHub/Sming/issues/1781

See Sming/Components/libsodium/Readme.md for further information.

Aside from the library, I fixed the partition table generation in startup.cpp for devices with flash sizes <= 1MB (should fix the issue reported in https://github.com/SmingHub/Sming/issues/1570#issuecomment-466719355) and added a definition of `__errno`, which does not exist in libc for some reasons, but is referenced by some functions in libsodium (via `errno`).

The


.gitmodules
```diff
@@ -118,3 +118,8 @@
 	path = Sming/Libraries/MCP_CAN_lib
 	url = https://github.com/coryjfowler/MCP_CAN_lib.git
 	ignore = dirty
+[submodule "Sming/Components/libsodium/libsodium"]
```
slaff: Please, place this submodule definition block after esptool2. Also change the fist line to:
`[submodule "libsodium"]`. This way looking at the gitmodules file the developers can easier figure out that this is a submodule that can be used in all architectures and is not an Arduino-style library.

Sming/Components/libsodium/component.mk
```diff
@@ -0,0 +1,103 @@
+# umm_malloc (custom heap allocation)
```
mikee47: comment

.gitmodules
mikee47: Should go at top (e.g. line 24) and rename to "libsodium"

Sming/Components/libsodium/Readme.md
```diff
@@ -0,0 +1,16 @@
+# libsodium
```
mikee47: Please rename file to `README.md` so doc. system picks it up.

.gitmodules
slaff: Another point. The code at the moment does not compile for the host architecture. You can repeat the following steps to reproduce the issue:

```
cd $SMING_HOME/../samples/HttpServer_FirmwareUpload
echo "COMPONENT_DEPENDS := libsodium" >> component.mk
make SMING_ARCH=Host ENABLE_GDB=1 ENABLE_LWIPDEBUG=1 ENABLE_MALLOC_COUNT=0
```

The compilation errors with:
```
Sming/Components/libsodium/libsodium/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c:30:31: error: implicit declaration of function ‘syscall’ [-Werror=implicit-function-declaration]
 #   define getrandom(B, S, F) syscall(SYS_getrandom, (B), (int) (S), (F))
                               ^
/home/slavey/dev/esp8266.dev.box/dev/Sming/Sming/Components/libsodium/libsodium/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c:265:18: note: in expansion of macro ‘getrandom’
         readnb = getrandom(buf, size, 0);
                  ^
cc1: all warnings being treated as errors
```

It would be great if `libsodium` works also on the Host architecture.

Sming/Components/libsodium/component.mk
```diff
@@ -0,0 +1,103 @@
...
+	$(LIBSODIUM_SRC)/crypto_stream/salsa2012/stream_salsa2012.c \
+	$(LIBSODIUM_SRC)/crypto_stream/salsa208/ref/stream_salsa208_ref.c \
+	$(LIBSODIUM_SRC)/crypto_stream/salsa208/stream_salsa208.c \
+	$(LIBSODIUM_SRC)/crypto_stream/xchacha20/stream_xchacha20.c \
+	$(LIBSODIUM_SRC)/randombytes/sysrandom/randombytes_sysrandom.c
+
+COMPONENT_INCDIRS := $(LIBSODIUM_ROOT)/src/libsodium/include
+
+# results from configure script run for --host=xtensa-lx106-elf (unused defines removed)
+COMPONENT_CFLAGS := -DSODIUM_STATIC -DHAVE_C_VARARRAYS=1 -DNATIVE_LITTLE_ENDIAN=1 -DHAVE_INLINE_ASM=1 -DHAVE_WEAK_SYMBOLS=1 -DHAVE_ALLOCA_H=1 -DHAVE_ALLOCA=1 -DCONFIGURED=1
```
slaff: Adding  -D_GNU_SOURCE=1 to the list of COMPONENT_CFLAGS makes the compilation work also under Host (Linux in my case).

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
```diff
@@ -106,7 +106,7 @@ extern "C" void ICACHE_FLASH_ATTR WEAK_ATTR user_pre_init(void)
 	static const partition_item_t partitions[] = {
 			PARTITION_ITEM(SYSTEM_PARTITION_BOOTLOADER,			0,					1),
 			PARTITION_ITEM(SYSTEM_PARTITION_RBOOT_CONFIG,		1,					1),
-			PARTITION_ITEM(SYSTEM_PARTITION_PROGRAM,			2,					MAX_PROGRAM_SECTORS - 2),
+			PARTITION_ITEM(SYSTEM_PARTITION_PROGRAM,			2,					(rfCalSector < MAX_PROGRAM_SECTORS ? rfCalSector : MAX_PROGRAM_SECTORS) - 2),
```
slaff: Revert this change. PR #1800 will take care of it.

mikee47:
Can we have a small, simple sample to check this build?

slaff:
> Can we have a small, simple sample to check this build?

The HttpServer_FirmwareUpload sample can be extended to support signed ROMS.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
```diff
@@ -134,3 +134,11 @@ namespace std {
         while(1);
     };
 }
+
+extern "C" {
+    __attribute__((weak)) int *__errno(void)
```
mikee47: Don't need this, add `#include <sys/errno.h>` somewhere.
(See `Arch/Esp8266/Components/gdbstub/gdbstub.cpp`.)

Sming/Components/libsodium/component.mk
```diff
@@ -0,0 +1,103 @@
+# umm_malloc (custom heap allocation)
+
+COMPONENT_SUBMODULES	:= libsodium
+
+LIBSODIUM_ROOT := $(COMPONENT_PATH)/libsodium
+LIBSODIUM_SRC := $(LIBSODIUM_ROOT)/src/libsodium
+
+COMPONENT_SRCFILES := \
```
mikee47: If you wanted to avoid listing all files individually, see `Sming/component.mk` for use of ListAllSubDirs.

aemseemann: In this case the file listing is intentional, because I excluded all files with architecture-specific assembly-optimized implementations.

.gitmodules
aemseemann: I put it under esptool2, as slaff proposed.

Sming/Components/libsodium/component.mk
aemseemann: removed leftover comment

Sming/Components/libsodium/Readme.md
aemseemann: done

.gitmodules
aemseemann: It was also my intention to make it available to all supported architectures. 
Adding _GNU_SOURCE fix the issue for me, too. However, It would be great if someone can check this out on a Windows host.
Offering a working solution for all possible realizations of the Host "architecture" can get quite involved, as it is not even restricted to x86-based Hosts.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
aemseemann: Done.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
aemseemann: This header is already included by <errno.h>. I don't see how including it explicitly should fix the linker error. At least in my toolchain built, sys/errno.h contains the following:

```
#define errno (*__errno())
extern int *__errno _PARAMS ((void));
```

I.e. to use errno, a definition of the `__errno` function has to be provided somewhere. Naturally, it should be in libc, but since Sming uses a stripped-down version of libc, it is not there, which is why I added it to startup.cpp (seemed the logical place, since it is a similar situation with the missing `__throw_bad_function_call`).

Please advise.

aemseemann:
I'll rebase on current develop and work on the example now.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
mikee47: Strange, it seems `errno` isn't actually used anywhere else (just the codes).
`libmicroc` has a definition for `__errno_location` which should probably be referenced in a header somewhere, but it isn't.

https://github.com/anakod/esp_microc/blob/master/errno/__errno_location.c

I'll raise another PR for this.

Sming/Components/libsodium/.patches/libsodium.patch
```diff
@@ -0,0 +1,1232 @@
...
++randombytes_sysrandom_stir(void)
++{
++}
++
++static void
++randombytes_sysrandom_buf(void * const buf, const size_t size)
++{
++    uint8_t *dst = buf;
++    for (size_t remaining = size; remaining > 0; --remaining) {
++        *dst++ = (uint8_t)(RANDOM_REG32 & 0xFF);
```
mikee47: I think `os_get_random()` would be a better choice.

Sming/Components/libsodium/.patches/libsodium.patch
```diff
@@ -0,0 +1,1232 @@
...
++    return RANDOM_REG32;
++}
++
++static void
++randombytes_sysrandom_stir(void)
++{
++}
++
++static void
++randombytes_sysrandom_buf(void * const buf, const size_t size)
```
mikee47: I think os_get_random() would be a better choice. It's also available in the emulator (and ESP32).

Sming/Components/libsodium/.patches/libsodium.patch
```diff
@@ -0,0 +1,1232 @@
...
+ #include "randombytes.h"
+@@ -108,7 +112,35 @@ randombytes_sysrandom_close(void)
+     return 0;
+ }
+ 
+-#else /* HAVE_SAFE_ARC4RANDOM */
++#elif defined(ARCH_ESP8266)
++
++static uint32_t
++randombytes_sysrandom(void)
```
mikee47: I think `os_random()` would be better. 

`os_random()` calls `r_rand()`, which calls `phy_get_rand()`. This reads `RANDOM_REG32` but also ensures the clock is enabled.

os_random() is provided in the emulator and ESP32. More background:

https://raw.githubusercontent.com/pfalcon/esp8266-re-wiki-mirror/master/Random_Number_Generator.mw

Note that lwip uses `r_rand()` directly to save an additional call. I wouldn't bother as we shouldn't be calling it in a loop anyway.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
aemseemann: Ok, I'll rebase again when https://github.com/SmingHub/Sming/pull/1801 is merged and then remove the definition in startup.cpp

Sming/Components/libsodium/.patches/libsodium.patch
mikee47: I'll do a PR for these.

aemseemann: Thanks for the information. I wasn't aware of the details and that the HWRNG is already used by the SDK. I replaced the calls as you suggested and also added your link to the documentation.

Sming/Arch/Esp8266/Components/esp8266/startup.cpp
slaff: > Ok, I'll rebase again when #1801 is merged

PR #1801 is merged in develop.

aemseemann:
I integrated signature verification into the HttpServer_FirmwareUpload example. 
It works on my ESP01 board, but I would still consider it work in progress.
The life cycle management for streams created in `fileUploadMapper` is kind of a mess. My debug logs suggest that streams are never destroyed, causing a potentially massive memory leak. Also, stream methods `init()` and `close()` are never invoked. The latter would be invoked from the destructor, but this probably is *after* the request completion callback (i.e. the `onUpload` method in this example) has returned. 

Second issue: Travis CI does not have the necessary python package to create the signing and verification keys. Can this package be installed or do you have any suggestion for a workaround?

mikee47:
Re. travis If you look in `.travis/install.sh` you can add required python packages in there.

aemseemann:
Travis issue fixed, thanks for the hint. Learned a lot today!

mikee47:
I'm having issues using the signtool with python 3, is there an updated version? Also, you could put the tool plus all of the make logic into the libsodium Component. Take a look at `Arch/Esp8266/Components/axtls-8266` which does something similar.

mikee47:
Patch for python 3. Breaks python 2 though :-(

```
diff --git a/samples/HttpServer_FirmwareUpload/signtool.py b/samples/HttpServer_FirmwareUpload/signtool.py
index 0a0562ff8..ff43b3331 100755
--- a/samples/HttpServer_FirmwareUpload/signtool.py
+++ b/samples/HttpServer_FirmwareUpload/signtool.py
@@ -39,5 +39,5 @@ extern "C" {
 
 static const uint8_t firmwareVerificationKey[''' + str(len(pk)) + '''] = {
-    ''' + ', '.join('0x%02X' % ord(b) for b in pk) + '''
+    ''' + ', '.join('0x%02X' % b for b in pk) + '''
 };
 #ifdef __cplusplus
@@ -109,5 +109,5 @@ def main():
         if args.keyfile:                
             with open(args.keyfile, 'wb') as privkeyfile:
-                privkeyfile.write(sk_seed.encode('Hex'))
+                privkeyfile.write(sk_seed)
             print("Private signing key written to '%s'." % args.keyfile)
         else:
@@ -115,6 +115,6 @@ def main():
     elif args.keyfile:
         print("Read signing key from '%s'" % args.keyfile)
-        with open(args.keyfile, 'r') as keyfile:
-            sk_seed = keyfile.read().decode('Hex')
+        with open(args.keyfile, 'rb') as keyfile:
+            sk_seed = keyfile.read()
     
     else:
```

mikee47:
We should be focused on Python 3 anyway - https://python3statement.org/

mikee47:
@aemseemann I've pushed some fixes to your repo https://github.com/aemseemann/Sming/pull/2
Builds for all arches, and runs in host emulator on both Linux and Windows.

aemseemann:
@mikee47: I specifically wrote signtool.py for python2, because the other tools in the build process (esptool.py, make terminal) also require python2 (which, btw, is a PITA for users of distros where "python" defaults to python3). I am very aware of the python2 deprecation notice and agree with you that Sming should move to python3 sooner than later. However, I would avoid a situation where both python2 and python3 is needed in the same build process. Please correct me if I misunderstood something.

> Also, you could put the tool plus all of the make logic into the libsodium Component.
This seems odd to me, because firmware signing is no feature of the libsodium component. I would rather create a new component that contains all the OTA upgrade stuff (classes like `SignedRbootOutputStream`, the signtool, make targets, and so on) and depends on libsodium. But IMO this is also content for a separate PR.

mikee47:
I'll raise an issue concerning Python as we need to agree some guidelines. For info, I'm running with python3 as default (linux 3.6.8, Windows 3.7.3) and both esptool and pyterm work fine.

> This seems odd to me, because firmware signing is no feature of the libsodium component. I would rather create a new component that contains all the OTA upgrade stuff (classes like `SignedRbootOutputStream`, the signtool, make targets, and so on) and depends on libsodium. But IMO this is also content for a separate PR.

Agreed.

aemseemann:
> For info, I'm running with python3 as default

My bad, I simply forgot to install PySerial and did not read the error messages carefully enough. Runs fine for me now, too.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -22,7 +132,7 @@ void onFile(HttpRequest& request, HttpResponse& response)
 	}
 
 	String file = request.uri.getRelativePath();
-
+	Serial.printf("Requested file: %s\n", file.c_str());
```
slaff: Please, remove this line.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -37,44 +147,35 @@ void onFile(HttpRequest& request, HttpResponse& response)
 
 int onUpload(HttpServerConnection& connection, HttpRequest& request, HttpResponse& response)
 {
+	Serial.printf("onUpload\n");
```
slaff: If you want to use debug code don't use `Serial.printf` but `debug_d` instead.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -37,44 +147,35 @@ void onFile(HttpRequest& request, HttpResponse& response)
...
-	uint8_t slot = bootConfig.current_rom;
-	slot = (slot == 0 ? 1 : 0);
-	Serial.printf("Firmware updated, rebooting to rom %d...\r\n", slot);
-	rboot_set_current_rom(slot);
-	System.restart(5); // defer the restart with 5 seconds to give time to the web server to return the response
-
-	response.sendFile("restart.html");
-
+	SignedRbootOutputStream* uploadStream = static_cast<SignedRbootOutputStream*>(file);
+	uploadStream->close();
```
slaff: Don't close the stream here. It will be closed later at the end of the request.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -3,10 +3,120 @@
 #include <Data/Stream/LimitedMemoryStream.h>
 #include <MultipartParser/HttpMultipartResource.h>
 #include <Data/Stream/RbootOutputStream.h>
+#include <sodium/crypto_sign.h>
+#include "FirmwareVerificationKey.h"
 
 HttpServer server;
 String lastModified;
 
+class SignedRbootOutputStream : public RbootOutputStream
```
slaff: Take that class out of  application.cpp. You can move it to a separate file in this sample OR if you think it is good enough to be reused move it to `Components/libsodium` . Similar to [MultipartParser](https://github.com/SmingHub/Sming/tree/develop/Sming/Components/MultipartParser). I would recommend you to do the latter.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -3,10 +3,120 @@
...
+{
+	const uint8_t* const verificationKey_;
```
slaff: Please don't use underscores as prefix or suffix in a variable, function/method or class names.

samples/HttpServer_FirmwareUpload/app/application.cpp
```diff
@@ -3,10 +3,120 @@
...
+			if(missingHeaderBytes_ > 0) {
+				const size_t chunkSize = std::min(size, missingHeaderBytes_);
+				memcpy(headerPtr_, data, chunkSize);
+				headerPtr_ += chunkSize;
+				missingHeaderBytes_ -= chunkSize;
+				data += chunkSize;
+				size -= chunkSize;
+				consumed += chunkSize;
+				if(missingHeaderBytes_ == 0) {
+					Serial.printf("Receive image for load address 0x%08X, slot starts at 0x%08X\n", header_.loadAddress,
```
slaff: Please, use `debug_d` for useful debug messages. For example the line above should become:
```
debug_d(("Receive image for load address 0x%08X, slot starts at 0x%08X\n", header_.loadAddress,startAddress_)
```

samples/HttpServer_FirmwareUpload/signtool.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python2
```
slaff: The README.md must contain brief information how to create signed ROM and how to use it.

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Done (Leftover from debugging).

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: removed (also leftover from debugging)

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: `uploadStream->close()` triggers the signature validation and therefore must be invoked *before* returning from `onUpdate` to decided if the received image is valid and the slot can be activated.
Did I miss something? According to https://github.com/SmingHub/Sming/blob/develop/docs/source/experimental/signed-ota.rst, the `close` method is the intended place for signature verification.

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Separate file it is. 
Moving it to libsodium isn't the right choice, because libsodium offers cryptographic functions, not firmware update functionality.

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Fixed.
(FYI: Unlike leading underscores, a trailing underscore does result in a reserved symbol. This  naming convention is quite common to denote private members (e. g. boost))

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Fixed.
However, since this is an example, I thought it would be beneficial to show some output during the upgrade, regardless of the debug settings.

samples/HttpServer_FirmwareUpload/signtool.py
aemseemann: The README.rst (already present in the folder) now contains the necessary instructions.

samples/HttpServer_FirmwareUpload/app/application.cpp
slaff: > uploadStream->close() triggers the signature validation

Not necessarily . It would be better if there is a `validate()` method that triggers the validation and sets the ok flag. Also the ok() method can be renamed to isValid(). The closing of the file is something separate and does not need to be coupled with the validation.  You can move the logic that deletes the first sector if the validation is not correct to the destructor.

> the close method is the intended place for signature verification.

I will change the documentation later in the week.

.travis/install.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex # exit with nonzero exit code if anything fails
 
+pip install --user PyNaCl
```
slaff: Concerning the python signing tool: 
- Create a file called requirements.txt that contains all required packages. Similar to the documentation requirements. The file should be in the same directory as the python signing tool.
- Replace the line above with
```
pip install  -r <path-to-signing-tool>/requirements.txt
```

samples/HttpServer_FirmwareUpload/app/application.cpp
slaff: > .. Learned a lot today!  ...
> .. but this should be done in a separate PR.

Wonderful! Thanks for the fantastic work so far.

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Moved validation code out of the `close` method. Still don't like it, because the `onUpload` callback should inform the user that *all* of the firmware update process has been completed successfully. Even if the signature matches, writing the flash can still fail, and as the last chunk of flash is only written in `RbootOutputStream::close` that must come before preparing the user response. 
I also don't like than an upcast from `ReadWriteStream` to `SignedRbootOutputStream` is necessary in `onUpload`. It may be an acceptable compromise for this example, but for a potential future component, I would prefer a more type-safe solution.

samples/HttpServer_FirmwareUpload/app/application.cpp
aemseemann: Thanks for your support as well!

.travis/install.sh
aemseemann: Used pip3 instead, same as for the doc packages.
However, Travis takes quite some time to install it. Seems like it is building something from source. 
Maybe it has something to do with pip being outdated: "You are using pip version 8.1.1, however version 19.2.1 is available."

mikee47:
May I suggest putting the requirements into the project's component.mk instead?
Plus couple of tags so custom targets show up for `make help`.

```
diff --git a/samples/HttpServer_FirmwareUpload/component.mk b/samples/HttpServer_FirmwareUpload/component.mk
index 74ac73b17..d78808aba 100644
--- a/samples/HttpServer_FirmwareUpload/component.mk
+++ b/samples/HttpServer_FirmwareUpload/component.mk
@@ -6,4 +6,6 @@ COMPONENT_DEPENDS := libsodium
 ENABLE_HTTP_SERVER_MULTIPART = 1
 
+##@Building
+
 web-pack:
 	$(Q) date +'%a, %d %b %Y %H:%M:%S GMT' -u > web/.lastModified
@@ -12,4 +14,7 @@ web-upload: web-pack spiffs-image-update
 	$(call WriteFlash,$(RBOOT_SPIFFS_0)=$(SPIFF_BIN_OUT))
 
+.PHONY: requirements
+requirements:
+	python -m pip install --user -r $(PROJECT_DIR)/requirements.txt
 
 SIGNTOOL := python $(COMPONENT_PATH)/signtool.py
@@ -17,5 +22,5 @@ SIGNING_KEY := $(COMPONENT_PATH)/signing.key
 VERIFICATION_HEADER := $(COMPONENT_PATH)/app/FirmwareVerificationKey.h
 
-$(SIGNING_KEY):	
+$(SIGNING_KEY): requirements
 	@echo "#########################################################"
 	@echo "# Generating new signing key for firmware update images #"
@@ -28,6 +33,5 @@ $(SIGNING_KEY):
 
 .PHONY: generate_signing_key
-generate_signing_key: 
-	@echo Generate a new firmware update signing key on explicit request.
+generate_signing_key: ##Generate a new firmware update signing key on explicit request.
 	$(Q) $(SIGNTOOL) --genkey --keyfile=$(SIGNING_KEY)
 
@@ -44,5 +48,5 @@ $(SIGNED_ROM0): $(RBOOT_ROM_0_BIN) $(SIGNING_KEY)
 
 .PHONY: signedrom
-signedrom: $(SIGNED_ROM0)
+signedrom: $(SIGNED_ROM0) ##Create signed ROM image
 
 
```

samples/HttpServer_FirmwareUpload/app/application.cpp
slaff: > I also don't like than an upcast from ReadWriteStream to SignedRbootOutputStream is necessary in onUpload. It may be an acceptable compromise for this example, but for a potential future component, I would prefer a more type-safe solution.

Feel free to suggest better solution as PR. This API is not final.

aemseemann: I'll try to come up with something, but it may take a couple of weeks.

What's your opinion on the current state of the example. Is it acceptable for this PR?

slaff: > What's your opinion on the current state of the example. Is it acceptable for this PR?

You are doing great and I am happy to welcome you to our team and thank you for your first accepted contribution :)

PR #1802, opened 2019-08-04
Refactor/abstract Wifi classes

mikee47:
The `StationClass`, `WifiEventsClass` and `AccessPointClass` classes are now virtual base classes.
The actual arch-specific implementations are `StationImpl`, `WifiEventsImpl` and `AccessPointImpl`.

Looking at Basic_SmartConfig sample it needs more work, the callback should be architecture-independent. (That'll be another breaking change.)

Breaking changes:

* `StationClass::getConnectionStatusName()` returns String instead of `const char*`.

Esp8266 'StationImpl':

* Use `addElement(BssInfo*)` instead of `add(BssInfo&)` to avoid additional copy operation
* Simplify config() logic, should be a little easier to follow
* Add `getMacAddr` method to retrieve raw MAC address

Added dependencies:

* ENABLE_WPS=1 requires ENABLE_ESPCONN=1
* ENABLE_SMART_CONFIG has been added as it too has dependencies

Host updates:

* Provides fixed list of AP's
* Neither WPS nor SmartConfig are supported and will generate a warning. (Was an error, but this causes dist-clean to fail.)


slaff:
@mikee47 Can you check the issues reported by Codacy?

mikee47:
@slaff If you're happy with this PR in general, there are some more breaking changes I'd like to implement, such as changing:

change `typedef Delegate<void(String ssid, uint8_t ssidLen, uint8_t bssid[6], uint8_t channel)> StationConnectDelegate;`
to `typedef Delegate<void(const String& ssid, uint8_t bssid[6], uint8_t channel)> StationConnectDelegate;`

change `typedef Delegate<void(uint8_t oldMode, uint8_t newMode)> StationAuthModeChangeDelegate;`
to something with enumerated types for the 'mode' parameter

and so on. The smartconfig stuff also needs rework.

This will break existing applications, but the alternative is to add additional delegates and deprecate the old ones.
Any thoughts?

slaff:
> This will break existing applications ...

Let's go with the breaking changes.  Of course we should describe them in our migration document.

slaff:
@mikee47 Ping me when the PR is ready and tested on your device(s).

mikee47:
@slaff Yes, tested with HttpServer_ConfigNetwork sample, Basic_SmartConfig.

PR #1806, opened 2019-08-07
Changes to the error handling in HttpServer 

slaff:
- The body parsers now return the number of parsed bytes
- This allows the server to trigger error as early as possible
- If there is an error `onMessageComplete` callback will still be called. In order to check if we are in error state `response.isSuccess()` can be used.
- On error `onMessageComplete` can provide custom response or return non-zero value to trigger the default error page.


slaff:
>  there is a potential memory leak if the request is not completed successfully.  ... The deallocation is done in the onMessageComplete callback, which is not invoked in case of parser errors or if the connection is interrupted in between.

@aemseemann That should be taken care of in this PR.

PR #1812, opened 2019-08-09
WiFi improvements

mikee47:
Update WiFi documentation

* Create doxygen `WiFi` group, and add station, ap and events to it.
* Add descriptions for Wifi disconnection reasons

MACAddress class:

* Bugfix: `operator!()` sense inverted
* Enforce array size checks - i.e. cannot instantiate MACAddress with anything other than uint8_t[6]
* Rename `getMacAddr()` -> `getMacAddress()`
* Make printable, so `Serial.print(mac)` works

Change `const MACAddress&` parameters in Delegates to `MACAddress`.

  I investigated the code which is generated for various cases where `MACAddress`, `IPAddress` and `String`
  types are passed by reference (`const MACAddress& mac`) or by value (`MACAddress mac`).

  For both `MACAddress` and `IPAddress` the code is virtually identical, so using the more complex `const&` syntax
  has no benefit. With `String` it *does* make a difference so they need to remain as references.

  Note that this works because the WifiEventsClass methods which invoke the callbacks don't re-use
  the constructed MACAddress or IPAddress objects, so the compiler doesn't have to worry about preserving
  the object and so doesn't care if the callback modifies them.

IPAddress class:

* Simplify IP address printing. IPAddress is printable, so `Serial.print(ip)` can be used instead of `Serial.print(ip.toString())`.
* Add `!=` operators. Notable omission. Also use lwip manipulation macros, will make it easier when/if the class gets updated to support IPv6.

Host implementation:

* Add some network messages to console so we can always see when IP address is allocated, etc.
* Add better check for Host WPS/SmartConfig support.
  Now throws an error compiling application if either is enabled, without breaking dist-clean.
  The previous version also added 'esp-wifi-check' files everywhere...

Review and update a couple of samples:

* Websocket_Client - bug in wsDisconnected, line 92 where msgTimer.stop() is called immediately after close(). This cancelled the 20-second retry timer because the wsDisconnected callback gets handled inside close().
* TcpClient_NarodMon (with translations)

### DIFF
--- a/Sming/Arch/Host/Platform/AccessPointImpl.cpp
+++ b/Sming/Arch/Host/Platform/AccessPointImpl.cpp
@@ -28,32 +28,32 @@ bool AccessPointImpl::config(const String& ssid, String password, WifiAuthMode m
 	return false;
 }
 
-IPAddress AccessPointImpl::getIP() const
+IpAddress AccessPointImpl::getIP() const
 {
 	return INADDR_NONE;
 }
 
-IPAddress AccessPointImpl::getNetworkBroadcast() const
+IpAddress AccessPointImpl::getNetworkBroadcast() const
 {
 	return INADDR_NONE;
 }
 
-IPAddress AccessPointImpl::getNetworkMask() const
+IpAddress AccessPointImpl::getNetworkMask() const
 {
 	return INADDR_NONE;
 }
 
-IPAddress AccessPointImpl::getNetworkGateway() const
+IpAddress AccessPointImpl::getNetworkGateway() const
 {
 	return INADDR_NONE;
 }
 
-bool AccessPointImpl::setIP(IPAddress address)
+bool AccessPointImpl::setIP(IpAddress address)
 {
 	return false;
 }
 
-MACAddress AccessPointImpl::getMacAddress() const
+MacAddress AccessPointImpl::getMacAddress() const
 {
 	return MACADDR_NONE;
 }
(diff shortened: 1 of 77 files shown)


Sming/Arch/Host/Platform/AccessPointImpl.cpp
```diff
@@ -53,7 +53,7 @@ bool AccessPointImpl::setIP(IPAddress address)
 	return false;
 }
 
-MACAddress AccessPointImpl::getMacAddr() const
+MACAddress AccessPointImpl::getMacAddress() const
```
slaff: Please, rename `MACAddress` to `MacAddress`. Since it is new no need to add any notes to the migration document.

Sming/Arch/Host/Platform/StationImpl.cpp
```diff
@@ -238,45 +243,33 @@ String StationImpl::getHostname() const
...
 
 	if(nif == nullptr) {
 		return MACADDR_NONE;
 	} else {
-		return nif->hwaddr;
+		return MACAddress(nif->hwaddr);
 	}
 }
 
 IPAddress StationImpl::getNetworkBroadcast() const
```
slaff: Rename `IPAddress` to `IpAddress`. Deprecate IPAddress and add a note in the migration documents about it.

slaff:
@mikee47 Do you want to add something more or shall I merge it?

mikee47:
It'll do for now I guess :-)

Issue #1813, opened 2019-08-09
Assertion in formUrlParser


mikee47:
Running `HttpServer_ConfigNetwork` under emulator (Windows or Linux) I get assertion failure in `HttpBodyParser.cpp`, line 43: `assert(state != nullptr);`. This is triggered from the `HttpServerConnection` destructor, which was modified in #1806:

```
	if(bodyParser) {
		bodyParser(request, nullptr, PARSE_DATAEND); ///< bang
	}
```

mikee47:
It doesnt crash any more, but it doesn't work either:

```
IP_CHANGE, ip: 192.168.13.10, netmask: 255.255.255.0, gateway: 192.168.13.1
1007 '/' registered
1015 '/ajax/get-networks' registered
1018 '/ajax/connect' registered
1021 '*' registered
5212315 HTTP parser error: HPE_CB_body
10394032 HTTP parser error: HPE_CB_body
15407573 HTTP parser error: HPE_CB_body
20419501 HTTP parser error: HPE_CB_body
25432939 HTTP parser error: HPE_CB_body
...
```

If I revert #1806 it works fine....

PR #1814, opened 2019-08-09
Fixes to body parser logic

slaff:
…racted.

Fixes: #1813.


slaff:
@mikee47 The HttpServer issues should be fix now. Can you check if the changes work for you?

mikee47:
Yes, that's back to normal now, thanks.

PR #1817, opened 2019-08-11
Automatic linker script generation

aemseemann:
Currently, for two-rom setups (on <=1MB devices), the second linker script has to be created manually. Although this quite easy, it is an unnecessary manual step when setting up a new project and it is also an additional complication for users new to the framework.

This PR integrates the linker script generation into the build process. 
It also tries to simplify the RBOOT configuration by automatically deriving the values of `RBOOT_TWO_ROMS` and `RBOOT_BIG_FLASH` from the provided ROM addresses, i.e. it now suffices to set `RBOOT_ROM1_ADDR` to the desired value.

Please consider this PR as a proposal. I am very open to suggestions on how to implement things differently, because I most certainly am not aware of all the details of the build system yet and the many thoughts behind it.

### DIFF
--- a/Sming/Components/rboot/.patches/rboot.patch
+++ b/Sming/Components/rboot/.patches/rboot.patch
@@ -1,11 +1,14 @@
 diff --git a/Makefile b/Makefile
-index 0b43474..7cb71f4 100644
+index 0b43474..5176f8f 100644
 --- a/Makefile
 +++ b/Makefile
-@@ -58,6 +58,16 @@ endif
+@@ -58,6 +58,19 @@ endif
  ifeq ($(RBOOT_IROM_CHKSUM),1)
  	CFLAGS += -DBOOT_IROM_CHKSUM
  endif
++ifneq ($(RBOOT_ROM0_ADDR),)
++	CFLAGS += -DBOOT_ROM0_ADDR=$(RBOOT_ROM0_ADDR)
++endif
 +ifneq ($(RBOOT_ROM1_ADDR),)
 +	CFLAGS += -DBOOT_ROM1_ADDR=$(RBOOT_ROM1_ADDR)
 +endif
@@ -19,7 +22,7 @@ index 0b43474..7cb71f4 100644
  ifneq ($(RBOOT_EXTRA_INCDIR),)
  	CFLAGS += $(addprefix -I,$(RBOOT_EXTRA_INCDIR))
  endif
-@@ -75,6 +85,10 @@ else ifeq ($(SPI_SIZE), 2Mb)
+@@ -75,6 +88,10 @@ else ifeq ($(SPI_SIZE), 2Mb)
  	E2_OPTS += -2048b
  else ifeq ($(SPI_SIZE), 4M)
  	E2_OPTS += -4096
@@ -30,9 +33,8 @@ index 0b43474..7cb71f4 100644
  endif
  ifeq ($(SPI_MODE), qio)
  	E2_OPTS += -qio
-
 diff --git a/rboot.c b/rboot.c
-index d622f97..00ccaeb 100644
+index d622f97..5e254fa 100644
 --- a/rboot.c
 +++ b/rboot.c
 @@ -13,6 +13,12 @@
@@ -48,10 +50,15 @@ index d622f97..00ccaeb 100644
  static uint32_t check_image(uint32_t readpos) {
  
  	uint8_t buffer[BUFFER_SIZE];
-@@ -258,7 +264,25 @@ static uint8_t calc_chksum(uint8_t *start, uint8_t *end) {
+@@ -257,8 +263,30 @@ static uint8_t calc_chksum(uint8_t *start, uint8_t *end) {
+ // created on first boot or in case of corruption
  static uint8_t default_config(rboot_config *romconf, uint32_t flashsize) {
  	romconf->count = 2;
++#ifdef BOOT_ROM0_ADDR
++    romconf->roms[0] = BOOT_ROM0_ADDR;
++#else
  	romconf->roms[0] = SECTOR_SIZE * (BOOT_CONFIG_SECTOR + 1);
++#endif
 +
 +#ifdef BOOT_ROM1_ADDR
 +	romconf->roms[1] = BOOT_ROM1_ADDR;
@@ -74,7 +81,7 @@ index d622f97..00ccaeb 100644
  #ifdef BOOT_GPIO_ENABLED
  	romconf->mode = MODE_GPIO_ROM;
  #endif
-@@ -306,100 +330,100 @@ uint32_t NOINLINE find_image(void) {
+@@ -306,100 +334,100 @@ uint32_t NOINLINE find_image(void) {
  	ets_delay_us(BOOT_DELAY_MICROS);
  #endif
  
@@ -204,7 +211,7 @@ index d622f97..00ccaeb 100644
  
  	// read boot config
  	SPIRead(BOOT_CONFIG_SECTOR * SECTOR_SIZE, buffer, SECTOR_SIZE);
-@@ -410,7 +434,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -410,7 +438,7 @@ uint32_t NOINLINE find_image(void) {
  #endif
  		) {
  		// create a default config for a standard 2 rom setup
@@ -213,7 +220,7 @@ index d622f97..00ccaeb 100644
  		ets_memset(romconf, 0x00, sizeof(rboot_config));
  		romconf->magic = BOOT_CONFIG_MAGIC;
  		romconf->version = BOOT_CONFIG_VERSION;
-@@ -433,10 +457,10 @@ uint32_t NOINLINE find_image(void) {
+@@ -433,10 +461,10 @@ uint32_t NOINLINE find_image(void) {
  
  		if (rtc.next_mode & MODE_TEMP_ROM) {
  			if (rtc.temp_rom >= romconf->count) {
@@ -226,7 +233,7 @@ index d622f97..00ccaeb 100644
  			temp_boot = 1;
  			romToBoot = rtc.temp_rom;
  		}
-@@ -447,10 +471,10 @@ uint32_t NOINLINE find_image(void) {
+@@ -447,10 +475,10 @@ uint32_t NOINLINE find_image(void) {
  	if (perform_gpio_boot(romconf)) {
  #if defined(BOOT_GPIO_ENABLED)
  		if (romconf->gpio_rom >= romconf->count) {
@@ -239,7 +246,7 @@ index d622f97..00ccaeb 100644
  		romToBoot = romconf->gpio_rom;
  		gpio_boot = 1;
  #elif defined(BOOT_GPIO_SKIP_ENABLED)
-@@ -462,7 +486,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -462,7 +490,7 @@ uint32_t NOINLINE find_image(void) {
  #endif
  		updateConfig = 1;
  		if (romconf->mode & MODE_GPIO_ERASES_SDKCONFIG) {
@@ -248,7 +255,7 @@ index d622f97..00ccaeb 100644
  			for (sec = 1; sec < 5; sec++) {
  				SPIEraseSector((flashsize / SECTOR_SIZE) - sec);
  			}
-@@ -474,7 +498,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -474,7 +502,7 @@ uint32_t NOINLINE find_image(void) {
  	// gpio/temp boots will have already validated this
  	if (romconf->current_rom >= romconf->count) {
  		// if invalid rom selected try rom 0
@@ -257,7 +264,7 @@ index d622f97..00ccaeb 100644
  		romToBoot = 0;
  		romconf->current_rom = 0;
  		updateConfig = 1;
-@@ -486,14 +510,14 @@ uint32_t NOINLINE find_image(void) {
+@@ -486,14 +514,14 @@ uint32_t NOINLINE find_image(void) {
  #ifdef BOOT_GPIO_ENABLED
  	if (gpio_boot && loadAddr == 0) {
  		// don't switch to backup for gpio-selected rom
@@ -274,7 +281,7 @@ index d622f97..00ccaeb 100644
  		// make sure rtc temp boot mode doesn't persist
  		rtc.next_mode = MODE_STANDARD;
  		rtc.chksum = calc_chksum((uint8_t*)&rtc, (uint8_t*)&rtc.chksum);
-@@ -504,7 +528,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -504,7 +532,7 @@ uint32_t NOINLINE find_image(void) {
  
  	// check we have a good rom
  	while (loadAddr == 0) {
@@ -283,7 +290,7 @@ index d622f97..00ccaeb 100644
  		// for normal mode try each previous rom
  		// until we find a good one or run out
  		updateConfig = 1;
-@@ -512,7 +536,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -512,7 +540,7 @@ uint32_t NOINLINE find_image(void) {
  		if (romToBoot < 0) romToBoot = romconf->count - 1;
  		if (romToBoot == romconf->current_rom) {
  			// tried them all and all are bad!
@@ -292,7 +299,7 @@ index d622f97..00ccaeb 100644
  			return 0;
  		}
  		loadAddr = check_image(romconf->roms[romToBoot]);
-@@ -543,7 +567,7 @@ uint32_t NOINLINE find_image(void) {
+@@ -543,7 +571,7 @@ uint32_t NOINLINE find_image(void) {
  	system_rtc_mem(RBOOT_RTC_ADDR, &rtc, sizeof(rboot_rtc_data), RBOOT_RTC_WRITE);
  #endif
  
--- a/Sming/Components/rboot/component.mk
+++ b/Sming/Components/rboot/component.mk
@@ -34,6 +34,8 @@ endif
 ### ROM Addresses ###
 # Make sure that your ROM slots and SPIFFS slot(s) do not overlap!
 
+CONFIG_VARS				+= RBOOT_ROM0_ADDR RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
+
 # Loation of first ROM (default is sector 2 after rboot and rboot config sector)
 RBOOT_ROM0_ADDR			?= 0x002000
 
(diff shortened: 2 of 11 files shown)


Sming/Components/rboot/.patches/rboot.patch
```diff
@@ -1,11 +1,14 @@
 diff --git a/Makefile b/Makefile
-index 0b43474..7cb71f4 100644
+index 0b43474..5176f8f 100644
```
mikee47: Would you rearrange this patch please, put rboot-api.c back at the end so comparison is easier?

mikee47: Would you rearrange this patch please, put rboot-api.c back at the end so comparison is easier?

Sming/Components/rboot/component.mk
```diff
@@ -21,37 +21,61 @@ endif
...
+
+ifeq ($(RBOOT_GPIO_ENABLED),0)
+ifneq ($(RBOOT_ROM2_ADDR),)
+$(warning RBOOT_GPIO_ENABLED is 0, RBOOT_ROM2_ADDR will be ignored)
+RBOOT_ROM2_ADDR :=
+endif
+endif
+
+# determine number of roms to generate
+ifneq ($(RBOOT_ROM1_ADDR),)
```
mikee47: As RBOOT_TWO_ROMS has been removed as a CONFIG_VAR, it would be helpful to add DEBUG_VARS += RBOOT_TWO_ROMS so we can check the logic.

Sming/Components/rboot/component.mk
```diff
@@ -21,37 +21,61 @@ endif
...
+RBOOT_TWO_ROMS := $(shell awk 'BEGIN { print ($(RBOOT_ROM0_ADDR) % 0x100000) == ($(RBOOT_ROM1_ADDR) % 0x100000) ? 0 : 1 }')
+else
+RBOOT_TWO_ROMS := 0
+endif
+
+# BIGFLASH mode is needed if at least one ROM addresses is exceeds the first 1MB of flash
+BIGFLASH_TEST := awk 'BEGIN { big=0; for(i = 1; i < ARGC; ++i) if(strtonum(ARGV[i]) > 0x100000) big=1; print big }'
```
mikee47: DEBUG_VARS += RBOOT_BIG_FLASH

Sming/Components/rboot/component.mk
```diff
@@ -75,7 +96,7 @@ COMPONENT_APPCODE		:= appcode rboot/appcode $(if $(RBOOT_EMULATION),host)
 APP_CFLAGS				+= -DRBOOT_INTEGRATION
 
 # these are exported for use by the rBoot Makefile
-CONFIG_VARS				+= RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
+CONFIG_VARS				+= RBOOT_ROM0_ADDR RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
```
mikee47: Should move these individually to CONFIG_VAR to point of declaration above

samples/HttpServer_FirmwareUpload/app/SignedRbootOutputStream.cpp
```diff
@@ -35,23 +34,21 @@ size_t SignedRbootOutputStream::write(const uint8_t* data, size_t size)
 				if(unexpectedMagic || unexpectedLoadAddress) {
 					setError(unexpectedLoadAddress ? "Unexpected load address. Try image for other slot."
 												   : "Invalid image received.");
-					return 0;
```
slaff: If there is an error just return 0; This way the request will be terminated earlier.

samples/HttpServer_FirmwareUpload/app/SignedRbootOutputStream.cpp
```diff
@@ -35,23 +34,21 @@ size_t SignedRbootOutputStream::write(const uint8_t* data, size_t size)
...
+		if(!errorFlag && (available > 0)) {
+			crypto_sign_update(&verifierState, static_cast<const unsigned char*>(data), available);
+			const size_t written = RbootOutputStream::write(data, available);
+			if(written != available) {
 				setError("Flash write failure");
 			}
 		}
 	}
-	return consumed;
+	return size; // Always pretend to consume everything. Otherwise the connection is closed before we get a chance to report our error message.
```
slaff: > Otherwise the connection is closed before we get a chance to report our error message.

Not really. The onMessageComplete callback will be called. In it you can decide what to do. If you return non-zero value the default value error page will be returned but if you return zero you can send custom error response.

For example the code below:
```
int onUpload(HttpServerConnection& connection, HttpRequest& request, HttpResponse& response)
{
	if(!response.isSuccess()) {
		debug_e("Request failed.");
		return -1;
	}

	ReadWriteStream* file = request.files["firmware"];
	if(file == nullptr) {
		debug_e("Something went wrong with the file upload");
		return 1;
	}

	SignedRbootOutputStream* uploadStream = static_cast<SignedRbootOutputStream*>(file);

	if(uploadStream->verifySignature(firmwareVerificationKey)) {
		// ...
	} else {
		response.code = HTTP_STATUS_BAD_REQUEST;
		response.setContentType(MIME_HTML);
		String html = "<H2 color='#444'>" + uploadStream->errorMessage + "</H2>";
		response.headers[HTTP_HEADER_CONTENT_LENGTH] = html.length();
		response.sendString(html);
	}
	return 0;
}
```

Can be changed to:

```
int onUpload(HttpServerConnection& connection, HttpRequest& request, HttpResponse& response)
{
	bool success = response.isSuccess();
        String errorMessage = "Bad Request";

        if(success) {
            ReadWriteStream* file = request.files["firmware"];
    	    if(file == nullptr) {
		debug_e("Something went wrong with the file upload");
                errorMessage = "<.....>";
		success = false;
	     }
        }

	if(success) {
             SignedRbootOutputStream* uploadStream = static_cast<SignedRbootOutputStream*>(file);
             if(uploadStream->verifySignature(firmwareVerificationKey)) {
		//...
	     }
             else {
                 success = false;
                 errorMessage =  uploadStream->errorMessage;
             }
        }
	
       if(!success) {
		response.code = HTTP_STATUS_BAD_REQUEST;
		response.setContentType(MIME_HTML);
		String html = "<H2 color='#444'>" + errorMessage + "</H2>";
		response.headers[HTTP_HEADER_CONTENT_LENGTH] = html.length();
		response.sendString(html);
       }

       return 0;
}
```

aemseemann: Thanks for the explanation, but unfortunately, I cannot to get it to work your way either.
Digging deeper through the HTTP server code, the reason seems to be that `HttpConnection::onTcpReceive` returns `false` in case of an HTTP parser error (and `SignedRbootOutputStream::write` returning something different from its `size` argument *is* considered a parser error):
```	
	int parsedBytes = http_parser_execute(&parser, &parserSettings, data, size);
	if(HTTP_PARSER_ERRNO(&parser) != HPE_OK) {
		// we ran into trouble - abort the connection
		onHttpError(HTTP_PARSER_ERRNO(&parser));
		return false;
	}
```
This causes the TCP connection to close, even if the `onMessageComplete` callback invoked from `onHttpError` has schedules an appropriate error response.

slaff: > This causes the TCP connection to close, even if the onMessageComplete callback ...

That should be taken care of in PR #1820.

aemseemann: Yeah that works. I'll rebase when yor PR is merged.

slaff: > Yeah that works. I'll rebase when yor PR is merged.

Great. The PR is merged so you can rebase.

aemseemann: Arrgh, my confirmation came too early (had the workaround still in place).

It does not work, because (note numbered comments)...

```
bool HttpConnection::onTcpReceive(TcpClient& client, char* data, int size)
{
	if(HTTP_PARSER_ERRNO(&parser) != HPE_OK) {
		setTimeOut(1);
		// if the parser is in error state then just ignore the incoming data.
		return false; // (3) kicks in for chunks received *after* graceful error handling
	}

	int parsedBytes = http_parser_execute(&parser, &parserSettings, data, size);
	if(HTTP_PARSER_ERRNO(&parser) != HPE_OK) {
		bool isRecoverable = onHttpError(HTTP_PARSER_ERRNO(&parser));
		if(!isRecoverable) {
			// we ran into trouble - abort the connection
			setTimeOut(1);
			return false;
		}
		// return true; <-- (1) Need this, or...
	}

	if(parser.upgrade) {
		return onProtocolUpgrade(&parser);
	} else if(parsedBytes != size) {
		return false; // (2) ...this kicks in
	}

	return true;
}
``` 
(1) missing `return true` in this line. Otherwise, method continues and...
(2) ...the partially consumed chunk is detected in this branch and handled ungracefully.

But even if these points are addressed and the connection stays open, any subsequently received chunk of data steps into.. 

(3) this condition, where the connection is again closed immediately.

slaff: @aemseemann Can you rebase this PR on the latest changes in `develop` and try again? It should work now as expected.

aemseemann: @slaff: Thanks for improving the error handling. Unfortunately, the connection is still closed immediately when uploading the "wrong" file. I'll further investigate it tomorrow.

slaff:
> Unfortunately, the connection is still closed immediately when uploading the "wrong" file

If you mean that the connection is closed right after the response is sent then that is the correct behaviour. See [HttpServer_FirmwareUpload](https://github.com/SmingHub/Sming/tree/develop/samples/HttpServer_FirmwareUpload) sample.

aemseemann:
I checked with Wireshark. Your are right, this is indeed what happens and the error message shows up as expected in Chromium.
However, Firefox (tested with v69.0.1) seems to be unable to handle [early responses](https://stackoverflow.com/questions/14250991/is-it-acceptable-for-a-server-to-send-a-http-response-before-the-entire-request) properly and shows a "Connection Reset" error instead, which kind of suggests to the user that the device is not behaving properly:
firefox-connection-reset

Since I would consider Firefox still significant enough to bother, I see three possible solutions:

1. Fix it in the network error handling code by silently ignoring data received after sending the error response.
2. Fix it in the _firmware_  of the example by pretending to consume everything (my original attempt)
3. Fix it in the _webinterface_ of the example, for example using an ajax-based upload. Not sure tough that these requests will behave differently when receiving an early response.

Which one would you prefer?

slaff:
> 3. Fix it in the webinterface of the example, for example using an ajax-based upload. Not sure tough that these requests will behave differently when receiving an early response.

> Which one would you prefer?

@aemseemann I would prefer solution 3.

aemseemann:
> > Which one would you prefer?
> 
> @aemseemann I would prefer solution 3.

I tried that. Unfortunately, ajax POST requests have the same problem in Firefox: Responses are not recognized if the server closes the connection early. This leaves us with option 1 or 2. 
To avoid further complicating the HTTP server code, I vote for option 2. Do you agree?

slaff:
> I tried that. Unfortunately, ajax POST requests have the same problem in Firefox: Responses are not recognized if the server closes the connection early.

Can you send me your sample Ajax code? I was expecting that in Firefox the ajax error callback will be triggered and the JavaScript code can report that the firmware was not accepted without any other visual issues.

aemseemann:
I used the test code below. The error callback is indeed invoked, but the error message from the ESP's response is not available. Without it, the page could only display that *something* went wrong, but I would prefer the actual cause.
```js
$(document).ready(function() {
    $('#uploadAjax').on('click', function (e) {
        var files = new FormData();
        files.append('firmware', $('#firmwarefile')[0].files[0]);

        $.ajax({
            type: 'post',
            url: '/upload',
            processData: false,
            contentType: false,
            data: files,
            success: function (response) {
                console.log("Success:");
                console.log(response);
            },
            error: function (err) {
                console.log("Error:");
                console.log(err);
            }
        });
    });
});
```

slaff:
> Without it, the page could only display that something went wrong, but I would prefer the actual cause.

@aemseemann Please make the sample use Ajax file upload and the stream class to report an error as early as an issue is detected. If there is an error during upload and the browser is Firefox there should be a message also explaining Firefox's behaviour. 

In the sample code there can be an information in the form of comments that explains how to work-around the issue with Firefox. But that should not be enabled by default and the developer should  be aware what he/she is doing.

aemseemann:
I can do that, but is it really your preferred option? I known I brought it up myself, but merely for completeness, not because I consider it a _good_ solution.
I find it rather impolite to display an error message like "_Oops, something went wrong, but we cannot tell you what exactly, because your browser (Firefox) does not behave the way we like, although our expectation is not based on any standard. Furthermore, this message proves that we know about it, but did not bother to implement a working solution_." (or maybe a little bit less explicit).

Since the whole HTTP server error handling discussion is rather off-topic in this PR, I created https://github.com/SmingHub/Sming/issues/1875.

aemseemann:
I rebased and reordered the commits a bit.
The error handling in `SignedRbootOutputStream::write` should now comply with the review change request and error messages in Firefox are still working.

Sming/Arch/Esp8266/Components/custom_heap/component.mk
```diff
@@ -7,17 +7,14 @@ COMPONENT_INCDIRS		:= umm_malloc/src umm_malloc/includes/c-helper-macros
...
 GLOBAL_CFLAGS			+= -DUMM_POISON_CHECK
 endif
 
-# Make copy of libmain and remove mem_manager.o module
-$(COMPONENT_RULE)$(LIBMAIN_SRC): $(SDK_LIBDIR)/libmain.a
-	$(info Enabling custom heap implementation)
-	$(Q) cp $^ $@
-	$(Q) $(AR) -d $@ mem_manager.o
+# remove mem_manager.o module from libmain of SDK
+define LIBMAIN_COMMANDS +=
```
slaff: @aemseemann Please, take out of this PR all changes that are not related to the main goal of this PR. If the removed changes are still needed for this PR then add them as new PRs and we will review and merge them before this one.

aemseemann: -> See #1959

The reason for this modifications is that the logic to prepare libmain in rboot/component.mk  became a bit unmanageable, so I felt it would be a good idea to make the whole libmain generation a bit more flexible

slaff:
@aemseemann Rebase this PR and ping me when you are ready.

aemseemann:
I tried it yesterday, but ran into trouble with alignment exceptions and a subsequent boot loop (`Fatal exception 9(LoadStoreAlignmentCause): epc1=0x40101044,...` where the PC points somewhere into `vPortFree`) . It seems to start with commit 1fb600ec38241d83b7a60dce2fb95698a98bd1b6 but unfortunately I hadn't much time yesterday to investigate it any further. Please be patient until next weekend.

aemseemann:
@slaff: The PR is now ready from my side. Successfully tested on an ESP01 board.

Sming/Components/rboot/component.mk
```diff
@@ -21,37 +21,67 @@ endif
...
+ifneq ($(RBOOT_ROM1_ADDR),)
+RBOOT_TWO_ROMS := $(shell awk 'BEGIN { print ($(RBOOT_ROM0_ADDR) % 0x100000) == ($(RBOOT_ROM1_ADDR) % 0x100000) ? 0 : 1 }')
+else
+RBOOT_TWO_ROMS := 0
+endif
+
+DEBUG_VARS 				+= RBOOT_TWO_ROMS
+
+# BIGFLASH mode is needed if at least one ROM address exceeds the first 1MB of flash
+BIGFLASH_TEST := awk 'BEGIN { big=0; for(i = 1; i < ARGC; ++i) if(strtonum(ARGV[i]) > 0x100000) big=1; print big }'
```
slaff: @aemseemann I am getting an error on my Linux box:

```
awk: line 2: function strtonum never defined
```

slaff: @mikee47  Can you suggest a simple check that should replace the awk code and work on Windows, Mac OS X, Linux (and optionally FreeBSD) ?

mikee47: I've seen `int()` used, but having other issues getting this to run. Line 178...

Sming/Components/rboot/component.mk
```diff
@@ -125,20 +151,50 @@ endif # RBOOT_EMULATION
...
+-include $2.config
+ifneq ($$(GEN_$3),$$($3))
+.PHONY: $2
+endif
+# Generate linker script from template
+$2: $(RBOOT_LD_TEMPLATE)
+	$$(info LDGEN $$@)
+	$$(Q) awk 'match($$$$0, /(^.*irom0_0_seg\s*:\s*)\S+/, m) { \
+			printf "%sorg = 0x%08X, len = 1M - 0x%X\n", m[1], 0x40200010 + $$($3) % 0x100000, $$($3) % 0x100000; next \
+		} 1 { print $$$$0 }' $$< > $$@
```
mikee47: mawk doesn't like this, as far as I can tell because it doesn't support the 3rd 'array' parameter.

Sming/Components/rboot/component.mk
mikee47: As I recall we had issues getting compatibility with mawk, I think if the standard were gawk we'd be fine.

aemseemann: Wasn't aware of the different flavours of awk... I'll try to get it to work with mawk and busybox awk. Hopefully the latter will be minimal enough to cover all forseable installations.

aemseemann: This is more complicated than I thought. It seems I can get it to work for `mawk '...'` and `gawk -n '...'` (or `awk -n` if that is an alias), but have no idea how to detect which one to use.
Since this is quite a core element of the build process, forcing the user to choose via `make AWK=...` is not acceptable either.

As alternatives to awk, how do you feel about...
- `$(( ... ))`  (or, more generally, which shells have to be supported?)
- `python -c '...'`

Personally, I would prefer the Python option, since it is probably the most platform independent solution and Python is already used by the build system.

slaff: The changes work on my Linux box. @mikee47 When you have time can you check if that works on your Windows box? I will check later today also the MacOS build.

mikee47: In MinGW, which uses gawk, it's not picking up the rom_base or rom_offset variables:

```
irom0_0_seg :                         org = 0x00000000, len = 1M - 0x10
```

According to https://web.mit.edu/gnu/doc/html/gawk_16.html variables are available within a BEGIN rule, but says nothing about their availablity within a match. I'll post back if I can find a working solution.

mikee47: Tried a variety different approaches and it's just not happy converting strings. Rather than attempting to filter `rboot_rom0.ld`, would it simplify things if we just generate the file entirely, or generated the required fragment and include it? Could we use the compiler to convert .h to .ld ?

mikee47: I'm mindful that at present the build system doen't provide any way for Components to customise the linker scripts, something to look at later I think.

aemseemann: Next attempt: I now moved all the math to the linker script itself, so awk does not have to do hex conversion and variables anymore. Should look like this for the first rom:
```
  irom0_0_seg : org = 0x40200010 + (0x2000 & 0xFFFFF), len = 1M - (0x2000 & 0xFFFFF) - 0x10
```

@mikee47, @slaff: Can you try again, please?

mikee47: Works a treat!

slaff: @mikee47 I am waiting for your approval to schedule this PR for merging?

slaff:
@aemseemann Please, rebase your PR.

aemseemann:
Rebase complete and successfully tested on real HW.

Issue #1816, opened 2019-08-11
Build doesn't find header file


kmihaylov:
Hello. Just updated to the latest develop. After make-distclean and make some headers could not be found. I'm not aware how the build system works. May you check?

Thanks!

```Building /opt/Sming/Sming/out/Esp8266/debug/lib/clib-libyuarel.a
make[1]: Nothing to be done for 'build'.

Building /opt/Sming/Sming/out/Esp8266/debug/lib/clib-Sming-ccc08e87d72c17c6c82f1e2405aa0d08.a
In file included from /opt/Sming/Sming/Core/FileSystem.cpp:11:0:
/opt/Sming/Sming/Core/FileSystem.h:18:26: fatal error: spiffs_sming.h: No such file or directory
 #include "spiffs_sming.h"
                          ^
compilation terminated.
/opt/Sming/Sming/component-wrapper.mk:155: recipe for target 'Core/FileSystem.cpp.d' failed
make[1]: *** [Core/FileSystem.cpp.d] Error 1
/opt/Sming/Sming/project.mk:357```

mikee47:
@kmihaylov What version have you updated from?

mikee47:
@kmihaylov There's probably an empty directory `Arch/Esp8266/Components/spiffs` which you'll need to delete.

kmihaylov:
@mikee47 
Wow thanks! It worked. I had to remove also rboot/rboot from Components. Now I see some deprecated functions. A lot of new stuff!
Thank you!

mikee47:
Have a read of https://sming.readthedocs.io/en/latest/upgrading/3.8-4.0.html

mikee47:
Also https://sming.readthedocs.io/en/latest/_inc/Sming/building.html - explanation of your problem under 'known issues'. Should really put that somewhere more obvious :-(

kmihaylov:
Also I had to install ```sphinx-common```, Ubuntu 18.04.
These links would be helpful on the Sming's github front page. I checked another link (sming-framework.readthedocs.io), obviously I was mistaken.

mikee47:
I guess once this makes it out of develop the front page will get updated. What's `sphinx-common`?

kmihaylov:
A package needed to build Sming, I believe it is related to the new documentation system. sphinx-build (or something similar) was missing.

mikee47:
[...]

PR #1818, opened 2019-08-11
Fix deprecated list plus minor doc updates

mikee47:
Address issues raised in #1816

* Some typedefs don't appear in dOxygen deprecated list unless `EXTRACT_ALL=YES`. Might be a better setting, can't find it.
* Fix docs. `clean` target so it doesn't require sphinx, thus dist-clean works without additional requirements
* Rename libsodium submodule README so doc. system picks it up
* Update Host readme with note on `DigitalHooks`
* Fix rboot reference in docs

### DIFF
--- a/Sming/Arch/Host/README.rst
+++ b/Sming/Arch/Host/README.rst
@@ -37,9 +37,7 @@ or later. If it’s older, execute these commands:
 Building
 --------
 
-Build the framework and application as usual, specifying :envvar:`SMING_ARCH` =Host. For example:
-
-::
+Build the framework and application as usual, specifying :envvar:`SMING_ARCH` =Host. For example::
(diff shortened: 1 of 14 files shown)


Sming/Arch/Host/README.rst
```diff
@@ -106,9 +102,7 @@ output at startup.)
 
 Note: For Windows users, ``putty`` is a good alternative to telnet. It
 has options to for things like carriage-return/linefeed translation
```
slaff: `has options to for` Do you mean: `has options *too* for` ?

slaff:
Do you want to add something before I merge this PR?

mikee47:
All done thanks.

PR #1821, opened 2019-08-18
Introduce Clock framework and Polled Timers

mikee47:
**Polled Timers** are a new feature, based heavily on the esp8266 Arduino project's PolledTimeout.h
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/PolledTimeout.h.

PR #1830 added *muldiv* and introduced the use of *std::ratio* which makes use of C++ language features
to perform compile-time calculations and parameter checking. This PR builds on that using the new
*NanoTime* library. I felt this warranted new development since existing solutions are either C-based
or not suited for embedded use. This also forms a core part of the Sming timer framework.

This PR moves the existing ElapseTimer and CycleCounter classes into Core. I've taken care to avoid
any extraneous dependencies in these timers which means they're appropriate for use in lower-level
code (Components, drivers, etc.).

The goal is to provide a consistent, logical framework for all timers.

More information can be found in the docs under 'timers'.


mikee47:
Quite a few structuraly rearrangements here but I'm feeling quite confident that it's the right direction in support of ESP32. Moving all the low-level stuff into separate drivers simplifies the class implementations without sacrificing performance or bloating the firmware.

This PR has been pulled out of a larger development with some additions for Sming, which I'll summarise briefly so you've got an idea where I'm going:

* Documentation on how Sming works (multitasking, interupts, memory, timers, tasks, etc.)
* Classes to support task management. i.e. a simple way to have one or more continuous background activities.
* I2S hardware driver. Pretty much a rewrite using an IDF-style API with Arduino support layer and some additional optimisations.
* Port of the ESP8266Audio libraries plus sample applications.
* Asynchronous I2C driver model. i.e. a proper device stack using classes to implement devices on a bus using asynchronous interrupt-based transfers.

Polled timers are something that can simplify and streamline driver code so it needs to be somewhere accessible, Core seems like the right place though as always totally open to suggestions.

mikee47:
Note before merging I'll need to update the migration docs. for this and expand on `timers.rst`.

slaff:
> Note before merging I'll need to update the migration docs. for this and expand on timers.rst.

Ping me when this PR is ready.

mikee47:
I'm splitting this PR as it's become quite complex.

slaff:
@mikee47 Is this PR tested and ready?

mikee47:
@slaff Yes indeed.

PR #1822, opened 2019-08-24
Add support for 64-bit number printing

mikee47:
Adds support to `Print` and `String` classes for `long long` and `unsigned long long` numbers using revised `stringconversion` routines.
`Print` class revised to use `stringconversion` routines for consistency with `String`.
Note that Print used its own code for this which produced hex in uppercase, rather than lowercase.

`stringconversion` updates:

* Fix incorrect `width` parameter passing in `ultoa_wp()`
* Move alias functions into header as static inline (removes redundant function calls)
* Eliminate code duplication by calling `ultoa_wp` from `ltoa_wp`
* Avoid un-necessary use of strlen in conversions
* Check `base` parameter and default to base 10 if invalid - this was in Print.cpp but moving it here ensures it's applied consistently
* Add 64-bit string conversion functions for `long long` and `unsigned long long`.


slaff:
@mikee47 Shall I merge it or you have something to add?

mikee47:
All done thanks.

PR #1825, opened 2019-08-29
Update toolchains to C++17

mikee47:
Updates framework to build using GCC 9.2.0 toolchain for C++17.

Toolchains are at https://github.com/earlephilhower/esp-quick-toolchain/releases.

See `docs/source/arch/esp8266/getting-started/eqt.rst` for installation details.

(Also at https://smingdev.readthedocs.io/en/dev-esp-quick-toolchain/arch/esp8266/getting-started/eqt.html)

Extract the toolchain to a suitable location, e.g. `opt/esp-quick-toolchain` or `C:\tools\esp-quick-toolchain`, and set `ESP_HOME` accordingly.

The main change for the new toolchain is that the core PROGMEM handling stuff is included in `sys/pgmspace.h`. See https://github.com/esp8266/Arduino/pull/5376 for more details.

Also includes de-duplication for NUL-terminated PSTR definitions.


mikee47:
OK, ready for review.

.gitmodules
```diff
@@ -71,7 +71,7 @@
 
 [submodule "Host.lwip"]
 	path = Sming/Arch/Host/Components/lwip/lwip
-	url = git://git.savannah.gnu.org/lwip.git
+	url = https://git.savannah.nongnu.org/git/lwip.git
```
slaff: @mikee47 Why is that change here?

.travis/build.sh
```diff
@@ -27,7 +27,6 @@ fi
 
 # Setup ARCH SDK paths
 if [ "$SMING_ARCH" == "Esp8266" ]; then
-	export ESP_HOME=$TRAVIS_BUILD_DIR/opt/esp-alt-sdk
 	export PATH=$PATH:$ESP_HOME/xtensa-lx106-elf/bin:$ESP_HOME/utils/
```
slaff: Where is the ESP_HOME set? If it is not done yet then also the PATH update will not work as expected.

Sming/Arch/Esp8266/Components/esp8266/ESP8266_NONOS_SDK.patch
```diff
@@ -0,0 +1,6 @@
+diff --git a/lib/libc.a b/lib/libc_.a
+rename from lib/libc.a
+rename to lib/libc_.a
+diff --git a/lib/libgcc.a b/lib/libgcc_.a
```
slaff: Why is the underscore renaming here?

Sming/Arch/Esp8266/Components/heap/component.mk
```diff
@@ -9,13 +10,13 @@ COMPONENT_SUBMODULES	:= umm_malloc
...
 COMPONENT_INCDIRS		+= umm_malloc/src umm_malloc/includes/c-helper-macros
 
-COMPONENT_DEPENDS		:= esp8266
-
 COMPONENT_VARS			+= UMM_POISON_CHECK
 ifeq ($(UMM_POISON_CHECK),1)
 GLOBAL_CFLAGS			+= -DUMM_POISON_CHECK
 endif
 
+COMPONENT_CFLAGS		+= -Wno-array-bounds
```
slaff: Why is that change needed?

.travis/build.sh
slaff: Sorry, my bad. I saw that it is now moved to .travis.yml.

slaff:
> OK, ready for review.

Looks good to me. Newer language features for a small price. Here is the comparison that I've done for one of my applications (same laptop, same Sming version and same SDK version):

Roughly 300 bytes less free RAM, 50 bytes less IRAM, and 4K bigger file size. 

Details below:

# OLD toolchain


## Memory Usage

```
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8580|    1408
    rodata|           ReadOnly Data (RAM)|    3FFE8580|    3FFE9880|    4864
       bss|      Uninitialized Data (RAM)|    3FFE9880|    3FFF0B18|   29336
      text|            Cached Code (IRAM)|    40100000|    40106C50|   27728
irom0_text|           Uncached Code (SPI)|    40202010|    4027C9FD|  502253
Total Used RAM : 35608
Free RAM : 46312
Free IRam : 5058
ESPTOOL2 out/Esp8266/debug/firmware/rom0.bin
```

## File size
524K for out/Esp8266/debug/firmware/rom0.bin

# New toolchain

## Memory Usage

```
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8680|    1664
    rodata|           ReadOnly Data (RAM)|    3FFE8680|    3FFE99C8|    4936
       bss|      Uninitialized Data (RAM)|    3FFE99C8|    3FFF0C40|   29304
      text|            Cached Code (IRAM)|    40100000|    40106C88|   27784
irom0_text|           Uncached Code (SPI)|    40202010|    4027D5FC|  505324
Total Used RAM : 35904
Free RAM : 46016
Free IRam : 5002
ESPTOOL2 out/Esp8266/debug/firmware/rom0.bin
```

## File size

528K for out/Esp8266/debug/firmware/rom0.bin

mikee47:
> Newer language features for a small price.

Plus the compilers are now consistent for all supported platforms (I'm assuming MacOS uses the Linux toolchain?). I'll be especially happy to be rid of GCC 4.8.5 as it's problematic.

.gitmodules
mikee47: #1988 refers. This same repo. is used as a submodule in esp8266 LWIP, using this URL. It didn't fix the problem so can pull this commit out and deal with that separately.

slaff: > so can pull this commit out and deal with that separately.

Yes, please.

Sming/Arch/Esp8266/Components/esp8266/ESP8266_NONOS_SDK.patch
mikee47: Renaming them eliminates the risk of their being mistakely picked up instead of the newlib ones.

Sming/Arch/Esp8266/Components/esp8266/esp_cplusplus.cpp
```diff
@@ -11,34 +11,9 @@ extern void (*__init_array_end)();
 void cpp_core_initialize()
 {
     void (**p)(void);
-    for (p = &__init_array_start; p != &__init_array_end; ++p)
+    for (p = &__init_array_start; p != &__init_array_end; ++p) {
```
mikee47: I note in the esp8266-arduino core constructors are called in reverse order. Not sure why...

Sming/Arch/Esp8266/Components/gdbstub/gdbstub-internal.h
```diff
@@ -12,6 +12,22 @@
 
 #pragma once
 
+#define XCHAL_CP0_SA_SIZE               0
```
mikee47: When I initially created this PR these were missing from the SDK. I'll re-check and use those if they've been added.

Sming/Arch/Esp8266/Components/heap/component.mk
mikee47: If you comment this out then try compiling with `ENABLE_CUSTOM_HEAP=1`:

```
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c: In function 'umm_init':
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:73:32: error: array subscript 1 is outside array bounds of 'char[1]' [-Werror=array-bounds]
   73 | #define UMM_BLOCK(b)  (umm_heap[b])
      |                       ~~~~~~~~~^~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:75:24: note: in expansion of macro 'UMM_BLOCK'
   75 | #define UMM_NBLOCK(b) (UMM_BLOCK(b).header.used.next)
      |                        ^~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:231:5: note: in expansion of macro 'UMM_NBLOCK'
  231 |     UMM_NBLOCK(block_1th) = block_last | UMM_FREELIST_MASK;
      |     ^~~~~~~~~~
In file included from s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:35:
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc_cfg.h:60:13: note: while referencing '_heap_start'
   60 | extern char _heap_start;
      |             ^~~~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:73:32: error: array subscript 1 is outside array bounds of 'char[1]' [-Werror=array-bounds]
   73 | #define UMM_BLOCK(b)  (umm_heap[b])
      |                       ~~~~~~~~~^~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:77:24: note: in expansion of macro 'UMM_BLOCK'
   77 | #define UMM_NFREE(b)  (UMM_BLOCK(b).body.free.next)
      |                        ^~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:232:5: note: in expansion of macro 'UMM_NFREE'
  232 |     UMM_NFREE(block_1th)  = 0;
      |     ^~~~~~~~~
In file included from s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:35:
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc_cfg.h:60:13: note: while referencing '_heap_start'
   60 | extern char _heap_start;
      |             ^~~~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:73:32: error: array subscript 1 is outside array bounds of 'char[1]' [-Werror=array-bounds]
   73 | #define UMM_BLOCK(b)  (umm_heap[b])
      |                       ~~~~~~~~~^~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:76:24: note: in expansion of macro 'UMM_BLOCK'
   76 | #define UMM_PBLOCK(b) (UMM_BLOCK(b).header.used.prev)
      |                        ^~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:233:5: note: in expansion of macro 'UMM_PBLOCK'
  233 |     UMM_PBLOCK(block_1th) = block_0th;
      |     ^~~~~~~~~~
In file included from s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:35:
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc_cfg.h:60:13: note: while referencing '_heap_start'
   60 | extern char _heap_start;
      |             ^~~~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:73:32: error: array subscript 1 is outside array bounds of 'char[1]' [-Werror=array-bounds]
   73 | #define UMM_BLOCK(b)  (umm_heap[b])
      |                       ~~~~~~~~~^~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:78:24: note: in expansion of macro 'UMM_BLOCK'
   78 | #define UMM_PFREE(b)  (UMM_BLOCK(b).body.free.prev)
      |                        ^~~~~~~~~
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:234:5: note: in expansion of macro 'UMM_PFREE'
  234 |     UMM_PFREE(block_1th)  = block_0th;
      |     ^~~~~~~~~
In file included from s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc.c:35:
s:/sandboxes/sming-dev/Sming/Arch/Esp8266/Components/heap/umm_malloc/src/umm_malloc_cfg.h:60:13: note: while referencing '_heap_start'
   60 | extern char _heap_start;
      |             ^~~~~~~~~~~
cc1.exe: all warnings being treated as errors
```

Sming/Arch/Esp8266/Components/esp8266/esp_cplusplus.cpp
slaff: @igrr @earlephilhower is there a reason for calling the C++ constructors in reverse order in `esp8266-arduino` ?

Sming/Arch/Esp8266/Components/esp8266/ESP8266_NONOS_SDK.patch
slaff: Ok, can you rename them to lib(x).a -> lib(x).orig.a  instead of underscore?

Sming/Arch/Esp8266/Compiler/ld/common.ld
```diff
@@ -173,11 +180,6 @@ SECTIONS
     /* debug_e() string pointers */
     *(.rodata._ZZ*log_string)
 
-    /* const references, mainly in templated code */
-    *(.rodata._ZN*)
-
-	*(.rodata._ZZN*)
-
```
mikee47: With the old toolchain this saved quite a bit of RAM, but with the new compiler it breaks because it also catches regular const strings. This feels generally unsafe so I've removed them. Will add some better ones.

Sming/build.mk
```diff
@@ -148,14 +147,27 @@ else
...
+GCC_VERSION			:= $(shell $(CC) -dumpversion)
+
+# Select C++17 if supported, defaulting to C++11 otherwise
+DEBUG_VARS			+= CPP_STD
+ifeq ($(GCC_VERSION),4.8.5)
+CPP_STD				:= c++11
+else
+CPP_STD				:= c++17
+endif
+CXXFLAGS			+= -std=$(CPP_STD)
```
mikee47: Note that I've elected to select C++17 for the current toolchains (except linux which doesn't support it). This has the potential to break some existing code, but AFAIK only insofar as it includes better checks and so not a bad thing IMHO. The alternative is to select C++17 for the new toolchain only. This could also be made user-selectable, but not sure that's necessarily a good move.

mikee47: .. although being able to select the standard would be useful for testing against C++20, etc.

mikee47:
Another thing to consider is where the toolchains live. At the moment they're pre-release, but perhaps before releasing Sming 4.0.1 we might consider adding them to our SmingTools repo. and pulling them from there.

mikee47:
I'll also add some documentation updates covering this PR.

Sming/Arch/Esp8266/Components/libc/component.mk
```diff
@@ -0,0 +1,13 @@
...
+
+ifeq ($(USE_NEWLIB),1)
+COMPONENT_SRCFILES		+= libc_replacements.c
+LIBDIRS					+= $(COMPONENT_PATH)/newlib
+EXTRA_LIBS				+= m c gcc
+else
+COMPONENT_SRCFILES		+= pgmspace.c
+LIBDIRS					+= $(COMPONENT_PATH)/lib
+EXTRA_LIBS				+= microc microgcc setjmp
+endif
```
mikee47: You may have spotted that `stdc++` has been removed from the list of libraries. One reason is that it contains allocation operators (new, delete, etc.) which conflict with our own. We don't use STL streams or a bunch of other stuff that it might be required for so leaving it out simplifies things. We can always revisit in the future if it's required for any other reason.

mikee47:
I elected to move the runtime libraries from `Compiler/lib` into the `libc` Component because we need a way to select them based on toolchain. The `Compiler/data` directly is a bit weird, that file should really go into the `Spiffs` Component. Which then just leaves the linker scripts. It feels like there should be somewhere more logical for them, but exactly where eludes me at the moment.

I mention this now because we're moving stuff around, but probably something to look at in another PR.

mikee47:
Integration testing. Do we want to leave the C++17 tests active, or perhaps make them conditional https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditionally-excluding-jobs

slaff:
> Do we want to leave the C++17 tests active

Is this PR containing such tests? If yes can you point me to the files? If no, but that is planned then yes - make them, the C++17 tests, run when the new toolchain is used.

mikee47:
> 
> 
> > Do we want to leave the C++17 tests active
> 
> Is this PR containing such tests? If yes can you point me to the files? If no, but that is planned then yes - make them, the C++17 tests, run when the new toolchain is used.

No, no specific tests. I was referring to the additional build in Travis and Appveyor which re-runs the build for both toolchains. If we're only considering the new toolchain experimental then at this stage we could disable the second stage for now.

slaff:
> If we're only considering the new toolchain experimental then at this stage we could disable the second stage for now.

No, better leave it.

Sming/build.mk
mikee47: I've added support for this and included note in documentation.

Sming/Arch/Esp8266/Components/gdbstub/gdbstub-internal.h
mikee47: No, these are still defined in core.h but oddly only after they're required. So still need them here.

mikee47:
For HostTests application:
```
               data     rodata   bss      text     irom0_text
Current:       1420     5876     27856    28827    439976
Old toolchain: 1420     6584     27856    28823    439241
New toolchain: 1804     6464     27856    29656    482632

               Used RAM    Free RAM    Free IRAM
Current:       35152       46768       3959
Old toolchain: 35860       46060       3963
New toolchain: 36124       45796       3130
```

The increased RAM usage is because GCC ignores section attributes in templated code.

I've had a brief look at the IRAM code and it appears the new compiler is producing
worse code than the old one. Reasons to be investigated.

slaff:
> but perhaps before releasing Sming 4.0.1 we might consider adding them to our SmingTools repo. and pulling them from there.

Yes, exactly.

Sming/Arch/Esp8266/Components/esp8266/esp_cplusplus.cpp
earlephilhower: I believe that was in @igrr's original work and not something I've touched, sorry.  Every example I see (well, they all look cut-n-pasted from the same source) does the initialization from [0]..[n].

While between translation units there's no guarantees, I would have thought that w/in a single file the order would be guaranteed.  May be a bug in the `esp8266-arduino` core.  Will open an issue there and we'll look into it.

mikee47: As per https://github.com/esp8266/Arduino/issues/6870 reverse-initialiation _is_ still required to ensure correct init priority. Will add test case to HostTests and fix.

Sming/Arch/Esp8266/Components/libc/include/sys/pgmspace.h
```diff
@@ -0,0 +1,174 @@
...
+
+#define PROGMEM __attribute__((aligned(4))) ICACHE_FLASH_ATTR
+
+// flash memory must be read using 32 bit aligned addresses else a processor exception will be triggered
+// order within the 32 bit values are
+// --------------
+// b3, b2, b1, b0
+//     w1,     w0
+
+#define pgm_read_with_offset(addr, res)                                                                                \
```
earlephilhower: There is a new version of this macro in the newlib folder which uses 1 fewer assembly instruction.

mikee47: Thanks, I'll need to review that module in some detail I think.

I note that you've used `asm` in those macros, but that won't compile with std=c++X, requires `__asm__` instead - I think because it uses extended syntax. (I've just added a #define asm __asm__ but it's a bit of a hack...)

Whilst I'm at it, our PROGMEM implementation includes `aligned((4))` but your implementation does not. Is there a specific reason for that?

earlephilhower: That's very odd.  The PROGMEM macro builds and runs fine w/ gcc=c++14 in the main Arduino repo.  Never looked into it, really, since it just worked.

We don't align PROGMEM, but we do align PSTRs and I've rewritten several of the _P string functions to use this to speed things up in newlib:
`  #define PSTR(s)            (__extension__({static const char __c[] __attribute__((__aligned__(4))) PROGMEM = (s); &__c[0];}))`

Also, related to PSTRs, recently we adjusted them to allow the linker to combine duplicated strings and save lots of memory in certain apps:  
https://github.com/esp8266/Arduino/pull/6565

I see now that *that* change did not filter back into newlib, darn it.  I need to do a PR there...

mikee47: The asm thing works because Arduino builds with `-std=gnu++11` but we're using `-std=c++11` and `-std=c++17` to try and keep with the standard.

The de-duplication is interesting, will definitely check that out thanks. My thought on the PROGMEM alignment was simply that if there's no justification for every using it unaligned then it does help to simplify things a little. My preference is to try to steer clear of 'raw' progmem stuff where possible. See #1974 for example, using counted flash strings with C++ objects. Plus there's https://github.com/Chris--A/PGMWrap. Lots of ways to leverage C++ and avoid the low-level stuff!

mikee47: @slaff I'm inclined to leave our sys/pgmspace.h as-is for this PR as it affects builds with existing (old) toolchains. The de-duplication can probably be applied in otger areas as well.

Sming/Arch/Esp8266/Compiler/ld/common.ld
```diff
@@ -156,27 +161,30 @@ SECTIONS
...
-    *(.rodata._ZTV*) /* C++ vtables */
-    *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom0.text.* .irom.text .irom.debug.*)
+    *(.irom0.literal .irom.literal .irom.text.literal .irom0.text .irom0.text.* .irom.text .irom.text.* .irom.debug.*)
 
     /* Generated libraries */
     *liblwip2.a:(.literal .text .literal.* .text.*)
     */clib-*.a:*(.literal .text .literal.* .text.* .stub .gnu.warning .gnu.linkonce.literal.* .gnu.linkonce.t.*.literal .gnu.linkonce.t.* .irom.debug.*)
 
+    /* Constant strings in flash (PSTRs) */
+    *(.irom0.pstr.*)
```
mikee47: This isn't actually necessary since `__fstr__` below catches it, but doesn't hurt.

mikee47:
You can see the latest doc. updates for this PR at https://smingdev.readthedocs.io/en/dev-esp-quick-toolchain/arch/esp8266/getting-started/index.html

slaff:
@mikee47 I am thinking about testing this PR on my devices at the end of the week. Before doing this let me ask you if you are ready with the PR, happy with the results and if you tested it on a real devices?

Sming/Arch/Esp8266/Components/libc/component.mk
avr39-ripe: Hello! I think that stdc++ is worth to keep. If there is any chance to do so. I use std::vector in my code and plan to switch to std:array all plain array. STL itself is good thing if used properly. Also there are libs that depend on STL flatbuffers as example. So do please keep this in Sming if possible.

mikee47: The STL shouldn't need the library as it's all templated code. It was only added in #1734 because of an issue with placeholders when compiling without optimisations.

It's probable that at least some parts of `libstdc++.a` will be required to support exception handling, RTTI, iostream, etc. but at present Sming doesn't use those.

Would you try some test compiles with your code to see if it builds OK without the library? Just remove the `stdc++` entry from `Arch/Esp8266/app.mk`. Post back it turns out there is stuff you do need from there then we can address that.

mikee47: Note: Host code gets build in GDB mode at `-O0` but as we've since updated Windows from GCC 5 to 8.2 it is no longer an issue.

mikee47:
@slaff It's all looking pretty good I think. Re-tested on on nodeMCU, still OK.

With the new toolchain I am expecting builds to be identical between linux/windows, but they're not (quite). If we can assure ourselves that the builds are identical it could save us a lot of testing time in the future. I'll post back once I've got more info.

mikee47:
At present the `mforce-l32` compiler flag is enabled and there's no systematic way to de-select it.

Given these test results, my preference would be to disable it by default:

For HostTests application:

```
                  data    rodata  bss     text    irom0_text
Old toolchain:    1420    6916    27872   29031   471152
no mforce-l32:    1420    6916    27872   28555   439288
New toolchain:    1804    6300    27872   28480   467928
no mforce-l32:    1804    6300    27872   28172   434980

                  Used RAM    Free RAM    Free IRAM
Old toolchain:    36208       45712       3755
no mforce-l32:    36208       45712       4231
New toolchain:    35976       45944       4306
no mforce-l32:    35976       45944       4614
```

mikee47:
Regarding `-mforce-l32`, the framework, libraries and samples do not require it and given the impact on memory consumption/performance I'm failing to see any benefit in using it. I would suggest that we remove it from makefiles completely. This doesn't prevent applications from enabling it for their code if required.

slaff:
> Regarding -mforce-l32, the framework, ... This doesn't prevent applications from enabling it for their code if required.

I was planning to answer you earlier today but yes, I completely agree with you. Just remove the mforce-l32 stuff.

mikee47:
Note that with GCC 9.2 we can no longer filter out the `__PRETTY_FUNCTION__`, `__FUNCTION__` and `__func__` strings into PROGMEM. These are added with `assert()` statements in debug builds, so they'll all go into RAM.

I've done some tests and the issue is also present in Linux GCC 9.2.1. See https://github.com/earlephilhower/newlib-xtensa/issues/5

earlephilhower:
Actually, I have set of patches that I've built or borrowed from other people that I apply to GCC.

If you can find the patch in GCC that started this, I can see if I can do a reverse-patch for gcc9 in the esp-quick-toolchain to undo it.

mikee47:
Re. PROGMEM and __FUNCTION__, etc. I've build the toolchain for 64-bit Linux with a patch incorporated, you can try it out from here https://github.com/mikee47/esp-quick-toolchain/releases/tag/3.0.1-patch

mikee47:
That last commit broke stuff. The libraries in question are `libc.a` and `libgcc.a` which need to be copied from the toolchain and symbol stripped. For example, `memcpy` needs to come out as that gets re-mapped onto `ets_mempcy` in ROM. I'd prefer a way to do this automatically rather than having to take copies.

mikee47:
Various issues now fixed:

1. PROGMEM must retain the alignment flag (STORE_ATTR) which is missing from the newlib version, so keeping our override definition in sys/pgmspace.h. This means we also need to provide PSTR() otherwise __STRINGIZE isn't defined.
2. Couple of gcc integer library functions ended up in IRAM, but they're available in ROM.
3. Another routine isn't in ROM and must be in IRAM.
4. Several floating point routines can be moved into flash - think it's reasonable to ban floating point code from interrupt routines.

The RingTonePlayer sample is a good test for this stuff since it's got more going on inside interrupt routines than most samples.
Tests added to HostTests to confirm all the farting around with flash strings is actually doing what it's supposed to.

Code needs to build now using the updated esp-quick-toolchain release. I've updated the sticky for this PR.

mikee47:
GNU5 pre-release means we can get rid of our libgcc.a copy. No change to build.

slaff:
@mikee47 Are you planning to add more changes to this PR or I should give it a try?

mikee47:
That's it. Everything seems stable now just waiting for a full toolchain release, though that shouldn't stop us!

slaff:
During linking I am having problems with calloc not found. Example: 
```
axtls-8266/compat/lwipr_compat.c:271: undefined reference to `calloc'
```

The linker command is:
```
mcu2: Linking out/Esp8266/debug/build/app_0.out
/opt/esp-quick-toolchain//xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc -Lout/Esp8266/debug/lib -L/path/to/Sming/Sming/out/Esp8266/debug/lib -L/path/to/Sming/Sming/Arch/Esp8266/Compiler/ld -L/path/to/Sming/Sming/Arch/Esp8266/Components/esp8266/ESP8266_NONOS_SDK/lib -Tout/Esp8266/debug/build/rom0.ld -Wl,--gc-sections -Wl,-Map=out/Esp8266/debug/build/app_0.map      -u Cache_Read_Enable_New  -Wl,-wrap,system_restart_local                -nostdlib -u call_user_start -Wl,-static  -Wl,--start-group  /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-Sming-a6e4e4c0e9ba907e2753b51279c9d7f5.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-jsvm.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-sming-arch-cb946bf523dff737a5f81d160e280fea.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-FlashString.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-spiffs-a03a03fea7fef360de03a27d2900eae5.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-http-parser.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-libb64.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-ws_parser.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-mqtt-codec.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-libyuarel.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-axtls-8266-df038c4910b7eb104978b6e3ef84e004.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-MultipartParser.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-libc.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-heap-97ffe5b05891a1ae82299f1c6294b46d.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-esp8266-5c761f91c6184d55e941aba5188d98b3.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-driver-108eee689840b4c0caae13629e7dd02e.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-fatfs.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-spi_flash.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-esp-open-lwip-e78c3e6cc8d0b73199a8dae97f56c284.a /path/to/Sming/Sming/out/Esp8266/debug/lib/clib-jerryscript.a /path2/to/mcu2/out/Esp8266/debug/lib/clib-App-9567291ec0289ce6e429d2349053e313.a -lm -lc -lgcc -lphy -lpp -lnet80211 -lwpa -lcrypto -lhal -lmain-30d55974682da5c04df76fdeb8a464ce -Wl,--end-group -o out/Esp8266/debug/build/app_0.out
```

The Xtensa GCC version is:
```
/opt/esp-quick-toolchain//xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc -v 
Using built-in specs.
COLLECT_GCC=/opt/esp-quick-toolchain//xtensa-lx106-elf/bin/xtensa-lx106-elf-gcc
COLLECT_LTO_WRAPPER=/opt/esp-quick-toolchain/xtensa-lx106-elf/bin/../libexec/gcc/xtensa-lx106-elf/9.2.0/lto-wrapper
Target: xtensa-lx106-elf
Configured with: /workdir/repo/gcc-gnu/configure --prefix=/workdir/xtensa-lx106-elf.x86_64 --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=xtensa-lx106-elf --disable-shared --with-newlib --enable-threads=no --disable-__cxa_atexit --disable-libgomp --disable-libmudflap --disable-nls --disable-multilib --disable-bootstrap --enable-languages=c,c++ --enable-lto --enable-static=yes --disable-libstdcxx-verbose
Thread model: single
gcc version 9.2.0 (GCC) 
```

What am I doing wrong?

slaff:
> What am I doing wrong?

I think I have found the issue: axTLS and umm_malloc.  You can reproduce the problem using the following set of commands 

```
cd $SMING_HOME/../samples/Basic_Ssl
make dist-clean
make ENABLE_CUSTOM_HEAP=1
```

Custom heap and axTLS don't play nice together with this PR. The problem is not reproducible in `develop` with the `old` 4.8 gcc toolchain.

slaff:
@mikee47  In addition it seems that you have done great job improving also the memory usage. It looks like the new toolchain not only provides new C++ standards but also decreases the RAM and IRAM usage giving us additional room to expand our Sming applications.

|         | Old           | New  | Free
| ------------- |:-------------:| -----:|-----:|
| RAM      | 46604 | 46964 | +360 |
| IRAM      | 5057      |   6230 | +1173 !!! |



# NEW

```
mcu2: Linking out/Esp8266/debug/build/app_0.out
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8680|    1664
    rodata|           ReadOnly Data (RAM)|    3FFE8680|    3FFE961C|    3996
       bss|      Uninitialized Data (RAM)|    3FFE9620|    3FFF0890|   29296
      text|            Cached Code (IRAM)|    40100000|    401067BC|   26556
irom0_text|           Uncached Code (SPI)|    40202010|    402734E8|  464088
Total Used RAM : 34956
Free RAM : 46964
Free IRam : 6230
```


## OLD
```
mcu2: Linking out/Esp8266/debug/build/app_0.out
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE8580|    1408
    rodata|           ReadOnly Data (RAM)|    3FFE8580|    3FFE975C|    4572
       bss|      Uninitialized Data (RAM)|    3FFE9760|    3FFF09F8|   29336
      text|            Cached Code (IRAM)|    40100000|    40106C51|   27729
irom0_text|           Uncached Code (SPI)|    40202010|    4027CA10|  502272
Total Used RAM : 35316
Free RAM : 46604
Free IRam : 5057
ESPTOOL2 out/Esp8266/debug/firmware/rom0.bin
```

slaff:
> Custom heap and axTLS don't play nice together with this PR. The problem is not reproducible in develop with the old 4.8 gcc toolchain.

@mikee47 Would you have time to look at this issue or you would prefer to have this PR merged and address the issue in a separate PR?

mikee47:
`calloc` is one of those functions which isn't used a great deal - only within axtls it seems!

When I built Basic_Ssl as you indicated I got a conflict error with `ICACHE_FLASH_ATTR` (now fixed) - hopefully you got that as well otherwise there's something really weird going on!

slaff:
> hopefully you got that

@mikee47 Thanks a lot! Basic_Ssl with Umm Malloc compiles now as expected.

mikee47:
Have you tested out GDB with the new toolchain? It includes the un-patched version which sends more stuff over the wire, but should not make any difference to how its used.

.travis.yml
```diff
@@ -23,10 +23,23 @@ jobs:
...
+
   - stage: build
     os: linux
-    env: SMING_ARCH=Esp8266 SDK_VERSION=3.0.1
+    env:
+    - SMING_ARCH=Esp8266
+    - SDK_VERSION=3.0.1
+    - ESP_HOME=$TRAVIS_BUILD_DIR/opt/esp-alt-sdk
+
+  - stage: build c++17
```
slaff: @mikee47 Can you try to use `build` as a stage name also here? It would be better to have the different builds running in parallel and not wait for `build` to finish in order for `build c++17` to start. I am not sure if travis will merge them under on stage and run them in parallel but it is worth trying.

mikee47:
I've attempted to test the custom heap on hardware, but it appears to be broken, but not by this PR. Investigating....

mikee47:
I've attempted to test the custom heap on hardware, but it appears to be broken... Investigating....

mikee47:
I've reverted to #1745 and still cannot get even Basic_Blink to work using custom heap. Very odd.

slaff:
> still cannot get even Basic_Blink to work using custom heap. Very odd.

```
make dist-clean
cd ../samples/Basic_Blink
make config-clean # !!!
make ENABLE_CUSTOM_HEAP=1 
```
I will try it too...

slaff:
> I will try it too...

Hm... I have compiled Basic_Blink and Basic_Serial with ENABLE_CUSTOM_HEAP=1  and both are just not starting for me. I can see that the rBoot is up and running but the application itself does not get started at all. And to make things worse there is no error message.

mikee47:
Same here. I have two thoughts:

1. UMM Malloc isn't getting initialised, perhaps some static data isn't zeroed out at startup (i.e. called before BSS is initialised)
2. Some code may be in the wrong section. I've whacked IRAM_ATTR on all the umm malloc code, no difference.

mikee47:
What's really bugging me is that it the original commit (1745) which was supposed to fix the custom heap doesn't work any more, which might support (1) above.

slaff:
Ok, add the parallel build change. I will merge this PR and then you can open a new PR related to umm_malloc.

PR #1829, opened 2019-09-05
Place IRAM and FLASH data in separate segments

mikee47:
Resolves section conflict issues when attempting to inline header methods, generally more of a problem with GCC 4.8.

Thanks to esp8266/Arduino#5116


slaff:
Did you test this change?

mikee47:
Yes, been running on real hardware without issue. Also benefits IRAM/flash usage as linker can discard unused bits.

(Note my initial PR failed because I pulled out the original commit for this, which I'd subsequently updated.)

PR #1830, opened 2019-09-11
Add `hw_timer` driver and clarify use of Timer1/Timer2

mikee47:
**Add `hw_timer` driver** and clarify distinction between Timer1 and Timer2.
    
PR #1810 modified the calculations for `usToTimerTicks()` and `timerTicksToUs()`,
but these are only applicable to Timer2 (which is a 32-bit up-counter).

HardwareTimer uses Timer1 (a 23-bit down-counter) so the conversion functions
are now static methods, now using a `muldiv()` function so times are rounded
rather than truncated for better accuracy. Overflows are also handled in a more
consistent fashion by using the maximal value rather than the modulus.

Both timers default to /16 prescale, but are independently configurable.
Note, however, that the Timer2 prescale must not be set directly as this drives
software timers which are used internally by the SDK. Instead, this is set using a
call to `system_timer_reinit`, which has been moved into the driver.
Startup code now calls `hw_timer_init` instead to ensure a consistent state.

HardwareTimer methods are now mainly in the header so the compiler can better
optimise calls. A `state` variable also improves speed as RAM accesses are faster
than I/O. The interval is stored as a tick value, rather than the time
in microseconds, to avoid un-necessary recalculations. This makes calls to restart
faster. A `setInterval` method has been added so pre-calculated tick values can
be set directly.

Rename `MAX_HW_TIMER1_INTERVAL_US` to `MAX_HW_TIMER1_INTERVAL` as it in ticks not microseconds.

Host emulation of timers now implemented in an improved `hw_timer` driver.

Added basic runtime support for rational number handling to complement std::ratio.

**Update documentation**
    
Add interrupts and timers pages
Add notes to upgrading page

**Restructure HostTests** using task queue, add rational number tests


slaff:
> 've no idea what codacy is upset about.

I've looked at the reported issues and will mark them as false-positive for now. Is the this PR tested on a real device and in the emulator :) ? Do you want to add something before merging it?

slaff:
> Update documentation

And thanks a lot for the new and very well explained stuff. I like it a lot.

mikee47:
Tested with LiveDebug, Servo samples, don't appear to be any others which use the HardwareTimer. I'll be adding loads more tests for this in the next PR.
I've reverted the previous change to common.h since it doesn't fix the codacy issue.
Good to go.

PR #1831, opened 2019-09-13
Unified Callback Timer API

mikee47:
**Add `CallbackTimer` class template** to support HardwareTimer, SimpleTimer and Timer.
    
- Contains common logic and checks for all callback timer types
- Templated for best performance (no VMT)
- Added `TimerCallback` supporting void* arg parameter (in addition to existing `InterruptCallback`)
- Templated methods added with compile-time checks on interval (so code won't compile if out of range)
- Added methods to support setting/checking/querying intervals directly in timer ticks
- Timer intervals stored internally in timer ticks, so querying the value confirms the actual time in use accounting for rounding, etc.

Note that delegate callbacks are supported only by the Timer class, which now also has an AutoDelete variant.

**Improve Host timer queue implementation** to handles multiple timers properly, fixes bug where all timers get cancelled instead of just one.

**Add `os_timer_arm_ticks()` function** so software timers are programmed in ticks instead of microseconds or milliseconds

- Used instead of `os_timer_arm_us` provides simpler and more flexible timer interface.
- Avoids un-necessary (and potentially inaccurate) time conversions
- Timers can be used with `USE_US_TIMER=0` for 3.2us tick resolution, providing basic range of 1'54" (with SimpleTimer)
- Default is still `USE_US_TIMER=1` for 0.2us tick resolution and reduced 0'7"9s range (without using Timer class)

**Update samples** to use improved timer API and add timers module to HostTests

**Update linker script** to apply IRAM section attributes for templated code

Templates ignore section attributes so requires entries in linker script for IRAM code.


slaff:
@mikee47 Tested and happy with this PR?

mikee47:
@slaff Yes, tested and happy :-)

PR #1835, opened 2019-09-17
Avoid using `WiringFrameworkDependencies.h` within framework

mikee47:
Too many low-level dependencies when using things like `String`, `Vector`, etc.


mikee47:
@slaff Nope, good to go thanks.

PR #1838, opened 2019-09-18
Fix issue with undefined millis() / micros()

### LINKED ISSUE #1837: Undefined `micros`
Some code crashes when attempting to use `micros()` because the linker generates a null value.

This bug was introduced by #1835. Adding `-u millis -u micros` to the linker fixes this problem, but then travis fails to build `HttpServer_ConfigNetwork` with SDK version 3:

```
out/Esp8266/debug/lib/libmain-rboot.a(mem_manager.o): In function `malloc':
(.text.malloc+0x8): multiple definition of `malloc'
/home/travis/build/SmingHub/Sming/Sming/Arch/Esp8266/Compiler/lib/libmicroc.a(malloc.o):C:\DEV\ESP8266\esp_microc/malloc/malloc.c:8: first defined here
```
This repeats for free, zalloc, calloc and realloc.

Clearly there is an underlying instability in how this is all linked. I'm looking to find a proper solution so the build is more robust, but any ideas most welcome.

mikee47:
Don't weaken millis() or micros(), and remove conflicting `time.o` functions from libmain.

Fixes #1837.

### DIFF
--- a/Sming/Components/rboot/component.mk
+++ b/Sming/Components/rboot/component.mk
@@ -119,6 +119,7 @@ COMPONENT_LDFLAGS		:= -u Cache_Read_Enable_New
 $(LIBMAIN_RBOOT_DST): $(LIBMAIN_SRC)
 	@echo "OC $@"
 	$(Q) $(OBJCOPY) -W Cache_Read_Enable_New $^ $@
+	$(Q) $(AR) d $@ time.o
 
 ifeq ($(RBOOT_BIG_FLASH),1)
 LIBMAIN					:= $(LIBMAIN_RBOOT)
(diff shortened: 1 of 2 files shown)


slaff:
> remove conflicting time.o

Please, check if we are also removing functions from time.o that are used elsewhere in the code.

mikee47:
If that were the case, samples wouldn't compile...

I've listed removed functions below. Note that SDK 1.5 doesn't have these at all, in any of the provided standard libraries.

The main intent appears to be sntp support. AFAIK Sming only uses these within axtls-8266 (gettimeofday, etc.), providing its own NTP support elsewhere. Note also that esp8266/arduino does its own thing, implementing versions in a core `time.cpp` file.

```
V2:

00000208 W _gettimeofday_r
00000138 T configTime
00000244 T getrealtimeofday
000000d0 T micros
00000094 T millis
00000024 T sntp_get_rtc_time
00000104 T sntp_update_rtc
000001dc T time
000000e8 T updateTime


V3:
000001e4 W _gettimeofday_r
00000138 T configTime
0000022c T getrealtimeofday
000000d0 T micros
00000094 T millis
00000024 T sntp_get_rtc_time
0000003c T sntp_update_rtc
000000e8 T updateTime

```

PR #1846, opened 2019-09-27
Host fixes plus default host serial port keyboard input

mikee47:
Fix `system_os_post()` return value
Fix bug when disarming timer at head of list (disables all timers).
Redirect keyboard input to UART0 if no telnet emulation specified.


slaff:
@mikee47 Do you have something to add here before I merge this PR?

mikee47:
Happy with this thanks.

PR #1848, opened 2019-09-27
Add `ensureCapacity()` method to `MemoryDataStream`

mikee47:
Revise internally to use offsets instead of pointers
Change member size variables from `int` to `size_t`


slaff:
@mikee47 Do you have something to add here before I merge this PR?

mikee47:
All good thanks.

PR #1850, opened 2019-09-28
Update interrupt functions

mikee47:
Updates to Interrupts module
    
* Use InterruptDelegate type
* Put delegate callback into separate function to save IRAM
* Use `ESP_MAX_INTERRUPTS` consistently

Refactor attachInterrupt code using inlines.
Be clear with distinction between ESP *GPIO_INT_TYPE* and Arduino interrupt *mode*.

Update `Basic_Interrupts` sample. Use function instead of lambda and explain reason in comments.

### DIFF
--- a/samples/Basic_Interrupts/app/application.cpp
+++ b/samples/Basic_Interrupts/app/application.cpp
@@ -10,6 +10,21 @@
 #define say(a) (Serial.print(a))
 #define newline() (Serial.println())
 
+static unsigned interruptToggleCount;
+
+void showInterruptToggleCount(uint32_t toggleCount)
+{
+	say("Toggle count hit ");
+	say(toggleCount);
+	say(", current value is ");
+	say(interruptToggleCount);
+	say("!");
+	newline();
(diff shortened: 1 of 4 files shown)


samples/Basic_Interrupts/app/application.cpp
```diff
@@ -26,29 +41,29 @@ void IRAM_ATTR interruptHandler()
...
-	++toggleCount;
-	if(toggleCount > MAX_TOGGLE_COUNTS) {
+	++interruptToggleCount;
+	if(interruptToggleCount > MAX_TOGGLE_COUNTS) {
+		System.queueCallback(showInterruptToggleCount, interruptToggleCount);
 		/*
-		 * Using a 'lambda function' we can easily include any code to be executed by the callback
-		 * without having to define a function elsewhere.
-		 * You could also pass a function pointer to queueCallback().
+		 * Note that `queueCallback` also supports std:;function arguments, so we can use lambdas,
```
slaff: Small typo: replace `std:;function` with `std::function`

PR #1851, opened 2019-09-28
Add delegate support to task queue

mikee47:
**Add `ENABLE_TASK_COUNT`** to control task counter, disable by default.
    
When stressed with periodic interrupts, the only way to get a reliable count
is to disable interrupts whilst updating the task counts.
As this is a diagnostic feature, leaving it disabled by default seems appropriate.

Fortunately, `noInterrupts()` already returns the existing interrupt level,
so we just need to add `restoreInterrupts()` so we can safely use it
from interrupt context.

**Fix 'ready' state tracking**
    
If no `onReady` handlers are set then `state` will never be set to ready.
Use the original `init_done_cb` callback to handle this separately.
Note: this gets called first before any queued task callbacks.

**Change standard TaskCallback parameter** to `void*`
    
Makes it easier to use and provides consistency with timer callbacks.
Avoids casting in many cases, and only requires `static_cast` for pointers.
Add `InterruptCallback` overload.
Note: Retaining existing `uint32_t` parameter overload.

**Add Delegate support** for System::queueCallback()

**Expand documentation** with notes on memory, multitasking, tasks


docs/source/information/tasks.rst
```diff
@@ -0,0 +1,20 @@
+Tasks
+=====
+
+Introduction
+------------
+
+If you need to perform any kind of background processing *task*, then you will need
+to consider how it to break it up so it executes in small slices to allow other
```
slaff: `consider how it to break it up`. Did you mean `consider how to break it up` ?

Issue #1841, opened 2019-09-19
build error: duplicate swap() 


etmmahi:
When building an application with the foolowing libs in  components.mk:
```ARDUINO_LIBRARIES := ArduinoJson6 Adafruit_SSD1306```

then the following code fails:
```
#include <SmingCore.h>
#include <Libraries/Adafruit_SSD1306/Adafruit_SSD1306.h>
#include <ArduinoJson.h>
```
with:
```
In file included from /home/martijn/espdev/Sming/Sming/Libraries/Adafruit_SSD1306/Adafruit_SSD1306.h:49:0,
                 from /home/martijn/espdev/projects3/bug01/app/application.cpp:3:
/home/martijn/espdev/Sming/Sming/Libraries/Adafruit_GFX/Adafruit_GFX.h:11:20: error: expected unqualified-id before '{' token
 #define swap(a, b) { int16_t t = a; a = b; b = t; }
                    ^
In file included from /home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/ArduinoJson/src/ArduinoJson/MsgPack/endianess.hpp:8:0,
                 from /home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/include/../ArduinoJson/src/ArduinoJson/MsgPack/MsgPackDeserializer.hpp:11,
                 from /home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/include/../ArduinoJson/src/ArduinoJson.hpp:36,
                 from /home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/include/../ArduinoJson/src/ArduinoJson.h:9,
                 from /home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/include/ArduinoJson.h:38,
                 from /home/martijn/espdev/projects3/bug01/app/application.cpp:4:
/home/martijn/espdev/Sming/Sming/Libraries/ArduinoJson6/ArduinoJson/src/ArduinoJson/Polyfills/utility.hpp:9:30: error: expected unqualified-id before '{' token
 inline void swap(T& a, T& b) {
                              ^
/home/martijn/espdev/Sming/Sming/component-wrapper.mk:155: recipe for target 'app/application.o' failed
```

It seems the swap() define in Adafruit_GFX.h is a (bad) duplicate.
Change the define into an inline function inside the Adafruit_GFX class fixes the problem.
```
[...]

PR #1855, opened 2019-09-30
Adafruit_GFX: Move swap macro to cpp file

etmmahi:
Fix for #1841


etmmahi:
There may be better fixes, but I tried to keep it as local as possible and follow what the original Adafuit_GFX did.  
Better fix would imply a more global fix including other components.
The same original swap style macro is also used in Adafruit_SSD1306 and ArduCam

Any suggestions?

Martijn

slaff:
Fixed in #1856.

PR #1852, opened 2019-09-29
Add experimental SSL to host emulator using axtls-8266

mikee47:

PR #1858, opened 2019-09-30
Fixes to SSL when working with sites sending 16K chunks of data.

slaff:
- Fixes missing debug information from axTLS when an application was compiled with SSL_DEBUG=1.
- Fixes for memory corruption on buffer resizing.
- Added message to explain why a condition has occurred.


mikee47:
The issue I raised in #1852 seems to have disappeared (probably unrelated). I've propsed #1861 to help test this sort of issue.

PR #1864, opened 2019-10-02
RingTone Player sample plus libraries

mikee47:
Used to give the I2S driver, etc. more of a thorough test as a networked application.
Includes graphs to evaluate system loading.


mikee47:
Note that this sample won't flash correctly as I've set SPIFF_SIZE too low, so use `make SPIFF_SIZE=0x40000`. I'll fix this in an update.

PR #1869, opened 2019-10-04
Simplify and refactor APA102 sample

mikee47:
Use the `col_t` structure instead of byte arrays, eliminate code duplication.
Sample refactored using state machine instead of long delays, WDT reset removed.

Untested on APA102 hardware but compiles and runs fine.


slaff:
Looks good to me. Let me check if there is someone that can test it on a real hardware.

PR #1870, opened 2019-10-04
Update servo library and sample

mikee47:
Move global variables into class as statics, minor updates (risk of symbol collision)
Stop hardware timer before clearing `started` flag and remove assert statement from ISR.
Pre-calculate times in clock ticks to avoid calculation overhead inside ISR


Sming/Libraries/Servo/Servo.cpp
```diff
@@ -7,79 +7,77 @@
...
-	if (channel_count > SERVO_CHANNEL_NUM_MAX) return false;
+	if(channel_count >= SERVO_CHANNEL_NUM_MAX) {
+		return false;
+	}
 	channels.add(channel);
 
 	ETS_INTR_LOCK();
 	getPins();
 	calcTiming();
 	ETS_INTR_UNLOCK();
```
mikee47: I don't like how this library globally disables interrupts whilst re-calculating values, which can take a (relatively) long time.
The solution which occurs to me is:
1. Double-buffer state information so interrupts only need to be disabled during a pointer swap operation;
2. Introduce an enable/disable Interrupts() method into HardwareTimer.

 I'm inclined to stick a pin in this for now and deal with this separately.

slaff:
@mikee47 Did you have the change to test with real hardware?

mikee47:
I checked it the output on a scope. That was a wee while ago though so I'll redo it. Just spotted some more stuff in the ServoChannel files as well. I'll ping you when I'm done.

slaff:
> I checked it the output on a scope. 

Wow - professional stuff :)

mikee47:
LOL. Most useful bit of kit I ever  bought; without it just poking around in the dark!

Sming/Libraries/Servo/ServoChannel.h
```diff
@@ -9,83 +9,90 @@
...
      */
 	uint32 getValue() const;
 
-    /** @brief  set value for output
+	/** @brief  set value for output
      *  @param  value time in us for actual output of this channel\n\r
      *  has to be smaller than max-min value
      *  @retval bool True on success
      */
 	bool setValue(uint32 value);
```
mikee47: I find this approach highly confusing. The logical expectation is that `minValue <= value <= maxValue` (a).
In reality it's `0 <= value <= (maxValue - minValue)` (b).

My preference would be to amend the API to define value as for (a), but that will break things.

The other approach would be to rename `minValue` and `maxValue`.

Thoughts?

slaff: > My preference would be to amend the API

Go for it. As far as I can see the impact will be relatively small within Sming. That will require good documentation too.

mikee47: OK, I've amended the API and revised the library so interrupts remain enabled. Scope output solid and servos work fine.

Sming/Libraries/Servo/Servo.cpp
mikee47: Done.

slaff:
> Scope output solid and servos work fine.

@mikee47  Something left before I merge it?

mikee47:
@slaff I've added a note on upgrade, that's it thanks.

PR #1878, opened 2019-10-09
Preparation for release 4.0.0 rc2

slaff:

PR #1881, opened 2019-10-11
Move Sming version information into `SmingVersion.h`

mikee47:
Define version components separately and include pre-release tag.


mikee47:
For release builds we have two options:

`#define SMING_PRE_RELEASE` (a)
or
`#define SMING_PRE_RELEASE ""` (b)

I'm thinking (a).

mikee47:
I'll re-submit to #1878

Issue #1882, opened 2019-10-12
Syntax Error


frankdownunder:
I use a newer compiler , but it is an easy one to fix. . . 

C+ /opt/Sming/Sming/Arch/Host/Components/hostlib/startup.cpp
In file included from /opt/Sming/Sming/Platform/Clocks.h:15,
                 from /opt/Sming/Sming/Core/PolledTimer.h:28,
                 from /opt/Sming/Sming/Platform/Timers.h:23,
                 from /opt/Sming/Sming/Arch/Host/Components/hostlib/startup.cpp:36:
/opt/Sming/Sming/Core/NanoTime.h: In static member function ‘static String NanoTime::TicksConst<Clock_, ticks_>::toString()’:
/opt/Sming/Sming/Core/NanoTime.h:608:10: error: could not convert ‘ticks_’ from ‘long long unsigned int’ to ‘String’
   return ticks_;
          ^~~~~~
SHOULD AVOID THE IMPLICIT CAST
   return String(ticks_);

make[1]: *** [/opt/Sming/Sming/component-wrapper.mk:156: startup.o] Error 1
make: *** [/opt/Sming/Sming/project.mk:359: hostlib-build] Error 2

slaff:
Should be fixed with PR #1883

Issue #1844, opened 2019-09-27
RC releases for Sming 4.0


slaff:
We will start Release Candidate (RC) releases before we officially announce version 4.0. The plan for the moment is to have the following RCs
- [x] RC1: latest code from develop. The release tag will be set in the `develop` branch. 
      Expected release date - 29th of September. 
- [x] RC2: Leave support only for SDK >= 3.0.1. Due to the security issues found recently the older SDKs will not be supported officially. 
     Expected release date - 15th of October.
- [x] RC3: Feature freeze and bug fixes only.

Expected release date for Sming 4.0.0 is beginning of November, 2019.

teicors:
Hi, just updated to latest version and in compiling my code, I receive this error:
/opt/sming/Sming/out/Esp8266/debug/lib/clib-ArduinoJson5.a(JsonObjectStream.o):(.rodata._ZTV16JsonObjectStream[vtable for JsonObjectStream]+0x38): undefined reference to `MemoryDataStream::seek(int)'

The json version is 5 into component.mk

slaff:
@mikee47 Can you take a look at that issue?

teicors:
after a make dtst-clean the configuration of the "project" is reset

/opt/sming/Sming/Arch/Esp8266/Components/esptool/esptool/esptool.py -p  -b 115200 write_flash -ff 40m -fm qio -fs 512KB 0x00000 out/Esp8266/debug/firmware/rboot.bin 0x002000 out/Esp8266/debug/firmware/rom0.bin 0x100000 out/Esp8266/debug/firmware/spiff_rom.bin

How to catch the error:I

> make SMING_ARCH=Esp8266 SPI_MODE=dio SPI_SIZE=4M COM_SPEED_ESPTOOL=1000000

[compiling ok]

> make dist-clean

> make flash

the error appers and the only way to flash the esp is  a

> make SMING_ARCH=Esp8266 SPI_MODE=dio SPI_SIZE=4M COM_SPEED_ESPTOOL=1000000

mikee47:
@teicors Can you reproduce the problem by modifying one of the sample applications? That way I can reproduce the error.

mikee47:
> after a make dtst-clean the configuration of the "project" is reset

Yes, that's by design. You can perform a clean build without resetting config like this:

```
make components-clean clean
make
```

mikee47:
@teicors Something to check. Do `git status` and make sure there aren't any old files from the previous build - those should be deleted.

[...]

Issue #1241, opened 2017-09-09
Windows QuickStart fails


fredimartins:
> C:\WINDOWS\system32>choco install sming -y -source 'https://www.myget.org/f/sming/'
> Chocolatey v0.10.8
> Installing the following packages:
> sming
> By installing you accept licenses for the packages.
> sming not installed. An error occurred during installation:
>  Unable to resolve dependency 'jre8 '.
> sming package files install completed. Performing other installation steps.
> The install of sming was NOT successful.
> sming not installed. An error occurred during installation:
>  Unable to resolve dependency 'jre8 '.
> 
> Chocolatey installed 0/1 packages. 1 packages failed.
>  See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
> 
> Failures
>  - sming (exited 1) - sming not installed. An error occurred during installation:
>  Unable to resolve dependency 'jre8 '.

kwis2:
with choco install -y sming on fresh system I got

> Chocolatey v0.10.8
> Installing the following packages:
> sming
> By installing you accept licenses for the packages.
> sming not installed. An error occurred during installation:
>  Unable to resolve dependency 'spiffy '.
> sming package files install completed. Performing other installation steps.
> The install of sming was NOT successful.
> sming not installed. An error occurred during installation:
>  Unable to resolve dependency 'spiffy '.
> 
> Chocolatey installed 0/1 packages. 1 packages failed.
>  See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).
> 
> Failures
>  - sming (exited 1) - sming not installed. An error occurred during installation
> :
>  Unable to resolve dependency 'spiffy '.
> 

and in log file 

> 2017-09-16 21:07:23,528 2976 [DEBUG] - _ Chocolatey:ChocolateyInstallCommand - Normal Run Mode _
> 2017-09-16 21:07:23,538 2976 [INFO ] - Installing the following packages:
> 2017-09-16 21:07:23,538 2976 [INFO ] - sming
> 2017-09-16 21:07:23,538 2976 [INFO ] - By installing you accept licenses for the packages.
> 2017-09-16 21:07:25,130 2976 [INFO ] - [NuGet] Attempting to resolve dependency 'eclipse-cpp '.
> 2017-09-16 21:07:25,300 2976 [INFO ] - [NuGet] Attempting to resolve dependency 'jre8 '.
[...]

PR #1883, opened 2019-10-12
Update windows manual installation instructions

mikee47:
Further to https://github.com/SmingHub/Sming/issues/1844#issuecomment-541169569 and #1241 I thought a good first step was to review the manual Windows installation process. I for one prefer the control a manual installation gives me.

I've rewritten the instructions and un-deprecated them.

* MinGW with GCC version 8.2.0 is installed. Note this is the original 32-bit branch, NOT the 64-bit port.
* Eclipse has advanced considerably from 'Luna' so references current version.
* Includes instructions for GIT installation.
* Code blocks now have a small 'copy' button to make it easier to use.

Note that the `EspressifESP8266DevKitAddon` is not required - this script can be replaced entirely with one line:

    mingw-get install mingw32-base-bin mingw-developer-toolkit-bin mingw32-gcc-g++-bin mingw32-pthreads-w32-dev mingw32-libmingwex

Also couple of minor code fixes, addressing #1882 which pops up as a result of advancing GCC to version 8.2.0.

I've tried this out on a Windows virtualbox installation, pretty straightforward I think.

### DIFF
--- a/Sming/Arch/Host/README.rst
+++ b/Sming/Arch/Host/README.rst
@@ -25,14 +25,8 @@ Ensure you are using relatively recent compilers, with 32-bit libraries availabl
 For Linux, you may require the ``gcc-multilib`` and ``g++-multilib``
 packages to build 32-bit executables on a 64-bit OS.
 
-For Windows, make sure your ``MinGW`` distro is up to date. If you run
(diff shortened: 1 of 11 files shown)


mikee47:
Note: Browseable version here https://smingdev.readthedocs.io/en/latest/arch/esp8266/getting-started/windows-manual.html

PR #1885, opened 2019-10-13
Update ArduinoJson6 to current version 12.0

mikee47:


Sming/Libraries/ArduinoJson6/component.mk
```diff
@@ -1 +1,2 @@
 COMPONENT_SUBMODULES := ArduinoJson
+COMPONENT_INCDIRS += ArduinoJson/src
```
slaff: Shouldn't this be `:=` instead of `+=` ?

mikee47: It's equivalent to `COMPONENT_INCDIRS := include ArduinoJson/src`. I can change it if you prefer?

slaff: Yes, please.

PR #1892, opened 2019-10-17
Handle case where WIFI_PWD is empty (open networks)

### LINKED ISSUE #1891: OpenWifi not working
Try example Basic_Wifi  with settings:

export WIFI_SSID="OpenNetwork"
export WIFI_PWD=""

The compile will fail because WIFI_PWD is not taken into the compile flags.
This happens due to the 
`ifdef WIFI_PWD 
`
check in component.mk

pse fix

Martijn

mikee47:
ifdef ignores empty values, must check if it's undefined.

Fixes #1891

### DIFF
--- a/Sming/component.mk
+++ b/Sming/component.mk
@@ -64,7 +64,7 @@ CONFIG_VARS				+= WIFI_SSID WIFI_PWD
 ifdef WIFI_SSID
 	APP_CFLAGS			+= -DWIFI_SSID=\"$(WIFI_SSID)\"
 endif
-ifdef WIFI_PWD
+ifneq ($(origin WIFI_PWD),undefined)
 	APP_CFLAGS			+= -DWIFI_PWD=\"$(WIFI_PWD)\"
 endif
 


slaff:
Please, leave only  commit  65ac1d1 in this PR.

mikee47:
One caveat to this PR. These are fine:

```
make WIFI_SSID=PrivateNetwork WIFI_PWD=reallySecurePassword
make WIFI_SSID=OpenNetwork WIFI_PWD=
```

If you then do this:

```
make WIFI_SSID=
```

It fails.

```
C+ /s/sandboxes/sming-dev/samples/Basic_WiFi/app/application.cpp
s:/sandboxes/sming-dev/samples/Basic_WiFi/app/application.cpp:6:0: error: "WIFI_PWD" redefined [-Werror]
 #define WIFI_PWD "PleaseEnterPass"
 ^
<command-line>:0:0: note: this is the location of the previous definition
cc1plus.exe: all warnings being treated as errors
```

This is in loads of samples. I'm not sure this requires a fix necessarily since I cannot think of any reason why one would do this.

```
#ifndef WIFI_SSID
#define WIFI_SSID "PleaseEnterSSID" // Put you SSID and Password here
#define WIFI_PWD "PleaseEnterPass"
#endif
```

slaff:
The SSID cannot be empty. Thus `make WIFI_SSID=`  is not correct and should fail. 

> I'm not sure this requires a fix

Nothing needs to be fixed here.

Issue #751, opened 2016-07-04
Using <cassert> doesn't work with Sming


cmarrin:
I'm using assert() by including <cassert>. This causes an undefined reference to __assert_func. The replacement from Sming is in esp_systemapi.h. I want to use the standard include because my app is cross-platform (esp/mac). I've solved this problem by including the following snippet in my app.cpp file:

```
void __assert_func(const char *file, int line, const char *func, const char *what)
{
    SYSTEM_ERROR("Assertion failed: (%s), function %s, file %s, line %d", what, func, file, line);
    while (1) ;
}
```

but it seems like Sming should support this in the core

slaff:
@cmarrin If you think that can help other developers please make a Pull Request and we will review it.

PR #1895, opened 2019-10-18
Use library `assert()` instead of redefined macro

mikee47:
See also #751


slaff:
We have to be careful with this PR because from now on the asserts will stop the abort the execution. The old behaviour was only to report failed assert and continue the execution.

mikee47:
It should help to focus where assert() should be used, i.e. logical coding errors rather than runtime issues.

PR #1896, opened 2019-10-19
Add `BasicHttpHeaders` class for in-place parsing

mikee47:
Used for fast processing of SSDP packets, for example.


mikee47:
I'll close this and re-submit with the UPnP stack.

Issue #1901, opened 2019-10-21
error: "WIFI_PWD" redefined [-Werror]


teicors:
Hi,
I updated sming to the latest version by git branch develop and I catch the error.
I did the following steps:

```
cd samples/HttpClient
make
make component-clean clean
make

```
and now I catch the error. 
the HttpClient is vanilla from git clone <Sming>. No one modification.
The same error on my projects.

mikee47:
See #1892. The fix is to ensure WIFI_SSID is defined. The error message could be clearer but in essence it's appropriate  I think.

The use of these variables is really only for basic samples and one-off applications. In deployable devices a more sophisticated mechanism for managing connection information is needed.

teicors:
Yes, I know. I put for my gadgets the ssid and password in an external file with the right value. The error is not related to a password empty, but to the duplicate (redefine) of password. In the example I expect an error like "SSID not defined" and not redefine of password.

slaff:
@mikee47 The following change can be done: In Sming/component.mk replace the setting of WIFI_PASS from:
```
ifdef WIFI_SSID
	APP_CFLAGS			+= -DWIFI_SSID=\"$(WIFI_SSID)\"
endif
ifneq ($(origin WIFI_PWD),undefined)
	APP_CFLAGS			+= -DWIFI_PWD=\"$(WIFI_PWD)\"
endif
```

to 

```

**Comment by teicors** [comment]
Not exactly.. an undefined WIFI SSID is not catched.
Now I don't report any duplicate, neither the WIFI SSID voluntary missed as verify the solution.

**Comment by mikee47** [comment]
@slaff That's a good idea, thanks.

@teicors 
> undefined WIFI SSID is not catched.

We can fix that for all the samples:

```
#ifndef WIFI_SSID
static_assert(false, "Please define WIFI_SSID and, if required, WIFI_PWD.");
#define WIFI_SSID "WIFI_SSID"
#endif
```

PR #1902, opened 2019-10-21
Provide better handling for undefined WIFI_SSID

mikee47:
Following from #1891, and addressing #1901, this PR aims to improve the handling of undefined WIFI_SSID in sample applications.

The following use cases are accounted for:

Undefined WIFI_SSID: Display a suitable error message (WIFI_PWD
Open network: WIFI_SSID defined, WIFI_PWD undefined
Secure network: Both WIFI_SSID and WIFI_PWD defined

Note that in all cases "" are not required when defining these values.

### DIFF
--- a/Sming/component.mk
+++ b/Sming/component.mk
@@ -64,9 +64,7 @@ CONFIG_VARS				+= WIFI_SSID WIFI_PWD
 ifdef WIFI_SSID
 	APP_CFLAGS			+= -DWIFI_SSID=\"$(WIFI_SSID)\"
 endif
-ifneq ($(origin WIFI_PWD),undefined)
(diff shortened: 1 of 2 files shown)


mikee47:
I've only modified the HttpClient sample. If this solution is acceptable then I'll go ahead and update the other 23 samples.

mikee47:
@slaff What are your thoughts on this approach? Shall I update the other samples?

slaff:
@mikee47 I would prefer the following something that is as close to the current expectations as possible. 

> Undefined WIFI_SSID: Display a suitable error message (WIFI_PWD

Keep the samples as before. Update the component.mk to set WIFI_PWD to empty string when WIFI_SSID is set and WIFI_PWD is not set. 

We can add later a new PR that forces the user to provide WIFI_SSID. But not now.

PR #1907, opened 2019-10-23
GCC 8.2 throws range warning attempting to copy out of FlashString.

mikee47:
To fix this requires re-defining the FlashString structure with a maximum-length array.
I've also added a private constructor to guard against inadvertently trying to instantiate this.

This change breaks the sample application which shows how to cast custom structures into a FlashString,
so a couple of new macros have been added to provide a cleaner way of doing this:
FSTR_STRUCT_PTR
DEFINE_FSTR_REF

The specific error reported is:

```
IMPORT_FSTR(fstr_master_key, PROJECT_DIR "/master.key");

memcpy_P(&master_key, fstr_master_key.data(), fstr_master_key.length());
^^
```

> r:/sming-ifs/Sming/Wiring/FakePgmSpace.h:157:40: error: 'void* memcpy(void*, const void*, size_t)' forming offset [5, 20] is out of the bounds [0, 4] of object 'fstr_master_key' with type 'const FlashString' [-Werror=array-bounds]

### DIFF
--- a/Sming/Wiring/FlashString.h
+++ b/Sming/Wiring/FlashString.h
@@ -95,7 +95,7 @@
  */
 #define DEFINE_FSTR(_name, _str)                                                                                       \
 	DEFINE_FSTR_STRUCT(_##_name, _str);                                                                                \
-	const FlashString& _name PROGMEM = _##_name.fstr;
+	DEFINE_FSTR_REF(_name, _##_name);
 
 /** @brief Define a FlashString for local (static) use
  *  @param _name variable to identify the string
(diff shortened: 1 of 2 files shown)


mikee47:
FlashStrings break with ArduinoJson version 6.13, specifically https://github.com/bblanchon/ArduinoJson/commit/6da6f921cd7667296f64ee4c641f61df13d24727.

MemberProxy now takes a copy of its key String, instead of a reference, and therefore attempts to instantiate FlashString. There are likely performance implications to this, which require further investigation.

The issue with FlashString is solved fairly easily by changing FlashString to a reference container, but  again that's a more significant change.

Closing this PR.

PR #1915, opened 2019-10-27
Deprecated DNSServer. Use DnsServer instead.

slaff:


mikee47:
@slaff This breaks Windows as it cannot distinguish between DNSServer and DnsServer.

slaff:
> This breaks Windows as it cannot distinguish between DNSServer.h and DnsServer.h

Hm... Strange how appveyor did not detect that. That will mean that also rBootOutput... and RbootOutput.. should have the same issue.

PR #1917, opened 2019-10-28
Fix coverity warnings

mikee47:
Also minor fix of duplicate debugf definition which prevents Host release build.


slaff:
@mikee47 Can you take out the commit related to DnsServer and put it into a new PR ?

PR #1920, opened 2019-10-30
Install current MinGW version for appveyor

mikee47:
Use pre-built release stored in separate github repository.

### DIFF
--- a/.appveyor/install.cmd
+++ b/.appveyor/install.cmd
@@ -1,5 +1,9 @@
 REM Windows install script
 
+rmdir /s /q c:\MinGW
+curl -LO https://github.com/mikee47/SmingTools/releases/download/1.0/MinGW.7z
+7z -oC:\ x MinGW.7z
+
 goto :%SMING_ARCH%
 
 :Esp8266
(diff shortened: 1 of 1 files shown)


.appveyor/install.cmd
```diff
@@ -1,5 +1,9 @@
 REM Windows install script
 
+rmdir /s /q c:\MinGW
+curl -LO https://github.com/mikee47/SmingTools/releases/download/1.0/MinGW.7z
```
slaff: @mikee47 Can you add the MinGW toolchain to https://github.com/attachix/sming-toolchains? Hopefully one day we can move all toolchains to SmingHub.

mikee47: OK, I've imported the repo to Attachix, rebased and changed the link.

mikee47: Looks like it didn't get the release properly. I'll recreate it.

mikee47: OK, should be fixed now.

slaff:
@mikee47 Finally we have better control over SmingHub. You should be now the administrator of https://github.com/SmingHub/SmingTools. Please, check if that is the case and then, whenever you have time, change the appveyor script to fetch the current MinGW from SmingHub/SmingTools

PR #1923, opened 2019-10-30
Bugfix: `flashmem_get_info()` only works if running from slot #0.

### LINKED ISSUE #1922: SPIFFS won't initialise when booting from second rBoot slot
For example, setting up a SPIFFS volume of 2MByte works fine when booting from slot #0 at 0x2000, but switching to slot #1 at 0x202000 fails to iniitalise.

Can result in `spiffs_get_storage_config()` reporting `the requested SPIFFS size is too big`.

The problem is with `flashmem_get_info()` which assumes that memory location 0x4020000 is the start of flash. However, rBoot remaps this address so it falls over. Instead, it should call flashmem_read at location 0. The sector at offset 0x200000 never gets properly initialised so returned data will be unpredictable.

mikee47:
Fixes #1922


mikee47:
LTS back-port?

PR #1927, opened 2019-10-31
Add deprecation warning to `flashmem_get_address()`

mikee47:
Added in #1490 but not currently used within the framework.

This function maps a PROGMEM address to the corresponding flash location, but doesn't work if program space has been remapped by rBoot.

I will fix this properly in a separate upgrade to Host flash memory handling and FlashStrings.


mikee47:
@slaff Any thoughts on this patch?

mikee47:
Since we don't use this in the framework (at present) but will shortly, I'll revert this patch but add a deprecation warning just in case anyone's used it.

Issue #1875, opened 2019-10-07
More flexible HTTP server error handling in case of bad body content


aemseemann:
There are basically two kinds of errors in an HTTP communication:
* A violation of the HTTP protocol, detected by the http_parser.
* Bad body data, i.e. the _application_ rejects some or all of the incoming data. This is done by letting the `bodyParser` or the `HttpResource::onBody` callback return a number less than the amount of bytes it was given to process.

Currently, Sming's HTTP server handles both error cases identically in the following way:
1. Invoke `HttpServerConnection::onHttpError,` which in turn invokes the `HttpServerConnection::onMessageComplete`, which invokes the `HttpRequest::onRequestComplete` to inform the application and give it an opportunity to prepare an error response.
2. In case an error response was prepared: Send it to the client while ignoring any further data received from the client during the transmission.
3. Close the TCP connection.

While this behaviour is fine for HTTP protocol violations, it may not always be appropriate for the 'bad body data' case for the following reasons:
* A client might not receive an early error response if it is still busy transmitting the request body by the time the server closes the connection. The standard does not mandate clients to listen for a response before the request is sent completely and browsers behave differently in that regard. Chrom(e|ium) does receive early responses. Firefox does not!
* Furthermore, since there is no HTTP protocol violation, the communication channel is still intact and no need to close it (assuming keep-alive mode).
* Conceptually: Bad body data is an application-level error and should be dealt with at the application level. The HTTP server should not step in here, at least not automatically.

As a remedy, I propose the following changes to the error handling behaviour for the 'bad body data' case:
- Invoke `HttpServerConnection::onHttpError `-> `HttpServerConnection::onMessageComplete` -> `HttpRequest::onRequestComplete` as usual, but...
- ...**do not close the connection after the (early) error response has been sent.** Instead ignore any further body chunks, i.e. do not invoke the `bodyParser` or `HttpRequest::onBody` for this request anymore. 
- At the end of the (partially ignored) request, use the 'close' and 'keep-alive' headers in the same way as for a successful request to decide if the underlying TCP connection should be closed or reused.

Since there might be situations where (application level code decides that) it is reasonable to close the connection as early as possible, the old behaviour should be activatible by a server configuration option or on a per-request basis.

What's your opinion? 
Would a PR for that have a chance to get accepted?

aemseemann:
Since the PR is now merged, I am closing this issue.
For future reference: The new behaviour can be activated by configuring the HTTP server as follows:
```cpp
[...]

PR #1928, opened 2019-10-31
Allow HTTP connections to ignore rejected body content (#1875)

aemseemann:
This PR allows the HTTP server to ignore invalid body content, i.e. content rejected by the body parser or the `HttpResource::onBody` callback, without closing the connection immediately (see #1875).
The old behaviour can be restored either for the whole server (`HttpServerSettings::closeOnBadBodyContent = true`) or per connection (`HttpServerConnection::setCloseOnBadBodyContent(true)`).

### DIFF
--- a/Sming/Core/Network/Http/HttpServerConnection.cpp
+++ b/Sming/Core/Network/Http/HttpServerConnection.cpp
@@ -33,6 +33,7 @@ int HttpServerConnection::onMessageBegin(http_parser* parser)
 	// and temp data...
 	reset();
 	bodyParser = nullptr;
+	hasContentError = false;
 
 	return 0;
 }
@@ -62,6 +63,10 @@ int HttpServerConnection::onMessageComplete(http_parser* parser)
 
 	if(bodyParser) {
 		bodyParser(request, nullptr, PARSE_DATAEND);
+	}
+
+	if(hasContentError) {
+		response.code = HTTP_STATUS_BAD_REQUEST;
 	}
 
 	if(resource != nullptr && resource->onRequestComplete) {
@@ -139,15 +144,28 @@ int HttpServerConnection::onHeadersComplete(const HttpHeaders& headers)
 
 int HttpServerConnection::onBody(const char* at, size_t length)
 {
+	if(hasContentError) {
+		return 0;
+	}
+
 	if(bodyParser) {
-		size_t consumed = bodyParser(request, at, length);
+		const size_t consumed = bodyParser(request, at, length);
 		if(consumed != length) {
-			return -1;
+			hasContentError = true;
+			if(closeOnContentError) {
+				return -1;
+			}
 		}
 	}
 
 	if(resource != nullptr && resource->onBody) {
-		return resource->onBody(*this, request, at, length);
+		const int result = resource->onBody(*this, request, at, length);
+		if(result != 0) {
+			hasContentError = true;
+			if(closeOnContentError) {
+				return result;
+			}
+		}
 	}
 
 	return 0;
--- a/Sming/Core/Network/Http/HttpServerConnection.h
+++ b/Sming/Core/Network/Http/HttpServerConnection.h
@@ -76,6 +76,11 @@ public:
 		return &request;
 	}
 
+	void setCloseOnContentError(bool close = true)
+	{
+		closeOnContentError = close;
+	}
+
 protected:
 	// HTTP parser methods
 
@@ -120,6 +125,8 @@ private:
 
 	const BodyParsers* bodyParsers = nullptr;	///< const reference ensures we cannot modify map, only look stuff up
 	HttpBodyParserDelegate bodyParser = nullptr; ///< Active body parser for this message, if any
+	bool closeOnContentError = false;
+	bool hasContentError = false;
 };
 
 /** @} */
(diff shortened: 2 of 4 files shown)


slaff:
> Travis complaining about coding style issues

Try the following commands:
```
cd $SMING_HOME
make cs
```

aemseemann:
I did that, but my clang-format is v9.0.0, which seems to generate a different output than v6.0 used in the Travis environment. (After 'make cs' I got diffs all around the source tree in places I never touched)

slaff:
@aemseemann Please, take the commit related to clang out of this PR and put it in a new PR. Thanks in advance.

aemseemann:
@slaff done, see #1929

Sming/Core/Network/Http/HttpServerConnection.cpp
```diff
@@ -139,15 +144,25 @@ int HttpServerConnection::onHeadersComplete(const HttpHeaders& headers)
 
 int HttpServerConnection::onBody(const char* at, size_t length)
 {
-	if(bodyParser) {
-		size_t consumed = bodyParser(request, at, length);
-		if(consumed != length) {
-			return -1;
+	if(!bodyContentError) {
```
slaff: Re-arrange a bit the if statements to avoid too many nested ifs and improve readability. For example that block here can be rewritten as:
```
     if(bodyContentError) {
            return 0;
     }

    if(bodyParser) {
          // ...
    }

    if(resource != nullptr && resource->onBody) {
       // ...
    }

    return 0;
```

Sming/Core/Network/Http/HttpServerConnection.cpp
```diff
@@ -33,6 +33,7 @@ int HttpServerConnection::onMessageBegin(http_parser* parser)
 	// and temp data...
 	reset();
 	bodyParser = nullptr;
+	bodyContentError = false;
```
slaff: Rename `bodyContentError` to `hasContentError`.

Sming/Core/Network/Http/HttpServerConnection.cpp
```diff
@@ -64,6 +65,10 @@ int HttpServerConnection::onMessageComplete(http_parser* parser)
 		bodyParser(request, nullptr, PARSE_DATAEND);
 	}
 
+	if(bodyContentError) {
+		response.code = HTTP_STATUS_BAD_REQUEST; // preset error response
```
slaff: Highly recommended: remove the comment as it does not add additional information about the logic and the flow.

Sming/Core/Network/Http/HttpServerConnection.h
```diff
@@ -120,6 +125,8 @@ class HttpServerConnection : public HttpConnection
 
 	const BodyParsers* bodyParsers = nullptr;	///< const reference ensures we cannot modify map, only look stuff up
 	HttpBodyParserDelegate bodyParser = nullptr; ///< Active body parser for this message, if any
+	bool closeOnBadBodyContent = false;
```
slaff: Please, rename `closeOnBadBodyContent` to `closeOnContentError`.

Sming/Core/Network/Http/HttpServerConnection.h
```diff
@@ -76,6 +76,11 @@ class HttpServerConnection : public HttpConnection
 		return &request;
 	}
 
+	void setCloseOnBadBodyContent(bool close = true)
```
slaff: Rename `setCloseOnBadBodyContent` to `setCloseOnContentError`.

Sming/Core/Network/HttpServer.h
```diff
@@ -29,6 +29,7 @@ typedef struct {
 	int keepAliveSeconds = 0;	  ///< default seconds to keep the connection alive before closing it
 	int minHeapSize = -1;		   ///< min heap size that is required to accept connection, -1 means use server default
 	bool useDefaultBodyParsers = 1; ///< if the default body parsers,  as form-url-encoded, should be used
+	bool closeOnBadBodyContent = false;
```
slaff: Please, make the default value to be `true`. On embedded device the default behaviour should be a bit more strict and trying to preserve resources. Further in the samples the users can adjust this so that for example Firefox does not show an error message. And in the sample(s) the contributor(s)  can and should add additional comments describing the advantages and disadvantages of having this set to `true`.

Sming/Core/Network/Http/HttpServerConnection.cpp
aemseemann: done

Sming/Core/Network/Http/HttpServerConnection.cpp
aemseemann: done

Sming/Core/Network/Http/HttpServerConnection.cpp
aemseemann: done

Sming/Core/Network/Http/HttpServerConnection.h
aemseemann: done

Sming/Core/Network/Http/HttpServerConnection.h
aemseemann: done

Sming/Core/Network/HttpServer.h
aemseemann: Done, but I have some mixed feelings about this. On the one hand, I understand your reasoning, especially for M2M applications. However, for HMI-focused applications, this default may lead to some unexpected and hard to track down behaviour...

Anyway, either default is fine for me, as long as it allows to resolve the open issue in #1817.

PR #1929, opened 2019-10-31
Make clang-format binary configurable.

aemseemann:
Allow overriding clang-format binary like `make CLANGFORMAT=path/to/my/clang-format cs`

### DIFF
--- a/Sming/build.mk
+++ b/Sming/build.mk
@@ -79,10 +79,17 @@ TOOLS_BASE		= $(SMING_HOME)/$(OUT_BASE)/tools
 USER_LIBDIR		= $(SMING_HOME)/$(OUT_BASE)/lib
 
 # Git command
+DEBUG_VARS	+= GIT
 GIT ?= git
 
 # CMake command
+DEBUG_VARS	+= CMAKE
 CMAKE ?= cmake
+
+# clang-format command
+DEBUG_VARS	+= CLANG_FORMAT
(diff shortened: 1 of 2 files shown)


Sming/build.mk
```diff
@@ -84,6 +84,8 @@ GIT ?= git
 # CMake command
 CMAKE ?= cmake
 
+CLANGFORMAT ?= clang-format
```
mikee47: I think `CLANG_FORMAT` would be more consistent. Also, add `DEBUG_VARS += CLANG_FORMAT`.

aemseemann: done.
I noticed that cmake and git are not in DEBUG_VARS. Should I add them, too?

mikee47: Think I left those out originally as it seemed unlikely they'd be changed, but I can't think of any reason not to add them as well. I've tried to stick with the convention of 'declaring' the variables before actually assigning them, like this:

```
# Git command
DEBUG_VARS += GIT
GIT ?= git

# CMake command
DEBUG_VARS += CMAKE
CMAKE ?= cmake

# clang-format command
DEBUG_VARS	+= CLANG_FORMAT
CLANG_FORMAT ?= clang-format
```

aemseemann: > Think I left those out originally as it seemed unlikely they'd be changed

I would have thought so, too, but one can never know. 
Changed it according to your suggestion.

PR #1930, opened 2019-11-01
Fix `pure virtual method called` error when running CI tests

### LINKED ISSUE #1905: Issues reported by our CI systems.
# AppVeyor CI

The host build fails with the following error.
```
c:/projects/sming-sb483/Sming/Arch/Host/Components/driver/hw_timer.cpp: In member function 'virtual void* CTimerThread::thread_routine()':
c:/projects/sming-sb483/Sming/Arch/Host/Components/driver/hw_timer.cpp:171:16: error: 'ETIMEDOUT' was not declared in this scope
    if(errno != ETIMEDOUT) {
                ^
make[1]: *** [hw_timer.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make: *** [driver-build] Error 2
Failed with error #1133 
```

See: https://ci.appveyor.com/project/slaff/sming-sb483/builds/28433616/job/8w1d4phc37w5u6wr

# Travis CI 
Quite often Travis CI is failing with the following problem (Full log can be seen here: https://travis-ci.org/SmingHub/Sming/jobs/601194512 )

```
<td><img src="back.gif" alt="[PARENTDIR]"></td>
<td colspan="5"><a href="/">Parent Directory</a></td>
</tr>
host_exit: returning 0
main: >> Normal Exit <<
host_flashmem_cleanup: Closed "out/Host/debug/firmware/flash.bin"
cleanup: Goodbye!
pure virtual method called
terminate called without an active exception
signal_handler: SIGABRT - usually caused by an abort() or assert()
/home/travis/build/SmingHub/Sming/Sming/Arch/Host/app.mk:46: recipe for target 'run' failed
```

I guess the issue is due to some race condition because if I restart the build multiple times at some point the error is not present. 

> pure virtual method called

@mikee47 Can that be the reason for the issue? Do you have idea why is that happening?

mikee47:
See #1905 for discussion.

Unsafe to call methods from virtual CThread destructor, so add `terminate()` method to all derived classes and call explicitly _before_ destroying the object.

Fix keyboard thread race condition.

Closes #1905

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -68,7 +68,7 @@ if [ "$TRAVIS_BUILD_STAGE_NAME" == "Test" ]; then
 
 	# Build and run tests
 	export SMING_TARGET_OPTIONS='--flashfile=$(FLASH_BIN) --flashsize=$(SPI_SIZE)'
-	$MAKE_PARALLEL tests || true
+	$MAKE_PARALLEL tests
 
 	# Build the documentation
 	mv $SMING_PROJECTS_DIR/samples ..
--- a/Sming/Arch/Host/Components/driver/hw_timer.cpp
+++ b/Sming/Arch/Host/Components/driver/hw_timer.cpp
@@ -34,10 +34,11 @@ public:
 		execute();
 	}
 
-	~CTimerThread()
+	void terminate()
 	{
 		state = terminating;
 		sem.post();
+		join();
 	}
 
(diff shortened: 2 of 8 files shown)


slaff:
@mikee47 Revert also a line that I have added to Travis CI:

https://github.com/SmingHub/Sming/blob/develop/.travis/build.sh#L71
```
$MAKE_PARALLEL tests || true
```

back to
```
$MAKE_PARALLEL tests
```

PR #1931, opened 2019-11-02
Add support for 'Expect: 100-continue' in HTTP server.

aemseemann:
When uploading files as multipart/form-data, browsers (and other HTTP clients like curl) may generate an [`Expect: 100-continue`](https://tools.ietf.org/html/rfc7231#section-5.1.1) header (depending on file size) and subsequently wait for a [`100 Continue`](https://tools.ietf.org/html/rfc7231#section-6.2.1) response from the server before continuing to send the body data. 

Although most clients will continue to send the body data anyway after a short timeout, this PR avoids  the timeout.

### DIFF
--- a/Sming/Core/Network/Http/HttpServerConnection.cpp
+++ b/Sming/Core/Network/Http/HttpServerConnection.cpp
@@ -101,16 +101,19 @@ int HttpServerConnection::onHeadersComplete(const HttpHeaders& headers)
 
 	if(resource != nullptr && resource->onHeadersComplete) {
 		error = resource->onHeadersComplete(*this, request, response);
+		if(error != 0) {
+			return error;
+		}
 	}
 
-	if(!error && request.method == HTTP_HEAD) {
-		error = 1;
+	if(request.method == HTTP_HEAD) {
+		return 1;
 	}
 
 	if(bodyParsers != nullptr && request.headers.contains(HTTP_HEADER_CONTENT_TYPE)) {
 		String contentType = request.headers[HTTP_HEADER_CONTENT_TYPE];
 		int endPos = contentType.indexOf(';');
-		if(endPos != -1) {
+		if(endPos >= 0) {
 			contentType = contentType.substring(0, endPos);
 		}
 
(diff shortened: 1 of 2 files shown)


slaff:
@aemseemann Did you test you PR?

Sming/Core/Network/Http/HttpServerConnection.cpp
```diff
@@ -103,34 +103,46 @@ int HttpServerConnection::onHeadersComplete(const HttpHeaders& headers)
...
-	if(!error && request.method == HTTP_HEAD) {
-		error = 1;
-	}
-
-	if(bodyParsers != nullptr && request.headers.contains(HTTP_HEADER_CONTENT_TYPE)) {
-		String contentType = request.headers[HTTP_HEADER_CONTENT_TYPE];
-		int endPos = contentType.indexOf(';');
-		if(endPos != -1) {
-			contentType = contentType.substring(0, endPos);
+	if(!error) {
```
slaff: @aemseemann Refactor this block and try to avoid many levels of nested ifs.

mikee47: I'd also prefer to see `if (endPos >= 0)` as its more robust

aemseemann:
@slaff: I did test this patch successfully on top of #1817 with a command line like `curl -F firmware=@$(RBOOT_ROM_0_BIN).signed http://$(BOARD_URL)/upload`. 
I also verified the `Expect:...` / `100 Continue` exchange with Wireshark and that the body transmission continues immediately. Without this patch, Wireshark timestamps show a delay between headers and body of approximately 1 second. This isn't much, but still avoidable by simply answering to the Expect header.

Sming/Core/Network/Http/HttpServerConnection.cpp
aemseemann: Done.

Hope that it's ok to skip all the body parser setup in case `error != 0`. Although this change is not strictly necessary for this PR, it seemed futile to prepare for a body that will never come.

slaff:
@aemseemann I don't remember if I have mentioned this: Thanks again for your contribution(s) and welcome to the Sming team :)

PR #1936, opened 2019-11-03
Expand host tests

mikee47:
Add tests for base64, datetime, http, url, serial, string and objectmap.

(Bunch of test code I had kicking about from earlier PRs.)

### DIFF
--- a/tests/HostTests/app/application.cpp
+++ b/tests/HostTests/app/application.cpp
@@ -72,10 +72,10 @@ void TestGroup::complete()
 	taskTimer.startOnce();
 }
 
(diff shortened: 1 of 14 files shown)


mikee47:
Codacy warning is false positive

mikee47:
Nope, even telling codacy the variable type it still complains. Code works as intended.

slaff:
@mikee47 As far as I can see codacy is complaining about the unused variable named `value`. Which seems to me indeed unused.

image

mikee47:
Setting `value = new TestClass;` invokes the `TestMap::Value` assignment operator which adds the new TestClass instance to the map. If it was truly unused then how could the map count be updated?

Issue #1937, opened 2019-11-04
HttpClient doesn't send a request


qdimka:
Hi, everyone.

I started using Sming a couple of days ago. It is the great tool.
But I have some problems.
I need to send the sensor data to thinkspeak. 
I am using this [sample](https://github.com/SmingHub/Sming/tree/lts-develop/3.8.x/samples/HttpClient_ThingSpeak).

I modified next line:
```cpp
void sendData()
{
   /* ... */
   url.Query["api_key"] = "my_key";
   /* ... */
}
```

Nothing to happened after first call thinkSpeak.downloadString. 
`onDataSent` wasn`t called. 
After the next call, I got the message "Already freed".

What do I do wrong?

```
phy ver: 972, pp ver: 10.1

mode : sta(5c:cf:7f:84:53:da)
add if0
f r0, scandone
state: 0 -> 2 (b0)
state: 2 -> 3 (0)
state: 3 -> 5 (10)
add 0
aid 2
cnt

connected with LETAIQD_2.4G, channel 6
dhcp client start...
ip:192.168.0.178,mask:255.255.255.0,gw:192.168.0.1
pm open,type:2 0
3fff01d0 already freed
3fff01d0 already freed
3fff01d0 already freed
3fff0028 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
3fff0018 already freed
```

mikee47:
Hi @qdimka, we'll look into this but as you've only just started using Sming I'd recommend moving to Sming 4.0.0 which has been released today.

mikee47:
I've located the reason for the 'already freed' message, I'll submit a PR to patch that (present in 4.0 as well). However, I've yet to discover why no response is received (success or failure).

mikee47:
Compiling the framework with DEBUG_VERBOSE_LEVEL=3 we get this:
```
26670300 Creating new HttpClientConnection
26670691 HttpClientConnection::connect: TCP state: -1, isStarted: 0, isActive: 0
26670940 HttpClientConnection::connecting ...
26675155 +TCP connection
26677257 connect to: api.thingspeak.com
26694713 DNS record found: api.thingspeak.com = 34.228.105.156
26695835 TCP connect result: 0
33893166 TCP connection error: -8
```

The -8 error code is `ERR_ABRT` (connection aborted).

mikee47:
[...]

PR #1938, opened 2019-11-04
bugfix: HttpConnection freed twice

mikee47:
Connection is owned by map so gets freed automatically when removed.

Issue raised in #1937.


mikee47:
+LTS

PR #1939, opened 2019-11-04
[WIP] Ensure request callback is invoked on client connection error

mikee47:
This PR addresses the apparent lack of response when HTTP connections fail, such as with invalid URLs (issue #1937).

When a HTTP request is issued a `completed` delgate is supplied, which receives a boolean `successful` parameter. This parameter is false if HTTP parsing fails or the remote connection returns an HTTP error code.

If, when sending the request, a failure occurs in the connection process then the callback is never invoked. This can happen if:

- The server address specified cannot be resolved (DNS failure)
- A TCP connection cannot be made to the server (no service on the specified port)

And probabaly others. These failures are permanent and if notifed the application can act on them promptly. The alternative is implementing a separate timeout mechanism and just giving up.

So this PR catches a TCP error condition and invokes the request `completed` delegate with a failure code.
The request remains on the waiting queue though, so the request will continue to be retried.

### DIFF
--- a/Sming/Components/Network/src/Network/Http/HttpClientConnection.cpp
+++ b/Sming/Components/Network/src/Network/Http/HttpClientConnection.cpp
@@ -61,6 +61,20 @@ void HttpClientConnection::reset()
 	HttpConnection::reset();
 }
 
+void HttpClientConnection::onError(err_t err)
+{
+	HttpConnection::onError(err);
+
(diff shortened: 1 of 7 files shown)


slaff:
@mikee47 ping me when ready with this PR so I can test it.

mikee47:
Closing until issue #2829 is scoped.

Issue #1940, opened 2019-11-04
HttpServer `maxActiveConnections` not honoured


mikee47:
This doesn't appear to be implemented in HttpServer. Is there any way to rate-limit TCP connections?

PR #1942, opened 2019-11-04
Enforce optional TCP connection limit

mikee47:
As per https://github.com/SmingHub/Sming/pull/1942#issuecomment-553451304, this PR includes a connection count check before accepting a TCP connection. If a connection limit has been specified and accepting the new connection would exceed this value, then the connection is aborted.

Discussion #1940

### DIFF
--- a/Sming/Core/Network/HttpServer.cpp
+++ b/Sming/Core/Network/HttpServer.cpp
@@ -30,6 +30,13 @@ void HttpServer::configure(const HttpServerSettings& settings)
 
 TcpConnection* HttpServer::createClient(tcp_pcb* clientTcp)
 {
+	if(connections.count() >= settings.maxActiveConnections) {
+		debug_w("HttpServer refusing connection, already have %u", connections.count());
+		return nullptr;
+	}
+
+	debug_i("HttpServer::createClient(), connections = %u", connections.count());
+
 	HttpServerConnection* con = new HttpServerConnection(clientTcp);
 	con->setResourceTree(&paths);
 	con->setBodyParsers(&bodyParsers);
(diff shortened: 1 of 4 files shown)


mikee47:
This is probably the wrong way to handle this situation as it's happening at the TCP level rather than HTTP.

My goal is to preserve resources but tell the prospective client 'please try again shortly'. We could accept the connection then issue an HTTP 429 response but that's more complex and not sure it's better.

Thoughts most welcome.

aemseemann:
In `HttpServer::createClient`, if the maximum amount of connections is reached you could `con->setResourceTree(&connLimitReachedResources);` with a `HttpResourceTree` instance specifically prepared to answer every request with HTTP 429 or something equally appropriate.

mikee47:
[503](https://httpstatuses.com/503) says "Some servers might simply refuse the connection." which implies that aborting the connection is an OK approach.

slaff:
> This is probably the wrong way to handle this situation as it's happening at the TCP level rather than HTTP.

In the TCP level in TcpServer.cpp there is a code that limits the number of connections based on free heap size. See below:

```C++
err_t TcpServer::onAccept(tcp_pcb* clientTcp, err_t err)
{
	// Anti DDoS :-)
	if(system_get_free_heap_size() < minHeapSize) {
		debug_w("\r\n\r\nCONNECTION DROPPED\r\n\t(%d)\r\n\r\n", system_get_free_heap_size());
		return ERR_MEM;
	}

#ifdef NETWORK_DEBUG
	debug_d("onAccept state: %d K=%d", err, connections.count());
	list_mem();
#endif

	if(err != ERR_OK) {
		//closeTcpConnection(clientTcp, nullptr);
		return err;
	}

	TcpConnection* client = createClient(clientTcp);
       // ...
```

In the same method  you can move the limitation based on active connections. This way any TcpServer based server can take advantage of the code and set limits based on heap memory and/or active connections.

PR #1945, opened 2019-11-04
Scheme missing from URL in `HttpClient_ThingSpeak` sample

mikee47:
Without this port doesn't get set, attempts to connect to port 0 which fails with ERR_ABORT.


mikee47:
+LTS

PR #1947, opened 2019-11-05
Disable optimsations entirely if building with ENABLE_GDB=1 on Host

mikee47:
I've found that `-Og` still performs some optimisations which makes step-through debugging difficult/impossible, so I propose disabling optimisations entirely when compiling for Host with ENABLE_GDB=1.


slaff:
@mikee47 It happened to me to think that the code compiled with -Og was not working as expected but it turned out to be something else. That said I haven't had a real issue with the flag up to now. Would it be possible to tell me how to reproduce the issue that you are experiencing and see how stepping through is working for me?

mikee47:
Try stepping through some ArduinoJson6 code. Templating gets heavily optimised.

Issue #1949, opened 2019-11-06
Unresolved variable in ds18s20.h


qdimka:
Sming 4.0.0 (from develop).
I`ve got a compile-time error "OneWire does not name a type".

I think OneWire.h should be here. 

https://github.com/SmingHub/Sming/blob/c07123139201f3912f13ff0f611edec68c75a14f/Sming/Libraries/DS18S20/ds18s20.h#L18

Because we have an unresolved variable here.

https://github.com/SmingHub/Sming/blob/c07123139201f3912f13ff0f611edec68c75a14f/Sming/Libraries/DS18S20/ds18s20.h#L133

Am I right? If my pull request isn`t correct, let me know.

PR #1950, opened 2019-11-06
Fix `OneWire does not name a type` error (#1949)

qdimka:
See #1949 for details

### DIFF
--- a/Sming/Libraries/DS18S20/ds18s20.h
+++ b/Sming/Libraries/DS18S20/ds18s20.h
@@ -15,7 +15,7 @@
  *  @ingroup    libraries
  *  @{
 */
-#include <Wire.h>
+#include <Libraries/OneWire/OneWire.h>
 
 #define MAX_SENSORS 4         ///< Maximum quantity of sensors to read
 
--- a/Sming/Libraries/DS18S20/ds18s20.h
+++ b/Sming/Libraries/DS18S20/ds18s20.h
@@ -15,7 +15,7 @@
  *  @ingroup    libraries
  *  @{
 */
-#include <Libraries/OneWire/OneWire.h>
+#include <Wire.h>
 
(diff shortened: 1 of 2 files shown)


Sming/Libraries/DS18S20/ds18s20.h
```diff
@@ -15,7 +15,7 @@
  *  @ingroup    libraries
  *  @{
 */
-#include <Wire.h>
+#include <Libraries/OneWire/OneWire.h>
```
mikee47: I think just a `class OneWire;` declaration would suffice as it's not used in the header, nor by client code.

qdimka: Thanks, @mikee47. I've added `class OneWire;` declaration

qdimka: Just a second. The variable `ds` is used in `ds18s20.cpp`. I still think we should use `OneWire.h` instead of `class OneWire;`

mikee47: Yes, but it already includes the header so it'll be happy :-)

mikee47: Just a question of not introducing a load of additional dependencies into application code.

mikee47: While you're at it, could you also remove the #include from `samples\Temperature_DS1820\app\application.cpp`. It's not needed and shouldn't be there; explains why the error you report hadn't surfaced already.

qdimka: >  remove the #include from `samples\Temperature_DS1820\app\application.cpp`. 

Yep, done.

mikee47: Thanks!

slaff:
@qdimka Thanks for your first contribution and welcome to the team of contributors :)

PR #1954, opened 2019-11-07
Add string tests from Arduino ESP8266 core

mikee47:
As a first step to #1951 I'm propsing the addition of these tests. This is a cut & paste job with minor amendments to fit the HostTests framework. As we don't yet have SSO and Sming doesn't have StringStream those tests have been commented out.

Bugfix: non-decimal negative #s should be as if they were unsigned


slaff:
@mikee47 do you have something to add here before I merge this PR?

PR #1956, opened 2019-11-08
Update Linux installation instructions

mikee47:
Remove details for SDK version, confusing.

### DIFF
--- a/docs/source/arch/esp8266/getting-started/linux.rst
+++ b/docs/source/arch/esp8266/getting-started/linux.rst
@@ -1,42 +1,35 @@
-******************
 Linux Installation
-******************
+==================
 
 .. highlight:: bash
 
 Pre-requisites
-==============
+--------------
 
-Ubuntu
-------
+Ubuntu::
 
-::
+   sudo apt-get install make unrar autoconf automake libtool libtool-bin gcc g++ gperf \
+      flex bison texinfo gawk ncurses-dev libexpat1-dev python sed python-serial \
+      python-dev srecord bc git help2man unzip bzip2
 
-   sudo apt-get install make unrar autoconf automake libtool libtool-bin gcc g++ gperf flex bison texinfo gawk ncurses-dev libexpat1-dev python sed python-serial python-dev srecord bc git help2man unzip bzip2
+Fedora::
 
-Fedora
-------
-
-::
-
-   sudo dnf install make autoconf automake libtool gcc gcc-c++ gperf flex bison texinfo gawk ncurses-devel expat-devel python sed pyserial srecord bc git patch unzip help2man python-devel
+   sudo dnf install make autoconf automake libtool gcc gcc-c++ gperf flex bison texinfo \
+      gawk ncurses-devel expat-devel python sed pyserial srecord bc git patch \
+      unzip help2man python-devel
 
 Prepare directory
-=================
+-----------------
 
-*We recommend /opt/ but you can use anything you want*
-
-::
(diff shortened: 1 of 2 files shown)


docs/source/arch/esp8266/getting-started/linux.rst
```diff
@@ -53,12 +46,21 @@ or 1.4.0:
...
 
    # This will take a while...
-   make VENDOR_SDK=1.5.4 STANDALONE=y
+   make STANDALONE=y
 
-The parameter ``VENDOR_SDK=1.5.4`` specifies that you will be using SDK
-version 1.5.4. Before changing this value to a newer version make sure
-to check that the SDK version that you have chosen is
-`supported <https://github.com/SmingHub/Sming#compatibility>`__.
+Here are the links for pre-compiled versions:
```
slaff: NONOS-SDK 1.5.4 is pretty old. A newer precompiled toolchain will be better.

mikee47: We need a repo!
https://github.com/SmingHub/Sming/issues/1905#issuecomment-547830862

slaff: @mikee47 Please accept the invite to join attachix's core team. After that you should be able to read, write, push, etc to https://github.com/attachix/sming-toolchains. We can start putting the different toolchains there.

mikee47: @slaff Are there issues with access to sminghub?

slaff: > sminghub?

Well yes. I am just a contributor added to SmingHub/Sming repo. And cannot create a new repository under SmingHub or add you as a core contributor to SmingHub/Sming. I also do not know for sure who is the owner of SmingHub. My wild guess is that this is not @anakod but a person who does not care so much for the success of Sming or, as strange as it may sound, might be even slightly offended by it. This is just a speculation from my side but in the last years we did not get much in  terms of support from the owner of the SmingHub account.

mikee47: Should we plan to move it then?

slaff: > Should we plan to move it then?

Well, if the problems start to mount then yes. But before that we should carefully communicate this with the community.

mikee47: Something else we might need/want to do in the near future is migrate some parts of Sming into separate repositories, e.g. some or all of Libraries, so it's a little annoying we don't have proper control over the main repo!

I'm happy to proceed using Attachix for this and the toolchains, but it does feel a little like we're be attaching a 'brand' to it. Maybe we should set up a separate github organisation specifically for Sming? All we'd need is an inspiring name to give it!

mikee47: Re. lack of support for SmingHub, it does seem odd. There is a lot of original work in Sming, but also a considerable amount which is refinement or development of other people's code and I do try to make sure that's properly credited. If anyone reading this feels slighted or under-mentioned please accept my sincere apologies.

mikee47:
For now, I've taken out all the 1.5.4 SDK references in the toolchain, and uploaded to SmingTools.

slaff:
> For now, I've taken out all the 1.5.4 SDK references in the toolchain, and uploaded to SmingTools.

Awesome, do you want to add something more before I merge this PR?

mikee47:
Good to go!

PR #1957, opened 2019-11-08
CStringArray improvements

mikee47:
Primarily this PR is intended to update CStringArray to use public String methods so it doesn't break when SSO is added. A unit test module has been added for this class which has thrown up some issues which have also been fixed.

CStringArray fixes:

- Make sure string count is maintained consistently for all operations
- Array always maintains final NUL character even if not provided
- Appending strings containing NUL characters are handled correctly, treated as an array
- Values may be empty

String fixes:

- String must be invalidated if assigned a null value.
- Rewrite `toLowerCase()` and `toUpperCase()` to handle binary strings
- Fix `equals`, `equalsIgnoreCase` and `compareTo` String methods to handle binary strings

Added features:

- Add `ignoreCase` option to allow case-sensitive searches (default is case insenstive)
- Add assignment and append operators

### DIFF
--- a/Sming/Core/Data/CStringArray.cpp
+++ b/Sming/Core/Data/CStringArray.cpp
@@ -13,30 +13,61 @@
 #include "CStringArray.h"
 #include <stringutil.h>
 
-bool CStringArray::add(const char* str, unsigned length)
+bool CStringArray::add(const char* str, int length)
 {
-	if(length == 0)
-		length = strlen(str);
-	if(!reserve(len + length + 1))
+	// Note: zero-length string is fine
+	if(str == nullptr) {
+		return true; // Nothing to append
+	}
+
+	int str_len = strlen(str);
+	if(length < 0) {
+		length = str_len;
+	}
+	auto len = this->length();
+
+	// Ensure trailing NUL is included
+	int addNul = (length == 0 || str[length - 1] != '\0') ? 1 : 0;
+	if(!setLength(len + length + addNul)) {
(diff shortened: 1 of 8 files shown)


mikee47:
@slaff I wasn't happy with CStringArray so have given it a more thorough review. Running it through some more focused tests has thrown up quite a few issues.

All: Please take a close look at the unit tests I've provided in this PR as they govern the acceptable behaviour of the class!

mikee47:
Reviewing previous related issues/PRs there's a conflict with PR #1785, which I now believe is wrong. My bad, sorry.

This was a fix for this condition, which would now (correctly) fail:

```
CStringArray test("abc\0", 4);
test.count() is 2
test[0] = "abc"
test[1] = (null) - should be ""
```

A consequence of the lack of integration tests, and in this instance because I was probably considering these \0 characters as _separators_, whereas they are _terminators_. I've included comments in header file to clarify this.

Including the NUL character provided by String, the above text is stored internally as "abc\0\0". That is a correctly double-null terminated string array containing one element. If we intended to create an array with two elements we would use "abc\0\0". Note that "abc" is considered equivalent to "abc\0" when constructing arrays, and is handled internally.

Note that this is an edge case and doesn't affect existing use within the framework. This PR should be portable to LTS.

slaff:
@mikee47 Is this PR tested and ready for merging?

mikee47:
Yes, all done.

PR #1959, opened 2019-11-09
Refactor creation of libmain

aemseemann:
Several components in the ESP8266 arch require slight modifications of libmain, e.g. 
- rboot: weaken symbol `Cache_Read_Enable_New`
- custom_heap: remove object `mem_manager.o` 
- Sming core for ESP8266: remove object `time.o`, because its functions are provided by the framework and would otherwise lead to multiple defined references.

The current libmain modification logic is spread across several Makefiles and tailored exactly to the uses described above, which makes it rather hard to maintain or extend.
This PR introduces a more flexible approach and in turn removes the undesired coupling of these components.

### DIFF
--- a/Sming/Arch/Esp8266/Components/custom_heap/component.mk
+++ b/Sming/Arch/Esp8266/Components/custom_heap/component.mk
@@ -13,8 +13,9 @@ GLOBAL_CFLAGS			+= -DUMM_POISON_CHECK
 endif
 
 # remove mem_manager.o module from libmain of SDK
-define LIBMAIN_COMMANDS +=
+define HEAP_LIBMAIN_COMMANDS
 @echo Enabling custom heap implementation
 $(Q) $(AR) -d $@ mem_manager.o
 
 endef
+LIBMAIN_COMMANDS += $(HEAP_LIBMAIN_COMMANDS)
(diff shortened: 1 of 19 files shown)


slaff:
@aemseemann Please, fix the failing AppVeyor test: https://ci.appveyor.com/project/slaff/sming-sb483/builds/28736412/job/7w3htx4uvpq8mbaq

aemseemann:
Fixing AppVeyor is easier said than done :(
At first, I suspected the issue is related to the different make versions used on Travis/Ubuntu (4.1) vs. AppVeyor (3.82). But even after building the older version from source, I could not reproduce the error, at least not on a Linux host.

I need to get my hands on a Windows machine first before investigating this any further.

mikee47:
@aemseemann I'll take a look, give me a mo.

aemseemann:
Correction: AppVeyor is using make version 3.8**1** and it *is* the reason for the build failure, since the multi-line variable feature I am using here was only introduced in 3.82.

@slaff: Do we have to stick with this ancient make version on AppVeyor?

mikee47:
@aemseemann I've suggested an alternative approach, pushed to your repo.

aemseemann:
@mikee47: Thank you very much! That did the trick. 
Although list-config doesn't expand the variables (`LIBMAIN_COMMANDS = $(RBOOT_LIBMAIN_COMMANDS) $(APP_LIBMAIN_COMMANDS)`) the expansion seems to work during the actual build.

mikee47:
> Although list-config doesn't expand the variables (`LIBMAIN_COMMANDS = $(RBOOT_LIBMAIN_COMMANDS) $(APP_LIBMAIN_COMMANDS)`) the expansion seems to work during the actual build.

Sorry, I should have taken that out - it's not terribly helpful!

mikee47:
I like this PR, good job thank you!

PR #1961, opened 2019-11-09
Make String object size configurable

mikee47:
SSO length is stored in 7 bits so can be at most 127 chars.
Revise tests to accommodate the largest size.


mikee47:
Continuing from #1951 the disadvantage of increasing string size is we can't use the space once the String switches out of SSO mode, so the additional allocation becomes an overhead.

One way to solve this is to make SSO_CAPACITY a template parameter so Strings can be defined with different values to suit requirements. That would require `String` to become a virtual base, however, which adds another 4 bytes to the size.

PR #1964, opened 2019-11-12
Fix `flashmem_get_address()` to account for rBoot mapping

mikee47:
Added in #1490, this function is to translate a memory pointer in flash to the corresponding physical address. Because rBoot can change the physical address mapping, this needs to be accounted for.

This address can then be used with `flashmem_read()` thus bypassing the cache. This is advantageous when reading infrequently-used data, especially if it's large, such as when using `IMPORT_FSTR` on files.

A working version is also included for the Host emulator.


Sming/Arch/Esp8266/Components/spi_flash/flashmem.c
```diff
@@ -17,6 +17,7 @@
 
 #include "include/esp_spi_flash.h"
 #include "espinc/peri.h"
+#include <rboot-api.h>
```
slaff: @mikee47 I like this PR, but what bothers me is the coupling between spi_flash and rBoot. At some point in time we can move to another bootloader or have a choice of bootloaders. I would prefer to have this code outside of `flashmem.c`. Every bootloader should have its own implementation. During linking time the linker can choose the right implementation of `flashmem_get_address`. 

Does that make sense to you?

mikee47: @slaff Yes, I agree, although it seems pretty unlikely that rBoot would get replaced for the ESP8266... Obviously the ESP32 is another matter.

What I would have preferred to do is query the hardware mapping register directly, but failed to find the relevant information; I included the `Cache_Read_Enable` document from Richard's blog for exactly this reason as so far it's the best research on the issue I've found so far.

For the moment I think co-operation with the bootloader is the safest way; I needed this fix for a deployed system so at some point I'll do some reverse-engineering on `Cache_Read_Enable` - unless someone else beats me to it :-) Then we can ditch the dependency and implement it 'properly' so it works independently of the bootloader.

slaff: > What I would have preferred to do is query the hardware mapping register directly

Is that what you are looking for: excerpt from [Cache_Read_Enable.mw](https://github.com/pfalcon/esp8266-re-wiki-mirror/blob/master/Cache_Read_Enable.mw) says "Implementation writes to two [[Memory_Map#dport0_.283FF00000h.E2.80.93.29|dport0 registers]] to configure page mapping."

mikee47:
Verification code:

```
#include <eagle_soc.h>
extern "C" void Cache_Read_Enable(uint8_t map, uint8_t p, uint8_t v);

#include <esp_spi_flash.h>

static uint32_t flashmem_get_address_CHECK(const void* memptr)
{
#define CACHE_FLASH_CTRL_REG 0x3ff0000C

	uint32_t addr = (uint32_t)memptr - INTERNAL_FLASH_START_ADDRESS;
	// Determine which 1MB memory bank is mapped
	uint32_t ctrl = READ_PERI_REG(CACHE_FLASH_CTRL_REG);
	uint8_t bank = (((ctrl >> 16) & 0x07) << 1) | ((ctrl >> 25) & 0x01);
	addr += 0x100000U * bank;
	return addr;
}

void IRAM_ATTR mmapCheck()
{
#define CACHE_FLASH_CTRL_REG 0x3ff0000C
#define CACHE_FLUSH_START_BIT BIT0
#define CACHE_EMPTY_FLAG_BIT BIT1
#define CACHE_READ_EN_BIT BIT8

	uint32_t ctrl[16];
	uint32_t addr[16];

	static const uint32_t testValue PROGMEM = 0;

	for(unsigned bank = 0; bank < 16; ++bank) {
		Cache_Read_Enable(bank % 2, bank / 2, 1);
		ctrl[bank] = READ_PERI_REG(CACHE_FLASH_CTRL_REG);
		addr[bank] = flashmem_get_address_CHECK(&testValue);
	}

	Cache_Read_Enable(0, 0, 1);

	for(unsigned i = 0; i < 16; ++i) {
		debug_i("ctrl[%u] = 0x%08x, BIT25 = %u, BITS16-18 = %2u, addr = 0x%08x", i, ctrl[i], (ctrl[i] & BIT25) ? 1 : 0,
				(ctrl[i] >> 16) & 0x07, addr[i]);
	}

	/*

ctrl[0]  = 0x04000102, BIT25 = 0, BITS16-18 =  0, addr = 0x0002e440
ctrl[1]  = 0x06000102, BIT25 = 1, BITS16-18 =  0, addr = 0x0012e440
ctrl[2]  = 0x04010102, BIT25 = 0, BITS16-18 =  1, addr = 0x0022e440
ctrl[3]  = 0x06010102, BIT25 = 1, BITS16-18 =  1, addr = 0x0032e440
ctrl[4]  = 0x04020102, BIT25 = 0, BITS16-18 =  2, addr = 0x0042e440
ctrl[5]  = 0x06020102, BIT25 = 1, BITS16-18 =  2, addr = 0x0052e440
ctrl[6]  = 0x04030102, BIT25 = 0, BITS16-18 =  3, addr = 0x0062e440
ctrl[7]  = 0x06030102, BIT25 = 1, BITS16-18 =  3, addr = 0x0072e440
ctrl[8]  = 0x04040102, BIT25 = 0, BITS16-18 =  4, addr = 0x0082e440
ctrl[9]  = 0x06040102, BIT25 = 1, BITS16-18 =  4, addr = 0x0092e440
ctrl[10] = 0x04050102, BIT25 = 0, BITS16-18 =  5, addr = 0x00a2e440
ctrl[11] = 0x06050102, BIT25 = 1, BITS16-18 =  5, addr = 0x00b2e440
ctrl[12] = 0x04060102, BIT25 = 0, BITS16-18 =  6, addr = 0x00c2e440
ctrl[13] = 0x06060102, BIT25 = 1, BITS16-18 =  6, addr = 0x00d2e440
ctrl[14] = 0x04070102, BIT25 = 0, BITS16-18 =  7, addr = 0x00e2e440
ctrl[15] = 0x06070102, BIT25 = 1, BITS16-18 =  7, addr = 0x00f2e440

	 */
}

```

Sming/Arch/Esp8266/Components/spi_flash/flashmem.c
mikee47: Thanks, the address was already in the readme, just need a little clarification so have updated the doc and posted some code below.

mikee47:
And this is what rBoot does, for comparison:

```
val = conf.roms[conf.current_rom];
val /= 0x100000;

rBoot_mmap_2 = val / 2;
rBoot_mmap_1 = val % 2;
		
Cache_Read_Enable(rBoot_mmap_1, rBoot_mmap_2, 1);
```

Sming/Arch/Host/Components/spi_flash/flashmem.cpp
```diff
@@ -160,6 +179,10 @@ uint32_t flashmem_write(const void* from, uint32_t toaddr, uint32_t size)
 
 uint32_t flashmem_read(void* to, uint32_t fromaddr, uint32_t size)
 {
+	if(readRealMem(to, fromaddr, size)) {
```
slaff: Just for my understanding: 

Let's say a developer has the following application that she/he wants to test on the host emulator:

1. App with two ROMs.
2. No matter which ROM is selected the app should be able to read data stored in the flash starting from the 3rd MB.
3. No matter which ROM is selected the app should be able to write data to the flash at an address starting from the 3rd MB.

That was working correctly up to now. Will it continue to work with this PR?

mikee47: Yes, it'll work exactly as before. Behaviour only changes for reads above 2GB so we're fine below that.

Both Linux and Windows reserve addresses above 2GB for system use so we'll never see them passed as pointers from applications. All that's happening is we're mapping the application's *real* pointer addresses into this unused space.

PR #1965, opened 2019-11-13
Adding precaching from Arduino for ESP8266.

slaff:
Function and macros for pre-loading code into the flash cache
 
 - It can be useful for code that accesses/uses SPI0 which is connected to the flash chip.
- Non interrupt handler code that is infrequently called but might otherwise
	 require being in valuable IRAM - such as bit-banging I/O code or some code
 	 run at bootup can avoid being permanently in IRAM.

Original code from this commit: https://github.com/esp8266/Arduino/commit/692e542d587145062ac8d90aae91cabb1b649f2e


mikee47:
This looks really interesting - maybe reclaim some IRAM! Some suggestions:

* If I understand this correctly, users should only require use of the `PRECACHE_ATTR` tag, so may I suggest taking that out and adding to `Arch/Esp8266/Components/esp8266/include/esp_attr.h`.
* Add a corresponding definition to `Arch/Host/Components/esp_hal/include/esp_attr.h`.
* Move the code and remaining header into the `spi_flash` Component.

The hooks for all this are in the arduino `core_esp8266_spi_utils.cpp` code but I would want to integrate that functionality into the SPI driver as it's going to be the arbiter for all the SPI stuff.

slaff:
> If I understand this correctly, users should only require use of the PRECACHE_ATTR tag

I might be wrong be I think the PRECACHE_ATTR is required for functions that will use inside them the Instruction cache preloading. But it is not enough. What else needs to be done is to mark the starting and the ending block of instructions with PRECACHE_START() and PRECACHE_END(). Excerpt from this commit: https://github.com/esp8266/Arduino/commit/60c8975feb1e81aa06d60aa0d0c84a6ee0b82f51

```C++
static SpiOpResult PRECACHE_ATTR
_SPICommand(volatile uint32_t spiIfNum,
            uint32_t spic,uint32_t spiu,uint32_t spiu1,uint32_t spiu2,
            uint32_t *data,uint32_t writeWords,uint32_t readWords)
{ 
  if (spiIfNum>1)
     return SPI_RESULT_ERR;

//..

  // preload any constants and functions we need into variables
  // Everything defined here must be volatile or the optimizer can
  // treat them as constants, resulting in the flash reads we're
  // trying to avoid
  uint32_t (* volatile Wait_SPI_Idlep)(SpiFlashChip *) = Wait_SPI_Idle;
  volatile SpiFlashChip *fchip=flashchip;
  volatile uint32_t spicmdusr=SPICMDUSR;

  if (!spiIfNum) {
     // Only need to precache when using SPI0
     PRECACHE_START();
     Wait_SPI_Idlep((SpiFlashChip *)fchip);
  }

  //..
  // Restore saved registers
  SPIREG(SPI0U) = oldSPI0U;
  SPIREG(SPI0U2)= oldSPI0U2;
  SPIREG(SPI0C) = oldSPI0C;

  PRECACHE_END();
  return (timeout>0 ? SPI_RESULT_OK : SPI_RESULT_TIMEOUT);
}
```

mikee47:
That's helpful, thanks. I'll make some changes and push to your repo later today, purely as suggestions.

slaff:
> I'll make some changes and push to your repo later today, purely as suggestions.

Thanks a lot. Feel free to modify the code as much as you want.

mikee47:
@slaff Sorry I haven't got around to this one. Next on the list, promise!

tests/HostTests/app/test-precache.cpp
```diff
@@ -0,0 +1,33 @@
+#include "common.h"
```
mikee47: #include <SmingTest.h>

PR #1981, opened 2019-11-25
Webcam stream and sample webcam web server.

slaff:
At the moment we have driver only for the fake/test driver:

image


mikee47:
I'm thinking this should be in Libraries, let's keep Components for core dependencies?

slaff:
> I'm thinking this should be in Libraries, let's keep Components for core dependencies?

But Libraries is only for Arduino compatible Libraries.

mikee47:
I think this is the more important definition:

*Components: Framework support code, not to be used directly by applications.*

From https://sming.readthedocs.io/en/latest/_inc/Sming/building.html#directory-layout.

Libraries is for everything else, i.e. libraries ported to (or written for) Sming. If you want to further sub-divide it then we could add another directory for 'non-arduino libraries', not sure what the advantage would be though.

slaff:
> Libraries is for everything else, i.e. libraries ported to (or written for) Sming

I like the idea and, if I am not mistaken, that will be the first pure-Sming library inside Libraries. I will try to make the recommended changes till the end of the week.

samples/WebcamServer/component.mk
```diff
@@ -1,5 +1,3 @@
-#ARDUINO_LIBRARIES := ArduCAM
-
-COMPONENT_DEPENDS := WebCam
+ARDUINO_LIBRARIES := WebCam
```
slaff: @mikee47 what bothers me is the name of the variable `ARDUINO_LIBRARIES`. Maybe we should deprecate it and add an alias called `LIBRARIES`. I am open to suggestions though.

mikee47: I think we should keep ARDUINO_LIBRARIES if only for the sanity of users!

We can't just keep adding stuff to the Libraries (and samples) directory indefinitely, it's feeling a bit unweildy already! I'll open an issue to discuss this.

PR #1984, opened 2019-11-26
Rename `custom-heap` Component as `heap`

mikee47:
This moves the `ENABLE_CUSTOM_HEAP` variable out of Sming and into the Component,
which simplifies the build logic.

This is also a pre-cursor to adding support for newlib 2 which the newer toolchains use.


slaff:
@mikee47 This PR looks good to me. Do you have something to add?

mikee47:
All done thanks.

PR #1987, opened 2019-11-27
Improve doxygen integration with documentation

mikee47:
Work towards a more streamlined and better integrated documentation.

Pre-built docs. are at https://smingdev.readthedocs.io/en/latest/

Summary of changes:

* Components specify code to be scanned by Doxygen in its component.mk file
* Add samples contained within Components - won't appear in main index
* Re-organise some doxygen groups
* With Doxygen integrated more use can be made of :c: and :cpp: domain references
* Eliminate duplication of documentation; it's in the .rst files or in the source comments but not both
* Remove excessive commenting, such as commenting individual overloaded methods when they can be treated as a group (e.g. WString.h)
* Resize sample images, reduces tree size by 8.85MB (24.69MB -> 15.84MB)
* Add `framework` documentation section to reflect source tree layout.

Notes:

* There are a lot of 'duplicate definition' warnings generated by Sphinx during the build. They don't cause any issues with the documentation. I have yet to find a way to deal with those, may have to just accept this as part of the build.
* Documentation of streaming classes needs a summary of how it fits in with Arduino Stream, Print, etc. and perhaps a list of classes. Pulling in documentation for every variant is probably unhelpful though.


mikee47:
Ready for review. There are no changes to the source code other than for documentation.

I've added the `framework` section which I named to avoid confusion with `API Documentation` for the 'legacy' doxygen documentation. I can't see a problem in having both, certainly at present we don't have a way to replicate the relationship diagrams from doxygen.

We now have in #1992 the ModbusMaster library containing its own sample, rather than adding it to the main samples directory. This structure is reflected in the generated documentation https://smingdev.readthedocs.io/en/latest/_inc/Sming/Libraries/ModbusMaster/index.html.

If there is general agreement with this approach then I can prepare a further PR to move the other samples out as per #1993.

mikee47:
Other things we could do:

* Rename `Arduino Libraries` to just `Libraries`
* Rename `API Documentation` to something else...

slaff:
> Rename Arduino Libraries to just Libraries

Yes, please.

> Rename API Documentation to something else...

Leave it for the moment.

slaff:
@mikee47 Is this PR ready? Shall I schedule it for merging?

mikee47:
Yes, nothing further thanks.

mikee47:
Oh, except there's a note at the top of the main README says docs are only for develop, but we have a release now! I'll fix that.

slaff:
Work towards a more streamlined and better integrated documentation.


Summary of changes:

* Components specify code to be scanned by Doxygen in its component.mk file
* Add samples contained within Components - won't appear in main index
* Re-organise some doxygen groups
* With Doxygen integrated more use can be made of :c: and :cpp: domain references
* Eliminate duplication of documentation; it's in the .rst files or in the source comments but not both
* Remove excessive commenting, such as commenting individual overloaded methods when they can be treated as a group (e.g. WString.h)
* Resize sample images, reduces tree size by 8.85MB (24.69MB -> 15.84MB)
* Add `framework` documentation section to reflect source tree layout.

Notes:

* There are a lot of 'duplicate definition' warnings generated by Sphinx during the build. They don't cause any issues with the documentation. I have yet to find a way to deal with those, may have to just accept this as part of the build.
* Documentation of streaming classes needs a summary of how it fits in with Arduino Stream, Print, etc. and perhaps a list of classes. Pulling in documentation for every variant is probably unhelpful though.

PR #1996, opened 2019-12-10
bugfix: Hangs at startup when custom heap enabled

### LINKED ISSUE #1994: Custom heap broken
Code built with ENABLE_CUSTOM_HEAP=1 doesn't run.

See https://github.com/SmingHub/Sming/pull/1825#issuecomment-563994482.

e.g. `Basic_Blink` sample:

`make flash ENABLE_CUSTOM_HEAP=1 COM_SPEED=74880`

builds and flashes, stops here:

```
python -m serial.tools.miniterm --raw --encoding ascii /dev/ttyUSB0 74880
--- Miniterm on /dev/ttyUSB0  74880,8,N,1 ---
--- Quit: Ctrl+] | Menu: Ctrl+T | Help: Ctrl+T followed by Ctrl+H ---

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x40100000, len 1116, room 16 
tail 12
chksum 0x0d
ho 0 tail 12 room 4
load 0x3ffe8000, len 576, room 12 
tail 4
chksum 0xdd
csum 0xdd

rBoot v1.4.2 - richardaburton@gmail.com
Flash Size:   32 Mbit
Flash Mode:   DIO
Flash Speed:  40 MHz

Booting rom 0 at 2000, load addr 30f90.

```

This was supposed to have been fixed in #1745, but that commit doesn't work either.

mikee47:
Crashes at startup if `umm_info` is called before heap initialised.
Fixes #1994.

### DIFF
--- a/Sming/Arch/Esp8266/Components/heap/README.rst
+++ b/Sming/Arch/Esp8266/Components/heap/README.rst
@@ -17,3 +17,14 @@ Configuration variables
 
 .. warning::
    Do not enable custom heap allocation and -mforce-l32 compiler flag at the same time.
+
+
+.. envvar:: UMM_FUNC_IRAM
+
+   Default: 1 (enabled)
+
+   Custom heap functions are stored in IRAM by default for performance reasons.
+
+   If you need the IRAM (about 1.5K bytes) then disable this option::
+   
+      make ENABLE_CUSTOM_HEAP=1 UMM_FUNC_IRAM=0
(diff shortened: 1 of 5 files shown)


mikee47:
Something I'm chewing over is whether to put all the heap functions into IRAM. (that's where they  are for non-custom heap.) The trade off:

* Normal heap: 5646
* Custom heap: 6486
* Custom heap IRAM: 5054

It's a fair chunk of IRAM which is why I'm humming and hawing. Any thoughts?

slaff:
> Any thoughts?

Put them in IRAM. If an application is very low on IRAM the developers can switch to the latest GCC and turn-off custom heap.

slaff:
That change:
```patch
 void *umm_info( void *ptr, int force ) {
+  if(umm_heap == NULL) {
+    umm_init();
+  }
```

should be proposed to the upstream repository: https://github.com/rhempel/umm_malloc.git

slaff:
@mikee47 Do you have something to add here?

mikee47:
Nope, all done.

Issue #1993, opened 2019-12-09
ECDHE-RSA for lightweight TLS handshake


cometurrata:
Hello,
it seems that Sming could benefit from enabling ECDHE-RSA instead of DHE-RSA. (Faster and lighter TLS).

https://security.stackexchange.com/questions/14731/what-is-ecdhe-rsa

Unfortunately axtls does not provide support for ECDHE, but mbed-tls does !

Any particular reason for not using mbed-tls ?
Has it already been discussed ?
Any hint of what I should start with ?
Do you think it could bring something useful ?

Thanks.

slaff:
> Any particular reason for not using mbed-tls ?

We have already started discussing this in #1713. What we lack is an abstraction that would allow us to switch during linking type between the different SSL implementations. 

> Any hint of what I should start with ?

@cometurrata if you want you can start with building that abstraction and adding axtls and mbed-tls implementations?

slaff:
@cometurrata PR #1999 will be merged soon. Once that is done you can add the new SSL adapter based on mbed-tls.

slaff:
@cometurrata the SSL adapters are now [merged](https://github.com/SmingHub/Sming/commit/41d03e8747b37e93f3f7d9e8068e3788bc2ee3ee)  in the `develop` branch. Bearssl supports ECDHE https://bearssl.org/support.html

Issue #1713, opened 2019-06-01
Future network support - LWIP and SSL


mikee47:
## LWIP versions

LWIP support is indicated by the `ENABLE_CUSTOM_LWIP` option:

0: Version provided with Esp8266 SDK
1: esp-open-lwip, sdk-1.5.0, 2015-12-30, last update 2016-05-26)
2: LWIP open version 2 (STABLE-2_0_2_RELEASE_VER, 2017-03-13, last update 2019-05-03)

At present the Host Emulator supports only LWIP2 (version 2.1, 2019-06-03).

ESP32-IDF uses version based on LWIP 2.0.3 (2017-09-15)

With the Esp8266 I've only really used the default, version 1, and not clear on the relative merits of the versions. Is there any reason we cannot standardise on the most recent LWIP 2 version ?

## SSL support

This is provided for the Esp8266 via AxTLS. I've started work adding this to the emulator, but a couple of issues have come to light:

1. The Arduino Esp2866 project supports both AxTLS and BearSSL, whilst ESP32 uses MBED TLS and the ESP8266 RTOS includes support for Wolf SSL.

I am not familiar with the relative merits of the various SSL/TLS implementations, however it seems clear that what is appropriate for the Esp8266 is perhaps rather limiting for the Esp32. That means moving any implementation-specific defintions or code into the related `Component`.

2. At present the framework has to be recompiled entirely to add SSL support, however only the modules in `Core/Network` are affected (plus the user application).

One solution to this would be to separate all the SSL stuff out into a separate library, so we wouldn't need a separate `libsmingssl.a`, just the regular `libsming.a`. All the SSL stuff would be in `libaxtls.a`.

slaff:
> One solution to this would be to separate all the SSL stuff out into a separate library, so we wouldn't need a separate libsmingssl.a, just the regular libsming.a. All the SSL stuff would be in libaxtls.a.

Great, that exactly what we should have for the SSL part.

> That means moving any implementation-specific defintions or code into the related Component.

Again true.

mikee47:
>
> All the SSL stuff would be in libaxtls.a.

slaff:
> Is there any reason we cannot standardise on the most recent LWIP 2 version ?

That is probably the best way to go in terms of maintenance effort. We have to check how stable and fast it is and what is the memory footprint. 
[...]

Issue #1444, opened 2018-09-24
AWS iot with Sming Certificate Problem


Vinimuller:
Hey! I'm trying to use AWS iot via MQTT, but it requires a certificate to make the connection. I have already used MQTT without certification, but I think that I'm having problems with the way that I'm passing the certificate and key to the function.
When I try to connect to the broker I get :

> please start sntp first !
Error: Invalid X509 ASN.1 file (X509 not ok)

Here is how I'm passing the certificate:

`	mqtt->addSslOptions(SSL_SERVER_VERIFY_LATER);

	unsigned int private_key_len = 1191, certificate_len = 861;
	unsigned char private_key[2000], certificate[2000];
	file_t cert,key;

	if(fileExist("cert.der")){

		cert = fileOpen("cert.der", eFO_ReadOnly);
		fileRead(cert,certificate,certificate_len);

	}else{
		Serial.println("CERT NOT FOUND");
	}

	if(fileExist("key.der")){
		
		key = fileOpen("key.der", eFO_ReadOnly);
		fileRead(key,private_key,private_key_len);

	}else{
		Serial.println("KEY NOT FOUND");
	}

	mqtt->setSslClientKeyCert(private_key, private_key_len,
							  certificate, certificate_len, NULL, true);

	fileClose(cert);
	fileClose(key);`

I'm running sming on Ubuntu16.04 and using ESP12F

slaff:
I guess your certificate is NOT in DER format. Open it with text editor and if you can read the content then it is in PEM (text format). Convert it to DER(binary format). Take a look here how this can be done: https://github.com/igrr/axtls-8266/blob/0c3a9f722f11799fbeda1f99f9d9ab77a82a4489/tools/make_certs.sh#L127

slaff:
> Do you know if I'm calling the functions in the right sequence?

Move the `connect` after you set the client certificate. In the code above this means to move `mqtt->connect("sdk-nodejs-2ada784d-7605-4f75-a3f2-c46dd3c84f40",true);` after `mqtt->setSslClientKeyCert...`

Vinimuller:
Moving the connect gives me the same output still.
Function:
````
void startMqttClient()
{
	procTimer.stop();

	if(!fileExist("cert.der")){
		Serial.println("cert not found");
	}

	if(!fileExist("key.der")){
		Serial.println("key not found");
	}
	
	mqtt->addSslOptions(SSL_SERVER_VERIFY_LATER);

[...]

PR #1999, opened 2019-12-12
Feature: ssl abstraction

slaff:
This is the initial draft for a SSL abstraction that will allow Sming users to change seamlessly between different  SSL implementations and without the need to recompile the whole framework.

This PR is still Work-In-Progress(WIP).

Related to #1993 and #1713.

We now have a single `ssl` Component which contains code to use the selected implementation. It generates a library variant as required, including when SSL is disabled. As the public interface is fixed, no recompilation of the framework is required when switching between implementations.

* [x] Review the new structure, especially the interface between Sming and the `ssl` Component. Does it make sense? Is should be logical and easy to follow - if not, fix it.
* [x] Simplify interface code as much as possible. In particular, avoid having Sming digging into internal parts of `Ssl::Session`. Change this into a class with methods.
* [x] The `lwirax` compatibility layer is now probably surplus to requirements - we don't need two levels of abstraction - so remove this and work with `axtls-8266` directly.
* [x] Basic_Ssl `onDownload` doesn't print certificate details. This is because axTLS disposes of them on handshake completion. We can (a) tell axTLS to keep hold of the certificates by setting `ssl->can_free_certificates = false`, or (b) construct `CertificateImpl` at that point and take a copy of the certificate data, (c) add an `onSslHandshakeComplete` callback for application use. Probably do (a) with an option flag.
* [x] Add BearSSL implementation. This will shake out any issues with the interface implementation.
* [x] Test/verify that SSL server works (both Host + ESP8266). See https://sming.readthedocs.io/en/latest/experimental/httpserver-ssl.html.
* [ ] Test if the SSL clients can send client certificates. Use #1444 for more realistic test.
* [ ] Test if the SSL server (at least on Host) can verify client certificates.
* [ ] Test if the SSL server (at least on Host) can set also CA certificate.
* [x] Revise application configuration layer using 'on SSL init' callback
* [x] Move certificate generation scripts, etc. into SSL component so they're not tied to any specific implementation.
* [ ] Check all file comment headers are correct and in place
* [x] Update the Sphinx documentation.


mikee47:
Following on from https://github.com/slaff/Sming/pull/40

> Instead of global functions, how about a factory class instead with createExtension(), createContext(), etc? This could be a global pointer SslImplementation* sslImpl which is null if SSL isn't enabled, then we could get rid of need for a dummy SSL component.

Further, let's shift everything into the project. Get rid of `ENABLE_SSL`, etc. then add this to the Project's component.mk file:

```
COMPONENT_DEPENDS += AxtlsSsl
```

The application then explicitly registers an SSL implementation:

```
// We'd like to use AXTLS thankyouverymuch.
static AxtlsSsl sslImpl;

void init()
{
  ...
  // Internally this sets a pointer to the implementation
  sming:ssl:register(&sslImpl);
  ...
}
```

If no implementation is registered, or it's null, then no SSL is used.

> Initially I had axtls-8266 and the AxtlsSsl implementation in one component. But then decided against it. axtls-8266 is containing Ssl functions and cryptographic functions, that are quite useful. If someone wants to use the cryptographic functions from axtls-8266 but prefer to have dummy ssl implementation having `AxtlsSsl` as a separate component makes this possible.

With the above approach, we can add multiple SSL Components to a project and pick the one to use, whilst allowing access to the code from other libraries. That means there's no problem having `axtls-8266` contained within `AxtlsSsl` which I feel is the correct approach.

Future work might include additional abstractions for the crypto stuff to avoid project's being tied to any specific implementation.

Finally, now that `AxtlsSsl` is fully decoupled from the Sming framework it can be moved into `Libraries`.

Sming/Core/Network/TcpConnection.cpp
```diff
@@ -42,16 +42,17 @@ bool TcpConnection::connect(const String& server, int port, bool useSsl, uint32_
 	ip_addr_t addr;
 
 	this->useSsl = useSsl;
-#ifdef ENABLE_SSL
+
```
mikee47: I do wonder about having two variables (`useSsl` and `sslOptions`). We could simplify the entire interface for this by passing an `Ssl::Options*` structure instead of these two parameters; if it's null, we're not using SSL. (See above notes regarding `Ssl::Options`.)

Sming/Core/Network/TcpConnection.h
```diff
@@ -273,24 +218,22 @@ class TcpConnection
...
+	SslConnection* ssl = nullptr;
+	SslExtension* sslExtension = nullptr;
 	bool sslConnected = false;
 	uint32_t sslOptions = 0;
 	SslKeyCertPair sslKeyCert;
 	bool freeKeyCertAfterHandshake = false;
 	SslSessionId* sslSessionId = nullptr;
-#endif
+
 	bool useSsl = false;
```
mikee47: Could wrap all of these up in a single structure or class:

```
struct SslSession {
  Ssl::Context* context = nullptr;
  Ssl::Connection* connection = nullptr;
  Ssl::Extension* extension = nullptr;
  Ssl::SessionId* sessionId = nullptr;

  bool connected = false;
  Ssl::Options options;
  Ssl::KeyCertPair keyCert;
  bool freeKeyCertAfterHandshake = false;

 ~SslSession()
  {
    delete context;
    delete connection;
    delete extension;
  }
};

SslSession* ssl;
```

Don't need `useSsl` any more since we just have a single check for ssl == nullptr. We also get some RAM back for non-SSL use. All (or most) of the related SSL code can also be moved into methods of this class.

samples/Echo_Ssl/app/application.cpp
```diff
@@ -7,6 +7,7 @@
  */
 
 #include <SmingCore.h>
+#include "Data/HexString.h"
```
mikee47: `#include <Data/HexString.h>` trying to standardise

samples/Echo_Ssl/app/application.cpp
```diff
@@ -25,51 +26,13 @@ TcpClient* client;
...
-		Serial.printf("SSL_AES256_SHA256");
-		break;
-
-	default:
-		Serial.printf("Unknown - %d", ssl_get_cipher_id(ssl));
-		break;
+	if(sessionId != nullptr && sessionId->isValid()) {
+		debugf("-----BEGIN SSL SESSION PARAMETERS-----");
+		debugf("%s", makeHexString(sessionId->getValue(), sessionId->getLength()).c_str());
+		debugf("-----END SSL SESSION PARAMETERS-----");
```
mikee47: Maybe add some of this debug stuff to the framework using `Print&` instead of debug functions.

samples/HttpClient/app/application.cpp
```diff
@@ -1,5 +1,5 @@
 #include <SmingCore.h>
-
+#include "Data/HexString.h"
 #include "Network/HttpClient.h"
```
mikee47: `<...>`

Sming/Core/Network/Ssl/Interface/SslCertificate.h
```diff
@@ -0,0 +1,69 @@
...
+    * @retval bool true on match, false otherwise
+    */
+	virtual bool matchPki(const uint8_t* hash) const = 0;
+
+	/**
+   * @brief Retrieve an X.509 distinguished name component
+   * @param name the desired distinguished name
+   * @retval the value for the desired distinguished name
+   */
+	virtual const String getName(const SslCertificateName& name) const = 0;
```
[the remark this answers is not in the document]
mikee47: Don't need `const` return type, just `virtual String getName(CertName name) const = 0`

mikee47:
Is there any accepted standard for C++ crypto/SSL library interfaces? I've found https://github.com/weidai11/cryptopp and there are likely boost libraries around.

General question: What sort of SSL/crypto code might we want to run on Sming? What's it written for? How can we simplify porting?

Also re. namespaces. A quick Sming search reveals ContentType, MallocCount, FSTR, NanoTime, PolledTimer, Json, RingTone, Profiling. So it seems logical that we'll pick `Sming::Ssl`, reserving lowercase convention for the STL.

If we end up namespacing everything in Sming (then using aliases for backward-compatibility) then it might even end up being `Sming::Tcp::Ssl`. It would then be logical to have all the classes renamed such as `Tcp::Client`, `Tcp::Connection`, etc. And would we have `Tcp::Http::Connection` or just `Http::Connection`?

mikee47:
Is there any accepted standard for C++ crypto/SSL library interfaces? I've found https://github.com/weidai11/cryptopp and there are likely boost libraries around.

General question: What sort of SSL/crypto code might we want to run on Sming? What's it written for? How can we simplify porting?

Also re. namespaces. A quick Sming search reveals ContentType, MallocCount, FSTR, NanoTime, PolledTimer, Json, RingTone, Profiling. So it seems logical that we'll pick `Sming::Ssl`, reserving lowercase convention for the STL.

If we end up namespacing everything in Sming (then using aliases for backward-compatibility) then it might even end up being `Sming::Tcp::Ssl`. It would then be logical to have all the classes renamed such as `Tcp::Client`, `Tcp::Connection`, etc. And would we have `Tcp::Http::Connection` or just `Http::Connection`?

Sming/Core/Network/TcpConnection.h
slaff: > Don't need useSsl any more since

useSsl is describing if the user requested SSL. For example when the URL proto is https instead of http.
sslSession is  describing if there is available Ssl implementation.  A tcp client with set ssl can use create SSL or non-SSL connection. If useSsl is true and there is no ssl implementation then the TCP connection should be aborted with also debug error message.

Sming/Core/Network/TcpConnection.cpp
slaff: > useSsl

Probably that should stay in TcpConnection. The rest should be moved to a struct as you have suggested.

Sming/Core/Network/Ssl/Interface/SslExtension.h
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * @brief Gets pointer to the internal object responsible for the SSL extensions.
+	 * 		  The SslExtension object is the one owning the internal object and it will be freed
+	 * 		  once the SslExtension object is freed.
+	 * @retval void *
+	 */
+	virtual void* getInternalObject() = 0;
+};
+
+/** @} */
```
mikee47: If I understand this correctly, the extension is just a set of options which can be applied to a context when creating a connection. I'm not sure it needs such an elaborate interface. In fact, I don't think this should be an interface class at all, just a regular class (or struct):

```
struct SslExtension {
  String hostName;
  SslExtensionFragmentSize fragmentSize;
};
```

This information is then assembled as required in the SslContext implementation.

Sming/Core/Network/Ssl/Interface/SslExtension.h
```diff
@@ -0,0 +1,67 @@
...
+enum SslExtensionFragmentSize {
+	// 0,1,2,3..6 corresponding to off,512,1024,2048..16384 bytes
+	eSEFS_Off,
+	eSEFS_512, // << 512 bytes
+	eSEFS_1K,  // << 1024 bytes
+	eSEFS_2K,
+	eSEFS_4K,
+	eSEFS_8K,
+	eSEFS_16K,
+};
```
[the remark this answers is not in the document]
mikee47: Maybe not... FragmentSize::512 wouldn't be valid. However, as a generic interface perhaps just specifying the size would be better, i.e. 0, 512, 1024, 16384, etc. and the implementation then rounds it to the nearest appropriate boundary.

Sming/Components/AxtlsSsl/src/SslContextImpl.cpp
```diff
@@ -0,0 +1,73 @@
...
+											  SslExtension* sslExtension)
+{
+	int clientfd = axl_append(tcp);
+	if(clientfd < 0) {
+		debug_d("SSL: Unable to add LWIP tcp -> clientfd mapping");
+		return nullptr;
+	}
+
+	SSL_EXTENSIONS* extensions = reinterpret_cast<SSL_EXTENSIONS*>(sslExtension->getInternalObject());
+	SSL* ssl = ssl_client_new(context, clientfd, sessionData, sessionLength, extensions);
```
mikee47: `sslExtension` needs a nullptr check. If it's not optional, parameter should be a reference. 
See comment about simplifying extension, so we'd put the code to create the SSL_EXTENSIONS structure here from that information.

Sming/Components/AxtlsSsl/src/SslExtensionImpl.h
```diff
@@ -0,0 +1,42 @@
...
+	bool setMaxFragmentSize(const SslExtensionFragmentSize& fragmentSize) override;
+
+	void* getInternalObject() override
+	{
+		return sslExtension;
+	}
+
+	virtual ~SslExtensionImpl()
+	{
+		free(sslExtension);
```
mikee47: Call to `ssl_ctx_free` frees all contained connections including any SSL_EXTENSIONS structure pointers, so freeing it in the destructor here isn't safe. Can remove this class entirely if we simplify as outlined above.

Sming/Components/AxtlsSsl/src/SslConnectionImpl.h
```diff
@@ -0,0 +1,62 @@
...
+
+	SslSessionId* getSessionId() override;
+
+	SslCertificate* getCertificate() override;
+
+	virtual ~SslConnectionImpl()
+	{
+		delete certificate;
+		delete sessionId;
+		certificate = nullptr;
```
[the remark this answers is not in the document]
slaff: > Shouldn't need to set values to null in a destructor. I normally don't,

I have a very dim memory of an issue reported from some of the static code analysis tools mentioning something about copying objects without explicit copy constructor ... But I might be wrong. For now I will remove the `= nullptr` lines.

Sming/Components/AxtlsSsl/src/SslContextImpl.cpp
```diff
@@ -0,0 +1,79 @@
...
+		return nullptr;
+	}
+
+	return new SslConnectionImpl(ssl);
+}
+
+bool SslContextImpl::loadMemory(const SslContextObject& memType, const uint8_t* data, size_t length,
+								const char* password)
+{
+	return (ssl_obj_memory_load(context, memType, data, length, password) != SSL_OK);
```
mikee47: == SSL_OK

Sming/Components/AxtlsSsl/src/SslContextImpl.cpp
```diff
@@ -0,0 +1,79 @@
...
+{
+	int clientfd = axl_append(tcp);
+	if(clientfd < 0) {
+		debug_d("SSL: Unable to add LWIP tcp -> clientfd mapping");
+		return nullptr;
+	}
+
+	SSL* ssl = ssl_server_new(context, clientfd);
+	if(ssl == nullptr) {
+		return nullptr;
```
mikee47: add `delete extensions;` - see implementation of `ssl_client_new`

Sming/Components/AxtlsSsl/src/SslContextImpl.cpp
```diff
@@ -0,0 +1,79 @@
...
+	SSL_EXTENSIONS* extensions = nullptr;
+	if(sslExtension != nullptr) {
+		extensions = ssl_ext_new();
+		ssl_ext_set_host_name(extensions, sslExtension->hostName.c_str());
+		ssl_ext_set_max_fragment_size(extensions, sslExtension->fragmentSize);
+	}
+
+	SSL* ssl = ssl_client_new(context, clientfd, sessionData, sessionLength, extensions);
+	if(ssl == nullptr) {
+		return nullptr;
```
mikee47: add `delete extensions;` - see implementation of `ssl_client_new`

slaff:
@mikee47 if you want take this PR and develop it further. I like your suggestions a lot. With the speed and spare time that I have it will take me ages to finish it (or at least not before February, 2020).

mikee47:
@slaff OK. If you want to push any outstanding stuff to your branch we can work from there?

slaff:
> If you want to push any outstanding stuff

Just take it as it is. Create a clone in your own repository of this branch. After that close this PR and open a new PR with your branch.

Sming/Core/Network/Ssl/Interface/SslConnection.h
```diff
@@ -0,0 +1,133 @@
...
+#define SSL_ALERT_UNSUPPORTED_CERTIFICATE 43
+#define SSL_ALERT_CERTIFICATE_EXPIRED 45
+#define SSL_ALERT_CERTIFICATE_UNKNOWN 46
+#define SSL_ALERT_ILLEGAL_PARAMETER 47
+#define SSL_ALERT_UNKNOWN_CA 48
+#define SSL_ALERT_DECODE_ERROR 50
+#define SSL_ALERT_DECRYPT_ERROR 51
+#define SSL_ALERT_INVALID_VERSION 70
+#define SSL_ALERT_NO_RENEGOTIATION 100
+#define SSL_ALERT_UNSUPPORTED_EXTENSION 110
```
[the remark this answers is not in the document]
mikee47: Looking at BearSSL, for example, error codes appear to be implementation-specific.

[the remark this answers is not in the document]
slaff: > Looking at BearSSL, for example, error codes appear to be implementation-specific.

We could find the base minimum of error codes that hopefully all implementations share and then map them per implementation.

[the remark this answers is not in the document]
mikee47: There's probably only a few codes that applications actually care about so that set should be quite small. I'll add a method for each implementation to convert their own error codes to string.

Sming/Core/Network/Ssl/SslCrypto.h
```diff
@@ -0,0 +1,36 @@
...
+#define MD5_SIZE 16
+#define SHA1_SIZE 20
+#define SHA256_SIZE 32
+
+// The following cryptographic functions should be present in every SSL implementation
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void hmac_md5(const uint8_t* msg, int length, const uint8_t* key, int key_len, uint8_t* digest);
```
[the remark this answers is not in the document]
mikee47: All these MD5 functions are provided in the ESP ROM. This is a good place to bring them out, only need to use axtls for Host support. See also https://github.com/SmingHub/Sming/pull/1999.

mikee47:
@slaff Could you rebase your PR? Made some progress.

slaff:
> Could you rebase your PR?

Yep, should be ready for you.

mikee47:
OK, so we've now got all the SSL stuff in a single Component so it can be worked on separately. A few points about the new structure:

1. The implementations are contained within the `ssl` component and generates a variant as required.
2. As the public interface is fixed, no recompilation of the framework is required when switching between implementations
3. No 'dummy' SSL is required, this is indicated by a null factory pointer.

Generally the code has just been moved around, I've avoided making any significant changes. Proposed next steps:

* Review the new structure, especially the interface between Sming and the `ssl` Component. Does it make sense? Is should be logical and easy to follow - if not, fix it.
* Simplify interface code as much as possible. In particular, avoid having Sming digging into internal parts of `Ssl::Session`. Change this into a class with methods.
* The `lwirax` compatibility layer is now probably surplus to requirements - we don't need two compatibility layers - so remove this and work with `axtls-8266` directly.
* Add BearSSL implementation. This will shake out any issues with the interface implementation.
* Check all file comment headers are correct and in place
* Review documentation in general

Sming/Core/Network/Ssl/Interface/SslContext.h
```diff
@@ -0,0 +1,137 @@
...
+		this->sessionCacheSize = sessionCacheSize;
+		return true;
+	}
+
+	/**
+	 * @brief Use to load into memory certificates, public and private keys, etc.
+	 * @param memType the type of the data. Example: public key, client certificate, etc
+	 * @param data - the data should be in DER format ( https://wiki.openssl.org/index.php/DER )
+	 * @param length
+	 * @param password - null terminated string
```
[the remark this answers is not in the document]
mikee47: Yes.

Sming/Core/Network/Ssl/Interface/SslExtension.h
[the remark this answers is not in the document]
mikee47: These are defined by SSL standard so keep as-is.

slaff:
> The lwirax compatibility layer is now probably surplus to requirements - we don't need two compatibility layers - so remove this and work with axtls-8266 directly.

[Lwirax](https://github.com/attachix/lwirax) is the "glue" between LWIP raw (async) mode and axTLS. We still need it.

mikee47:
> 
> > The lwirax compatibility layer is now probably surplus to requirements - we don't need two compatibility layers - so remove this and work with axtls-8266 directly.
> 
> [Lwirax](https://github.com/attachix/lwirax) is the "glue" between LWIP raw (async) mode and axTLS. We still need it.

I've integrated the necessary bits into `ConnectionImpl.cpp`. If you could rebase when #2009 is merged and I'll push to your repo. from https://github.com/mikee47/Sming/tree/feature/ssl/no-lwirax

slaff:
@mikee47 Thanks a lot for the Session.cpp/h implementation! With your help the SSL abstraction might arrive earlier than expected :)

slaff:
> If you could rebase when #2009 is merged

Done.

slaff:
> fix `make cs` - todo: how to make this pluggable?

We can just add `.cs` file with 0 bytes to every directory where the the coding style rules should be applied. The coding style rules should be valid for the requested directories AND their sub-directories. And then with a simple `find $SMING_HOME/../  -name '.cs' -type f` command we should be able to get a list of all directories where the CS rules should be applied.

mikee47:
@slaff Thanks! That would never have ocurred to me in a milllon years. LOL.

I'm looking at BearSSL now using the [esp8266-arduino](https://github.com/esp8266/Arduino/tree/master/tools/sdk/ssl) directly seems best.

mikee47:
Codacy warning: should have used delete[] instead of delete. Will fix it in next PR.

mikee47:
@slaff Have you any pointers for an SSL server sample application? As far as I can see the samples are all client-focussed.

slaff:
> Have you any pointers for an SSL server sample application?

Take one of the HttpServer samples and try to make the following changes: https://github.com/SmingHub/Sming/wiki/How-to-use-SSL-in-HttpServer

kpishere:
@mikee47 So I'm trying to use your fork with the bear-ssl branch.  My code is not far off from the SSL Basic_Ssl client example.  The factory does not appear to initialize correctly as the stack trace shows this :

3fffff70:  3ffec3a0 3fff0x4025ce24: Ssl::Session::onConnect(tcp_pcb*) at /Volumes/case-sensitive/Sming/**Sming/Components/ssl/src/Session.cpp:82** (discriminator 4)
0x4021b04d: tcp_connect at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/tcp.c:764
0x40105bde: wDev_ProcessFiq at ??:?
0x4024a292: TcpConnection::internalOnConnected(signed char) at /Volumes/case-sensitive/Sming/Sming/Core/Network/TcpConnection.cpp:403
0x4024a585: TcpConnection::internalOnDnsResponse(char const*, ip_addr*, int) at /Volumes/case-sensitive/Sming/Sming/Core/Network/TcpConnection.cpp:530
0x4000050c: ?? ??:0
0x4021a228: pbuf_free at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/pbuf.c:1104
0x4000050c: ?? ??:0
0x4021a7c8: tcp_seg_free at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/tcp.c:331
0x4024a2d5: TcpConnection::internalConnect(IpAddress, unsigned short)::{lambda(void*, tcp_pcb*, signed char)#1}::_FUN(void*, tcp_pcb*, signed char) at /Volumes/case-sensitive/Sming/Sming/Core/Network/TcpConnection.cpp:386
 (inlined by) _FUN at /Volumes/case-sensitive/Sming/Sming/Core/Network/TcpConnection.cpp:388
0x4021f851: tcp_process at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/tcp_in.c:728
 (inlined by) tcp_input at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/tcp_in.c:340
0x4021a3aa: pbuf_alloc at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/pbuf.c:1104
0x4021d301: ip_input at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/core/ipv4/ip.c:559
0x4021dc95: ethernet_input at /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip/lwip/netif/etharp.c:1379
0x4022baf7: pp_tx_idle_timeout at ??:?

When I build it, it looks good.  .. 

Mac-mini:vfdTest kevinpeck$ make

vfdTest: Invoking 'all' for Esp8266 (debug) architecture

Building /Users/kevinpeck/Documents/Sming/vfdTest/out/Esp8266/debug/lib/clib-App-ae8e272eaadc92c61df0481c20b413ce.a
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/Calendar.cpp
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/Configuration.cpp
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/SerialReadLine.cpp
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/application.cpp
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/vfdisplay.cpp
C+ /Users/kevinpeck/Documents/Sming/vfdTest/app/webserver.cpp
CC /Volumes/case-sensitive/Sming/Sming/Components/rboot/appcode/rboot-overrides.c
CC /Volumes/case-sensitive/Sming/Sming/Components/rboot/rboot/appcode/rboot-api.c
CC /Volumes/case-sensitive/Sming/Sming/Components/rboot/rboot/appcode/rboot-bigflash.c
AS /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/gdbstub/appcode/gdbstub-entry.S
C+ /Volumes/case-sensitive/Sming/Sming/Arch/Esp8266/Components/gdbstub/appcode/gdb_hooks.cpp
AR /Users/kevinpeck/Documents/Sming/vfdTest/out/Esp8266/debug/lib/clib-App-ae8e272eaadc92c61df0481c20b413ce.a
E2 /Users/kevinpeck/Documents/Sming/vfdTest/out/Esp8266/debug/firmware/rboot.bin
Prepare libmain
LDGEN out/Esp8266/debug/build/rom0.ld
vfdTest: Linking out/Esp8266/debug/build/app_0.out
      data|        Initialized Data (RAM)|    3FFE8000|    3FFE85A0|    1440
    rodata|           ReadOnly Data (RAM)|    3FFE85A0|    3FFE9604|    4196
       bss|      Uninitialized Data (RAM)|    3FFE9608|    3FFF0420|   28184
      text|            Cached Code (IRAM)|    40100000|    40106817|   26647
irom0_text|           Uncached Code (SPI)|    40202010|    402704F2|  451810
Total Used RAM : 33820
Free RAM : 48100
Free IRam : 6139
ESPTOOL2 out/Esp8266/debug/firmware/rom0.bin
Checking for spiffs files
files directory exists. Creating out/Esp8266/debug/firmware/spiff_rom.bin
Creating rom 'out/Esp8266/debug/firmware/spiff_rom.bin' of size 0x40000 (262144) bytes.
Adding files in directory 'files'.
Added 'index.html' to spiffs (2769 bytes).
Added 'index.js' to spiffs (826 bytes).
Using BEARSSL SSL implementation

mikee47:
We need a mechanism for applications to configure all the SSL session options. At present we have a few items stored within an HttpRequest: `sslOptions`, `sslFingerprints` and `sslKeyCertPair`. These all end up in the Session anyway,so a better way I think would be to replace those with a single callback delegate, something like this:

`using SslInitDelegate = Delegate<void(Ssl::Session& session, void* param)>`.

Applications then set:

`request->onSslInit(mySslInitCallback, &request)`.

Or something similar. This gets called by the SSL layer before a connection is created.

It's difficult to tell what additional options we might want to set in the future, so something like this means we only need to update Session to support those and expose them for application use.

slaff:
> so something like this means

Agreed. Go for it.

kpishere:
@mikee47 I can add that i would'nt bother supporting max fragment length across ssl libraries. just implement record size limit. when it becomes available in various libraries. ref : [https://tlswg.org/tls-record-limit/draft-ietf-tls-record-limit.html](url)

mikee47:
Whilst server mode works OK for axTLS, the TCP send buffers just fill up with BRSSL. Could do with some help tracking that one down....

slaff:
> Could do with some help tracking that one down....

I was able to progress a bit so that the server now delivers small pages. Unfortunately for bigger pages the BearSSL engine sets the status to closed in `ssl_hs_server.c` line: 1413

```c
			case 43: {
				/* fail */

	br_ssl_engine_fail(ENG, (int)T0_POPi());
	T0_CO();

				}
				break;

```

before the complete content is delivered.

mikee47:
@slaff Thanks for trying with this. I've finally pinned it down to two things:

1. Need bi-directional buffering for server mode
2. BRSSL can handle partial writes; this differs from axTLS but the TCP logic handles it fine.

slaff:
> Could do with some help tracking that one down....

Fixed. See my latest db0bd16
Merge state
Add more commits by pushing to the feature/ssl-abstraction branch on slaff/Sming.

This branch has not been deployed
No deployments
Some checks haven’t completed yet
2 pending checks

continuous-integration/appveyor/pr Pending — Waiting for AppVeyor build to complete
Details

continuous-integration/travis-ci/pr Pending — The Travis CI build is in progress
Details
This branch has no conflicts with the base branch
Merging can be performed automatically.
 or view .
@slaff
 
 
> Could do with some help tracking that one down....

Fixed. See my latest commit.

Sming/Components/ssl/BearSsl/BrConnection.cpp
```diff
@@ -201,6 +198,14 @@ int BrConnection::runUntil(InputBuffer& input, unsigned target)
 				return BR_ERR_IO;
 			}
 
+			if(wlen == 0) {
+				if(target & BR_SSL_SENDAPP) {
+					return len;
```
mikee47: That looks dodgy; we're saying "yes, 'len' bytes were sent" even though they weren't...

mikee47:
(I've pushed what I think is the 'proper' fix to your repo.)

slaff:
> I've pushed what I think is the 'proper' fix to your repo

I will take a look at it shortly and probably merge the proper fix ;).

slaff:
With the latest commits from @mikee47  the initial abstraction is pretty much ready. 
What is missing is 

Documentation
--------------------
Documentation explaining how to 
- [ ] migrate applications
- [ ] how to write newer SSL adapters

Testing
---------
- [ ] functional tests: client certificates, setting server CA certificates
- [ ] nice to have: comparison between axTLS and BearSSL - memory footprint, async operations and speed.

Final polishing
-------------------
- [ ] correct headers
- [ ] doxygen documentation
- [ ] consistency: consistent way of naming the different SSL adapters. For example `ENABLE_SSL=Axtls` and `ENABLE_SSL=Bearssl` only. Similar to the naming that we have for the SMING_ARCH variables.
- [ ] final review

slaff:
@mikee47 I am planning to merge this PR. At the moment is needs a bit of polishing that can be done also later. Do you have something to add before I merge it?

mikee47:
@slaff I haven't prepared anything, but I'm happy to give it a quick review to see if anything sticks out.

mikee47:
OK, I've pushed some changes to your repo. Checked the documentation build and made some changes to that so it's all in the SSL component.

Sming/Components/ssl/include/Network/Ssl/Session.h
```diff
@@ -0,0 +1,178 @@
...
+ */
+enum FragmentSize {
+	eSEFS_Off, ///< Let SSL implementation decide
+	eSEFS_512, ///< 512 bytes
+	eSEFS_1K,  ///< 1024 bytes
+	eSEFS_2K,
+	eSEFS_4K,
+	eSEFS_8K,
+	eSEFS_16K,
+};
```
mikee47: Still niggles. The best I can think of is this:

```
enum class FragmentSize {
	Default, ///< Let SSL implementation decide
	B512, ///< 512 bytes
	K1,  ///< 1024 bytes
	K2,
	K4,
	K8,
	K16,
};
```

kpishere: When I see this, I'm imagining that FramentSize today refers to the max_fragment_length of today (https://tools.ietf.org/html/rfc6066) and in the future, being library dependent as they update, will refer to record_size_limit of tomorrow (https://tools.ietf.org/html/rfc8449).  Record size limit can be any value where fragment size is only an enumeration of valid sizes.  The enumeration of sizes looks granular enough to be to be usable and would translate just fine between the transition.  If I were to split hairs, the enumeration name isn't generic between the two standards -- your call.

mikee47: At the application level, what information should this convey? If the library handles the detail, my thought is that we just want to limit the buffer size  so we have some control over RAM usage. That being the case, we'd call this `maxBufferSize` and it would just be a hint. Does that cover it, or am I missing something?

mikee47: At the application level, what information should this convey? If the library handles the detail, my thought is that we just want to limit the buffer size  so we have some control over RAM usage. That being the case, we'd call this `maxBufferSize` and it would just be a hint. Does that cover it, or am I missing something?

slaff: > maxBufferSize ... K1,  ///< 1024 bytes

Agreed, make a PR and I will merge it.

Issue #2001, opened 2019-12-13
Sming `No Wifi` build option


mikee47:
I recently came across https://basicengine.org/.

> The BASIC Engine is a very low-cost single-board home computer with advanced 2D color graphics and sound capabilities, roughly comparable to late-1980s or early-1990s computers and video game consoles. It can be built at home without special skills or tools and using readily available components for under 10 Euros in parts, or mass-produced for even less."

It's an extraordinary project, but of interest here is its use of the ESP8266 without WiFi stack to release a ton of RAM. It uses `https://github.com/pvvx/SDKnoWiFi.git` to achieve this. There's also an [Arduino Port](https://github.com/uli/Arduino_nowifi.git).

I think it'd be rather handy to have this as a build option for Sming, but of course it would have to be done properly. SDKnoWiFi has a ton of low-level definitions and essentially re-implements the low-level support code for task queues, timers, etc. which the SDK normally does. I may end up doing that but as we'd need 100% compatibility with the existing SDK it seems like a lot of work and a little risky. (Although most of the code is floating about in various places it needs tidying up, checking, testing, etc.)

I've investigated some alternative approaches with the aim of hooking or patching the offending startup code within `libmain.a` itself, so that the linker discards all the wifi stuff we don't need and gives us the RAM back.

This seems simple enough - just replace calls to `wifi_station_start` and so on with no-ops or calls to dummy functions, but patching ELF object files is rather more involved and I'm disinclined to do that manually; I want a scripted approach.

slaff:
> just replace calls to wifi_station_start and so on with no-ops

So I guess you are not looking for the solution in NoWifiSDK: https://github.com/pvvx/SDKnoWiFi/blob/master/Makefile#L91-L100?

One thing that comes to my mind is that when the name and the signature of a function are known then you can do the following:
- describe all functions that need to be replaced with their unique name and function signature (params + return value). Put them in a yaml file (manual work)
- group the functions by signature. Let's call this function group (FG). (manual work - in the same yaml file)
- weaken the functions from the list based on their names  (can be automated )
- create new file with one noop function per FGs.  (can be automated)
- in the same file alias all functions to their FG noop functions (can be automated)
```
    int wifi_do_something(int x, int y) __attribute__((alias("noop_int_int_int")));
```
- during link time if the noWifiSDK was selected then include the noop file.

mikee47:
Made a start on this, https://github.com/mikee47/Sming/tree/feature/no-wifi.

[...]

PR #2004, opened 2019-12-15
No-WiFi build option

mikee47:
Applications which don't require WiFi can benefit from increased available flash and RAM. This PR allows all non-networking samples to be built with `DISABLE_WIFI=1`. See #2001 for discussion.

There are various items (listed below) which need addressing, but it's pretty functional as-is so I'm proposing we merge it as an experimental feature.

As an example, `ScreenTFT_ILI9163C`:

Memory | WiFi | No-WiFi | Saving
--- | --- | --- | ---
Used flash | 209615 | 60151 | 149464
Free RAM | 54191 | 79564 | 25373
Free IRAM | 6466 | 28021 | 21555

Memory savings should be consistent for any given sample. See the README for more details in the esp_no_wifi component.

Things to investigate/fix:

* [x] Timer calibration - running HostTests shows a disparity between wifi/no-wifi timed results, values are higher (by about 25%)
* [ ] Sleep functions need implementing
* [x] The SDK calls `sleep_reset_analog_rtcreg_8266()` at startup, but this breaks the code. Known issue: do this before initialising PLL.
* [x] LiveDebug sample doesn't work if compiled with `ENABLE_GDB=0`.
* [ ] Evaluate power consumption to ensure device operating as expected and initialisation is correct
* [ ] Implement watchdog init, etc.
* [ ] Fix analogue read functions - phy needs initialising
* [x] Check output from `os_random` is truly random (check entropy) - depends on PHY being initisalised correctly

Issues which would be nice to fix, but don't seem to affect functionality:

* `user_pre_init()` doesn't work, specifically partition registration fails.
* Loading/saving/restoring of system parameters. These are generally RF parameters so probably not required.

**Base changes**

This PR migrates some more headers from the ESP8266 RTOS SDK as they provide necessary register definitions for implementing various low-level functions:

* Remaining dependencies on peri.h removed from UART driver
* Move `peri.h` into `Core/esp8266_peri.h` - these definitions may be required by ported arduino libraries or existing applications, but in general they should be avoided in favour of drivers and other APIs provided by the framework.
* Headers in `esp8266` component override those from SDK, including eagle_soc.h
* 'missing' definitions moved from `esp_systemapi.h` into `esp_missing.h`. These are only required for older SDK versions.


Sming/Arch/Esp8266/Components/esp_no_wifi/user_interface.c
```diff
@@ -0,0 +1,540 @@
...
+  43:	0020c0               	memw
+  46:	0558                	l32i.n	a5, a5, 0
+  48:	0020c0               	memw
+  4b:	c06652               	s32i	a5, a6, 0x300
+  4e:	441b                	addi.n	a4, a4, 1
+  50:	744040               	extui	a4, a4, 0, 8
+  53:	e43487               	bltu	a4, a8, 3b <system_rtc_mem_write+0x37>
+  56:	120c                	movi.n	a2, 1
+  58:	f00d                	ret.n
+*/
```
mikee47: This is an example where I would like to just be able to use the implementation from libmain, but it's not in a separate section. This disassembly clearly shows there are no relocations or external dependencies so in theory I should be able to pull this symbol out as a separate object file. Any clues as to how I can do this?

mikee47: NB. I've added the `sdk-disassemble` build target so the full disassembly can be generated. The above example will be in `sdk/main/user_interface.asm`.

slaff:
> I should be able to pull this symbol out as a separate object file. Any clues as to how I can do this?

objcopy should be your best friend. Something like this should work.
```
objcopy --strip-all --keep-symbol system_rtc_mem_write  originalFile  newFile
```

mikee47:
I've had limited success using `objcopy`. For this to work, the function must not have any relocations as these get stripped - this either leads to an `invalid property table` error at link time, or it links successfully but crashes because the relocation is left at 0 / null.

We can keep relocations using `--strip-unneeded-symbols` but then that includes everything, so back to square one.

When I use `objcopy -S -K system_rtc_mem_write -K system_rtc_mem_read`, the output object file contains a lot more than just these two functions, although it does only retain the requested symbols. It works, but it's not optimal.

Is there a way I can specify the address/offset range for extraction? For example, start at `system_rtc_mem_write-0x4` and extract 0x5A bytes, retaining any relocations encountered?

slaff:
Or may be this can help: https://stackoverflow.com/questions/50651610/extract-raw-machine-code-of-function-from-o-object-file ?

mikee47:
I've managed to improve on the objcopy approach by filtering out sections *and* symbols - the helpfully named `*UND*` catches the relocation entries so they don't get discarded!

slaff:
@mikee47 Is this PR no longer work in progress and ready for testing from adventurous Sming fans?

mikee47:
@slaff Yes, let's get it out there, I'm hoping others can help sort out the deficiencies.

mikee47:
For info, generated about 512K of random data by calling `os_get_random` and dumping it via GDB to a local file. The randomness tester from https://www.fourmilab.ch/random/ produced these results:

```
Entropy = 7.999625 bits per byte.

Optimum compression would reduce the size of this 517120 byte file by 0 percent.

Chi square distribution for 517120 samples is 268.34, and randomly would exceed this value 27.09 percent of the times.

Arithmetic mean value of data bytes is 127.4527 (127.5 = random).
Monte Carlo value for Pi is 3.151555937 (error 0.32 percent).
Serial correlation coefficient is -0.000179 (totally uncorrelated = 0.0).
```

Results also confirmed via https://servertest.online/entropy.

PR #2007, opened 2019-12-18
Some improvements to multipart parser

aemseemann:
This PR addresses some issues with the current multipart parser implementation.
- Allow part headers to span multiple TCP packets (similar to how HTTP headers are handled). It is very unlikely for this issue to occur for requests with only one part, since all the headers are probably part of the first TCP packet. However, requests with multiple parts may fail randomly if a part header just happens to span across a TCP packet boundary. 
- Allow construction of the ``MultipartParser`` instance to fail, e.g. when the 'boundary' part of the Content-Type header is missing.
- Make parser state (``multipart_parser_t``) a member of ``MultipartParser`` to avoid additional dynamic memory allocation.
- Allow (and ignore) preamble data. Although preamble data is rarely used in practice, it has been [observed](https://github.com/civetweb/civetweb/issues/470) that some clients add an additional newline before the message body, which, technically, qualifies as preamble data.


Sming/Components/MultipartParser/MultipartParser.cpp
```diff
@@ -62,83 +62,113 @@ size_t formMultipartParser(HttpRequest& request, const char* at, int length)
...
 }
 
-MultipartParser::~MultipartParser()
+MultipartParser *MultipartParser::create(HttpRequest& request)
 {
-	multipart_parser_free(parser);
-	parser = nullptr;
+	if(request.headers.contains(HTTP_HEADER_CONTENT_TYPE)) {
+		// Content-Type: multipart/form-data; boundary=------------------------a48863c0572edce6
+		int startPost = request.headers[HTTP_HEADER_CONTENT_TYPE].indexOf(_F("boundary="));
```
mikee47: Optimisation: `indexOf(_F("boundary="), 9)`

Sming/Components/MultipartParser/MultipartParser.cpp
```diff
@@ -62,83 +62,113 @@ size_t formMultipartParser(HttpRequest& request, const char* at, int length)
...
-	return multipart_parser_execute(parser, at, length);
+	return multipart_parser_execute(&parserEngine, at, length);
 }
 
 int MultipartParser::partBegin(multipart_parser_t* p)
 {
 	GET_PARSER();
 
+	parser->headerName = nullptr;
+	parser->headerValue = nullptr;
```
mikee47: Consider `.setLength(0);` instead to retain existing allocations.

mikee47:
> 
> ...the entire framework needs rebuilding when `ENABLE_HTTP_SERVER_MULTIPART` changes.

I've pushed a commit to https://github.com/mikee47/Sming/commit/ca584130f8a08db9bdf36a38b4dd3df6e4030cc1.

Sming/Components/MultipartParser/MultipartParser.cpp
aemseemann: Is there any downside in using `indexOf(FS("boundary="))` instead?
At least for small strings and thanks to your SSO, `WString` should not do a heap allocation, so both `_F` and `FS` with its implicit `operator WString` will essentially do a `memcpy_aligned`.

Sming/Components/MultipartParser/MultipartParser.cpp
```diff
@@ -62,83 +62,113 @@ size_t formMultipartParser(HttpRequest& request, const char* at, int length)
...
 	return 0;
 }
 
-int MultipartParser::readHeaderValue(multipart_parser_t* p, const char* at, size_t length)
+int MultipartParser::processHeader()
 {
-	GET_PARSER();
-
-	if(parser->headerName == _F("Content-Disposition")) {
+	if(headerName == _F("Content-Disposition")) {
```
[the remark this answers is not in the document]
aemseemann: Flipping FlashString <-> WString to avoid heap allocation. 
Nice trick, have to remember that.

Sming/Components/MultipartParser/MultipartParser.cpp
```diff
@@ -62,83 +62,113 @@ size_t formMultipartParser(HttpRequest& request, const char* at, int length)
...
+
 	return 0;
 }
 
 int MultipartParser::readHeaderName(multipart_parser_t* p, const char* at, size_t length)
 {
 	GET_PARSER();
 
-	parser->headerName.setString(at, length);
+	if (parser->headerValue) {
```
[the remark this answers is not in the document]
aemseemann: Using `headerValue != String::empty` should be fine here. As far as I remember, header values must not be empty.

Sming/Components/MultipartParser/MultipartParser.cpp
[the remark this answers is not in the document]
mikee47: Also short-circuits comparison if lengths don't match.

Sming/Components/MultipartParser/MultipartParser.cpp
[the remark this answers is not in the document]
mikee47: Or `headerValue.length() != 0`

Sming/Components/MultipartParser/MultipartParser.cpp
mikee47: Good point. With SSO `indexOf(F("boundary="))` is also an option.

Sming/Components/MultipartParser/MultipartParser.h
```diff
@@ -14,33 +14,44 @@
...
 
 #include "multipart-parser/multipart_parser.h"
 
 class MultipartParser
 {
+	MultipartParser(HttpRequest& request, const String& boundary);
 public:
-	explicit MultipartParser(HttpRequest* request);
-	~MultipartParser();
+	static MultipartParser *create(HttpRequest& request);
```
slaff: @aemseemann Please create a subdirectory `src` under MultipartParser. Move HttpMultipartResource.cpp, HttpMultipartResource.h, MultipartParser.cpp and MultipartParser.h to `src`. Add new empty `.cs` file in the `src` directory.  Modify the component.mk file to have
```
COMPONENT_SRCDIRS := src
COMPONENT_INCDIRS := src
```
After that from $SMING_HOME run `make cs`.
```

File: Sming/Components/MultipartParser/multipart-parser.patch [hunk 6]
```diff
@@ -78,8 +94,98 @@ index 981dabb..52e0200 100644
...
+         if (c != p->multipart_boundary[p->index]) {
+-          return i;
++          p->state = s_preamble;
++          --i;
++        } else {
++          p->index++;
++        }
++        break;
++
++      case s_preamble:
```
slaff: That code should be tested. My very quick test resulted in getting stuck here.

Sming/Components/MultipartParser/multipart-parser.patch
```diff
@@ -36,39 +36,55 @@ index 981dabb..52e0200 100644
...
- }
- 
+-                               strlen(boundary) +
+-                               strlen(boundary) + 9);
+-
+-  strcpy(p->multipart_boundary, boundary);
+-  p->boundary_length = strlen(boundary);
+-  
+-  p->lookbehind = (p->multipart_boundary + p->boundary_length + 1);
++  p->multipart_boundary = boundary;
```
slaff: @aemseemann This is C code. And in C you should use strcpy to copy strings. Here you are making a copy of a pointer that gets invalidated at the time when you actually have to access `multipart_boundary`.

aemseemann: The intention *is* to copy the pointer and not the string content (to avoid an additional allocation).

It is the caller's responsibility to ensure that the content stays stays valid (and unchanged) throughout the lifetime of the `multipart_parser`. This is ensured, because the caller is the `MultipartParser` constructor and both `multipart_parser` and the `boundary` string are both members of `MultipartParser`.

Sming/Components/MultipartParser/multipart-parser.patch
aemseemann: Strange, it works for me (tested with the `otatool.py upload` from PR#2006, modified to include some preamble data).
Can you describe your test setup?
(The preamble data must end with '\r\n' before the first boundary)

aemseemann:
Moving sources to 'src' and code formatting is done.

I am still thinking about @mikee47's [proposal](https://github.com/SmingHub/Sming/pull/2007#issuecomment-568039426). Maybe we can go a step further: Instead of just avoiding the SmingCore recompilation, MultipartParser could be converted into a Library. Then remove `ENABLE_HTTP_SERVER_MULTIPART` and make it the user's responsibility to do the following call during web server configuration:
```cpp
server.setBodyParser(MIME_FORM_MULTIPART, formMultipartParser);
```

Sming/Components/MultipartParser/multipart-parser.patch
slaff: @aemseemann  can you try the following: Compile HttpServer_FirmwareUpload in debug mode for the Host
```bash
cd $SMING_HOME/../sample/HttpServer_FirmwareUpload
make dist-clean
make clean
make SMING_ARCH=Host ENABLE_GDB=1 ENABLE_LWIPDEBUG=1 ENABLE_MALLOC_COUNT=0
make flash
```

Once that finishes if you have valgrind run the application by calling
```bash
# setup your virtual network interface. see: https://sming.readthedocs.io/en/latest/_inc/Sming/Arch/Host/Components/lwip/index.html
make valgrind
```

At least on my system valgrind starts to report issues with the `multipart_boundary`. Example:
```
==23134== Invalid read of size 1
==23134==    at 0x809CC14: multipart_parser_execute (multipart_parser.c:112)
```

You can also run the application in a visual debugger and start monitoring the bytes starting at the memory address that `multipart_boundary` points to. As far as I can see the first bytes are overwritten.

aemseemann: Can confirm the bug, thanks for the explanation!
I accidentally took the address from the `MultipartParser` constructor argument, which was shadowing the `boundary` member. This is fixed now.

However, valgrind discovered an issue with `TcpConnection::checkSelfFree`:
```
==8800== Invalid read of size 1
==8800==    at 0x12C374: checkSelfFree (TcpConnection.h:191)
==8800==    by 0x12C374: TcpConnection::checkSelfFree() (TcpConnection.h:189)
==8800==    by 0x12CD26: TcpConnection::internalOnPoll() (TcpConnection.cpp:507)
```
The reason is in TcpConnection.cpp:500pp
``` cpp
err_t TcpConnection::internalOnPoll()
{
	//if (tcp->state != ESTABLISHED)
	//	return ERR_OK;

	sleep++;
	err_t res = onPoll();
	checkSelfFree(); /// <<< instance was deleted inside `onPoll`
	debug_tcp_ext("<poll");
	return res;
}
```
Could be fixed with:
```cpp
	err_t res = onPoll(); 
	if (res != ERR_TIMEOUT) {
		// on ERR_TIMEOUT, checkSelfFree is already called inside checkSelfFree
		checkSelfFree();
	}
```
However, I lack the in-depth knowledge of the low level classes and relying on the error code seems quite fragile.

slaff:
> Can confirm the bug, thanks for the explanation!

@aemseemann Awesome, thanks a lot for fixing it :)

> and make it the user's responsibility to do the following call during web server configuration:

Yes, let's do it. Add the related changes as part of this PR.

> The reason is in TcpConnection.cpp:500pp

Create a separate PR for this issue.

aemseemann:
MultipartParser conversion into a Library is done.

The `TcpConnection` fix will have to wait untill later this week.

slaff:
@aemseemann This PR is scheduled for merging. Once the small changes are committed I will take care to merge it in `develop`.

PR #2008, opened 2019-12-20
Add `COMPONENT_RELINK_VARS`

mikee47:
Behaves just like ``COMPONENT_VARS`` except dependent Components are not rebuilt.
This is appropriate where the public interface (header files) are not affected
by the variable setting, but the library implementation still requires a variant.

For example, previously `make ENABLE_CUSTOM_HEAP=1` would create a new library variant for not only `heap`, but also `sming-arch`, `Sming` and `App` (the application itself) which all depend on it. This is not necessary as it's only the heap implementation which changes - not the API.

The following build variables have been changed accordingly:

Esp8266 driver: ENABLE_CUSTOM_PWM
Esp8266 heap: ENABLE_CUSTOM_HEAP, UMM_POISON_CHECK, UMM_FUNC_IRAM
Host heap: ENABLE_MALLOC_COUNT
Host hostlib: LWIP_SERVICE_INTERVAL
spiffs: SPIFF_FILEDESC_COUNT

**Other changes**

* Add `CalculateVariantHash` build function - avoids duplication and potentially useful elsewhere
*  Remove extraneous definition of `MB_RESPONSE_TIMEOUT` from ModbusMaster sample make file. This is defined in the `ModbusMaster` library component.mk file.


Sming/Libraries/ModbusMaster/samples/generic/component.mk
```diff
@@ -1,7 +1,2 @@
 ARDUINO_LIBRARIES := ModbusMaster
 DISABLE_SPIFFS = 1
-
-# Configurable Modbus response timeout
-COMPONENT_VARS += MB_RESPONSE_TIMEOUT
```
slaff: Where did that go?

mikee47: It's defined in the ModbusMaster library's `component.mk` file.

PR #2010, opened 2019-12-27
Feature: Applying coding style to selected directories

slaff:
- If a directory in the framework should be included in the coding style check it must have an empty '.cs' file.
- If an application also wants to follow Sming's coding style rules then its `component.mk` file should have `CS_DIRS := app lib/core`. Make sure to replace `app`  and `lib/core` with existing directories in your application where the coding style rules should be applied.

### DIFF
--- a/Sming/build.mk
+++ b/Sming/build.mk
@@ -303,23 +303,3 @@ endef
 	$(Q) $(call TryApplyPatch,$*,$(*F).patch)
 	$(Q) touch $@
 
-##@Developer
-
-# Recursive wildcard search
-# $1 -> list of directories
-# $2 -> file extensions filters (using % as wildcard)
-define rwildcard
-	$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $2,$d))
-endef
-
-# Files that should follow our coding standards
-CS_DIRS		= $(shell find $(SMING_HOME)/../ -name '.cs' -type f  | xargs dirname)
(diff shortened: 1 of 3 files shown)


Sming/build.mk
```diff
@@ -303,3 +303,23 @@ endef
...
+CS_DIRS		= $(shell find $(SMING_HOME)/../ -name '.cs' -type f  | xargs dirname)
+CS_FILES	= $(call rwildcard,$(CS_DIRS:=/*),%.cpp %.h %.c)
+
+.PHONY: cs
+cs: ##Apply coding style to all core files
+	$(if $(V),$(info Applying coding style to $(words $(CS_FILES)) files ...))
+	@for FILE in $(CS_FILES); do \
+		$(CLANG_FORMAT) -i -style=file $$FILE; \
+	done
+
```
mikee47: When running `make cs` from a project, I wouldn't expect (or want) all of the framework code to be formatted as well as the project. I think it's OK to have `make cs` for both project and framework, but the behaviour differs depending on where it's run from:
* From SMING_HOME: Format only framework code (including all samples, contained Components, Libraries, etc.) so behaviour is essentially unchanged.
* From project: Format only project code (if requested)
This leaves third-party libraries which are not part of the framework. I'm thinking perhaps a separate make target would be appropriate for those
Also, we shouldn't have any make targets in build.mk, though shared macros/functions are fine.

mikee47:
I've pushed some changes to your repo. for consideration.

slaff:
@mikee47 Do you have something to add to this PR before I merge it?

mikee47:
Nope, nothing to add.

PR #2011, opened 2019-12-30
ArduinoJsonStream::isFinished() returns wrong value

mikee47:
Shouldn't have to call `readMemoryBlock` before getting correct value.


slaff:
@mikee47 Is this PR tested and ready for merging?

mikee47:
Yes, tested ArduinoJson5 with Basic_WebskeletonApp_LTS.

PR #2014, opened 2020-01-04
Revise fingerprint handling and add basic Crypto support library

mikee47:
The purpose of this PR is to simplify and improve fingerprint handling, and by implication how hashes and such are handled within the framework. I've also added a mechanism for selecting SSL cipher suites.

**crypto library**

To this end I've added a separate crypto library with the aim of creating a more consistent and robust C++ interface to commonly used routines, both by SSL and by application code.

The library is self-contained, using hash implementations based on AxTLS and Bear SSL, or ESP8266 ROM routines where available.

I've implemented HMAC as a class template. It supports keys larger than the block size.

**Fingerprint handling** has been rationalised by moving the validation itself into the validator, which
are now all class-based. The certificate implemention now just returns the hashes it supports.

One thing I've noticed is that whilst SHA1 and SHA256 fingerprints are commonly reported by browsers (e.g. firefox) when inspecting the certificate, the PKI fingerprint is not readily available. Also, there is no easy way to get BearSSL to obtain this fingerprint.

**Cipher Suite selection**

I've added a mechanism for selecting the cipher suites for BearSSL. These are defined as flash arrays of Suite identifiers. I've defined two standard sets:

- `basic` (default) intended to provide equivalent behaviour for comparing Axtls with BearSSL (and other implementations)
- `full` to provide maximum functionality, defined by the SSL implementation.

Only BearSSL uses these at present - Axtls just uses `basic` and ignores the setting. 

Goals for this PR are:

1. Ease of use for users. Minimal knowledge of SSL, crypto, etc.
2. Early verification. Having strongly-typed fingerprints and hashes makes code easier to follow and allows the compiler do do more checking, especially that they're the right size.
3. Simplification. Using templates, for example, reduces the amount of code we need to maintain.
4. Isolation of applications and other Compnents from the specific crypto implementations.
5. Avoid, where possible, messing about with pointers, arrays and C-style code. Think 'objects' instead.

TODO:

* [x] Review ROM hash routines and add additional tests and benchmarking to confirm correct operation and evaluate relative performance.
* [x] Write crypto hash wrappers for BearSSL.
* [x] Take another look at the validator/fingerprint interface. Can use a single `bool Certificate::getFingerprint(FingerprintType, Fingerprint&)` method instead of multiple overrides.


mikee47:
One further thought about all this is that there seems to be little value in re-implementing hashing and certificate checking in each SSL implementation. If the SSL library itself provides the main support for this then the implementation code can be reduced.

One way to do that is just to pass the raw certificate data directly into a separate module. That would also have the benefit that we only need to generate hashes for those fingerprints we intend to validate; at present we have to pre-emptively generate all of them.

mikee47:
You can review the documentation changes at https://smingdev.readthedocs.io/en/feature-crypto/

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,116 @@
...
+
+	void update(const String& str)
+	{
+		static_cast<Context*>(this)->update(str.c_str(), str.length());
+	}
+
+	/*
+	 * Update from a fixed array or struct, but not a pointer - base method handles that, with size
+	 */
+	template <typename T> typename std::enable_if<!std::is_pointer<T>::value, void>::type update(const T& data)
```
aemseemann: Use `enable_if_t` and `is_pointer_v` instead.
Maybe also constrain on `is_pod`?

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * @brief Update hash over a given block of data
+	 * @{
+	 */
+	void update(const Blob& blob)
+	{
+		static_cast<Context*>(this)->update(blob.data, blob.length);
+	}
+
+	void update(const String& str)
```
aemseemann: Overload unnecessary, since `Blob` has an implicit conversion constructor from `String`

Sming/Components/crypto/include/Crypto/Md5.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "HashContext.h"
+
+namespace Crypto
+{
+using Md5Hash = HashValue<MD5_SIZE>;
```
aemseemann: Is this necessary? It creates an alternative to `Md5::Hash`, which raises the question which one is preferred.

Sming/Components/crypto/include/Crypto/Md5.h
```diff
@@ -0,0 +1,40 @@
...
+{
+public:
+	using Hash = Md5Hash;
+
+	Md5()
+	{
+		init();
+	}
+
+	void init()
```
aemseemann: Appears redundant to constructor.
Instead of
```cpp
Md5 md5;
...
md5.init()
```
the user could simply write
```
Md5 md5;
...
md5 = Md5();
```

Sming/Components/crypto/include/Crypto/Md5.h
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	using HashContext::update;
+
+	void update(const void* data, size_t size)
+	{
+		MD5_Update(&context, static_cast<const uint8_t*>(data), size);
+	}
+
+	void final(Hash& hash)
```
aemseemann: Return `Hash` instead of argument?

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * @brief Class template for a Hash value, essentially just an array of bytes
+ */
+template <size_t size> struct HashValue {
+	uint8_t data[size];
+
+	static constexpr size_t length = size;
+
+	bool copy(const void* src)
```
aemseemann: Should be constructor, i.e.
```
auto hash = HashValue(data);
```
instead of
```
HashValue hash;
hash.copy(data)
```

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,116 @@
...
+#include <axtls-8266/crypto/crypto.h>
+#include <WString.h>
+#include <Data/HexString.h>
+
+namespace Crypto
+{
+/**
+ * @brief
+ */
+struct Blob {
```
aemseemann: This would have been a perfect opportunity to use `std::span`. Unfortunately, it's only available in c++20.

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,116 @@
...
+	String toString(char separator = '\0') const
+	{
+		return makeHexString(data, size, separator);
+	}
+};
+
+/**
+ * @brief Class template for a Hash implementation 'Context'
+ */
+template <class Context, class Hash> class HashContext
```
aemseemann: Looks like CRTP (`class Context` is the derived class). Isn't that overly complex? 

On first sight, it should be possible to avoid the inheritance altogether by making `Context` an "engine" class that is embedded as a private member inside HashContext.

After all, the goal is to create a common interface for different Hash algorithms, so all instantiations of HashContext should provide the same public interface.

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: I'd prefer not to have this but ran into difficulty having it defined directly inside the class. Suggestions?

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: For now, keeping to C++11 so have to stick with existing traits.
Re-checking `is_pod` appears to be deprecated in C++20. I _think_ we need `is_std_layout`?
https://stackoverflow.com/questions/6496545/trivial-vs-standard-layout-vs-pod

mikee47: (I know, I've used it in FlashString - something to fix.)

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: I'll pass the hash size to the class template instead, thanks.

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: I figure the `init`, `update` and `final` sequence is more familiar to existing crypto users.

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: Adding constructors breaks POD and aggregate initialisation which we need for things like this:

```
static constexpr Crypto::MD5::Hash myHash PROGMEM = { ... };
```

I had this issue with FlashString.

Sming/Components/crypto/include/Crypto/Md5.h
aemseemann: Yeay, but in C++, it's usually the constructors job to do the 'init'. It creates ambiguity if it is always necessary to call `init` after construction or only for reused objects.

Sming/Components/crypto/include/Crypto/Md5.h
aemseemann: Could you please elaborate on the problems you encountered?
I would have expected `using Hash = HashValue<MD5_SIZE>;` inside the class should work fine.

Sming/Components/crypto/include/Crypto/HashContext.h
aemseemann: Ok, didn't know that.

Sming/Components/crypto/include/Crypto/HashContext.h
aemseemann: Handn't noticed that yet. `is_std_layout` seems right then.

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: Took me a bit of time, but I get what you mean. Excellent idea. Thanks.

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: I've re-jigged using your Engine idea, much better. This issue has now evaporated :-)

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: https://sming.readthedocs.io/en/latest/_inc/Sming/Components/FlashString/object.html#aggregate-initialization

mikee47: Too me more time than I'd care to admit working up a solid solution to that!

mikee47: Because of this particular issue, I added additional checks to the DEFINE macros in FlashString:

```
#define FSTR_CHECK_STRUCT(name)                                                                                        \
	static_assert(std::is_pod<decltype(name)>::value, "FSTR structure not POD");                                       \
	static_assert(offsetof(decltype(name), data) == sizeof(uint32_t), "FSTR structure alignment error");

```

Both of these issues caused me headaches. I think Crypto is simple enough we can omit this, but something to bear in mind just in case.

aemseemann: Thanks for the link, I did read the cppreference page in the meantime. That would have thrown me off quite a while, too.

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: I see your point. With the 'Engine' revision it makes less sense keeping init, so I'll remove it.

mikee47: I'm going to add a method back, but called `reset()` to resolve the ambiguity.

It makes the code very clunky otherwise, for example:

```
// Reset (init) on construction
Crypto::Sha1 certificateSha1;

// (1) Reset by assignment: need to specify the type again
certificateSha1 = Crypto::Sha1();

// (2) Reset by method: don't care about the type
certificateSha1.reset();
```

Using (2) means we can change the hash algorithm simply by amending the declared type, everything else remains the same.

mikee47:
For info, I've been checking heap usage with the Basic_SSL sample. Here are some stats using malloc_count.

 Item | Axtls | Bearssl
 --- | --- | ---
  Free bytes | 46808 | 47368
  Used bytes | 4297 | 4297
  Peak used bytes | 30477 | 28581
  Allocation count | 835 | 693
  Total used bytes | 104891 | 71893

Note that axTLS fails to complete with standard heap, so both examples are with ENABLE_CUSTOM_HEAP=1.

Sming/Components/crypto/include/Crypto/Md5.h
aemseemann: That's a much better name for the method. Now it adds value over the constructor.

aemseemann: FYI: One could use brace initialization to avoid repeating the type:
```
certificateSha1 = {};
```
However, `reset()` is clearly more expressive in this case.

mikee47:
Some timing comparisons, connection times in milliseconds measured between IP address acquisition to `onDownload()` callback:

Attempt | Axtls | Bearssl (Basic) | Bearssl (Full)
--- | --- | --- | ---
#1 | 3779 | 3717 | 5049
#2 | 3814 | 3170 | 5090
#3 | 3475 | 3148 | 5064

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: I've removed this method, it's just asking for trouble.

Sming/Components/crypto/include/Crypto/Md5.h
mikee47: Makes more sense now it's in the Md5Engine class, not called by users.

slaff:
@mikee47 What is the current state of this PR? Is it ready and tested on a real device + host? Are you happy with the design and results so far?

mikee47:
@slaff It's OK. Works fine on both host and ESP8266, have some further updates to come for the ROM functions, with updated tests and benchmarking. I also intend to write MD5/SHA1 wrappers for BearSSL to remove the code duplication and allow it to take advantage of the faster ROM routines. Neither of those things affects the overal architecture though, so if you have any thoughts on that most welcome.

**Other stuff to consider**, but don't intend to address in this PR:

At present the code just links in all the supported hashes and encryption schemes, even if they're not used. I'd prefer to see a more unified approach:

* User specifies in general terms which connection types to compile, via component.mk and/or appcode. This will control linking of the corresponding hashes and encryption routines.
* Each connection can further restrict which connection types to use, via ssl init callback.


BearSSL uses a fixed-size buffer for the suites (BR_MAX_CIPHER_SUITES=48_ entries) which get checked during handshake. We could save 96 bytes and might gain more flexibility if that was done via callback - involves modifying the T0 code and recompiling it.

Sming/Components/crypto/include/Crypto/HashContext.h
```diff
@@ -0,0 +1,149 @@
...
+	{
+		engine.init();
+	}
+
+	/**
+	 * @name Calculate hash on a single block of data
+	 * @retval Hash
+	 * @{
+	 */
+	static Hash calculate(const void* data, size_t length)
```
aemseemann: Just a thought: To take advantage of all `update` overloads, both `calculate` methods could be replaced by a variadic template:
```cpp
template <typename... Ts> static calculate(Ts&&... args) {
	HashContext<Engine> ctx;
	ctx.update(std::forward<Ts>(args)...);
	return ctx.hash();
}
```

Sming/Components/crypto/include/Crypto/Sha256.h
```diff
@@ -0,0 +1,35 @@
...
+class Sha256Engine
+{
+public:
+	using Hash = HashValue<Sha256Engine>;
+	static constexpr const char* name = "SHA256";
+	static constexpr size_t size = SHA256_SIZE;
+
+	void init()
+	{
+		SHA256_Init(&context);
```
aemseemann: `ESP_` prefix for ESP8266 architecture?

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
```diff
@@ -35,10 +35,11 @@ int BrClientConnection::init()
 	br_ssl_client_set_default_rsapub(&clientContext);
 
 	// X509 verification
-	x509Context = new X509Context([this]() { return context.getSession().validateCertificate(); });
-	br_ssl_engine_set_x509(getEngine(), *x509Context);
+	delete certificate;
+	certificate = new BrCertificate([this]() { return context.session.validateCertificate(); });
```
aemseemann: could use `std::unique_ptr` instead.

Sming/Components/ssl/include/Network/Ssl/Certificate.h
```diff
@@ -65,28 +66,31 @@ class Certificate
...
-	 * @brief Check if certificate fingerprint (SHA1) matches the one given.
-	 * @param hash - SHA1 fingerprint to match against
-	 * @retval bool true on match, false otherwise
+	 * @name Obtain certificate fingerprint
+	 * @param fingerprint On success, returned fingerprint
+	 * @retval bool true on success, false if fingerprint /not available
+	 * @{
 	 */
-	virtual bool matchFingerprint(const uint8_t* hash) const = 0;
+	virtual bool getFingerprint(Fingerprint::Cert::Sha1& fingerprint) const
```
aemseemann: Unfortunately, c++17 is not the default yet :(
If it were, this interface could have been simplified to
```cpp
virtual std::optional<Fingerprint::Cert::Sha1> getFingerprint() const
```
(A advantage of return-by-value over by-reference parameters is that the returned type does not need a default constructor, nor does it have to support any kind of 'invalid state', which is usually achieved by the default constructor)

Sming/Components/ssl/include/Network/Ssl/Fingerprints.h
```diff
@@ -10,124 +10,70 @@
...
-	bool setSha1_P(const uint8_t* cert, unsigned length)
-	{
-		return setValue(certSha1, SHA1_SIZE, cert, length);
-	}
-
-	/** @brief Set the SHA256 fingerprint
-	 *  @param cert data to copy
-	 *  @param length for checking
-	 *  @retval bool false on length check or allocation failure
+union Fingerprint {
```
aemseemann: According to the documentation comments above, it appears the same could have been achieved by using nested Namespaces, without the distraction created by the union and the restrictions implied on the contained types.

Sming/Components/crypto/component.mk
```diff
@@ -0,0 +1,3 @@
+COMPONENT_LIBNAME :=
+COMPONENT_DEPENDS	:= axtls-8266
```
aemseemann: This dependency should probably depend on SMING_ARCH (unnecessary for ESP8266, as it uses ROM functions)

Sming/Components/crypto/include/Crypto/Md5.h
```diff
@@ -0,0 +1,44 @@
+#pragma once
+
+#include "HashContext.h"
+
+#ifdef ARCH_ESP8266
+#include <esp_crypto.h>
+#define CRYPTO_DEF(f) ESP_MD5_##f
+#else
+#define CRYPTO_DEF(f) MD5_##f
```
aemseemann: The `#else` case should also include a header where the `MD5_` functions are defined.

aemseemann:
> At present the code just links in all the supported hashes and encryption schemes, even if they're not used.

Just an idea, maybe not suitable for every use case: Provide a non-backwards compatible TLS1.3 configuration for BearSSL via a config option. I've heard that they got rid of many of the less secure algorithms in this protocol version.

docs/source/upgrading/4.0-4.1.rst
```diff
@@ -20,148 +20,7 @@ SSL enabled clients and servers.
...
-
-  COMPONENT_DEPENDS += axtls-8266
-
-You can use also one SSL adapter, for example Bearssl, with another cryptographic library, for example Axtls.
-
-SSL namespace
-~~~~~~~~~~~~~
-All SSL related classes and types are organized in a separate namespace called `Ssl`. For example you should use `Ssl::KeyCertPair`
-instead of `SslKeyCertPair` and `Ssl::Fingerprints` instead of `SslFingerprints`.
+See :component:`ssl` for further details.
```
slaff: Can you hint the user to go directly to the SSL upgrade document ( https://sming.readthedocs.io/en/feature-crypto/_inc/Sming/Components/ssl/upgrade.html )?

Sming/Components/crypto/include/Crypto/Sha256.h
mikee47: SHA256 is implemented in libcrypto.a, and benchmarks slower than the axtls version so no need to handle it differently. I'll add a document to summarise.

Sming/Components/ssl/include/Network/Ssl/Fingerprints.h
mikee47: I started with namespaces but we don't get nested formatting... I'm undecided about having multiple `Certificate::getFingerprint` methods, we could replace that with a single virtual method which takes a type enum plus Fingerprint union.

I'll take another look at this; won't affect the public interface, just the connection between the validator and the certificate.

mikee47: Adding fingerprint type enum back also gives us an efficient way for ValidatorList to track which fingerprint validators have been added. For BearSSL, we then only generate those fingerprints. For example, there's no point in validating both Cert::SHA1 and Cert::SHA256, we'd only ask for one of those. And if Pki::SHA256 isn't needed, we don't calculate it. That should speed things up considerably.

slaff:
@mikee47 Sorry for bothering you. I guess you have been quite busy lately. Do you think this PR will be ready by the end of the month? I would like to prepare for the new 4.1.0 release.

mikee47:
These are the benchmarks output from HostTests on a node MCU.

The Crypto hashes should be the fastest implementations available:

MD5: ROM
SHA1: ROM
SHA224: BearSSL - the only implementation available. An axTLS implementation would likely be faster.
SHA256: axTLS
SHA384: BearSSL
SHA512: BearSSL

In addition, the SHA384 and SHA512 context buffers are very large (c. 640 bytes) so best avoided.

```
Benchmark Crypto Hashes.
md5: count=100, total=53827us, min=536us, max=580us, average=538us
sha1: count=100, total=95000us, min=948us, max=993us, average=950us
sha224: count=100, total=206146us, min=2059us, max=2260us, average=2061us
sha256: count=100, total=174255us, min=1740us, max=2015us, average=1742us
sha384: count=100, total=452213us, min=4521us, max=4651us, average=4522us
sha512: count=100, total=452307us, min=4522us, max=4566us, average=4523us

>> Benchmark ESP Hashes
esp.MD5: count=100, total=53476us, min=534us, max=562us, average=535us
esp.SHA1: count=100, total=94867us, min=948us, max=960us, average=949us

>> Benchmark axTLS Hashes
ax.MD5: count=100, total=63424us, min=631us, max=962us, average=634us
ax.SHA1: count=100, total=230031us, min=2297us, max=2408us, average=2300us
ax.SHA256: count=100, total=174252us, min=1740us, max=2007us, average=1742us
ax.SHA384: count=100, total=498127us, min=4978us, max=5281us, average=4981us
ax.SHA512: count=100, total=497531us, min=4975us, max=4999us, average=4975us

>> Benchmark BearSSL Hashes
br.md5: count=100, total=92378us, min=921us, max=1166us, average=924us
br.sha1: count=100, total=117937us, min=1178us, max=1347us, average=1179us
br.sha224: count=100, total=206153us, min=2059us, max=2272us, average=2061us
br.sha256: count=100, total=205987us, min=2059us, max=2083us, average=2060us
br.sha384: count=100, total=452498us, min=4521us, max=4935us, average=4525us
br.sha512: count=100, total=452275us, min=4522us, max=4546us, average=4523us

>> Benchmark HMAC
md5: count=100, total=60507us, min=604us, max=723us, average=605us
sha1: count=100, total=109794us, min=1097us, max=1143us, average=1098us
sha224: count=100, total=233773us, min=2336us, max=2494us, average=2338us
sha256: count=100, total=191413us, min=1910us, max=2276us, average=1914us
sha384: count=100, total=527917us, min=5277us, max=5519us, average=5279us
sha512: count=100, total=527961us, min=5279us, max=5318us, average=5280us

>> Benchmark HMAC functions
ESP_hmac_md5: count=100, total=59508us, min=595us, max=607us, average=595us
ESP_hmac_sha1: count=100, total=108774us, min=1088us, max=1093us, average=1088us
ax_hmac_md5: count=100, total=70834us, min=705us, max=997us, average=708us
ax_hmac_sha1: count=100, total=249286us, min=2491us, max=2637us, average=2493us
ax_hmac_sha256: count=100, total=190511us, min=1901us, max=2275us, average=1905us

```

mikee47:
The Crypto library now depends on both axtls-8266 and bearssl-esp8266. Also, axtls-8266 depends on crypto so make gives a circular dependency warning. I've updated the build makefiles to handle this.

The net effect is that we now have all of these libraries pulled in. The only way around that is to take copies of the appropriate implementations into the Crypto library so it's stand-alone.

mikee47:
I've also decided not to bother trying to work in the ROM hmac functions. They're not that complicated and the C++ interface is much more flexible with insignificant performance difference: it's the hash functions which are critical.

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
mikee47: Could probably use that in a lot of places in the framework... I'm not sure about starting now...

Sming/Components/crypto/component.mk
mikee47: We need axtls for other hashes... plus bearssl now...

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
slaff: Please not here and not now. This is definitely work that should be organized in a separate PR. If we decide to go for std:unique_ptr, std::shared_ptr and family we should apply those change to the complete Sming Library (Core + Core Components + Sming only libraries).

mikee47:
> take copies of the appropriate implementations into the Crypto library so it's stand-alone.

I'm leaning towards this approach, but not intending to do anything further with this PR until you've had a chance to review.

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
mikee47: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#rsmart-smart-pointers

Sming/Core/Network/SmtpClient.cpp
```diff
@@ -175,11 +175,9 @@ void SmtpClient::onReadyToSendData(TcpConnectionEvent sourceEvent)
 	case eSMTP_SendAuthResponse: {
 		// Calculate the CRAM-MD5 response
 		//     base64.b64encode("user " +hmac.new(password, base64.b64decode(challenge), hashlib.md5).hexdigest())
-		uint8_t digest[MD5_SIZE] = {0};
-		hmac_md5((const uint8_t*)authChallenge.c_str(), authChallenge.length(), (const uint8_t*)url.Password.c_str(),
-				 url.Password.length(), digest);
 
-		String token = url.User + ' ' + makeHexString(digest, MD5_SIZE);
+		auto digest = Crypto::HmacMd5::calculate(authChallenge, url.Password);
```
mikee47: Data/password are transposed. It's too easy to do this with potentially catestrophic results.
Simplest way I can think of to mitigate this problem generally is to make passwords/keys strongly typed so we have to do this:

```
Crypto::HmacMd5::calculate(Crypto::Password(password), data);
```

aemseemann: I would prefer to name it `Crypto::Secret(...)` instead.

How about adding a non-static `calculate` method?
```
Crypto::HmacMd5(password).calculate(data)
```

mikee47:
OK, so hash routines pulled into `crypto` and no longer depends on axtls-8266 or bearssl-esp8266. Implementations tweaked, benchmark times in microseconds for hashing a 2336-byte block of plaintext in RAM:

Algorithm | crypto | ESP | axTLS | Bear SSL | Notes
--- | --- | --- | --- | --- | ---
md5 | 538 (554) | 535 | 638 | 922 | ROM routine (software)
sha1 | 950 (1127) | 949 | 2295 | 1181 | ROM routine (software)
sha224 | 1845 | |  | 2032 | Modified axTLS
sha256 | 1842 | | 1762 | 2029 | axTLS
sha384 | 4505 | | 4973 | 4526 | Bear SSL
sha512 | 4506 | | 4967 | 4515 | Bear SSL

Sming/Components/crypto/include/Crypto/HashContext.h
mikee47: NB. `std::array` doesn't have a constructor for exactly this reason.

mikee47:
Last commit unrolls the byte encode/decode functions which increases speed slighly but also adds about 3K code size. Not convinced it's worth it but have a think.

slaff:
@mikee47 Is this PR tested on a real device? If yes is it ready for merging or would like to polish it a bit more?

Sming/Components/ssl/BearSsl/BrConnection.h
```diff
@@ -84,15 +92,11 @@ class BrConnection : public Connection
...
-	br_ssl_engine_context* getEngine() const
-	{
-		return const_cast<BrConnection*>(this)->getEngine();
-	}
+private:
+	void setCipherSuites(const CipherSuites::Array* cipherSuites);
 
 private:
-	uint8_t* buffer = nullptr;
+	std::unique_ptr<uint8_t> buffer;
```
aemseemann: Should be
```cpp
std::unique_ptr<uint8_t[]> buffer;
```
(Although it does not make a difference for PODs)

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
```diff
@@ -33,14 +31,93 @@ int BrClientConnection::init()
...
+		return getLastError();
+	}
 
-	// X509 verification
-	x509Context = new X509Context([this]() { return context.getSession().validateCertificate(); });
-	br_ssl_engine_set_x509(getEngine(), *x509Context);
+	return startHandshake();
+}
+
+template <class HashCtx> void resetHash(std::unique_ptr<HashCtx>& ctx, bool create)
```
aemseemann: Could go in an anonymous namespace

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
```diff
@@ -33,14 +31,93 @@ int BrClientConnection::init()
...
-	return startHandshake();
+	certificate.reset(new BrCertificate);
+	x509Decoder.reset(new X509Decoder(&certificate->subject, &certificate->issuer));
+
+	auto& types = context.session.validators.fingerprintTypes;
+	resetHash(certSha1Context, types.contains(Fingerprint::Type::CertSha1));
+	resetHash(certSha256Context, types.contains(Fingerprint::Type::CertSha256));
+}
+
+template <class HashCtx> void updateHash(HashCtx& ctx, const uint8_t* buf, size_t len)
```
aemseemann: anonymous namespace?

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
```diff
@@ -33,14 +31,93 @@ int BrClientConnection::init()
...
+		return;
+	}
+
+	x509Decoder->push(buf, len);
+
+	updateHash(certSha1Context, buf, len);
+	updateHash(certSha256Context, buf, len);
+}
+
+template <class FP, class HashCtx> bool getFp(FP& fp, HashCtx& ctx)
```
aemseemann: This seems unused.

Sming/Components/ssl/include/Network/Ssl/Certificate.h
```diff
@@ -65,28 +66,17 @@ class Certificate
...
-
 	virtual ~Certificate()
 	{
 	}
 
 	/**
-	 * @brief Check if certificate fingerprint (SHA1) matches the one given.
-	 * @param hash - SHA1 fingerprint to match against
-	 * @retval bool true on match, false otherwise
+	 * @name Obtain certificate fingerprint
```
aemseemann: `@brief` ?

Sming/Components/ssl/BearSsl/BrClientConnection.cpp
mikee47: AFAIK templates won't be visible externally anyway, so am I missing something or is it just good practice?

mikee47: Thank you for pointing this out, noted for reference use of anonymous namespaces instead of static or internal stuff. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnonexported-entities

aemseemann: I'm not so sure about that. Take the following demo code, for example:
```cpp
// test.cpp
#include <iostream>
#include <typeinfo>

template <typename T>
void testFunc()
{
    std::cout << "testFunc called for type " << typeid(T).name() << "\n";
}

template void testFunc<int>();

```
```cpp
// main.cpp
#include <stdio.h>

template <typename T> void testFunc();

int main(int argc, char *argv[]) 
{
    testFunc<int>();
    
    return 0;
}
```
The command sequence below gives me no "undefined reference" error, so the explicitly instantiated template is visible to other translation. (Implicitly instantiated versions of `testFunc` are visible, too.)
```
g++ -o test.o -c test.cpp
g++ -o main.o -c main.cpp
g++ test.o main.o -o main
```
(BTW: clang++ gives the same result.)

mikee47: Thank you. I can confirm for this specific case as follows:

1. Amend template function declaration in `BrClientConnection.cpp` so it emits a symbol:

```
template <class HashCtx>
__attribute__((noinline)) void updateHash(HashCtx& ctx, const uint8_t* buf, size_t len)
```

2. Add some code to application.cpp:

```
namespace Ssl {
template <class HashCtx> void updateHash(HashCtx& ctx, const uint8_t* buf, size_t len);
}

void init()
{
  // ...
  std::unique_ptr<Crypto::Sha1> ctx;
  Ssl::updateHash(ctx, nullptr, 0);
}

```

And it compiles. Wrap the template declaration in an anonymous namespace and it doesn't.

Conclusion: anonymous namespace a good idea.

mikee47:
@slaff OK, this is tested and working on nodeMCU.

Bear SSL loops the hashing over a full set of 8 rotations, using a lookup table for the constants. I've adopted this using templates for all SHA2 implementations as I think it provides the best balance between code size and speed.

Unwinding the loops fully might reasonably be expected to offers the fastest performance, but it generally makes things worse, as does -O2. I can only conclude that this is because the code size within the main loops is too big and the instruction cache cannot handle it all. This is especially true for the 64-bit SHA2 implementations, so code size is more important than loop unrolling.

The 32-bit SHA2 code benchmarks faster if the loops are fully unrolled, but I believe this is misleading because in practice there will be lots of other code running through the cache which will slow everything down.

Templated sha2::process, lookups in flash:
    rodata|           ReadOnly Data (RAM)|    3FFE8080|    3FFE843C|     956
irom0_text|           Uncached Code (SPI)|    40202010|    4022B6B0|  169632
sha224: count=100, total=187563us, min=1875us, max=1929us, average=1876us
sha256: count=100, total=187540us, min=1875us, max=1891us, average=1875us
sha384: count=100, total=419138us, min=4183us, max=4991us, average=4191us
sha512: count=100, total=418506us, min=4185us, max=4216us, average=4185us

For comparison, lookups in RAM:
    rodata|           ReadOnly Data (RAM)|    3FFE8080|    3FFE87C0|    1856
irom0_text|           Uncached Code (SPI)|    40202010|    4022B334|  168740
sha224: count=100, total=172074us, min=1720us, max=1828us, average=1721us
sha256: count=100, total=171996us, min=1720us, max=1735us, average=1720us
sha384: count=100, total=407162us, min=4065us, max=4749us, average=4072us
sha512: count=100, total=406678us, min=4066us, max=4120us, average=4067us

slaff:
The purpose of this PR is to simplify and improve fingerprint handling, and by implication how hashes and such are handled within the framework. I've also added a mechanism for selecting SSL cipher suites.

**crypto library**

To this end I've added a separate crypto library with the aim of creating a more consistent and robust C++ interface to commonly used routines, both by SSL and by application code.

The library is self-contained, using hash implementations based on AxTLS and Bear SSL, or ESP8266 ROM routines where available.

I've implemented HMAC as a class template. It supports keys larger than the block size.

**Fingerprint handling** has been rationalised by moving the validation itself into the validator, which
are now all class-based. The certificate implemention now just returns the hashes it supports.

One thing I've noticed is that whilst SHA1 and SHA256 fingerprints are commonly reported by browsers (e.g. firefox) when inspecting the certificate, the PKI fingerprint is not readily available. Also, there is no easy way to get BearSSL to obtain this fingerprint.

**Cipher Suite selection**

I've added a mechanism for selecting the cipher suites for BearSSL. These are defined as flash arrays of Suite identifiers. I've defined two standard sets:

- `basic` (default) intended to provide equivalent behaviour for comparing Axtls with BearSSL (and other implementations)
- `full` to provide maximum functionality, defined by the SSL implementation.

Only BearSSL uses these at present - Axtls just uses `basic` and ignores the setting. 

Goals for this PR are:

1. Ease of use for users. Minimal knowledge of SSL, crypto, etc.
2. Early verification. Having strongly-typed fingerprints and hashes makes code easier to follow and allows the compiler do do more checking, especially that they're the right size.
3. Simplification. Using templates, for example, reduces the amount of code we need to maintain.
4. Isolation of applications and other Components from the specific crypto implementations.
5. Avoid, where possible, messing about with pointers, arrays and C-style code. Think 'objects' instead.

Issue #2026, opened 2020-01-15
DISABLE_SPIFFS flag not working


etmmahi:
When in the example Basic_rBoot in the component.mk

```
## option to completely disable spiffs
DISABLE_SPIFFS		?= 1
```

the DISABLE_SPIFFS is set, this is not reflected in the compiler flags during a build.

```
make config-clean
make clean
make VERBOSE=1
```

pse add something like:
-DDISABLE_SPIFFS=$(DISABLE_SPIFFS)
to the compiler flags

Martijn

mikee47:
https://sming.readthedocs.io/en/latest/_inc/Sming/Components/spiffs/index.html

etmmahi:
This is a snippet from the rboot sample. It can expect DISABLE_SPIFFS to be set 

```
#ifndef DISABLE_SPIFFS
			Serial.println("  ls - list files in spiffs");
			Serial.println("  cat - show first file in spiffs");
#endif
```
But the flag is not given to the GCC compiler.

Furthermore the 'readthedocs' says  

 DISABLE_SPIFFS

    0
        Disable filesystem generation
    1 (default)
        Enable filesystem generation 

Where setting  DISABLE_SPIFFS=1 means **enable** ?? A bit confusing.

Martijn

mikee47:
Ah, yes the docs are wrong. The point is is only determines if a filesystem image is generated for upload to the device. It doesn't disable SPIFFS. (Yes, I agree the name is confusing - historical.)

etmmahi:
The basic_rboot sample expects the flag to be set or not (conditional code with #ifdef / #endif) 
Can you add an automatic -DDISABLE_SPIFFS=$(DISABLE_SPIFFS) to the compiler flags when it is set?
The the basic-rboot sample is correct

mikee47:
Yes, that would be logical. I did that for `Basic_WebSkeletonApp`, can't think why I didn't apply it universallly.

PR #2027, opened 2020-01-15
#define `DISABLE_SPIFFS` for application use, correct documentation

mikee47:
Addresses #2026


Sming/Components/spiffs/component.mk
```diff
@@ -15,8 +15,9 @@ $(COMPONENT_RULE)$(SPIFFY):
 ## Application
 
 # This controls filesystem generation, it doesn't actually disable SPIFFS support in the application
-CACHE_VARS			+= DISABLE_SPIFFS
+CONFIG_VARS			+= DISABLE_SPIFFS
 DISABLE_SPIFFS		?= 0
+APP_CFLAGS			+= -DDISABLE_SPIFFS=$(DISABLE_SPIFFS)
```
slaff: Maybe we should deprecate DISABLE_SPIFFS and for consistency use ENABLE_SPIFFS.  ENABLE_SPIFFS by default should be 1.

mikee47: Maybe something like `GENERATE_SPIFFS_FILESYSTEM`? I should really get partition tables sorted because that will make this variable obsolete. May I suggest leaving it for now?

PS. LOL. I've just added `DISABLE_WIFI`. I did ruminate about whether it should be `ENABLE_WIFI` but I figured the default of OFF makes more sense.

slaff: > May I suggest leaving it for now?

Ok, that's fine. 

> I've just added DISABLE_WIFI. I did ruminate about whether it should be ENABLE_WIFI but I figured the default of OFF makes more sense. 

Well ... well... how about ENABLE_WIFI=0 (the default of course will be 1) , same as ENABLE_MALLOC_COUNT=0.

PR #2030, opened 2020-01-23
Add iterator support to CStringArray

mikee47:
Make it simpler to use and more efficient

### DIFF
--- a/Sming/Components/malloc_count/malloc_count.cpp
+++ b/Sming/Components/malloc_count/malloc_count.cpp
@@ -1,5 +1,5 @@
 /******************************************************************************
- * malloc_count.c
+ * malloc_count.cpp
  *
  * malloc() allocation counter based on http://ozlabs.org/~jk/code/ and other
  * code preparing LD_PRELOAD shared objects.
@@ -353,6 +353,8 @@ void operator delete[](void* ptr)
 	mc_free(ptr);
 }
 
+#if __cplusplus >= 201402L
+
 void operator delete(void* ptr, size_t)
 {
 	mc_free(ptr);
@@ -362,6 +364,8 @@ void operator delete[](void* ptr, size_t)
 {
 	mc_free(ptr);
 }
+
+#endif
 
 extern "C" char* WRAP(strdup)(const char* s)
 {
--- a/Sming/Core/Data/CStringArray.cpp
+++ b/Sming/Core/Data/CStringArray.cpp
@@ -33,7 +33,7 @@ bool CStringArray::add(const char* str, int length)
 	}
 
 	// Append the content
-	auto buf = begin() + len;
+	auto buf = String::begin() + len;
 	memcpy(buf, str, length);
 
 	// Append NUL if one wasn't provided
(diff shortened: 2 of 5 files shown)


slaff:
Looks good to me. @aemseemann do you have any comments on this PR?

aemseemann:
Sorry for not responding earlier - was stuck at work the whole evening yesterday.

Unfortunately, there are problems with this iterator implementation, which are not revealed by the tests.
At first I was quite confused how test lines like `REQUIRE(it1 == it2)` even compile, because the iterator class lacks operators `==` and `!=` with another iterator instance. But then I realized, the implicit `operator bool()` conversion kicks in (confirmed with gdb), such that `REQUIRE(it1 == it2)` effectively becomes `REQUIRE((bool)it1 == (bool)it2)`. Thus, two non-end iterators would compare equal, although they point to different elements.

Further thoughts:
* An iterator should also have an `operator->()`.
* For completeness, there could be an `operator--` and `operator--(int)`
* for stdlib compatibility, `CStringArray` could have an alias `using const_iterator = Iterator;`

I noticed that private members in `Iterator` have trailing underscores which seems to violate
[Naming conventions](https://smingdev.readthedocs.io/en/latest/contribute/coding-style.html): 
> Variable names must not start with, end with or contain underscore (_) or dollar sign ($) characters.

However, I would be strongly in favour of changing that rule to allow trailing underscores for private (and maybe also protected) data members.

mikee47:
@aemseemann Thanks for taking the time to review, much appreciated. Couple of points:


    An iterator should also have an operator->().
    For completeness, there could be an operator-- and operator--(int)
    for stdlib compatibility, CStringArray could have an alias using const_iterator = Iterator;

mikee47:
@aemseemann Thanks for taking the time to review, and for spotting the opeartor== goof. Some questions:

> An iterator should also have an operator->().

I'm not clear on what this is for. Does it return a const char** ? What's it's purpose?

> For completeness, there could be an operator-- and operator--(int)

I decided against making this a reverse/random iterator because it's inefficient, and would be no better than a regular for/next loop with index. The class is still fairly basic as it doesn't support insertion/deletion/updating, but perhaps if that ever happens we might consider this as well. Unless you have a particular reason for adding this?

> for stdlib compatibility, CStringArray could have an alias using const_iterator = Iterator;

Sure, why not! Is this the correct way to implement iterators? Or should I be using a std::forward_iterator template, do you think?

aemseemann:
> > An iterator should also have an operator->().
> 
> I'm not clear on what this is for. Does it return a const char** ? What's it's purpose?

`operator->` would usually return the same as `operator*`. However, I see now that since this iterator's `operator*` is not returning an object reference, the arrow operator would not make sense, so just ignore that comment.

> > For completeness, there could be an operator-- and operator--(int)
> 
> ...Unless you have a particular reason for adding this?

No  particular reason. Lets keep it forward-only until the need for bidirectionality arises.





> Sure, why not! Is this the correct way to implement iterators? Or should I be using a std::forward_iterator template, do you think?

The `const_iterator` typedef is just a convenience for template algorithms designed to work with stdlib containers, however, its usefulness diminishes with the `auto` keyword and range-based for loops.

For the second part: Do you mean deriving from `std::iterator` with category `std::forward_iterator_tag`? I would not do this here, as `CStringArray::Iterator` behaves a bit special in that it does not refer to actual objects, so there is no `value_type` or `reference_type`.

PR #2031, opened 2020-01-24
MDNS component: initial version.

mikee47:
This is an initial attempt to provide a simple, consistent API for advertising devices via mDNS.
It works for both host and esp8266 devices.

todo:

* [x] Bypass espconn_ calls when using open lwip, call lwip directly
* [x] Enable mDNS for lwip2 on esp8266.

The version of lwip2 being used is very out of date (14 June 2017), some major work required there to get it up to speed. Lots of updates to mDNS. These can be seen in the Host implementation which is current.  Future work may involve expanding the library to support service discovery API provided by lwip2.

Note: It would be nice to have a single lwip component which provides support for all architectures. However, that's probably not going to happen as we'll likely need to patch lwip differently for each one.


Sming/Components/mdns/include/Network/Mdns.h
```diff
@@ -0,0 +1,82 @@
...
+	 */
+	using ServiceTextDelegate = Delegate<CStringArray(const Service& svc)>;
+
+	/**
+	 * @brief Initialise the responder
+	 * @param hostname
+	 * @param getServiceText Callback invoked for a service to obtain TXT information
+	 * @retval bool true on success
+	 */
+	bool init(const String& hostname, ServiceTextDelegate getServiceText);
```
aemseemann: Could the `getServiceText` argument be moved to `addService` or a separate method?
I guess "hostname.local" discovery will be sufficient for many people, so let's not confuse them with service records.

mikee47: Agreed. I think what's confusing me is that the old `mdns_info` interface for LWIP1 mixes up host discovery with the PUCK service. I had to dig around to find out what that was!

slaff:
@mikee47 What is the state of this PR?

mikee47:
@slaff Hi, yes haven't made any further progress on this. I was looking at bringing LWIP2 up to date but that's a big job and haven't had time. As this is an optional Component it could be merged as-is so further contributions could be made? It should have some value in providing a starting point for a single, more consistent interface to use MDNS services.

slaff:
> As this is an optional Component it could be merged as-is

Let's do it. Please add an empty `.cs` in the component source code directory Sming/Components/mdns/ so that the proper coding style is applied. And ping me once the this is done so that I can merge it.

Thanks for your work @mikee47 !

slaff:
@mikee47 Can you check why the travis CI build is failing?

PR #2038, opened 2020-02-01
Fix MB_RESPONSE_TIMEOUT variable

kmihaylov:
This PR makes direct use of MB_RESPONSE TIMEOUT in ModbusMaster.cpp instead of assigning the value of MB_RESPONSE_TIMEOUT to ku16MBResponseTimeout in ModbusMaster.h

This prevents error when compiling  ModbusMaster/samples/generic

### DIFF
--- a/Sming/Libraries/ModbusMaster/ModbusMaster.patch
+++ b/Sming/Libraries/ModbusMaster/ModbusMaster.patch
@@ -12,15 +12,16 @@ index cdcc534..f6f93d4 100644
        u8MBStatus = ku8MBResponseTimedOut;
      }
 diff --git a/src/ModbusMaster.h b/src/ModbusMaster.h
-index 8ab0d52..c335790 100644
+index 8ab0d52..b568a2c 100644
 --- a/src/ModbusMaster.h
 +++ b/src/ModbusMaster.h
-@@ -255,7 +255,7 @@ class ModbusMaster
+@@ -254,9 +254,6 @@ class ModbusMaster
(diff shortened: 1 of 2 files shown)


Sming/Libraries/ModbusMaster/ModbusMaster.patch
```diff
@@ -1,13 +1,26 @@
...
+index 8ab0d52..c335790 100644
 --- a/src/ModbusMaster.h
 +++ b/src/ModbusMaster.h
 @@ -255,7 +255,7 @@ class ModbusMaster
      static const uint8_t ku8MBReadWriteMultipleRegisters = 0x17; ///< Modbus function 0x17 Read Write Multiple Registers
      
      // Modbus timeout [milliseconds]
 -    static const uint16_t ku16MBResponseTimeout          = 2000; ///< Modbus timeout [milliseconds]
-+    static const uint16_t ku16MBResponseTimeout          = MB_RESPONSE_TIMEOUT; ///< Modbus timeout [milliseconds]
++    //static const uint16_t ku16MBResponseTimeout          = 2000; ///< Modbus timeout [milliseconds]
```
mikee47: Perhaps just delete these lines, rather than comment them out?

kmihaylov: I kept them to preserve consistency with the original repo, since it is not my work.

slaff: > with the original repo, since it is not my work.

Here we are talking about the patch file so it is fine to remove the commented line. @kmihaylov ping me when you have committed this change so that I can review and merge this PR.

PR #2041, opened 2020-02-03
NTPClient: ignore invalid timestamps

### LINKED ISSUE #2040: NTPClient accepts malformed responses
It looks like `NTPClient` is accepting malformed packages. Sometimes we receive NTP responses which are interpreted as `2036-02-07T06:28:16Z` (which is the uint32 NTP rollover date).

It looks like the the received packet contains a all-zero timestamp which is processed as a valid packet. I've tried to capture these invalid packets on the used gateway but failed to do so - therefore I'm not sure if it's either a bogus response, an unsupported protocol version, an transition error on the WiFi or an firmware bug. Nevertheless, other libraries have seen these problems, too. They just add a check for `timestamp == 0` to circumvent this (see https://github.com/gmag11/NtpClient/pull/77).

Do you think adding such a check is the right way to fix this, or is it worth investigating whats the cause for this malformed packets?

fooker:
This fixes #2040 by checking against `0` timestamps.

### DIFF
--- a/Sming/Core/Network/NtpClient.cpp
+++ b/Sming/Core/Network/NtpClient.cpp
@@ -122,6 +122,7 @@ void NtpClient::onReceive(pbuf* buf, IpAddress remoteIP, uint16_t remotePort)
 	debug_d("NtpClient::onReceive(%s:%u)", remoteIP.toString().c_str(), remotePort);
 
 	stopTimer();
+	auto retry = [this]() { startTimer(NTP_RESPONSE_TIMEOUT_MS); };
 
 	// We do some basic check to see if it really is a ntp packet we receive.
 	// NTP version should be set to same as we used to send, NTP_VERSION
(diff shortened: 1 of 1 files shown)


Sming/Core/Network/NtpClient.cpp
```diff
@@ -132,28 +132,34 @@ void NtpClient::onReceive(pbuf* buf, IpAddress remoteIP, uint16_t remotePort)
 	uint8_t ver = (versionMode & 0b00111000) >> 3;
 	uint8_t mode = (versionMode & 0x07);
 
-	if(mode == NTP_MODE_SERVER && (ver == NTP_VERSION || ver == (NTP_VERSION - 1))) {
-		//Most likely a correct NTP packet received.
+	if(mode != NTP_MODE_SERVER || (ver != NTP_VERSION && ver != (NTP_VERSION - 1))) {
+		// Got a dodgy packet so treat it as we would a response failure
```
slaff: I like the "reflowing" of the code. Can you please change this original comment `// Got a dodgy packet so treat it as we would a response failure` to `// Got a dodgy packet - ignore it and retry` ?

slaff:
@fooker

Some things to fix before we can merge this PR
1. Rebase your PR against the lastest `develop` branch.
2. Fix the coding style:
```
@@ -143,7 +143,7 @@ void NtpClient::onReceive(pbuf* buf, IpAddress remoteIP, uint16_t remotePort)
 	pbuf_copy_partial(buf, &timestamp, sizeof(timestamp), 40); // Copy only timestamp.
 	timestamp = ntohl(timestamp);
 
-	if (timestamp == 0) {
+	if(timestamp == 0) {
 		// Got a invalid timestamp - retry
 		startTimer(NTP_RESPONSE_TIMEOUT_MS);
 		return;'
```

You can run:
```
cd $SMING_HOME; make cs
```

3. Fix any issue reported by travis, codacy and appveyor.

mikee47:
@fooker Don't worry about appveyor - rebase will fix that.

fooker:
@mikee47 I really like the idea using a lambda for code deduplication. I've adopted it.
@slaff rebased to develop and fixed formatting (by implementing suggestion from above).

Issue #1991, opened 2019-12-08
ModbusMaster in Sming


kmihaylov:
Hello. During OpenFest I had quick conversation with @slaff about modbus that I used in Sming.

I was able to use ModbusMaster (https://github.com/nomis/ModbusMaster/tree/fixes-2.0.1) without any special modifications.

However during my research I also found that @mikee47 has his own implementation of modbus master (https://github.com/SmingHub/Sming/compare/develop...mikee47:feature/modbus)

What opinion have the core developers regarding modbus master implementation (mike47's and/or ModbusMaster)?

PS: for modbus slave I use modbusino with 1-3 major changes that I can prepare for PR.
PS2: Sming ModbusMaster firmware is currently working on my desk together with Kinco HMI for video shooting and PR.

mikee47:
I think it's down to what your application needs. Here are some specifics about my own RTU implementation:

* Fully asynchronous. Timeouts and response handling are handled via callback. No polling, loops, WDT issues.
* Minimal memory usage. Uses hardware serial buffer.
* Does not attempt to provide streaming support: modbus-RTU is a control protocol, not a streaming protocol.

I use this as part of a device control stack which talks to a load of other devices as well. That includes DMX512 controllers sitting on the same wires (using a different driver), plus devices on other busses. The approach is closer to how device stacks work in a real OS so it's capable of handling the load without affecting system responsiveness. Other implementations are just too generic and when you start to load up the application it just doesn't work well enough.

At some point I'll probably include this as part of a seprate `IO Control` library.

kmihaylov:
I would like to see your ```IO Control``` as part of Sming. I guess it could be used for modbus slave also?

mikee47:
Doesn't have slave at present but definitely something to add. I'll get around to it eventually! Things have changed a lot since #1431.

mikee47:
I like this PR as it gives an easy introduction to modbus and can try stuff out, and port existing applications over more easily. Then we can have a more integrated approach if it's required. Lots of options, good.

slaff:
> At some point I'll probably include this as part of a separate IO Control library.

Go @mikee47 Go! If the IO Control Library makes it into this release then I already have a code name for the release :)

kmihaylov:
Serial.onTransmitComplete() issues, detals here: #2043 

Meanwhile I'm not sure about the UART status register values. I tried to copy->paste @mikee47's solution, but the TX FIFO is always empty.

```
[...]

Issue #1645, opened 2019-03-11
UART TX_FIFO_EMPTY interrupt on start of transmission


egr8086:
I working on modbus RTU implementation using Sming and HardwsreSerial. Before uart transmission i need to pull up TX_ENABLE pin (GPIO 0 in my case) and pull it down after transmission complete. But txCompleteDelegate() calls on begin of transmission.

Here is initialization:
```c++
MbMaster::MbMaster(HardwareSerial & port, int baud, int8_t txEnGpio) : 
port(port), baud(baud), txEnGpio(txEnGpio) 
{
	port.begin(baud);
	port.onTransmitComplete(TransmitCompleteDelegate(&MbMaster::txCompleteDelegate, this));
	port.onDataReceived(StreamDataReceivedDelegate(&MbMaster::rxIsrDelegate, this));
	port.setTxBufferSize(256);
	port.systemDebugOutput(false);

	pinMode(txEnGpio, OUTPUT);
	digitalWrite(txEnGpio, 0);

	state = MB_IDLE;
	poll_timer.initializeUs(50, TimerDelegate(&MbMaster::poll, this)).start();
}
```
txCompleteDelegate:
```c++
void MbMaster::txCompleteDelegate(HardwareSerial& serial)
{
	if(serial.getUart()->uart_nr == port.getUart()->uart_nr) {
		state = MB_TX_COMPLETE;
		digitalWrite(txEnGpio, 0);
		last_rx_tick = tick_counter_us_x50;
	}
}
```
![UART_TX](https://user-images.githubusercontent.com/19286536/54116980-8a22d600-43f8-11e9-82be-9962fa7b2b97.png)

setTxWait(true) doesn't help.

mikee47:
@egr8086 Thanks for this, I'm working on some serial driver changes ATM so will look at this also.

I have a generic modbus RTU driver for Sming here:

https://github.com/mikee47/Sming/tree/feature/modbus/Sming/Services/Modbus

It uses a serial driver interrupt callback (see `Modbus::uartCallback`) to ensure the TX control line is switched promptly. This isn't set up until `Modbus::execute()` is called, so isn't affected by the issue you describe.

It would be great to have a generic driver for Sming, so perhaps we could work on this together?

Example:

```
Modbus modbus;
ModbusTransaction mbt;

// Process transaction result
void callback(modbusTransaction& mbt)
{
  if(mbt.status == MBE_Success) {
    // All good, deal with result data
    ...
  } else {
[...]

PR #2043, opened 2020-02-05
[Modbusino] - simple modbus rtu library

kmihaylov:
This PR introduces Modbusino. Component and sample with modifications for Sming.

The most important modification is the replacement of Serial.flush() with Serial.clear().
I removed some code that was spamming the line in case of MB errors.
I also added some minimal delays for the interface IC to ensure proper line detection when transmitting data.

Four environment variables exist for the component and one config variable is for the application (the modbus address).


Sming/Libraries/modbusino/README.rst
```diff
@@ -0,0 +1,27 @@
+modbusino RTU Library (modbus slave)
+====================================
+
+The author's project is at https://github.com/stephane/modbusino
+
+modbusino is lightweight RTU modbus slave library that supports 'read holding registers' and 'write multiple registers' functions.
+
+Four environment variables deifne the RS485 interface hardware pins and pre- and post-transmission delay.
```
slaff: Replace `deifne` with `define`. Make sure to run a spell checker.

kmihaylov: To be honest my Eclipse's spell checker doesn't work. I'm investigating.

mikee47:
I've had a longer think about this and I think to do this properly involves integrating master and slave functionality in the same library as there's a lot of common code involved. Rewriting the modbusino library as I suggested above is probably not the best way forward so happy with this PR as-is, minor points notwithstanding.

Sming/Libraries/modbusino/samples/generic/README.rst
```diff
@@ -0,0 +1,13 @@
+Modbusino RTU generic sample
+============================
+
```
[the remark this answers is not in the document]
kmihaylov: Do you mean to set the default style to bash (first line)?

Sming/Libraries/modbusino/README.rst
```diff
@@ -0,0 +1,27 @@
+modbusino RTU Library (modbus slave)
+====================================
+
+The author's project is at https://github.com/stephane/modbusino
```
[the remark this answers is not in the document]
kmihaylov: Ok, I'm removing it.

Sming/Libraries/modbusino/README.rst
```diff
@@ -0,0 +1,27 @@
...
+
+modbusino is lightweight RTU modbus slave library that supports 'read holding registers' and 'write multiple registers' functions.
+
+Four environment variables define the RS485 interface hardware pins and pre- and post-transmission delay.
```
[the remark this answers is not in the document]
kmihaylov: OK!

Sming/Libraries/modbusino/README.rst
```diff
@@ -0,0 +1,27 @@
...
+
+
+.. envvar:: MB_PRE_TX_DELAY
+
+   With :envvar:`MB_PRE_TX_DELAY` (in microseconds) the delay between changing the RE pin level and sending data can be set. Used to ensure proper line detection by the master device.
+
+
+.. envvar:: MB_POST_TX_DELAY
+
+   :envvar:`MB_POST_TX_DELAY` is the time in microseconds between the end of data transmission and releasing the line.
```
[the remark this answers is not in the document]
kmihaylov: OK!

kmihaylov:
Thank you for your corrections, these README files would be good templates.

Your opinion about the polling changed my mind about how hard would this to be implemented. If you're not in a hurry I may try these changes (in few weeks probably)?

My idea was to include some stuff while your library is getting prepared. I thought it supports both master and client? Maybe I'm in a mistake!

mikee47:
When transmitting, you can express the pre/post delays in terms of characters
then use the 'transmit complete' serial callback to set RE inactive again.

This means you don't need to wait around for the transmission to complete,
and you no longer require pre/post delay variables since the delays are now tied to the baud rate.

You could also use this approach in ModbusMaster.

```
void ModbusinoSlave::setup(long baud)
{
    Serial.begin(baud);

    // When any transmission completes, set RE line inactive
    Serial.onTransmitComplete(
        [](HardwareSerial &) { digitalWrite(RS485_RE_PIN, !RS485_TX_LEVEL); });
}

static void send_msg(uint8_t *msg, uint8_t msg_length)
{
    uint16_t crc = crc16(msg, msg_length);

    msg[msg_length++] = crc >> 8;
    msg[msg_length++] = crc & 0x00FF;

    // Set RE active, add a one-character delay before and after message
    digitalWrite(RS485_RE_PIN, RS485_TX_LEVEL);
    constexpr uint8_t NUL{0};
    Serial.write(NUL);
    Serial.write(msg, msg_length);
    Serial.write(NUL);

    // RE is set inactive by Serial transmit-complete callback
}

```

mikee47:
> My idea was to include some stuff while your library is getting prepared. I thought it supports both master and client?

https://github.com/SmingHub/Sming/issues/1991#issuecomment-563244700

kmihaylov:
OK let me give it a try and I'll try to make a commit today.

kmihaylov:
I think I'm getting closer to removing loops and use callbacks instead. The current state of the code successfully receives write multiple registers (0x10) and replies. Probably I'm doing something terrible with the dataPtr pointer, because the result isn't filled.

On the master device I'm getting timeout errors, but they're probably due some bug in the master. They appeared when I changed the delays with NUL's method. It seems that the loops in the master are going to me removed too... If I get a little bit more efficient.

mikee47:
It's the pre-delay causing the timeout, master interprets it as first character of message. What happens if you just remove the first NUL after asserting TX_EN ?

slaff:
@kmihaylov What is the state of this PR?

kmihaylov:
@slaff, there was still a problem with the timing. I'll try to finish the task in the first half of the week.

kmihaylov:
I found problem both in modbusino and in ModbusMaster. It is related to ```onTransmitComplete``` callback in modbusino. In ModbusMaster it isn't used, but is similar -> the introduced (posttransmission) delay might be not enough and the RE pin might be switched before the actual serial transmission is finished.

It's just the program that goes faster than the UART.

#1645

kmihaylov:
The latest update of ```modbusino``` library introduces the following updates (I may forget something):
1. Remove NUL character from the beginning of a message (suspected to trigger TX FIFO empty interrupt that would release the TX line).
2. Add a NUL characted to the end of the message.
3. The length of the received message is now not strictly compared (==), but 'at least' (>=). The message length is calculated from the information contained in the message frame, and not by the length of the UART hardware.
4. The RX buffer wasn't cleared on successful message. This is now fixed. In the past, under certain circumstances proper work of the driver was observed due to the introduced delays, that would trigger a timeout and would clear the RX buffer, which is not how the program should work.

Unfortunately too many errors somehow led to proper work of the previous versions. I hope that this release is much more stable and production-ready.

slaff:
> Unfortunately too many errors somehow led to proper work of the previous versions. I hope that this release is much more stable and production-ready.

@kmihaylov does it mean that this PR is ready for review and merging?

kmihaylov:
Yes.

slaff:
> Yes.

@mikee47 @kpishere Can you review this PR and test it with a real device?

kmihaylov:
If boards are needed maybe I can help.

Sming/Libraries/modbusino/samples/generic/app/application.cpp
```diff
@@ -0,0 +1,24 @@
+#include <SmingCore.h>
+#include <Debug.h>
+#include <Modbusino.h>
+
+#define ARRLEN 3
+uint16_t mbDataArray[ARRLEN] = {0,0,0};
+ModbusinoSlave MbSlave(MB_SLAVE_ADDR, mbDataArray, ARRLEN);
```
slaff, line 7: Rename `MbSlave` to `mbSlave` to match the [coding style](https://sming.readthedocs.io/en/latest/contribute/coding-style.html).

Sming/Libraries/modbusino/samples/generic/app/application.cpp
```diff
@@ -0,0 +1,24 @@
+#include <SmingCore.h>
```
slaff, line 1: Add a new empty file named '.cs' inside of this folder 'Sming/Libraries/modbusino/samples/generic/app/application.cpp' and run the following commands to apply the coding style rules:
```
cd $SMING_HOME # change this to %SMING_HOME% if you are using Windows
make cs
``` 
Make sure to include the new .cs file in this PR.

slaff:
@mikee47 Any comments here?

Sming/Libraries/modbusino/samples/generic/README.rst
```diff
@@ -0,0 +1,16 @@
...
+.. highlight:: bash
+
+The sample code provides three element uint16_t array used for the modbus slave registers. These values are printed each second through UART1.
+mbSlaveLoopTimer checks for new data each 10 milliseconds (Only checks for new data. Serial communication is handled by hardware.)
```
mikee47, line 7: Needs updating to reflect actual operation i.e. register values are printed only when a valid WriteMultipleRegisters command is received, no timer.

Sming/Libraries/modbusino/samples/generic/README.rst
```diff
@@ -0,0 +1,16 @@
...
+
+Several environment variables (envvar:`RS485_RE_PIN` :envvar:`RS485_TX_LEVEL` :envvar:`MB_PRE_TX_DELAY` :envvar:`MB_POST_TX_DELAY`) can be used for configuration.
```
mikee47, line 9: `MB_PRE_TX_DELAY` and `MB_POST_TX_DELAY` don't exist!

Sming/Libraries/modbusino/component.mk
```diff
@@ -0,0 +1,19 @@
...
+RS485_TX_LEVEL ?= HIGH
+COMPONENT_CXXFLAGS += -DRS485_TX_LEVEL=$(RS485_TX_LEVEL)
+
+COMPONENT_VARS += MB_PRE_TX_DELAY
+MB_PRE_TX_DELAY ?= 100
+COMPONENT_CXXFLAGS += -DMB_PRE_TX_DELAY=$(MB_PRE_TX_DELAY)
+
+COMPONENT_VARS += MB_POST_TX_DELAY
+MB_POST_TX_DELAY ?= 1000
+COMPONENT_CXXFLAGS += -DMB_POST_TX_DELAY=$(MB_POST_TX_DELAY)
```
mikee47, line 19: Not used - remove

Sming/Libraries/modbusino/README.rst
```diff
@@ -0,0 +1,37 @@
...
+   Default: 100 (microseconds).
+
+   Time to wait after RE pin level change, before writing serial data. Used to ensure proper line detection by the master device.
+
+
+.. envvar:: MB_POST_TX_DELAY
+
+   Default: 1000 (microseconds).
+
+   Delay after data transmission, before RE pin level change.
```
mikee47, line 37: Not used - remove

kmihaylov:
I'll fix these once I get back to the city, because I'm currently locked out. BTW I'm using 115200, probably that is why I didn't notice the delays you have observed. Maybe I can switch to 9600 and try the interrupt.

mikee47:
It's not actually related to baud rate. Below, the blue trace is _RE_ and red trace is slave response. I see either this:

image

And sometimes this:

image

The delay is caused by the `mbPrint()` callback - commenting it out eliminates the delay. This is because on these occasions the RECEIVE handler is getting called before transmitComplete.

If the master tries to send another command within this 5ms period then it'll get lost, otherwise it may not be an issue.

kmihaylov:
> I note in modbusinoSlave.cpp you've got `uartCallback` commented-out in (lines 66+ and 96+). If you were having trouble getting this to work note that `uartCallback` is invoked in interrupt context so line 66 would require `IRAM_ATTR` adding to the function.

Sure I have. How could I forget `IRAM_ATTR` ! Just did as you propose, however the interrupt did not fire. Before I had some behavior like the interrupt is called after the transmission of the first NUL character (when I didn't use `IRAM_ATTR`). Should I force enable interrupts?

mikee47:
> ... however the interrupt did not fire.

Because the call to `Serial.onDataReceived` resets the interrupt callback. For this to work you'd need to handle receives in your ISR as well. Operation seems good enough without this so I'd suggest just removing the commented-out lines.

> Before I had some behavior like the interrupt is called after the transmission of the first NUL character (when I didn't use `IRAM_ATTR`). Should I force enable interrupts?

Interrupt occurs when the transmit FIFO becomes empty. When idle, that will (probably) happen as soon as you write the first character because it gets moved (almost) immediately from the FIFO into the output shift register. A simple way to mitigate this is to set a flag when the message has been written; ignore any TX_EMPTY interrupt unless that flag has been set.

kmihaylov:
@mikee47 @slaff 
I removed the unused setUartCallback function. Also I added a note to the component's readme describing the RE delay. I understand that better is to replace onTransmitComplete with ISR, so maybe the next update would do so...
Meanwhile I'm sticking with Mike's recommendation to leave it this way.

slaff:
> Meanwhile I'm sticking with Mike's recommendation to leave it this way.

@kmihaylov If the PR is ready in the current form I will schedule it for merging.

kmihaylov:
Let me compile and flash it just to be sure and I'll write if everything works in 15 minutes.

kmihaylov:
Yes, it works as expected. I do not observe RE delays on the default conditions for ModbusMaster and modbusino, as they exist now in the PR (115200). I remind that Mike is using another testing approach.

slaff:
> I remind that Mike is using another testing approach.

@mikee47 Do you want to do a final test before I merge this PR?

mikee47:
@slaff Looks fine, happy to merge.

PR #2044, opened 2020-02-05
Allow to specify ROM size

fooker:
This adds an optional parameter allowing to specify the ROM size of each slot. The default value of this settings mimics the current behavior.

### DIFF
--- a/Sming/Components/rboot/component.mk
+++ b/Sming/Components/rboot/component.mk
@@ -39,19 +39,22 @@ endif
 
 CONFIG_VARS				+= RBOOT_ROM0_ADDR RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
 
-# Loation of first ROM (default is sector 2 after rboot and rboot config sector)
+# Location and size of first ROM (default is sector 2 after rboot and rboot config sector)
 RBOOT_ROM0_ADDR			?= 0x002000
+RBOOT_ROM0_SIZE			?= (0x100000 - $(RBOOT_ROM0_ADDR))
 
 # The parameter below specifies the location of the second ROM.
 # You need a second slot for any kind of firmware update mechanism.
 # Leave empty if you don't need a second ROM slot.
 RBOOT_ROM1_ADDR			?=
+RBOOT_ROM1_SIZE			?= (0x100000 - $(RBOOT_ROM1_ADDR))
 
 # The parameter below specifies the location of the GPIO ROM.
 # It is only used when RBOOT_GPIO_ENABLED = 1
 # Note that setting this parameter will only configure rboot.
 # The Sming build system does not create a ROM for this slot.
 RBOOT_ROM2_ADDR			?=
+RBOOT_ROM2_SIZE			?= (0x100000 - $(RBOOT_ROM2_ADDR))
 
 ifeq ($(RBOOT_GPIO_ENABLED),0)
 ifneq ($(RBOOT_ROM2_ADDR),)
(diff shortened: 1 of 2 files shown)


slaff:
@fooker Please see the issues reported in Travis. Also make sure that at least the Basic_Blink and Basic_rBoot samples compile locally on your computer.

fooker:
> @fooker Please see the issues reported in Travis. Also make sure that at least the Basic_Blink and Basic_rBoot samples compile locally on your computer.

@slaff: Sure, I first wanted to collect some feedback before investing time in polishing stuff (Maybe I should have added a WIP or RFC tag)

Sming/Components/rboot/component.mk
```diff
@@ -162,24 +165,30 @@ ifndef RBOOT_EMULATION
...
 $2: $(RBOOT_LD_TEMPLATE)
 	$$(info LDGEN $$@)
 	$$(Q) $(AWK) 'match($$$$0, /^.*irom0_0_seg[ \t]*:[ \t]*/) { \
-			printf "%sorg = 0x40200010 + ($$($3) & 0xFFFFF), len = 1M - ($$($3) & 0xFFFFF) - 0x10\n", substr($$$$0, RSTART, RLENGTH); next \
+			printf "%sorg = 0x40200010 + ($$($3) & 0xFFFFF), len = ($$($4) & 0xFFFFF) - 0x10\n", substr($$$$0, RSTART, RLENGTH); next \
 		} 1 { print $$$$0 }' $$< > $$@
-	$$(Q) echo GEN_$3 := $($3) > $2.config
+	$$(Q) rm -f $2.config
+	$$(Q) echo 'GEN_$3 := $($3)' > $2.config
+	$$(Q) echo 'GEN_$4 := $($4)' > $2.config
```
mikee47: `>>`

Sming/Components/rboot/component.mk
```diff
@@ -39,19 +39,22 @@ endif
 
 CONFIG_VARS				+= RBOOT_ROM0_ADDR RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
 
-# Loation of first ROM (default is sector 2 after rboot and rboot config sector)
+# Loation and size of first ROM (default is sector 2 after rboot and rboot config sector)
```
mikee47: `Location`

mikee47:
@aemseemann Any thoughts?

Sming/Components/rboot/component.mk
fooker: Right

Sming/Components/rboot/component.mk
fooker: Done

aemseemann:
A general note: The ROM image is composed of more than just what the linker puts into 'irom0_0_seg', i.e. the final image may still exceed the size specified by `RBOOT_ROMx_SIZE`. 

If the goal is to detect flash memory overflows, it would be more reliable to compare the size of the resulting image file (see `RBOOT_ROM_x_BIN`) to the configured value instead of adjusting a section in the linker script.

Bonus: It would be nice if the build system also detects misconfigured sizes, e. g. if `RBOOT_ROM_x_BIN + RBOOT_ROMx_SIZE` crosses a 1MB boundary.

Sming/Components/rboot/component.mk
```diff
@@ -39,19 +39,22 @@ endif
 
 CONFIG_VARS				+= RBOOT_ROM0_ADDR RBOOT_ROM1_ADDR RBOOT_ROM2_ADDR
 
-# Loation of first ROM (default is sector 2 after rboot and rboot config sector)
+# Location and size of first ROM (default is sector 2 after rboot and rboot config sector)
 RBOOT_ROM0_ADDR			?= 0x002000
+RBOOT_ROM0_SIZE			?= (0x100000 - $(RBOOT_ROM0_ADDR))
```
aemseemann: Note: `RBOOT_ROMx_ADDR` may be larger than 0x100000 (for devices with more than 1MB of flash memory). Hence this calculation yields a negative value.

Sming/Components/rboot/component.mk
```diff
@@ -162,24 +165,30 @@ ifndef RBOOT_EMULATION
...
 endif
+ifneq ($$(GEN_$4),$$($4))
+.PHONY: $2
+endif
 # Generate linker script from template
 $2: $(RBOOT_LD_TEMPLATE)
 	$$(info LDGEN $$@)
 	$$(Q) $(AWK) 'match($$$$0, /^.*irom0_0_seg[ \t]*:[ \t]*/) { \
-			printf "%sorg = 0x40200010 + ($$($3) & 0xFFFFF), len = 1M - ($$($3) & 0xFFFFF) - 0x10\n", substr($$$$0, RSTART, RLENGTH); next \
+			printf "%sorg = 0x40200010 + ($$($3) & 0xFFFFF), len = ($$($4) & 0xFFFFF) - 0x10\n", substr($$$$0, RSTART, RLENGTH); next \
```
aemseemann: Since `$4` is a size, I'd crop at 0x100000 instead of masking by 0xFFFFF.

It would be even better if the build system ensures that only reasonable sizes can be used and no modification is necessary in the linker script.

Sming/Components/rboot/component.mk
```diff
@@ -39,19 +39,22 @@ endif
...
 # Leave empty if you don't need a second ROM slot.
 RBOOT_ROM1_ADDR			?=
+RBOOT_ROM1_SIZE			?= (0x100000 - $(RBOOT_ROM1_ADDR))
 
 # The parameter below specifies the location of the GPIO ROM.
 # It is only used when RBOOT_GPIO_ENABLED = 1
 # Note that setting this parameter will only configure rboot.
 # The Sming build system does not create a ROM for this slot.
 RBOOT_ROM2_ADDR			?=
+RBOOT_ROM2_SIZE			?= (0x100000 - $(RBOOT_ROM2_ADDR))
```
aemseemann: This size is not used in the build system, because the build system never generates an image for slot2 (the address is only used for configuring the bootloader).
If there is no sanity-checking the memory layout (i.e. check if slots do not overlap with each other or reserved flash areas), `RBOOT_ROM2_SIZE` should be removed IMO.

slaff:
The latest changes in `develop` make this PR unnecessary.

PR #2050, opened 2020-02-20
Feature: distance measurement using vl53l0x sensor

slaff:


samples/Distance_Vl53l0x/app/application.cpp
```diff
@@ -0,0 +1,51 @@
+#include <SmingCore.h>
+#include <Adafruit_VL53L0X.h>
+
+Adafruit_VL53L0X lox = Adafruit_VL53L0X();
```
mikee47: `Adafruit_VL53L0X lox;` is sufficient.

samples/Distance_Vl53l0x/app/application.cpp
```diff
@@ -0,0 +1,51 @@
...
+	lox.setLongRangeMode(true);
+
+	if(!lox.begin()) {
+		Serial.println(F("Failed to boot VL53L0X"));
+		while(1) {
+		}
+	}
+
+	Serial.println(F("VL53L0X API Simple Ranging example\n\n"));
+	loopTimer.initializeMs(100, loop).start();
```
mikee47: Could use `loopTimer.initializeMs<100>(loop).start();`

PR #2055, opened 2020-04-05
Updated NON-OS SDK to 3.0.3

slaff:


mikee47:
I note that version.h is wrong, shows 3.0.1, perhaps worth patching to 3.0.3 to avoid confusion?

PR #2059, opened 2020-04-14
ModbusMaster fixes

kmihaylov:
This PR introduces the following fixes to ModbusMaster:
0. ModbusMaster is patched to use single ```_serial->write()``` to send the whole message instead of 'for loop'.
1. ModbusMaster is now patched so as to send single NUL character after the message in order to ensure TX line release after the whole message is sent.
2. Delays in the generic sample are commented out, but left for reference (the example works well without the delays).
3. In ```mbLoop()``` the variable ```result``` was used instead of ```mbResult``` when reporting the error code. ```result``` is removed and the proper ```mbResult``` is used.
4. In ```mbLogReceive()```  debug message of the RX ADU is added.
5. In ```mbLogTransmit``` ADU is changed to "TX ADU" to clarify the message.

### DIFF
--- a/Sming/Libraries/ModbusMaster/ModbusMaster.patch
+++ b/Sming/Libraries/ModbusMaster/ModbusMaster.patch
@@ -1,8 +1,8 @@
 diff --git a/src/ModbusMaster.cpp b/src/ModbusMaster.cpp
-index cdcc534..ecb03a5 100644
+index cdcc534..b08a335 100644
 --- a/src/ModbusMaster.cpp
 +++ b/src/ModbusMaster.cpp
-@@ -756,15 +756,13 @@ uint8_t ModbusMaster::ModbusMasterTransaction(uint8_t u8MBFunction)
+@@ -756,15 +756,12 @@ uint8_t ModbusMaster::ModbusMasterTransaction(uint8_t u8MBFunction)
    // flush receive buffer before transmitting request
    while (_serial->read() != -1);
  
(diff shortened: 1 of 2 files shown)


Sming/Libraries/ModbusMaster/samples/generic/app/application.cpp
```diff
@@ -47,12 +46,12 @@ void mbLoop()
 void preTransmission()
 {
 	digitalWrite(RS485_RE_PIN, HIGH);
-	delayMilliseconds(2);
+	//delayMilliseconds(1);
```
slaff, line 49: If the delay is not needed just remove that line.

Sming/Libraries/ModbusMaster/samples/generic/app/application.cpp
```diff
@@ -47,12 +46,12 @@ void mbLoop()
...
 }
 
 void postTransmission()
 {
-	delayMicroseconds(500);
+	//delayMicroseconds(500);
```
slaff, line 54: Remove the commented line too.

Sming/Libraries/ModbusMaster/ModbusMaster.patch
```diff
@@ -1,8 +1,27 @@
...
+   {
+     _preTransmission();
+   }
+-  for (i = 0; i < u16ModbusADUSize; i++)
+-  {
+-    _serial->write(u8ModbusADU[i]);
+-  }
++  constexpr uint8_t NUL{0};
++  _serial->write(u8ModbusADU, u16ModbusADUSize);
++  _serial->write(NUL);
```
slaff, line 20: Remove `constexpr uint8_t NUL{0};`. And then replace `_serial->write(NUL);` with `_serial->write((uint8_t)0);

Sming/Libraries/ModbusMaster/samples/generic/app/application.cpp
kmihaylov: Should I remove also 
```
		/*
		for(i = 0; i < nrOfRegisters; i++) {
			buffer[i] = mbMaster.getResponseBuffer(i);
			debugComPort.printf("Reg %d: %d\r\n", i, buffer[i]);
		}
		*/
```
then?

slaff: The pasted and commented code is not part of your current changes. If this if-block is replaced by something else or is not useful at all then you can also remove it.

kmihaylov: I'll leave it. Improved example should take into account the number of registers to read/write. Maybe I'll do it later.

PR #2072, opened 2020-05-18
Prevent name clash with macro in ESP-IDF.

slaff:
Preparation for ESP32 support.


mikee47:
@slaff Using `ALIGNUP4` (or `ALIGNUP32`) would be less ambiguous, I think, and easier to introduce 16/64-bit versions. I can push this to the `FlashString` library to avoid patching, can't think of a good reason not to.

slaff:
> Using ALIGNUP4 (or..

Let it be ALIGNUP4. I was thinking initially to use ESP-IDF's ALIGNUP but for the moment we don't have the need for it. If that need emerges in the future we will just introduce in Sming ESP-IDF's ALIGNUP and change the code to use it.

>  I can push this to the FlashString library to avoid patching, can't think of a good reason not to.

This would be great! Just tell me when you are done so I can update the submodule SHA.

PR #2073, opened 2020-05-18
IpAddress fixes to work with LWIP v1, LWIP 2 with IPv4 and LWIP 2 wit…

slaff:
…h IPv4+IPv6.

Preparation for the ESP32 support.

### DIFF
--- a/Sming/Wiring/IpAddress.cpp
+++ b/Sming/Wiring/IpAddress.cpp
@@ -22,26 +22,15 @@
 
 void IpAddress::fromString(const String& address)
 {
-	const char* p = address.c_str();
-	for(unsigned i = 0; i < sizeof(ip4_addr_t); ++i) {
-		operator[](i) = strtol(p, const_cast<char**>(&p), 10);
-		if (*p++ != '.')
-			break;	// Missing '.' or end of input string
-	}
+    ipaddr_aton(address.c_str(), &(this->address));
 }
 
 size_t IpAddress::printTo(Print& p) const
 {
-    size_t n = 0;
-    for (unsigned i = 0; i < sizeof(ip4_addr_t) - 1; i++) {
-        n += p.print(operator[](i), DEC);
-        n += p.print('.');
-    }
-    n += p.print(operator[](3), DEC);
-    return n;
+    return p.print(toString());
 }
 
 String IpAddress::toString() const
 {
-	return ip4addr_ntoa(&address);
+    return ipaddr_ntoa(&address);
 }
--- a/Sming/Wiring/IpAddress.cpp
+++ b/Sming/Wiring/IpAddress.cpp
@@ -32,5 +32,7 @@ size_t IpAddress::printTo(Print& p) const
 
 String IpAddress::toString() const
 {
-    return ipaddr_ntoa(&address);
+    char text[16];
+    ipaddr_ntoa_r(&address, text, 16);
+    return text;
 }
--- a/Sming/Wiring/IpAddress.cpp
+++ b/Sming/Wiring/IpAddress.cpp
@@ -23,7 +23,7 @@
 void IpAddress::fromString(const String& address)
 {
 	const char* p = address.c_str();
-	for(unsigned i = 0; i < 4; ++i) {
+	for(unsigned i = 0; i < sizeof(ip4_addr_t); ++i) {
 		operator[](i) = strtol(p, const_cast<char**>(&p), 10);
 		if (*p++ != '.')
 			break;	// Missing '.' or end of input string
@@ -33,7 +33,7 @@ void IpAddress::fromString(const String& address)
 size_t IpAddress::printTo(Print& p) const
 {
     size_t n = 0;
-    for (unsigned i = 0; i < 3; i++) {
+    for (unsigned i = 0; i < sizeof(ip4_addr_t) - 1; i++) {
         n += p.print(operator[](i), DEC);
         n += p.print('.');
     }
@@ -43,12 +43,5 @@ size_t IpAddress::printTo(Print& p) const
 
 String IpAddress::toString() const
 {
-    String res;
-    for (unsigned i = 0; i < 4; i++)
-    {
-    	if (i)
-    	  res += '.';
-    	res += operator[](i);
-    }
-    return res;
+	return ip4addr_ntoa(&address);
 }
--- a/Sming/Wiring/IpAddress.cpp
+++ b/Sming/Wiring/IpAddress.cpp
@@ -22,7 +22,6 @@
 
 void IpAddress::fromString(const String& address)
 {
-	this->address.addr = 0;
 	const char* p = address.c_str();
 	for(unsigned i = 0; i < 4; ++i) {
 		operator[](i) = strtol(p, const_cast<char**>(&p), 10);
@@ -44,9 +43,8 @@ size_t IpAddress::printTo(Print& p) const
 
 String IpAddress::toString() const
 {
-	String res;
-    res.reserve(sizeof(address) * 4);
-    for (unsigned i = 0; i < sizeof(address); i++)
+    String res;
+    for (unsigned i = 0; i < 4; i++)
     {
     	if (i)
     	  res += '.';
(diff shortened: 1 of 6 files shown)


Sming/Wiring/IpAddress.cpp
```diff
@@ -44,9 +43,8 @@ size_t IpAddress::printTo(Print& p) const
 
 String IpAddress::toString() const
 {
-	String res;
-    res.reserve(sizeof(address) * 4);
```
mikee47, line 48: `reserve` call is an optimisation, should remain `res.reserve(sizeof(ip4_addr_t) * 4);`

Sming/Wiring/IpAddress.cpp
```diff
@@ -44,9 +43,8 @@ size_t IpAddress::printTo(Print& p) const
...
-    for (unsigned i = 0; i < sizeof(address); i++)
+    String res;
+    for (unsigned i = 0; i < 4; i++)
     {
     	if (i)
```
mikee47, line 49: Missing braces { }

Sming/Wiring/IpAddress.h
```diff
@@ -47,75 +53,83 @@ class IpAddress : public Printable
...
+#endif
 
-#if LWIP_VERSION_MAJOR == 2
-	IpAddress(ip_addr_t address)
+	IpAddress(uint8_t first_octet, uint8_t second_octet, uint8_t third_octet, uint8_t fourth_octet)
 	{
-		this->address = address;
+		IP_ADDR4(&address, first_octet, second_octet, third_octet, fourth_octet);
+	}
+
```
mikee47, line 77: Should move this back to line 50, minimise diff.

Sming/Wiring/IpAddress.h
```diff
@@ -47,75 +53,83 @@ class IpAddress : public Printable
...
 	bool operator==(const uint8_t* addr) const
 	{
 		return *this == IpAddress(addr);
 	}
 
-	bool operator!=(const IpAddress& addr) const
+	bool operator!=(const IpAddress& otherAddress) const
 	{
-		return !ip_addr_cmp(&address, &addr.address);
+		return !ip_addr_cmp((ip_addr_t *)&address, (ip_addr_t* )&otherAddress);
```
mikee47, line 132: Could write as `return !ip_addr_cmp(&address, &otherAddress.address);` to avoid cast

Sming/Wiring/IpAddress.h
```diff
@@ -132,36 +146,38 @@ class IpAddress : public Printable
...
 	{
-		return ip_addr_netcmp(&address, &addr.address, &mask.address);
+		return ip4_addr_netcmp((ip4_addr_t* )&address, (ip4_addr_t* )&addr, (ip4_addr_t* )&mask);
 	}
 
 	// Overloaded index operator to allow getting and setting individual octets of the address
 	uint8_t operator[](int index) const
 	{
-		if(unsigned(index) >= sizeof(address)) {
+		if(unsigned(index) >= 4) {
```
mikee47, line 155: `sizeof(ip4_addr_t)` instead of magic number - see also other instances.

Sming/Wiring/IpAddress.h
```diff
@@ -27,6 +27,12 @@
 #define LWIP_IP_ADDR_T const ip_addr_t
 #else
 typedef struct ip_addr ip_addr_t;
+typedef ip_addr_t ip4_addr_t;
+#define IP_ADDR4(IP, A, B, C, D) IP4_ADDR(IP, A, B, C, D)
```
mikee47, line 31: Why do we need IP_ADDR4? Looking at the IDF code it appears to be related to IPv6, in which case perhaps leave this change until it's needed? Or perhaps we should add v6 support at this stage, thus justifying the changes?

Sming/Wiring/IpAddress.h
slaff: > sizeof(ip4_addr_t) instead of magic number

Hm.. the size of IPv4 address is constant while the size of ip4_addr_t struct might not be...

Sming/Wiring/IpAddress.h
slaff: > Why do we need IP_ADDR4?

Because it creates an address in which the IPv4 part is initialized using the octets that we provide.  In LWIP 1/2 with IP4 only it does not make much sense but in LWIP 2 with IPv4 + IPv6 it sets the correct bytes plus sets the type of the whole ip_addr_t structure to be IPv4 address. Or  you have meant something else?

Sming/Wiring/IpAddress.cpp
```diff
@@ -22,35 +22,15 @@
...
-	String res;
-    res.reserve(sizeof(address) * 4);
-    for (unsigned i = 0; i < sizeof(address); i++)
-    {
-    	if (i)
-    	  res += '.';
-    	res += operator[](i);
-    }
-    return res;
+    return ipaddr_ntoa(&address);
```
mikee47, line 35: Prefer `ipaddr_ntoa_r`

Sming/Wiring/IpAddress.h
```diff
@@ -49,73 +55,81 @@ class IpAddress : public Printable
...
 	{
-		this->address = address;
+		ip_addr_copy_from_ip4(address, addr);
 	}
 #endif
 
 	IpAddress(const uint8_t* address)
 	{
-		IP4_ADDR(&this->address, address[0], address[1], address[2], address[3]);
+		IP_ADDR4(&this->address, address[0], address[1], address[2], address[3]);
```
mikee47, line 85: Could change to `IpAddress(const uint8_t[4])` to clarify. Could probably also do with comment for constructor.
```
/*
@param address IP4 address octets
*/
```

Sming/Wiring/IpAddress.h
```diff
@@ -49,73 +55,81 @@ class IpAddress : public Printable
...
+	operator ip_addr_t*()
 	{
-		return reinterpret_cast<ip_addr*>(&address);
+		return &address;
 	}
 
-#if LWIP_VERSION_MAJOR == 2
-	operator ip_addr_t*()
+#if LWIP_VERSION_MAJOR == 2 && LWIP_IPV6
+	operator ip4_addr_t const()
```
mikee47, line 109: Should this be `operator ip4_addr_t() const` ?

Sming/Wiring/IpAddress.h
```diff
@@ -132,36 +146,38 @@ class IpAddress : public Printable
 
 	bool compare(const IpAddress& addr, const IpAddress& mask) const
 	{
-		return ip_addr_netcmp(&address, &addr.address, &mask.address);
+		return ip4_addr_netcmp((ip4_addr_t* )&address, (ip4_addr_t* )&addr, (ip4_addr_t* )&mask);
```
mikee47, line 149: `ip_addr_netcmp(&address, &addr.address, &mask.address)`

Sming/Wiring/IpAddress.h
```diff
@@ -132,36 +146,38 @@ class IpAddress : public Printable
...
+		if(unsigned(index) >= sizeof(ip4_addr_t)) {
 			abort();
 		}
-		return reinterpret_cast<uint8_t*>(&address)[index];
+
+		return reinterpret_cast<uint8_t*>(&ip_2_ip4(&address)->addr)[index];
 	}
 
 	// Overloaded copy operators to allow initialisation of IpAddress objects from other types
 	IpAddress& operator=(const uint8_t* address)
```
mikee47, line 172: `uint8_t address[4]`

Sming/Wiring/IpAddress.h
```diff
@@ -49,73 +55,81 @@ class IpAddress : public Printable
...
 	}
 
-	bool operator!=(const IpAddress& addr) const
+	bool operator!=(const IpAddress& otherAddress) const
 	{
-		return !ip_addr_cmp(&address, &addr.address);
+		return !ip_addr_cmp(&address, &otherAddress.address);
 	}
 
 	bool operator!=(const uint8_t* addr) const
```
mikee47, line 135: `const uint8_t addr[4]`

Sming/Wiring/IpAddress.h
```diff
@@ -49,73 +55,81 @@ class IpAddress : public Printable
...
 #endif
 
-	bool operator==(const IpAddress& addr) const
+	bool operator==(const IpAddress& otherAddress) const
 	{
-		return ip_addr_cmp(&address, &addr.address);
+		return ip_addr_cmp(&address, &otherAddress.address);
 	}
 
 	bool operator==(const uint8_t* addr) const
```
mikee47, line 125: `const uint8_t addr[4]`

slaff:
@mikee47 Thanks for the review suggestions and the patience!

PR #2075, opened 2020-05-19
Allow every architecture to set the flash memory start address.

slaff:


mikee47:
What is the FlashString patch for?

slaff:
> What is the FlashString patch for?

To remove the `#ifndef ARCH_HOST` checks when isFlashPtr is used.  If you want add them to FlashString directly and I will use the latest FlashString version.

mikee47:
Problem is getting Host builds to place 'flash' data in a separate segment, so how do we implement `isFlashPtr` for that?

slaff:
> Problem is getting Host builds to place 'flash' data in a separate segment

Well, architectures that don't need to know where the "flash data" is stored don't define FLASH_MEMORY_START_ADDR. In that case FLASH_MEMORY_START_ADDR is 0 and then the isFlashPtr check always succeeds. Thus there is no need for an exception for the host architecture.

mikee47:
> isFlashPtr check always succeeds

Hence if you build and run FlashString test application it goes pop in `ObjectBase.cpp` line 78:

```
assert(!isFlashPtr(this));
```

This happens if you try to instantiate a FlashString instance:

```
FlashString empty;
```

This is perfectly valid, creates a reference to a NULL FlashString.

My point is that these checks cannot be validated for Host, so your proposed patch is dangerous.

slaff:
> My point is that these checks cannot be validated for Host, so your proposed patch is dangerous.

Ok, I will remove the patch.

slaff:
> Ok, I will remove the patch.

Done.

PR #2091, opened 2020-06-29
Fix/sdcard byte order

mopac:
I couldn't get SDCard to work on a Wemos D1 mini with an SD Card adaptor. Always gave the same error. SD Card Error: FF.
I hooked up the scope and discovered that the SD Card was receiving SPI messages but never responsed.
I hooked up the logic analyser to analyse the problem. Had to slow down the SPI to allow my analyser to work.
Image SPI_MSBFIRST.jpg is the result.
You can see first 80 dummy clock cycles; then 2 dummy reads, then the 6 byte CMD0 which is required to initialise the SDcard. 
Then 11 dummy reads while the host waits for a response from the SD card. No response so the cycle repeats (not shown)
The problem is the order of the bytes. The correct command is 40 00 00 00 00 95 (command ID, 32 attribute bits, checksum)
What is actually shown is 00 00 00 40 95 00. The byte order is wrong on a 32 bit boundary.

The fix is to initialise SPI with byte order set to LSBFISRT.

Image SPI_LSBFIRST.jpg shows the 80 dummy clock cycles, followed by 2 dummy reads then the correct CMD0. Now the SD responds with 01. Which prompts the host to 
send CMD8 requesting the card type. Everything works fine now.

I changed sdcard.cpp SDCard_begin function to take byte order and frequency limit as variables [SDCard_begin(uint8 PIN_CARD_SS, uint8 byte_order, uint32 freq_limit)]
This allows application.cpp to choose the byte order as well as set frequency limit.
Frequency limit is useful because from various related posts, it appears some configurations become unreliable at 40MHz.
It also allows me an easy way to slow things down so my logic analyser will work.

The other change was to change the test file that is written and read from 'test' to 'test.txt'. The former could be written but not read.
No idea why but test.txt works for both read and write

Ray
SPI_LSBFIRST
SPI_MSBFIRST


samples/SDCard/README.rst
```diff
@@ -1,4 +1,27 @@
 SD Card
 =======
 
-SDCard/FAT file usage and write benchmark.
+I couldn't get SDCard to work on a Wemos D1 mini with an SD Card adaptor. Always gave the same error. SD Card Error: FF.
```
slaff, line 4: Please, remove this text from here. This file should contain helpful information describing what the sample is doing.
As such this information is very good as part of the PR ticket but not quite useful in the README file.

Sming/Libraries/SDCard/SDCard.cpp
```diff
@@ -38,19 +38,29 @@ Descr: Low-level SDCard functions
...
+uint8 SPI_byte_order;		/* The order the SPI bytes are sent */
+uint32 SPI_freq_limit;		/* The max frequency the SPI should run */
+uint32 SPI_init_freq;		/* SPI frequency usined for init */
 
 
 #define SCK_SLOW_INIT 10
 #define SCK_NORMAL 0
 
-void SDCard_begin(uint8 PIN_CARD_SS)
+void SDCard_begin(uint8 PIN_CARD_SS, uint8 byte_order, uint32 freq_limit)
```
slaff, line 49: Please, use camelCase for the variable names. For example `byteOrder` instead of `byte_order`. 
Once you are ready with the changes make sure to run the coding style checker: https://sming.readthedocs.io/en/latest/contribute/coding-style.html

Sming/Libraries/SDCard/README.md
```diff
@@ -1,3 +1,27 @@
 # SD Card
 
 Low-level support code for accessing SD Cards using FATFS.
+
+I couldn't get SDCard to work on a Wemos D1 mini with an SD Card adaptor. Always gave the same error. SD Card Error: FF.
```
slaff, line 5: Please, remove this text from here. Here we should have helpful information about the SDCard library itself. Explanation of issues with the library is good as part of a PR ticket information but does not belong here.

slaff: Please, leave the images as part of the ticket and remove them from this commit. Remove `samples/SDCard/SPI_MSBFIRST.JPG`  and `samples/SDCard/SPI_LSBFIRST.JPG`.

samples/SDCard/README.rst
mopac: Removed

Sming/Libraries/SDCard/SDCard.cpp
mopac: changed to camelCase and run through clang

Sming/Libraries/SDCard/README.md
mopac: text and images removed

Sming/Libraries/SDCard/SDCard.h
```diff
@@ -11,10 +11,10 @@ Descr: Low-level SDCard functions
 #include <SmingCore.h>
 #include "SPISoft.h"
 
-void SDCard_begin(uint8 PIN_CARD_SS);
+void SDCard_begin(uint8 PIN_CARD_SS, uint8 byteOrder, uint32 freqLimit);
```
mikee47, line 14: Use `uint8_t` rather than the non-standard `uint8`, simililarly `uint32` -> `uint32_t`

Sming/Libraries/SDCard/SDCard.h
```diff
@@ -11,10 +11,10 @@ Descr: Low-level SDCard functions
...
 
 //extern SPISoft *SDCardSPI;
```
mikee47, line 16: Remove unused code

Sming/Libraries/SDCard/SDCard.cpp
```diff
@@ -33,27 +33,39 @@ Descr: Low-level SDCard functions
...
-SPIBase *SDCardSPI = NULL;
-uint8 SPI_CS;				/* SPI client selector */
+#include "fatfs/diskio.h" /* Declarations of disk I/O functions */
 
+FATFS* pFatFs = NULL; /* FatFs work area needed for each volume */
+SPIBase* SDCardSPI = NULL;
+uint8 SPI_CS;		  /* SPI client selector */
+uint8 SPI_byteOrder;  /* The order the SPI bytes are sent */
+uint32 SPI_freqLimit; /* The max frequency the SPI should run */
+uint32 SPI_init_freq; /* SPI frequency usined for init */
```
mikee47, line 43: Private/internal variables should either be `static` or, preferably, be defined within an un-named namespace.
The code might be clearer if these were within a structure:

```
namespace {

struct SpiConfig{
  uint8_t cs;
  uint8_t byteOrder;
  uint8_t freqLimit;
  uint8_t initFreq;
};

SpiConfig spi;

};
```


Variables names should be camel-case, especially `SPI_CS` which looks like a #defined constant.
Also, in C++ `nullptr` is preferred to `NULL`.

Sming/Libraries/SDCard/SDCard.cpp
```diff
@@ -33,27 +33,39 @@ Descr: Low-level SDCard functions
...
+		SPI_init_freq = freqLimit; /* Test if the limit frequency is higher that the init frequency */
+	else
+		SPI_init_freq = 4000000; /* It is useful to debug at a lower speed so the logic analyser can catch everything */
 
-	if(!SDCardSPI)
-	{
+	if(freqLimit <= 40000000)
+		SPI_freqLimit = freqLimit; /* Test to see if frequency is too high */
+	else
+		SPI_freqLimit = 40000000; /* Some SD cards struggle at 40MHz */
```
mikee47, line 66: Braces `{ }` should be used consistently with conditional statements:

```
if (...) {
  ...
} else {
  ...
}
```

Sming/Libraries/SDCard/SDCard.cpp
```diff
@@ -83,7 +93,7 @@ void SDCard_begin(uint8 PIN_CARD_SS)
 /*-------------------------------------------------------------------------*/
 /* Platform dependent macros and functions needed to be modified           */
 /*-------------------------------------------------------------------------*/
-static inline void dly_us (UINT n)	/* Delay n microseconds (avr-gcc -Os) */
+static inline void dly_us(UINT n) /* Delay n microseconds (avr-gcc -Os) */
```
mikee47, line 96: private functions can also be placed inside nameless namespace as alternative to `static`

Sming/Libraries/SDCard/SDCard.h
mopac: done

Sming/Libraries/SDCard/SDCard.h
mopac: done

slaff:
Note to myself: We should replace the outdated and no longer supported SDCard library with something better as for example https://github.com/greiman/SdFat

PR #2106, opened 2020-09-11
CS5460 Arduino library

kmihaylov:
This PR adds CS5460 library from this author: https://github.com/xxzl0130/CS5460
CS5460 IC is widely found in socket power meters. It can calculate RMS values and is used for electrical measurements voltage, current, power factor, energy, etc.
The provided sample is tested, however due to the dangerous voltage (my test board measures 230VAC) I couldn't provide quality video.

Please note that CS5460 requires calibration, use of (internal) filters, reset line use is more than recommended.

The component is build with the default MISO pin set to 12 (default on ESP-12F).


Sming/Libraries/CS5460/CS5460.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/CS5460.h b/CS5460.h
+index 0fbbb5b..fe7ff1d 100644
+--- a/CS5460.h
++++ b/CS5460.h
+@@ -1,7 +1,8 @@
+ #ifndef __CS5460_H__
+ #define __CS5460_H__
+ 
+-#include <arduino.h>
```
slaff, line 9: Can you try to replace `<arduino.h>` with `<Arduino.h>` and see if the library compiles and works?

slaff (commented):
The PR looks good to me. Please, check the comment that I have posted.

Sming/Libraries/CS5460/CS5460.patch
kmihaylov: Thanks for updating it!

mikee47:
@kmihaylov Just noticed there's no README for this library, would you add one please? This is all we get on the doc. page:
https://sming.readthedocs.io/en/latest/_inc/Sming/Libraries/CS5460/index.html

See https://sming.readthedocs.io/en/latest/contribute/documentation.html#readme-files for further details.

kmihaylov:
Sure. I'll need few days to react. Also the setting of the filters in the example is incorrect.

PR #2114, opened 2020-10-16
Feature: Simple Service Discovery Protocol (SSDP)

slaff:
Initial implementation for service discovery in Sming. Can be used to announce service(s) and search based on searchType (deviceType). Comes with an example that demonstrates a simple workflow.


slaff:
@mikee47 Can you help me improve the memory usage in the SSDP component? I will use the proposed changes from you not only to improve this component but also to create a small document describing best practises in improving memory usage.

mikee47:
@slaff You may want to take a look at https://github.com/mikee47/Sming-UPnP which 
incorporates an SSDP layer. I intentionally wrote it all with callbacks to keep memory usage down.

slaff:
> You may want to take a look at https://github.com/mikee47/Sming-UPnP

@mikee47 Let's merge you UpnP stuff then?

mikee47:
@slaff OK, There's four separate libraries in Sming-UPnP: `RapidXML`, `SSDP`, `UPnP` and `HueEmulator` so I'll split them out with those names. Let me know if you'd prefer different ones?

slaff:
The names of the components look good to me :)

slaff:
Closing in favour of the more advanced PR #2115.

PR #2119, opened 2020-10-19
Apply String move semantics to reduce memory re-allocations

mikee47:
* Apply coding style to WString
* Add `String::getBuffer` and `String::setBuffer` to support general move operations
* Optimise `readString` method for `MemoryDataStream`
* Add MemoryDataStream(String&&) constructor: heap allocation required only if String is using SSO mode.
* Provide MemoryDataStream::readString override to perform move when entire content of stream is requested.
* Add overloads to various classes to support r-values.
* Use overload instead of optional length parameter for String constructor and setString().

### DIFF
--- a/Sming/Core/Data/Stream/MemoryDataStream.cpp
+++ b/Sming/Core/Data/Stream/MemoryDataStream.cpp
@@ -10,6 +10,14 @@
 
 #include "MemoryDataStream.h"
 
+MemoryDataStream::MemoryDataStream(String&& string)
+{
+	auto buf = string.getBuffer();
+	buffer = buf.data;
+	size = buf.length;
+	capacity = buf.size;
+}
+
 bool MemoryDataStream::ensureCapacity(size_t minCapacity)
 {
 	if(capacity < minCapacity) {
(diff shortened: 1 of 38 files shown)


slaff:
@mikee47 Maybe you can port also this change from Arduino: https://github.com/esp8266/Arduino/commit/c24109fd57fc1a291805ed32629f7b073cc0d109 ?

Sming/Core/Data/Stream/MemoryDataStream.cpp
```diff
@@ -82,3 +90,23 @@ int MemoryDataStream::seekFrom(int offset, unsigned origin)
...
+	// Move content into the String object
+	ensureCapacity(size + 1); // Ensure size < capacity
+	String s;
+	if(s.setBuffer({buffer, capacity, size})) {
+		buffer = nullptr;
+		readPos = 0;
+		size = 0;
+		capacity = 0;
+	}
+	return s;
```
mikee47, line 111: I'm on the fence about this behaviour. It's probably correct in 99% of situations, but the safer option might be to add an overload such as `size_t readString(String& s);`

slaff: > but the safer option might be to add an overload such as size_t readString(String& s);

Got for it as long as the safer option does not impact compatibility, memory usage and speed.

slaff:
@mikee47 Ping me when this PR is ready.

mikee47:
@slaff OK, I'm done with this now!

PR #2116, opened 2020-10-17
Documentation fixes 

mikee47:
@kmihaylov Re #2106 I've included a basic README for the CS5460 library, if you'd like to add anything you could do it here!

PR #2120, opened 2020-10-20
Revert change to upgrade notes in #2116

mikee47:
Premature as PR hasn't been submitted yet! This will end up in 4.1 -> 4.2.


slaff:
>  This will end up in 4.1 -> 4.2.

Just add a new 4.1-4.2 migration file and move this change there.

PR #2124, opened 2020-10-23
Fix appveyor crashing issue and use latest MinGW

mikee47:
See also #2086 

I've been able to reproduce this problem on my dev. machine. The problem occurs during build of HostTests, specifically the `bearssl-esp8266` Component. It can be reproduced thus:

```
cd %SMING_HOME%\..\tests\HostTests
make bearssl-esp8266-clean
make bearssl-esp8266-build
```

(I add -j20 to speed things up, but it makes no difference.)

Now, if I remove *any* 3 files from the build (e.g. by renaming .c -> .cXX) and repeat the above commands, the problem does not occur. This puts the number of files passed to the `ar` tool at 292.

My first attempt was to invoke separate commands for each object file. The first 40 objects were added, then it failed!

So then I looked at path lengths. On my test build system, here's one of the commands:

`ar rcsP /s/sandboxes/sming-tmp/Sming/out/Host/debug/lib/clib-bearssl-esp8266.a bearssl/src/ec/ec_secp521r1.o`

So I did this:

```
subst Z: s:\sandboxes\sming-tmp
set SMING_HOME=Z:\Sming
```

And repeated the test. Success!

However, this only fixes the issue for appveyor unless users manually do the above.

So, part (2) of this fix is to change the makefiles to to generate a `build.ar` script. This keeps the command line small and actually fixes the problem, rather than works around it.

### DIFF
--- a/.appveyor/build.cmd
+++ b/.appveyor/build.cmd
@@ -1,11 +1,12 @@
 REM Windows build script
 
-set SMING_HOME=%APPVEYOR_BUILD_FOLDER%\Sming
+subst Z: %APPVEYOR_BUILD_FOLDER%
+set SMING_HOME=Z:\Sming
 
 if "%build_compiler%" == "udk" set ESP_HOME=%UDK_ROOT%
 if "%build_compiler%" == "eqt" set ESP_HOME=%EQT_ROOT%
 
-cd %SMING_HOME%
+cd /d %SMING_HOME%
 gcc -v
 
 set MAKE_PARALLEL=make -j2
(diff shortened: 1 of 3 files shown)


slaff:
@mikee47 Looks good to me :) Windows is full with hidden surprises :) Shall I merge this PR or you would like to add something more?

mikee47:
@slaff I'm happy as this is a fairly simple fix and doesn't appear to have broken anything else :-) Nothing to add.

PR #2126, opened 2020-10-27
Fix submodule fetching and add duplicate Component check

mikee47:
Fix submodule fetching to use recursion to reduce maintenance and development effort.

* Add `submodules` target

    `make submodules` from a project directory fetches only those submodules required for the project.
    (If run from SMING_HOME fetches every registered submodule.)
    
    Recursively fetch submodules, then run the actual build step in a sub-make.
    This ensures the source tree is complete before the sub-make begins.

    The .patches directory now no longer requires a copy of the Components component.mk.
    This was previously required to establish the full chain of sub-dependencies in one go.

    A patched component.mk is only required if a Component needs to change any settings and does not
    contain a copy itself.

* Add `fetch` target to main Makefile
    
    `make fetch` to list all Components which have submodules.

    `make fetch BMP180 UPnP` and so on to fetch all submodule for that component.

    Dependencies are not followed.

    This allows samples to be accessed which are contained therein.
    Dependencies will be handled when those are built.

* Add MAKE_CLEAN variable

    Components can use this to eliminate conflicts during clean operations.

* Add MAKE_DOCS variable

    Components can use this to eliminate conflicts during documentation build, since all component.mk
    files are pulled in regardless of SMING_ARCH setting.

* Add duplicate Component check

  Usually there should only be one Component with the same name in the list of searched Component directories.
  This could be intended, for example a user library taking precedence over another.
  It could also cause problems, for example if a Component was moved but the original remains.
  Warn about it but continue the build.


slaff:
@mikee47 Are you ready with this PR?

mikee47:
@slaff Yep.

PR #2131, opened 2020-10-29
BitSet enhancements

mikee47:
* Make BitSet constexpr-compatible

    Allows construction of fixed values which may be used as pre-defined masks, e.g. FileOpenFlags.

* Add `size` template parameter

    Primary use case is enumerated types which only use a subset of the Storage.
    For example, uint32_t can store 32 values but an enum may only have 15.
    Attempting to use ~ operator would produce an incorrect result without this information.

    Use this to provide `domain()` (set containing all values).

* Add static asserts to check template parameters are compile time

* Add methods to correspond with std::bitset

    Also tested and fixed for old compiler (linux)

* Add page to framework docs

* Add bitset module to HostTests for integration testing

* Add `toString` support

### DIFF
--- a/Sming/Core/Data/BitSet.h
+++ b/Sming/Core/Data/BitSet.h
@@ -4,7 +4,7 @@
  * http://github.com/SmingHub/Sming
  * All files of the Sming Core are provided under the LGPL v3 license.
  *
- * BitSet.h
+ * BitSet.h - Strongly typed sets of values
  *
  * @author: 2020 - Mikee47 <mike@sillyhouse.net>
  *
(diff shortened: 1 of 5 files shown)


Sming/Core/Data/BitSet.h
```diff
@@ -13,96 +13,201 @@
...
 
 	/**
-	 * @brief Remove value from set
+	 * @brief Remove elements from set
 	 */
-	BitSet& operator-=(BitSet rhs)
+	BitSet& operator-=(const BitSet& rhs)
 	{
-		value &= ~rhs.value;
+		value_ &= ~rhs.value_;
```
slaff, line 192: @mikee47 Please, remove leading or ending underscores.

slaff:
@mikee47 Ready?

mikee47:
@slaff OK, following further checks against my IFS branch this is working fine now. All done!

PR #2132, opened 2020-10-29
HttpClient improvements. Added support for servers that do not support connection reuse.

slaff:
- Add onClose callbacks to TcpConnection. Useful for retries and reacting on closing a connection. Will not be called if there was no connection at all.
- HttpClient now works with servers that don't support Tcp connection re-usage.


mikee47:
@slaff Sorry for not reviewing, I did look through and nothing stuck out but didn't get around to testing it. I'll definitely give the UPnP stuff a proper check though!

PR #2136, opened 2020-11-03
Initial DIAL (http://www.dial-multiscreen.org/) client.

slaff:
Allows Sming applications to control remote monitors/TVs. 
@piperpilot For example a Sming application can open YouTube on a smart TV and display a video when the BBQ is ready.


mikee47 (commented):
I've made a few suggestions via commits (take 'em or leave 'em). On namespaces should we use `DIAL` for consistency?

slaff:
> I've made a few suggestions via commits (take 'em or leave 'em). 

Thanks. All looks good to me.

> On namespaces should we use DIAL for consistency?

Better not. Better have the same coding style for namespaces as for classes. Which reminds me that `mDns` should be renamed to `Mdns` and it is probably a good idea to do it before the 4.2 release.

mikee47:
> 
> > On namespaces should we use DIAL for consistency?
> 
> Better not. Better have the same coding style for namespaces as for classes. Which reminds me that `mDns` should be renamed to `Mdns` and it is probably a good idea to do it before the 4.2 release.

I think the main reason for upper-camelcase names is consistency to avoid silly typos, although that should be negated using eclipse, etc. It feels more important that the name is *correct*, which for network protocols means `TCP`, `mDNS`, `UPnP`, `SSDP`, etc. There's a good reason you picked `mDNS` in the first place!

May I suggest we simply add namespace aliases, i.e. `namespace Mdns = mDns` and could even add `MDNS` and `mdns`? I can do the same with SSDP/UPnP.

mikee47:
@slaff I'm doing some revisions to UPnP library which should help to streamline this PR. Should have them out today.

Sming/Libraries/DIAL/sample/app/application.cpp
```diff
@@ -7,7 +7,9 @@
 #define WIFI_PWD "PleaseEnterPass"
 #endif
 
-static Dial::Client client;
+namespace
```
slaff, line 10: @mikee47 I will remove the anonymous namespace. If we apply this change it should be applied to all samples for consistency. Which can be part of a separate PR.

mikee47:
@slaff The Dial::Client is very interesting as it demonstrates behaviour that UPnP::ControlPoint should handle. Specifically, tracking unique service names and fetching descriptions. I could pull that out now, or leave it for another PR?

slaff:
> I could pull that out now, 

@mikee47 Do you mean to create a separate repository for the Dial client? If yes, I will make this. I wanted to change also the naming of one flash variable... let me do it first...

slaff:
> Specifically, tracking unique service names and fetching descriptions. I could pull that out now, or leave it for another PR?

Or you have meant to put the tracking and fetching of the description from Dial::Client into UPnP::ControlPoint ? If yes - let's finish with this PR and you can make your changes in a separate PR for it. Is that ok for you?

mikee47:
@slaff I've kind of done it already :-) I'll push the changes and see what you think - also found an important bug in TCP.

mikee47:
@slaff Thank you for your work in getting this control point stuff started, looking really good! I've opened an [issue](https://github.com/mikee47/Sming-UPnP/issues/1) to discuss and track changes necessary before Sming 4.2 release.

slaff:
@mikee47 Shall I wait for you or better merge it as it is in `develop` and the move it into a separate repository?

mikee47:
@slaff I think go ahead and move it, feels like a good place to shove a stick in the mud.

PR #2138, opened 2020-11-04
Several null/range checks and minor fixes

mikee47:
* ESP8266 `StationImpl::getSSID()`: Return invalid String for station SSID if not set

* m_printHex: If no data, just print the tag. Existing behaviour is to show nothing at all.

* Add nullptr handling in `String::equalsIgnoreCase`

* Fix copyFrom - may end early as readMemoryBlock() may return 0 but doesn't mean its finished

* BitSet fixes

    toString should return empty string for empty set, not invalid one
    Fix BitSet operators using is_convertible so regular enums don't get caught

* StreamTransformer nullptr check


slaff:
@mikee47 Tests are failing. See the Host build for details.

PR #2139, opened 2020-11-04
Update Streams and File functions to use `SeekOrigin` instead of `SeekOriginFlags`

mikee47:
Name is more appropriate (i.e. not a flag!)
SeekOrigin is strongly typed.

This should help to avoid silly errors using the wrong value, and also be more consistent as there are multiple 'C' defintions for the same thing.


slaff:
Looks good to me. Can you also also a note in the migration 4.1 -> 4.2 guide concerning this change?

slaff:
@mikee47  Ready with this one?

slaff:
@mikee47  can you rebase it please?

PR #2115, opened 2020-10-16
UPnP framework

mikee47:
The framework contains three libraries:

* RapidXML: Parsing, serialising and de-serialising XML data
* SSDP: Simple Service Discovery Protocol server
* UPnP: Framework to build UPnP compliant device stacks. Written from scratch with reference to the documented standards. The `Basic_UPnP` sample gives an idea of what can be done. See the library README for further details,

The main purpose for this library was to provide an Alexa-compatible interface. It was important to provide a smooth end-user experience so that device discovery is reliable and painless. As such, device discovery and presentation is essentially complete, tested and working.

Also included is the `Basic_Alexa` sample with `HueEmulator` library.

Devices will appear in Windows explorer under 'Network'. You can find more comprehensive test tools at https://www.meshcommander.com/upnptools (Windows only - _Device Spy_ is the most interesting! See #2114 README for notes on testing in Linux.

Both samples will run in Host emulator mode.

Additional changes:

* Add `isLocal()` methods to AccessPoint & Station classes
* Add additional toString() functions/methods
* Make SMING_ARCH available to code and use it to generate Hue device names
* HTTP Headers
    Add `Accept` to standard header fields
    Move HTTP header field name handling into separate unit, and add BasicHttpHeaders
* Fix HTTP content type field parsing
    
    May contain spaces which must be removed, for example:
    
            CONTENT-TYPE: text/xml ; charset="utf-8"\r\n

    Also map `application/xml` (as well as `text/xml`) to `MIME_XML`

* Fix LimitedMemoryStream, and add `getStreamPointer` method
    
    Attempting to write data larger than available space discards all of it.
    Correct behaviour is to write as much as possible, but still discard the remainder.
    
    Also, seek operation (for reading) is now bi-directional between the start of the buffer and the current write position.

TODO:

* [x] Review #2114 and merge appropriate content, e.g. Basic_SSDP sample
* [x] Handling of service requests and generating responses is not fully complete. Review and re-test, fix as required.
* [x] Document any known issues which require further development. Update `TODO.rst`.
* [x] Document BitSet class
* [x] Add home page for Basic_UPnp sample showing device stack with links to their individual presentation pages.
* [x] Document how to implement service requests (see Wemo device)

Issue #2015, opened 2020-01-05
String conversion functions / methods


mikee47:
There are numerous places within the framework where we have functions and methods to convert typed values into strings. A few examples:

* `NanoTime::Frequency::toString()` and other classes
* `IpAddress::toString()`
* `MacAddress::toString()`
* `Ssl::getCipherSuitName(CipherSuite id)`

It appears this is not the most appropriate way to do this:

http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class

They should, in fact, be declared as regular functions. This one's got it right (not mine!):

* `ContentType::toString(MimeType mime)`

I can see the sense of this. It would make for a much more uniform API just typing `toString(` then getting a selection of type conversions in the IDE. It also avoids making up names for things.

There are proabably other conversions which could benefit from this appropriate, but perhaps if there is agreement we can add `toString()` functions. In many cases to avoid breaking things they'd be in addition to existing class methods, but if they've been added since 4.0.1 release we can just replace them.

What do you think? Worth doing?

mikee47:
Forgot to mention, cruicially this would work uniformly with any type of data, including enums which cannot of course have methods.

aemseemann:
I like the idea, in particular because the standard library is also moving into that direction with `std::size(...)` and friends. It also resembles the common pattern of having overloads of `operator<<(std::ostream& s, ...)` for all types that may need a textual representation.

Before this is rolled out to the entire code base, an aspect to consider is which return type to allow for `toString`. Always a Wiring String or would it be fine to return a `const char *` or `FlashString`, too? Shouldn't matter in most cases thanks to implicit conversion constructors, just method chaining constructs like `toString(...).otherMethod()` would not work universally.

mikee47:
IMHO best to standardise on WString for consistency, and as you mention method chaining (although I'm not a real fan of that idiom). In actual use it'll most likely end up in a WString expression anyway. Whilst `const char*` is convenient for use in printf expressions I'm trying to move away from using it (except in debug_x calls) in favour of Print methods as it's more robust.

It should also be clear that, IMHO, `toString` should produce a readable string. For example:

```
using BinaryData = std::array<uint8_t, 16>;

String toString(const BinaryData& data)
{
  return makeHexString(data.content()
}
```

Whilst we could also produce a `String` object from the binary data, that's should be done by something else:

```
[...]

PR #2140, opened 2020-11-04
Add ContentType overload functions and strong enums

mikee47:
* Add `ContentType` overload functions to return MimeType instead of String
* Add strong enums with appropriate `toString` overloads for `MimeType`, `HttpMethod`, `HttpStatus`, `HttpError`, `HttpHeaderFieldName`.
* Fix `HttpRequest` and `HttpResponse` toString method declarations, add `toString()` functions and for `Url`.
* Move some related code into headers, optimise String operations

### DIFF
--- a/Sming/Core/Network/Http/BasicHttpHeaders.h
+++ b/Sming/Core/Network/Http/BasicHttpHeaders.h
@@ -141,14 +141,9 @@ public:
 	/**
 	 * @brief Obtain response status
 	 */
-	http_status status() const
+	HttpStatus status() const
 	{
-		return http_status(parser.status_code);
-	}
-
-	String statusText() const
-	{
-		return httpGetStatusText(status());
+		return HttpStatus(parser.status_code);
 	}
 
 	/**
--- a/Sming/Core/Network/Http/BasicHttpHeaders.h
+++ b/Sming/Core/Network/Http/BasicHttpHeaders.h
@@ -128,22 +128,14 @@ public:
 	/**
 	 * @brief Obtain request method
 	 */
-	http_method method() const
+	HttpMethod method() const
 	{
-		return http_method(parser.method);
+		return HttpMethod(parser.method);
 	}
 
-	void setMethod(http_method method)
+	void setMethod(HttpMethod method)
 	{
 		parser.method = unsigned(method);
-	}
-
-	/**
-	 * @brief Obtain text for request method
-	 */
-	const char* methodStr() const
-	{
-		return http_method_str(method());
 	}
 
 	/**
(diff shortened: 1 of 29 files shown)


mikee47:
@slaff In this PR I've removed functions from `HttpCommon.h` (only added in #2115) because `http_errno`, `http_status` and `http_method` are all weak types. Re. #2015 these should be redefined using `enum class` but perhaps in a separate PR.

Sming/Core/Network/Http/HttpRequest.h
```diff
@@ -119,6 +119,11 @@ class HttpRequest
 		return this;
 	}
 
+	String methodStr()
```
slaff, line 122: Please, change the name of the method to `String getMethod()`. And add a doc block.

mikee47: I don't think `getMethod()`  is appropriate as it implies we're getting the enumerated type. I considered `getMethodString()` and `methodString()` but we already have `BasicHttpHeaders::methodStr()`. Maybe a better way to deal with this is to define `HttpMethod` as a strong enum (shadowing `http_method`) and add a regular `toString` function for it?

slaff: > ... Maybe a better way to deal with this is to define HttpMethod as a strong enum (shadowing http_method) and add a regular toString function for it?

Yes, let's do it this way. Will C++ at some point complain that there is a function `String toString(X x)` that is so heavily overloaded?

mikee47: Nope. Each and every instance of `toString` has a different (mangled) name **provided** the types are strong.

mikee47:
Could do a similar thing for http_status, however it's going to break a lot of applications (more so than HttpMethod)

Sming/Core/Network/Http/BasicHttpHeaders.h
```diff
@@ -138,14 +138,6 @@ class BasicHttpHeaders : public HttpHeaderFields
 		parser.method = unsigned(method);
 	}
 
-	/**
-	 * @brief Obtain text for request method
-	 */
-	const char* methodStr() const
```
slaff, line 144: > Could do a similar thing for http_status, however it's going to break a lot of applications (more so than HttpMethod)

Nooo... it's too much ... The commit `Re-define HttpMethod as strong enum` is a good idea but I would like to delay this for a later release as it will break already all Http related apps. Can you do a minimal thing: rename `methodStr()` to `methodAsString()`. Or `getMethodText()` ?

mikee47: I have a solution in mind which won't break things... bear with.

mikee47: @slaff OK, have added HttpMethod, HttpStatus and HttpError seems OK... Whadya think?

slaff: >  Whadya think?

Thanks, looks much better now. I will look at it closer later today.

slaff:
@mikee47 Looks good to me. Is this PR ready for merging?

mikee47:
@slaff Yep, all ready.

PR #2141, opened 2020-11-05
Updated RapidXML to latest version.

slaff:


mikee47:
@slaff Why not just wrap this up in your DIAL PR?

slaff:
> Why not just wrap this up in your DIAL PR?

Ok, I will do it.  The change will be part of #2136.

mikee47:
Just thinking whilst this is still all on the develop branch it keeps things a bit simpler.

PR #2148, opened 2020-11-07
Core `Task` class to support co-operative multitasking

mikee47:
Sming provides a task queue for applications to use. It can be used to defer processing from an interrupt service routine, but can also be used to provide a form of (co-operative) multi-tasking.

This PR introduces a `Task` class to provide a simple way to create a task which runs continuously, or can be started and stopped as required.


slaff:
@mikee47 Can you rebase this PR?

PR #2150, opened 2020-11-09
Hardware SPI library

mikee47:
Class-based asynchronous SPI driver stack for Sming.
Currently ESP8266-only, but will be extended to support ESP32.
Also contains demo for S1D13781 TFT controller.


samples/ScreenTFT_S1D13781/app/app.cpp
```diff
@@ -0,0 +1,287 @@
+#include <SmingCore.h>
```
slaff, line 1: - Please rename this file to app/application.cpp to be consistent with the other samples that we have.
- Remove the anonymous namespace. Adding an anonymous namespace to all samples should happen in a later release.

mikee47: OK, I've moved the sample into the library and stripped out all the RAM stuff so it's much simpler.

slaff:
@mikee47 Ready with this PR?

mikee47:
@slaff yep, thanks.

PR #2156, opened 2020-11-25
Add pong delegate support for websoket connection and resource

avr39-ripe:
Add ability to call delegate on pong frame reception. This is useful for
closing stalled connections. Websocket server can send ping frames periodically and check for pong back from clients. On pong reception server can reset some counters attached to websocket connection via setUserData. Also server periodically checks for counter not reaching some minimum value for every connection and close connections that do not reply to ping request assuming them as stalled. Connection may stall if some hard network problem occurs and there is no proper way to close client connection nor by websocket protocol or tcp.
Remove websocket state change before calling close() in
WebsocketConnection destructor. Without this change if connection was
removed by WebsocketRecource, onDisconnect delegate for Websocket
connection will not run.

### DIFF
--- a/Sming/Core/Network/Http/Websocket/WebsocketConnection.h
+++ b/Sming/Core/Network/Http/Websocket/WebsocketConnection.h
@@ -73,7 +73,6 @@ public:
 
 	virtual ~WebsocketConnection()
 	{
-		state = eWSCS_Closed;
 		close();
 	}
 
@@ -214,7 +213,14 @@ public:
 	{
 		wsBinary = handler;
 	}
-
+	/**
+	 * @brief Sets the callback handler to be called when pong reply received
+	 * @param handler
+	 */
+	void setPongHandler(WebsocketDelegate handler)
+	{
+		wsPong = handler;
+	}
 	/**
 	 * @brief Sets the callback handler to be called before closing a websocket connection
 	 * @param handler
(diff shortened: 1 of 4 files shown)


Sming/Core/Network/Http/Websocket/WebsocketConnection.h
```diff
@@ -73,7 +73,6 @@ class WebsocketConnection
 
 	virtual ~WebsocketConnection()
 	{
-		state = eWSCS_Closed;
```
slaff, line 76: What is the reason to remove the change of state here?

avr39-ripe: if close method got connection in eWSCS_Closed state it do not fire onDisconnect delegate and this is not good. close() method change state by itself, and as far as I understand changing state to eWSCS_Closed in close() method do not hurt anything.

avr39-ripe: https://github.com/SmingHub/Sming/blob/08c9eb18ac672f6e974d77885b82b56d972fe6f3/Sming/Core/Network/Http/Websocket/WebsocketConnection.cpp#L259 Here is it, on line 269 there is delegate call, but it inside if(state != eWSCS_Closed) ...

PR #2149, opened 2020-11-07
[WIP] Connecting to Amazon Web Services

mikee47:
This PR is intended to provide a guide to building and connecting devices using AWS IOT.

The requirement is to allow Sming applications to publish their services via AWS IOT cloud so they can be found by Alexa devices.

For basic devices the HueEmulator provides a much simpler solution as devices such as the Echo Dot can connect to these locally without any additional setup.

This project is incomplete but perhaps others have got a bit further and can assist?

PR #2157, opened 2020-11-28
MqttClient revisions

mikee47:
Move callbacks into PROGMEM
Fix issues identified by Codacy
Fix pingRepeatTime setting - must always be <= keepAlive
Handle pings using PolledTimer


mikee47:
@slaff I've pulled out the MqttClient revisions from #2149 and made some further improvements. I've checked using the `MqttClient_Hello` sample, setting the message interval to 50 seconds and pings, etc. all look fine. However, I'm unable to connect to the `attachix.com` addresses in that sample so I used `test.mosquitto.org` - should the sample be updated?

slaff:
>  so I used test.mosquitto.org - should the sample be updated?

@mikee47 Yes, please.

PR #1579, opened 2019-01-25
Installable File System

mikee47:
This is a work in progress. See `Basic_IFS` sample.

TODO:

* [x] Test/fix ESP32 build (WIP)
* [x] Add `virtual IFileSystem::mkdir()` and `IFileSystem::makedirs()`
* [x] Review framework and use host filesystem where appropriate
~~* [ ] Implement `fscopy` utility~~
* [x] Add partition support
* [x] Review and tidy documentation
~~* [ ] Tidy up test application~~

PR #2164, opened 2020-12-02
Add dataset templating support

mikee47:
This PR extends templated streaming to support formatting of large datasets.
It is used by #1579 to provide a mechanism for streaming directory listings in various formats.

Upgrade TemplateStream

* Malformed template data can cause lockup waiting for next input when there isn't any.
* Add output enable control to support filtering
* Allow spaces in tag names, increase max length to 128
* Add `evaluate` method to enable support for extended tag parsing
* Add optional `owned` parameter to TemplateStream constructor for situations where caller is responsible for allocation
* Allow TemplateStream to be rewound: implement `seekFrom`
* Add option to use double braces {{ }} - necessary for JSON

Add class-based support for escaping/formatting strings (html, json, etc)

* Base Formatter class, headers and standard/html/json within Format namespace

Add SectionStream class

* Presents each marked section within a source stream as a separate stream
* Sections are identified by index and can be selected to allow repeating of sections, etc. as required

Add SectionTemplate class to support dataset reporting

* Adds basic expression parsing with conditionals, built-in functions, etc.
* Uses SectionStream to support typical reports with a header/content/footer. The content section is repeated for each record.

Add separate TemplateStream test module, with more tests


Sming/Core/Data/Stream/SectionTemplate.h
```diff
@@ -0,0 +1,179 @@
...
+	XX(_repeat, "{!repeat:A:B} Repeat A, number of iterations is B")                                                   \
+	XX(_kb, "{!kb:A} Convert A to KB")                                                                                 \
+	XX(_ifdef, "{!ifdef:A}block{/if} emit block if A is not zero-length")                                              \
+	XX(_ifndef, "{!ifdef:A}block{/if} emit block if A is zero-length")                                                 \
+	XX(_ifeq, "{!ifeq:A:B} emit block if A == B")                                                                      \
+	XX(_ifneq, "{!ifneq:A:B} emit block if A != B")                                                                    \
+	XX(_ifgt, "{!ifgt:A:B} emit block if A > B")                                                                       \
+	XX(_iflt, "{!iflt:A:B} emit block if A < B")                                                                       \
+	XX(_ifge, "{!ifge:A:B} emit block if A >= B")                                                                      \
+	XX(_ifle, "{!ifle:A:B} emit block if A <= B")                                                                      \
```
slaff, line 52: Wow, this "simple" templating mechanism starts to get more and more powerful and will one day beat PHP :)

Sming/Core/Data/Stream/SectionTemplate.h
```diff
@@ -0,0 +1,179 @@
...
+	XX(record, "{$record} Current record index")
+
+/**
+ * @brief Provides enhanced template tag processing for use with a SectionStream
+*/
+class SectionTemplate : public TemplateStream
+{
+public:
+	enum class Command {
+		_unknown = 0,
```
slaff, line 75: Remove leading and trailing underscores where possible.

mikee47:
```
 * @note Command tags are prefixed with _ to allow use of reserved words
 * in the Command enumeration. This represents the ! prefix in actual use.
```

mikee47: I could change this to use some other character but as they're all wrapped up in an `enum class` it should be pretty safe.

slaff: > I could change this to use some other character

Yes, please.

Sming/Core/Data/Stream/SectionTemplate.h
mikee47: This kind of evolved but is there a standard templating language that we should be using?

slaff: > This kind of evolved but is there a standard templating language that we should be using?

Hm... this is a very good question. It is better, adoption-wise, to use something that is already easy and popular.  I can suggest [Twig](https://twig.symfony.com/) used in the PHP world,  [Jinja for C++](https://github.com/hughperkins/Jinja2CppLight) or [Nunjucks](https://mozilla.github.io/nunjucks/). There are plenty of templating engines out there and I am open for suggestions.

mikee47: On reflection I'm inclined to leave this 'engine' as-is.

The main reason I wrote this initially was to find a generic way to produce a directory listing in JSON format. I initially used a custom stream class, but then I realised templating could offer a more flexible solution. The browser can then do all the formatting, etc. to produce reports. However, I got carried away of course and the conditional stuff got added!

So the use cases are rather different I think. It looks like Jinja or Nunjucks could be added if the need arises for those more advanced features. They seem fairly consistent in using the {% ... %} escaping, however the command syntax here is much simpler so I think keeping {! ...} helps to differentiate it.

slaff: > On reflection I'm inclined to leave this 'engine' as-is.

Agreed, we can always improve it if there is need for it.

slaff (approved):

Issue #1676, opened 2019-04-27
ESP-IDF Partition Table support


mikee47:
Further to #1333 porting Sming to Esp32 (and other platforms) the next roadblock we're likely to run into is the use of partition tables. This is an ESP-IDF feature, provided for the *Esp32* and *Esp8266 RTOS* platforms. See [Here](https://docs.espressif.com/projects/esp-idf/en/latest/api-guides/partition-tables.html) for an overview.

Instead of accessing flash memory directly (e.g. using `flashmem_read`), a module/component (such as SPIFFS) uses the Partition API (e.g. `esp_partition_read`). It prevents inadvertent access to other areas of flash (configuration, application, etc.). It also offers proper modular support, so an application could have multiple SPIFFS partitions, etc. if required. Custom flash partitions can be defined for other uses.

As such partitions offer significant benefits for ease of use, flexibility, modularisation and customisation for applications so should form a core part of the Sming framework.

Adopting this mechanism is a pre-requisite for Esp32 support, and opens the door for easy adoption of other features like [nvs_flash](https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/storage/nvs_flash.html).

I've opened this Issue so we can explore exactly how this will work for Sming, and how it can support user projects.

## Proposal

At present there is no unified way to describe the flash partition layout for Sming, instead relying on various pre-compiler definitions to set addresses.

What I propose is implementing/updating the following _components_ for the Esp8266 platform, using the *esp_partition* API instead of #defined values. Reference code is available in the [Esp32 IDF](https://github.com/espressif/esp-idf/tree/master/components) and [Esp8266 RTOS](https://github.com/espressif/ESP8266_RTOS_SDK/tree/master/components).

* *partition_table* Application (`parttool.py`) for creating the binary partition table image from a .csv file
* *spi_flash* Partition API plus IDF-compatible flash memory functions. This could be a little tricky because the IDF functions have the same names as the non-OS functions, but behave differently. For example, the IDF `spi_flash_read()` is equivalent to `flashmem_read()` as it takes care of alignment issues.
* *esptool_py* use instead of existing *esptool* (for all build environments)
* *spiffs* (update) use `spiffsgen.py` instead of `spiffy` to provide more flexible building options. Flash memory access is via *esp_partition_*' calls instead of direct flash access (partitions always at offset 0).
* *rboot* (update) will simplify: `rboot-overrides.c` removed, separated completely from *spiffs*

[...]

PR #2130, opened 2020-10-28
Revisions to SPI classes

mikee47:
Refactor, tidy and implement SPIClass for ESP32.

External interfaces haven't changed but internal ones have.
Tested with `ScreenTFT_ILI9340-ILI9341`, `SDCard` and `Radio_nRF24L01` samples.

**bugfix**: correct CPU multipler, should be 1M (was 10M)
   
**Naming / comments / style**
  
* Remove extraneous debug statement in setClock()
* Get rid of redundant/duplicate comments
* Don't repeat method name in brief
* Use cstdint types

**SPI code revisions to simplify and improve performance**

* SPISettings passed by reference. SPIClass no longer keeps a copy of these, no point.
* Clock register value is pre-calculated and stored in settings. Code simplified for readability.
* bugfix: exception (division by zero) raised in some cases, e.g. freq = 27M.
* BeginTransaction() and prepare() just set up registers anyway as there is now minimal overhead in doing this.

    
**Implement SPIClass for Esp32**

Has three available SPI modules:

* VSPI works reliably
* HSPI works with SPI display but not with SDCard (even at low clocks) but it does work with remapped pins (e.g. `SPIClass(SpiBus::HSPI, SpiPins{18, 19, 23})`). Suspect involvement with RTC/IO but as yet unable to pin this down.
* FSPI crashes: needs more work (it's shared with Flash).

PR #2171, opened 2020-12-03
Partition tables

mikee47:
Implement class-based API for Sming with support for all architectures.
See #1676 for discussion.

Summary of changes:

* Create new `Storage` component and port low-level Esp32 partition support
* Use `gen_esp32part.py` tool as basis for new `hwconfig.py` tool. Integrate into build system. Add new targets.
* Implement a C++ API. `Partition::SubType::App` and `Partition::SubType::Data` are strong enums used to identify all the standard types, but the base API still uses type/subtype so custom types can be accommodated.
* Revise esp8266 sector layout - see updated documentation. This places all the critical information at the start of flash, whereas previously it was at the end. This avoids issues where flash size is set incorrectly.
* Custom `sysParam` data partition sub-type used for esp8266 system parameters and RF calibration data.
* Add support for 'external' devices (e.g. SPIRAM, supplementary flash, EEPROM, etc.) using a custom `Storage` partition type. Applications register a `Storage::Device` object implementation with the Partition API to register these external devices with a C++ interface for them.
* Disentangle SPIFFS from rBoot - OTA upgrading uses partition API.
* Drive SPIFFS image creation from partition table entries - more flexible, multiple images.
* Generate build variables directly from hardware config file (see `out/.../hwconfig.mk`). Preserve existing build variables where possible, document changes.
* Document hardware configuration with schema and incorporate validation. This can assist with fixing malformed configurations and supplements checks included in the partition tool.
* Add `Basic_Storage` sample to more thoroughly demonstrate usage
* Update documentation
* [x] Test (HostTests) on all arches
* [x] Remove `DISABLE_SPIFFS`, `SPI_SIZE` and `SPIFF_SIZE` build variables from all samples, including those in submodules.
* [x] Include samples and tests within submodules to CI
* [x] Add check to ensure image size will fit within partition

For other PR:

* Revise `rBoot` and `OtaUpgrade` to make more effective use of partitions.
* Update FlashString to support external data storage by Partition?
* Add NVS Component based on or using ESP-IDF for storing configuration data.
* Switch to spiffsgen.py (IFS)


Sming/Components/Partition/src/Arch/Esp8266/base.csv
```diff
@@ -0,0 +1,3 @@
+# Name,   Type, SubType, Offset, Size,    Flags
```
slaff, line 1: Maybe it will be better to have this ARCH specific files under Sming/Arch/Esp8266/... . This way every architecture will have all arch specific configurations in one place.

Sming/Components/Partition/src/Arch/Host/base.csv
```diff
@@ -0,0 +1,3 @@
+# Name,   Type, SubType, Offset, Size,    Flags
```
slaff, line 1: This configuration file should go under Sming/Arch/Host as well as the C/C++ files.

Sming/Components/Partition/src/include/Partition.h
```diff
@@ -0,0 +1,227 @@
...
+	XX(Ota15, ESP_PARTITION_SUBTYPE_APP_OTA_15, "OTA #15")                                                             \
+	XX(Test, ESP_PARTITION_SUBTYPE_APP_TEST, "Test application")
+
+#define PARTITION_DATA_TYPE_MAP(XX)                                                                                    \
+	XX(Ota, ESP_PARTITION_SUBTYPE_DATA_OTA, "OTA selection")                                                           \
+	XX(Phy, ESP_PARTITION_SUBTYPE_DATA_PHY, "PHY init data")                                                           \
+	XX(Nvs, ESP_PARTITION_SUBTYPE_DATA_NVS, "NVS")                                                                     \
+	XX(CoreDump, ESP_PARTITION_SUBTYPE_DATA_COREDUMP, "Core Dump data")                                                \
+	XX(NvsKeys, ESP_PARTITION_SUBTYPE_DATA_NVS_KEYS, "NVS key information")                                            \
+	XX(EfuseEm, ESP_PARTITION_SUBTYPE_DATA_EFUSE_EM, "eFuse emualtion")                                                \
```
slaff, line 64: "eFuse emualtion" - > "eFuse emulation"

Sming/Components/Partition/src/include/Partition.h
```diff
@@ -0,0 +1,227 @@
...
+	XX(EspHttpd, ESP_PARTITION_SUBTYPE_DATA_ESPHTTPD, "ESPHTTPD")                                                      \
```
slaff, line 65: > XX(EspHttpd, ESP_PARTITION_SUBTYPE_DATA_ESPHTTPD, "ESPHTTPD")

Do we need this?

samples/Basic_Serial/app/application.cpp
```diff
@@ -239,6 +240,29 @@ void handleCommand(const String& command)
 	}
 }
 
+void listPartitions()
```
slaff, line 243: Partitioning should have its own sample but I guess you have already planned this and the code in Basic_Serial is only for initial tests.

mikee47:
I dislike the .csv format. It's easy to make mistakes and inflexible. Entries have to be in the correct order.

Instead of this:

```
# Name,   Type, SubType, Offset,   Size,    Flags
phy_init, data, phy,     0x3000,   0x01000,
nvs,      data, nvs,     0x4000,   0x04000,
app,      app,  factory, 0x008000, 0xf8000,
spiffs,   data, spiffs,  0x100000, 65536,
```

I'd prefer this:


```
{
	"Esp8266": [
		{
			"address": "0x3000",
			"size": "0x1000",
			"label": "phy_init",
			"type": "data",
			"subtype": "phy"
		},
		{
			"address": "0x4000",
			"size": "0x4000",
			"label": "nvs",
			"type": "data",
			"subtype": "nvs"
		},
		{
			"address": "0x8000",
			"size": "0xf8000",
			"label": "app_0",
			"type": "app",
			"subtype": "factory"
		},
		{
			"address": "0x100000",
			"size": "0x10000",
			"label": "spiffs",
			"type": "data",
			"subtype": "spiffs"
		}
	],
    "Esp32": [
        ...
    ],
    "Host": [
        ...
    ]
}
```

It's clearer, multi-arch and we can add other fields as required (e.g. comments).

One file contains all appropriate settings. The build system would generate this for the user application,
which can then be edited as required.

The address is the most important field, so that goes first.
Parsing this to produce the correct .csv file is trivial, and we can sort the entries if needed.

Generating this from our fixed configuration variables is also much simpler.

The partition tool(s) could probably do with a rewrite...

Sming/Components/Partition/src/Arch/Esp8266/base.csv
mikee47: I think this whole approach needs clarifying and simplifying. I have some thoughts which I'll set out below.

Sming/Components/Partition/src/include/Partition.h
mikee47: No idea, this is simply laying out the supported esp32 partition types. If someone wants to use it then it should be there, but at the very least it's a 'reserved' type.

samples/Basic_Serial/app/application.cpp
mikee47: Yes, just being lazy. It was the first sample to hand which used SPIFFS so just used it for basic testing. I'll probably add something to hostTests for this once.

mikee47:
Instead of just a partition table configuration file, let's make it part of a 'hardware configuration' file. Post updated.

mikee47:
For PR #2171 I propose that the information provided by some build variables (see below) is placed instead in a hardware configuration file attached to the project (and under revision control).
Each file is specific to a hardware variant.
May I propose `.hw` as the file extension ? For example, `widget.hw`. We could have `widget32.hw` for an esp32 version, etc.

This would be supported by a python tool, based on or in addition to `gen_esp32part.py`. Suggest `hwconfig.py` ?

## Pre-configuration


Instead of defining build variables directly in a project, we add a 'pre-configuration' step which reads `widget.hw` and generates `hwconfig.mk` containing the appropriate information. It also generates the binary partition table.

Example:

    make configure widget.hw

The build variables are still present, but managed in a more flexible way.

The generator also allows the makefile logic to be simplified, e.g. by performing numeric calculations,
hex/decimal conversions, etc. anything which is a pain to do in a makefile.

The .hw file allows different sets of these variables to be generated depending on the hardware being targeted.

Following pre-configuration, project building proceeds as usual. The tool only needs to be run again if any of the information in the .hw file changes: this could be handled automatically by the build system after initial configuration.


## Existing projects

Use `hwconfig.py` to generate `widget.hw` file from existing project settings.
Those settings should then be removed from the project's component.mk file.


## Other information widget.hw might contain:

-   Build information to create/update the image file for the partition content, e.g. command line, makefile target, etc.
-   Description of the target hardware
-   Build options relating to that hardware
-   Anything which needs to change if switching between different hardware variants

## Standard configs

We can provide some standard configs available via the build system. Example:

    make create-config SMING_ARCH=Esp8266 TEMPLATE=spiffs NAME=widget.hw

We'd then edit the new `widget.hw` file and then run `make configure widget.hw`


## Affected build variables


- PARTITIONS_CSV

    Taken from esp32. Not required.

    I considered adding `HARDWARE_CONFIG` to indicate the .hw file to use, however by adding a pre-configuration step this is not necessary.


- PARTITION_TABLE_OFFSET

    Store this in the .hw file.

    This is adjustable to allow more room for the boot sector / config.
    Everything following the partition table is defined the partition table entries.

    For example, with esp8266:

        0x0000 rBoot boot sector
        0x1000 rBoot config sector
        0x2000 PARTITION_TABLE
        ... entries


- DISABLE_SPIFFS

    It *does not* mean a SPIFFS partition is not available, only that it won't be initialised with data.

    Very confusingly named as it only controls SPIFFS image generation.
    Not required as partition table provides this information.


- RBOOT_SPIFFS_0, RBOOT_SPIFFS_1, SPIFF_SIZE

    Specified in the partition table.

- SPIFF_FILES, SPIFF_BIN

    Used by SPIFFS image builder. The partition table can contain the appropriate makefile build command to generate the image, and it will use this. Example:

        "spiffs1": {
            "address": "0x200000",
            "size": "0x10000",
            "type": "data",
            "subtype": "spiffs",
            "filename": "$(FW_BASE)/spiff_rom.bin",
            "content": "files",
            "make": "spiffsgen"
        },

    The build system can append `PARTITION=spiffs1` automatically to the make command so the partition information can be queried (using parttool.py). For example, our `spiffsgen` target queries the partition table for `content` to identify
    what goes in the image, and `filename` so it knows where to write it.


- RBOOT_ROM_0
- RBOOT_ROM0_ADDR
- RBOOT_ROM_1
- RBOOT_ROM1_ADDR
- RBOOT_ROM2_ADDR (deprecate, move to .hw file)

    Goes in the related partition table entry. Pre-configuration will contain appropriate build targets. Example:

        "app_main": {
            "address": "0x8000",
            "size": "0xf8000",
            "type": "app",
            "subtype": "factory",
            "filename": "$(FW_BASE)/rom0.bin",
            "make": "fwimage"
        },


- SPI_SIZE

    esptool can (usually) auto-detect flash size so this shouldn't be required.
    As the critical sectors are now at the start of flash we don't need it for that either.
    It should go in the .hw file anyway.

mikee47:
For info. I'm currently porting gitext://gotocommit/85cb1e8ca37302d282b041fc5768a2d248e08a32/components/nvs_flash for Sming. It's mainly C++, which is nice :-)

mikee47:
@slaff The CodeQL is currently failing, probably because it's trying to use python2...

mikee47:
@slaff I'm generally happy now with the general API and structure for partition management. However, there are some things which will break existing code:

* Manual SPIFFS functions have been removed as they are incompatible with partitions.
* rBoot (needs further work) must be driven from the partition table with RBOOT_ROM0_ADDR, etc. removed.

I've thought about possible ways to just deprecate some of this but I don't think it's possible. However, updating applications should be reasonably straightforward and partitions should make life easier with no need to mess around with raw flash.

Any thoughts?

slaff:
>  However, there are some things which will break existing code:

Ok, if we will be breaking things then let me suggest something:
- Let's finish PR #2130 and merge it to 4.2.0. It would be great if there is SPI support for ESP32.
- Once this is done we will release 4.2.0 and tag it also as LTS.
- Let's move this PR and #1579 to release 4.3.0.

This way more conservative users can stay on 4.2.0 and we can introduce breaking changes in  4.3.x.

Does it sound like a plan to you?

mikee47:
@slaff Yes, good plan. My view is that partitions should make the build a lot less complicated and easier to use so we can focus on ensuring that happens!

slaff:
> However, there are some things which will break existing code:

Time to break the existing code and write a migration guide.

mikee47:
@slaff OK, ready for review, etc.

mikee47:
Documentation preview here https://smingdev.readthedocs.io/en/feature-partition-tables/

samples/FtpServer_Files/component.mk
```diff
@@ -0,0 +1,2 @@
+HWCONFIG := spiffs
+DISABLE_SPIFFS := 1
```
slaff, line 2: Aren't these two lines conflicting? Isn't the first line saying - use spiffs, here is the hardware configuration and the second one saying, well actually tell the app that we are not using spiffs? 

Can you remove the second line and leave the first one?

samples/HttpClient_Instapush/component.mk
```diff
@@ -1,2 +1,3 @@
+HWCONFIG := spiffs
 DISABLE_SPIFFS = 1
```
slaff, line 2: This sample is using spiffs - just remove line 2. The original component.mk was wrong.

samples/HttpClient_ThingSpeak/component.mk
```diff
@@ -1 +1,2 @@
+HWCONFIG := spiffs
 DISABLE_SPIFFS = 1
```
slaff, line 2: Remove line 2.

samples/HttpServer_Bootstrap/component.mk
```diff
@@ -0,0 +1,2 @@
+HWCONFIG := spiffs
+DISABLE_SPIFFS := 1
```
slaff, line 2: Remove line 2.

samples/Basic_Blink/component.mk
```diff
@@ -29,11 +29,16 @@
...
 
-## Configure flash parameters (for ESP12-E and other new boards):
-# SPI_MODE := dio
+## Configure hardware
+# Default is 'standard' (no spiffs), can also provide your own
+#HWCONFIG := spiffs
 
-## SPIFFS options
+## SPIFFS options (deprecated: use HWCONFIG)
+##	0: (default) use `HWCONFIG=standard` as default
```
slaff, line 37: The documentation does not match the meaning of the values. For example DISABLE_SPIFFS=1 is `HWCONFIG=standard`.

slaff (commented):
Looks awesome to me. Small configuration changes are needed for some of the samples.

samples/FtpServer_Files/component.mk
mikee47: I'd like to get rid of [DISABLE_SPIFFS](http://localhost:8000/_inc/Sming/Components/spiffs/index.html?highlight=disable_spiffs#envvar-DISABLE_SPIFFS) entirely as it's misleading - it means "don't generate a filesystem image". Try running `make DISABLE_SPIFFS=0` and see what happens... (answer: "Unable to open directory 'files'").

This behaviour can be overridden by customising the 'spiffs0' partition so could be removed in future...

samples/HttpClient_Instapush/component.mk
mikee47: Another way to look at this is:

```
HWCONFIG := spiffs    # I want a SPIFFS partition please
DISABLE_SPIFFS = 1    # BUT I don't want any default filesystem image (i.e. leave it un-initialised)
```

samples/Basic_Blink/component.mk
mikee47: Thanks. I'll fix this.

mikee47: Should be something like this:

```
## SPIFFS filesystem options (requires HWCONFIG=spiffs)
## 0: (default) Build filesystem image from content of 'files' directory
## 1: Leave spiffs partition un-initialised
```

samples/FtpServer_Files/component.mk
mikee47: Perhaps a better solution to this is to get rid of DISABLE_SPIFFS and replace it with `SPIFF_FILES=` instead. This has the same effect. (Try `make DISABLE_SPIFFS=0 SPIFF_FILES=`)

Sming/Components/Storage/README.rst
```diff
@@ -0,0 +1,335 @@
...
+
+
+Partition maps
+--------------
+
+This is a concise view of your flash partitions. Display it like this::
+
+   make map
+
+For the :sample:`Basic_Storage` samnple application, we get this:
```
slaff, line 159: Typo `samnple` -> `sample`.

samples/FtpServer_Files/component.mk
mikee47: Correction. DISABLE_SPIFFS and SPIFF_FILES have subtly different behaviour.

`SPIFF_FILES=` Create empty SPIFFS image (`make flash` will write this)
`DISABLE_SPIFFS = 1` Don't create a filesystem image (`make flash` will not write anything to the partition)

slaff: > I'd like to get rid of DISABLE_SPIFFS entirely

Agreed. Add also error message error when DISABLE_SPIFFS=1 is used. Something like:
`$(error  "DISABLE_SPIFFS=1 is no longer allowed. You can fix the issue by deleting this option from your component.mk file or config.")`

mikee47:
@slaff Couple of push requests to fix this PR:

https://github.com/slaff/Arduino_TensorFlowLite/pull/1
https://github.com/slaff/Sming-DIAL/pull/5

slaff:
> Couple of push requests to fix this PR:

Merged. If you want update the submodules for these two Libraries to point to the latest version.

samples/Basic_Blink/component.mk
mikee47: Removed entirely

slaff:
@mikee47 Do you want to add something more before I merge this PR?

mikee47:
@slaff Few more fixes, but that should be it unless I find anything else.

mikee47:
@slaff OK, good to go.

slaff:
@mikee47 Is it possible to change the location of the partition table? I can see that it is possible to set `partition_table_offset` to my desired location. Example:

```json
{
    "partition_table_offset": "0x3fa000", // <-- HERE
    "name": "SmartGuard V1 Partition Table",
    "base_config": "spiffs",
    "devices": {
        "spiFlash": {
            "size": "4M"
        }
    },

```

But after that the calculations are not accounting for this. For example I get:
```
Partition 'rom0' @ 0x00002000-0x000fefff must be located after @ 0x003fb000
```

And also the `map` target shows some strange results:

```
Partition map:
Device            Start       End         Size        Type      SubType   Name              Filename
----------------  ----------  ----------  ----------  --------  --------  ----------------  ------------
spiFlash          0x00000000  0x003f9fff       4072K                      Boot Sector       
spiFlash          0x003fa000  0x003fafff          4K                      Partition Table
...
```

mikee47:
@slaff Nope. The partition table offset is adjustable to make more (or less!) space for the bootloader but must be before any defined partitions. This is a requirement for ESP32 so we need to enforce it generally. Is that a problem?

mikee47:
I'll look at improving the handling of this and adding a note in the documentation.

slaff:
> Is that a problem?

No, not really.... I was hoping to be able to put the partition somewhere else so that I can have bigger roms. With the current state I should decrease the maximum rom size with 4K which can be a problem in the future...

mikee47:
I guess you're referring to the restricted memory paging capability of the ESP8266?

You've probably already figured that moving `phy_init` and `sys_param` means `rom0` can move down to 0x3000. Getting that 4K back could be done if we combine the boot sector and partition table. For example:

* rboot.bin: 1712 bytes (typical)
* partitions.bin: 3072 bytes (but 256 bytes would be adequate for most cases, supporting 8 partitions)

So there's easily enough space to squash them together like this. If you feel there's a use case for this then perhaps open an issue and we can take a look at how it might be achieved?

slaff:
> You've probably already figured that moving phy_init and sys_param means rom0 can move down to 0x3000.

Yep.

> Getting that 4K back could be done if we combine the boot sector and partition table ...  If you feel there's a use case for this then perhaps open an issue

It's ok as it is for now. Thanks for pointing out what is the `partition_table_offset` for.

PR #2177, opened 2020-12-09
Updated DIAL client library and sample…

slaff:
Updated DIAL client library and sample to match the latest UPnP changes in develop.

- [ ] Test against a real DIAL device.

### DIFF
--- a/Sming/Wiring/WString.cpp
+++ b/Sming/Wiring/WString.cpp
@@ -653,6 +653,12 @@ bool String::startsWith(const String& prefix, size_t offset) const
 	return memcmp(&cbuffer()[offset], prefix_buffer, prefix_len) == 0;
 }
 
+bool String::endsWith(char suffix) const
+{
+	auto len = length();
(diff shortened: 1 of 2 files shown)


mikee47:
Bug in UPnP library:

```
diff --git a/src/ControlPoint.cpp b/src/ControlPoint.cpp
index c190592..ce10153 100644
--- a/src/ControlPoint.cpp
+++ b/src/ControlPoint.cpp
@@ -170,6 +170,7 @@ void ControlPoint::onNotify(SSDP::BasicMessage& message)
 
 			return 0;
 		});
+		sendRequest(request);
 		return;
 	}
 
```

mikee47:
@slaff Also, we shouldn't have to mess about fetching descriptions, etc. Instead request a device and use that. I'll take a look if you like?

slaff:
> Bug in UPnP library:

Ok - make a PR with the latest UPnP version and I will merge it.

> I'll take a look if you like?

Yep, go for it.

PR #2183, opened 2020-12-18
SPIFFS buffer alignment wrong - may cause UNALIGNED exception

mikee47:
Redefine buffers using their correct types to ensure appropriate alignment.

SPIFFS_mount definition should really be corrected.


slaff:
@mikee47 Did you test this PR on a real device: either Esp8266 or Esp32 ?

mikee47:
@slaff The problem fixed occurs on the Esp8266, so yes definitely tested on hardware. (Esp32 doesn't care)

PR #2184, opened 2020-12-18
Host CI and debug code improvements

mikee47:
## Use lwip github mirror

savannah goes weird from time to time, this will help with CI reliability

## Add nothrow versions of new() for esp8266

e.g. `var = new (std::nothrow) uint8_t[123];`

## Update malloc_count
    
Enforce ENABLE_MALLOC_COUNT setting to disable hooks but retain API
Add nothrow_t new()

## Update SmingTest

Add REQUIRE2, REQUIRE_EQ and REQUIRE_NEQ macros

## Optimise m_printHex

Use buffer for `m_printHex`, can be very slow especially in Host

### DIFF
--- a/Sming/System/m_printf.cpp
+++ b/Sming/System/m_printf.cpp
@@ -245,50 +245,67 @@ void m_printHex(const char* tag, const void* data, size_t len, int addr, size_t 
 		return;
 	}
 
+	// tag + ": " + address(*8) + ' ' + hex(*2) + ' ' + char + "\r\n"
+	auto bufSize = taglen + ((bytesPerLine ?: len) * 4) + 16;
+	auto lineBuffer = new char[bufSize];
+
 	size_t offset = 0;
 	while (offset < len) {
-		if (taglen) {
+		auto linePtr = lineBuffer;
+		if (taglen != 0) {
 			if (offset == 0) {
-				m_nputs(tag, taglen);
-				m_putc(':');
+				memcpy(linePtr, tag, taglen);
+				linePtr += taglen;
+				*linePtr++ = ':';
+				*linePtr++ = ' ';
 			}
 			else {
-				for (size_t  i = 0; i <= taglen; ++i)
-					m_putc(' ');
+				memset(lineBuffer, ' ', taglen + 2);
+				linePtr += taglen + 2;
 			}
-			m_putc(' ');
 		}
 
-		if (addr >= 0)
-			m_printf("%08X ", addr);
+		if (addr >= 0) {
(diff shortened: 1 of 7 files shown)


Sming/System/m_printf.cpp
```diff
@@ -245,50 +245,67 @@ void m_printHex(const char* tag, const void* data, size_t len, int addr, size_t
 		return;
 	}
 
+	// tag + ": " + address(*8) + ' ' + hex(*2) + ' ' + char + "\r\n"
+	auto bufSize = taglen + ((bytesPerLine ?: len) * 4) + 16;
+	auto lineBuffer = new char[bufSize];
```
slaff, line 250: Why don't you declare `auto lineBuffer = new char[bufSize];` as `char lineBuffer[bufSize]={0}`? This way the delete on line 310 is not needed and also we avoid potential future  memory leaks when someone decide to put a return statement somewhere in this function without releasing lineBuffer.

mikee47: I my mind variable stack allocations are a bit dodgy. Did a quick search and found this https://www.geeksforgeeks.org/variable-length-arrays-in-c-and-c/

mikee47: Re. memory leaks I can change it to use `std::unique_ptr<>`

slaff: If I am not mistaken I have seen this recommended in some of the C/C++ code analyzers: 
`char lineBuffer[bufSize]{0}` instead of `auto lineBuffer = new char[bufSize];` when the size of the buffer is constant and will not change.  I cannot remember exactly where but I have seen this but I have found this link: https://rules.sonarsource.com/c/RSPEC-984.

mikee47: The reason I prefer heap allocation here is that passing `bytesPerLine = 0` essentially makes the size unbounded. In general, large/variable memory allocations should be on the heap. Also, don't need to zero-initialise the buffer here.

slaff: Ok, let it be this way. If this needs to be changed we can alway do it later.

slaff (approved):

Sming/System/m_printf.cpp
mikee47: Couple of thoughts for a possible future PR:

* Use `String` for the line buffer.
* Add callback parameter for printing each line.
* Re-implement as class elsewhere in framework; number of parameters is unwieldy.
* `m_printHex` can use this class internally (no public dependency on String).

PR #2188, opened 2020-12-30
Fix ESP32 digital/analogue implementations

mikee47:
Pins with analogue (RTC) functionality need extra attention.

* Add ANALOG pin mode, set internally via analogRead (approach taken by arduino32)
* Ensure RTC pins are configured correctly for digital mode. Input pullups need to be enabled/disabled in both primary and RTC IO blocks.
* Internally, prefer use of gpio_ll_XX and rtcio_ll_XX functions as these map to macros which access hardware directly (as appropriate for ESP32, ESP32S2, etc) whilst avoiding RTOS locks, etc. which Sming does not (should not) require.


slaff (commented):
Looks good to me.  Have to test it tomorrow on the ESP8266(s) and ESP32(s) that are lying around.

PR #2203, opened 2021-01-27
Fix samples location on CI system.

slaff:


mikee47:
I think it would be preferable to add `python-requirements` to each sub-make for the `make samples` target. Less of a hack!

slaff:
> I think it would be preferable to add python-requirements to each sub-make for the make samples target. Less of a hack!

I agree with you. Do you want to add a quick PR with that change?

mikee47:
I'm out ATM but can incorporate this into my `update/getting-started` branch, which I hope to push later today.

slaff:
> I'm out ATM but can incorporate this into my

Great! Thanks a lot.

slaff:
Superseded by #2204

PR #2210, opened 2021-01-31
Fix bug in `smg_uart_set_tx()`

### LINKED ISSUE #2208: Switching to GPIO2 as TX on UART0 after initializing does not work.
I'm using a H801 LED driver board, which uses GPIO2 to transmit UART0 data. The following works:

```C++
Serial.begin(115200, SERIAL_8N1, SERIAL_FULL, 2);
```

While this does not:    

```C++
Serial.begin(115200, SERIAL_8N1, SERIAL_FULL, 1); // Same as Serial.begin(115200);
Serial.setTx(2);
```

In both cases I receive the bootloader output (at 74480 baud), but only the first version produces anything when calling `Serial.write()`.

I've tried both with the current develop (bdf1d31db839385a8e1075f901b9cd95d6fb538d) and master (4.2.0, 7bc3603e22ca91c36d0b6b1bcfa96b5801afa325) branches. Toolchain is https://github.com/SmingHub/SmingTools/releases/download/1.0/x86_64-linux-gnu.xtensa-lx106-elf-e6a192b.201211.tar.gz

mikee47:
Sets pin on wrong uart. Closes #2208.

### DIFF
--- a/Sming/Arch/Esp8266/Components/driver/uart.cpp
+++ b/Sming/Arch/Esp8266/Components/driver/uart.cpp
@@ -910,9 +910,9 @@ void smg_uart_swap(smg_uart_t* uart, int tx_pin)
 bool smg_uart_set_tx(smg_uart_t* uart, int tx_pin)
 {
 	if(uart != nullptr && uart->uart_nr == UART0 && smg_uart_tx_enabled(uart)) {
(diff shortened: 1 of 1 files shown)


mikee47:
@slaff LTS bugfix

PR #2216, opened 2021-02-09
Storage fixes

mikee47:
* HOST_PARAMETERS missing from valgrind and gdb cmdline
* RingTone sample requires HWCONFIG=spiffs
* Storage missing nullptr check
* Wrong device name in error message in `Device::loadPartitions`


mikee47:
@slaff Is there any difference between appveyor run here and https://ci.appveyor.com/project/mikee47/sming/builds/37674484.
They're both on the same commit, pity the system isn't smart enough to pick this up!

slaff:
> Is there any difference between appveyor run here and

I guess no. 

> same commit, pity the system isn't smart enough to pick this up

This is something that Travis and AppVeyor still don't pick up. Such deduplication would have been quite helpful and resource saving.

PR #2218, opened 2021-02-12
Handle SPIFFS metadata more elegantly

mikee47:
**IFS changes**

Define SPIFFS_OBJ_META_LEN as build variable so it's under user control
Don't read or write metadata if defined meta length is < 16
Add magic to start of SPIFFS metadata, set dirty flag when initialised
Newly created metadata is automatically flushed to storage, but only if file is open in write mode
Add filesystem info attribute to indicate if metadata is supported

Fix block size assignment for SPIFFS to support use with other device types
Define global `IFS::Host::fileSystem&` for easier use
Fix Basic_IFS sample handling for FWFS_HYBRID=1

**Replace `spiffy` with `spiffsgen` tool from ESP IDF**

Generate default metadata for each file, provided meta len >= 16
Add check in HostTests to compare spiffy output with spiffsgen.
Output messages for `samples` and `component-samples` target to make log easier to follow

**spiffsgen.py changes**

* Add `SPIFFS_ALIGNED_OBJECT_INDEX_TABLES` setting as page indices must be word-aligned
* Don't write image file if errors occur
* Set `--use-magic` and `--use-magic-len` defaults to False
* Allow spiffsgen `base_dir` to be empty
* Remove leading / from filenames


Sming/Components/spiffs/spiffsgen.py
```diff
@@ -0,0 +1,562 @@
...
+            name = self.name.encode()
+            img += name + (b'\x00' * (self.build_config.obj_name_len - len(name)))
+
+            if self.build_config.meta_len >= 16:
+                # Build default IFS metadata
+                magic = 0xE3457A77
+                mtime = self.mtime
+                attr = 0
+                flags = 0xff
+                userRole_none = 0x00
```
slaff, line 230: Can you remove this line? LGTM complains that the local variable userRone_none is never used:
```
The value assigned to local variable 'userRole_none' is never used.
```

PR #2100, opened 2020-08-12
Feature: build new SDK 4.0.x libraries on first run.

slaff:
Related to: #2097.

PR #2221, opened 2021-02-19
Catch missing python modules and print message

mikee47:
See https://github.com/SmingHub/Sming/pull/2100#issuecomment-781745695

### DIFF
--- a/Sming/Components/Storage/Tools/hwconfig/config.py
+++ b/Sming/Components/Storage/Tools/hwconfig/config.py
@@ -5,7 +5,13 @@
 import os, partition, storage
 from common import *
 from builtins import classmethod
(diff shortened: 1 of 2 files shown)


Sming/Components/Storage/Tools/hwconfig/config.py
```diff
@@ -5,7 +5,13 @@
 import os, partition, storage
 from common import *
 from builtins import classmethod
-from rjsmin import jsmin
+
+try:
+    from rjsmin import jsmin
+except ModuleNotFoundError as err:
```
slaff, line 11: @mikee47 Can you replace `ModuleNotFoundError` with `ImportError`. It is also compatible with older python versions.

PR #2222, opened 2021-02-22
Install library python prerequisites by default.

slaff:


mikee47:
There was discussion previously about mixing installation stuff up with a build. In particular, it would break any custom python installation the user might choose. See https://github.com/SmingHub/Sming/pull/2019#issuecomment-574377010

mikee47:
What we could do is ensure that modules aren't imported until they're actually used - that would avoid issues with `make help`, etc.

slaff:
>  it would break any custom python installation the user might choose

We can make things get more complicated and use `virtualenv` which will contain only the packages for Sming. But from my experience people get confused pretty quickly...

> Already done that!

What do you mean here?

mikee47:
The argument here (which I tend to agree with) is the separation of installation and building steps. Having stuff installed 'in the background' is not appropriate.

By 'already done that' I meant moving the module import into the code itself, but the error in question (missing `rjsmin`) is with code which already does that. Removing rjsmin entirely would be one solution but my feeling is that supporting comments makes dealing with JSON more user-friendly, and it's the simplest approach I can think of.

Having said all this there is clearly still an issue because on a fresh clone running `make python-requirements` won't work! I'll get a fix sorted out for that.

mikee47:
This works:

```
make -C $SMING_HOME python-requirements
```

But again it should be done as part of the Sming installation. #2240 should help a bit with this .

mikee47:
I'm thinking with `rjsmin` it should be baked-in rather than an installed package. Easy enough to do and should avoid this problem entirely.

slaff:
Superseded by PR #2223.

PR #2223, opened 2021-02-22
Fix issues with missing `rjsmin` python package

mikee47:
Add rjsmin.py directly to source (suggested by author https://github.com/ndparker/rjsmin)
Catch missing `jsonschema` and generate warning, but allow compilation to continue

IFS: Apply the same changes, plus use `strerror` instead of `strerror_r` as it's more standard.

### DIFF
--- a/Sming/Components/Storage/Tools/hwconfig/config.py
+++ b/Sming/Components/Storage/Tools/hwconfig/config.py
@@ -3,15 +3,9 @@
 #
 
 import os, partition, storage
+from rjsmin import jsmin
 from common import *
 from builtins import classmethod
-
-try:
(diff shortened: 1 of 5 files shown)


slaff:
Looking at the logs here I have a feeling that I have merged this PR soon (probably a minute before a new force-push, but might be wrong). @mikee47 Can you check if all your changes made it into the `develop` branch?

mikee47:
> 
> Looking at the logs here I have a feeling that I have merged this PR soon (probably a minute before a new force-push, but might be wrong). @mikee47 Can you check if all your changes made it into the `develop` branch?

@slaff Yes, all OK.

PR #2224, opened 2021-02-22
Fix IFS attribute handling and other improvements

mikee47:
**Fixes**

Fix directory enumeration on Host
Ensure 'read-only' flag is honoured in SPIFFS open()
Change `setattr` to use filename instead of handle
Fix mis-identification of FWFS named objects, broke Data24 content (i.e. files >= 64K)
Fix issue with `strerror_r` which may occur with some compiler versions
Rename `truncate` -> `ftruncate` for consistency (operates on file handle)

**Improvements**

Define Compression information in structure, add `setCompression` method
Display numeric values for unknown File Attribute, User Role or Compresssion type.
Add FileSystem::readContent methods for simpler file content processing using callback
Add compression information to SPIFFS metadata
Use extended attributes to support metadata on Host
Add maxNameLength, maxPathLength to filesystem info
Update builder to store MD5 hash with files. This can be accessed via new `fcontrol()` method.
Remove `File` namespace, replacing with `File` object instead. This provides a simple wrapper for managing a file handle and complements `Directory` (now moved into library, where it should have been originally!)

**Housekeeping**

Move extra methods out of `IFileSystem` and into separate `FileSystem` descendent (keeps base class pure)
Remove `IFileSystem::isfile` method
Add `NameStat` copy constructor, assignment operator
Remove `FileStat.clear()` and use static intialiser instead (more C++)
Remove obsolete/unused code
Update file headers

**Integration testing**

Develop fstest, adding file hash checks to verify integrity.
Generate a large random file to test Data24 content.


mikee47:
@slaff Bit of a hefty move-around here but feels a little better structured.

Sming/Arch/Esp32/app.mk
```diff
@@ -11,7 +11,7 @@ LDFLAGS	+= \
 
 
 .PHONY: application
-application: $(CUSTOM_TARGETS) $(TARGET_BIN)
+application: $(TARGET_BIN)
```
slaff, line 14: @mikee47 Why are you removing CUSTOM_TARGETS from an application? For example the SSL enabled applications still need it to create a private key and certificate.

slaff: Ok, I see where it is used. Sorry, forget about this comment.

PR #2225, opened 2021-02-22
Revert removed spiffs-image-clean

Matzz:
It was removed in #2171 but is required by `spiffs-image-update`


mikee47 (approved):
Yes, this should have remained! Thanks.

mikee47:
Note also that we could probably do with targets to clean/re-build other partitions, e.g. `make part-image-update`, `part-image-clean`.

slaff:
> Note also that we could probably do with targets to clean/re-build other partitions, e.g. make part-image-update, part-image-clean.

Sounds like a new PR to me :)

PR #2226, opened 2021-02-23
Fix handling of `.gz` files in HttpResponse

slaff:
Most browsers expect correct mime type of data otherwise they refuse to load CSS, JavaScript and so on.

### DIFF
--- a/Sming/Core/Network/Http/HttpResponse.cpp
+++ b/Sming/Core/Network/Http/HttpResponse.cpp
@@ -77,14 +77,14 @@ bool HttpResponse::sendFile(const String& fileName, bool allowGzipFileCheck)
 		if(fileStats(fnCompressed, stat) >= 0) {
 			debug_d("found %s", stat.name);
 			stat.compression.type = IFS::Compression::Type::GZip;
-			stat.name = IFS::NameBuffer((char*)fileName.c_str(), fileName.length(), fileName.length());
+			stat.name = IFS::NameBuffer(const_cast<String&>(fileName));
 			return sendFile(stat);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


Sming/Core/Network/Http/HttpResponse.cpp
```diff
@@ -77,7 +77,7 @@ bool HttpResponse::sendFile(const String& fileName, bool allowGzipFileCheck)
 		if(fileStats(fnCompressed, stat) >= 0) {
 			debug_d("found %s", stat.name);
 			stat.compression.type = IFS::Compression::Type::GZip;
-			stat.name = IFS::NameBuffer(fnCompressed);
+			stat.name = IFS::NameBuffer((char*)fileName.c_str(), fileName.length(), fileName.length());
```
mikee47, line 80: Could also write this as:

```
stat.name = IFS::NameBuffer(const_cast<String&>(fileName));
```

mikee47:
Could we devise some form of loopback network interface so we can catch this with integration testing?

slaff:
>  Could we devise some form of loopback network interface so we can catch this with integration testing?

Do we have some framework/recommended way to do integration tests? Or I should run commands in the container where architecture is set to Host.

If we can create a tap network interface on the CI system then it should be possible. I will take care of this. My plan is:
- build HttpServer_Bootstrap for the Host architecture
- fetch with curl CSS and JavaScript files and check if their mime type is as expected.

mikee47:
Possibly simpler, would an instance of HttpClient and HttpServer be able to communicate within the same application? If so, then could add a module to HostTests.

PR #2227, opened 2021-02-23
Initial code for google-cast client

slaff:
This library allows you to communicate with Chromecast dongles or smart TVs that support the Google Cast Protocol.


mikee47 (commented):
I've done a review and opened a PR on your repo.

The sample runs under Host for me OK, but I had to comment out some stuff in `onMessage` so guess that's still work in progress. I get `{"type":"CLOSE"}` response when I target my chromecast dongle, so guess it's doing something!

Perhaps the complexities of processing a message could be handled using a wrapper class which manages a contained `extensions_api_cast_channel_CastMessage`. Further, that could first be done for nanopb, with the `pbEncodeData` and `pbDecodeData` helpers integrated into a `Protobuf::Stream` class.

These classes could then deal with memory allocation, de-allocation, getting/setting data, etc. in a safer and more user-friendly way.

mikee47:
Also, it would probably be easier working with this in a separate library?

slaff:
> Also, it would probably be easier working with this in a separate library? i.e. submodule

Yes, once this is at least working with the basic sample I will move it to a submodule as DIAL and others.

mikee47:
Found another useful reference source https://github.com/home-assistant-libs/pychromecast/tree/master/pychromecast

slaff:
@mikee47 Let's merge your improvements. First can you create a PR with the protobuf/nanopb improvements in SmingHub. After that I will rebase on develop and you can create a PR against slaff/Sming with the further improvements to the code. 


>  I get {"type":"CLOSE"} response when I target my chromecast dongle, so guess it's doing something!

Do you have an idea which google cast version is your dongle using? It seems to me like it is just rejecting the connection due to wrong protocol. 

> Found another useful reference source https://github.com/home-assistant-libs/pychromecast/tree/master/pychromecast

We can add this also to the documentation.

mikee47:
> 
> @mikee47 Let's merge your improvements. First can you create a PR with the protobuf/nanopb improvements in SmingHub. After that I will rebase on develop and you can create a PR against slaff/Sming with the further improvements to the code.

OK.

> 
> > I get {"type":"CLOSE"} response when I target my chromecast dongle, so guess it's doing something!
> 

I've discovered this is a standard response to an error condition. Sending the request immediately after a Connect doesn't work. However, sending the launch request when a PONG is received works fine.

mikee47:
Re. device discovery we could use UPnP for that...

slaff:
> Re. device discovery we could use UPnP for that... DIAL stuff

Hm... The last time I was looking at the docs they were saying that Google Cast Protocol v2 uses mdns to find the devices. Do you mean that all Google Cast devices support also UPnP so that we can use this instead?

mikee47:
Yes, it's probably not standard and MDNS is going to be more efficient. Currently the `MDNS` component doesn support discovery, that does need adding. All the chromecast-capable devices I have (several) are reported via UPnP with `<deviceType>urn:dial-multiscreen-org:device:dial:1</deviceType>`, so it would at least allow the device to be specified by friendly name.

slaff:
> Currently the MDNS component doesn support discovery, that does need adding.

Some time ago I have started work on this but never tested it and it's still raw. But feel free to use it and enhance it.

https://github.com/SmingHub/Sming/compare/develop...slaff:feature/mdns-query

slaff:
> After that I will rebase on develop and you can create a PR against

@mikee47 This PR is rebased on latest `develop`. Your changes from feature/channel-cast are also merged. You can now get the latest version of this PR/branch and add the rest on top of it.

mikee47:
@slaff Have pushed to your repo, got it playing back a video using the media URL from the pychromecast video2 example.

I'm getting more accustomed to how all this fits together. The library https://github.com/jhenstridge/cast-app is interesting as it goes a step further to define a channel as the `source/destination` pair, then within that channel interfaces which correspond to the particular namespace. Only when the TCP connection has been established to the Chromecast device is a channel created, with the default `Connection` and `Hearbeat` interfaces. The `Media` interfaces is constructed only when requested.

However, it's not clear to me yet whether this is entirely necessary as AFAIK only one application can be active at a time, hence only one interface of any given type. Something to chew on I think.

slaff:
> got it playing back a video using the media URL from the pychromecast video2 example.

Awesome, thanks! Will test it tonight with my TV.

mikee47:
@slaff Thanks for the MDNS code, just whipping it into shape so expect PR shortly.

mikee47:
@slaff Taken the liberty of rebasing, added a PR to your repo. proposing client discovery

mikee47:
Note that ESP32 will require #2248.

slaff:
@mikee47 When you have time compile the latest version from this PR and check if it is working on your multiple Google Cast v2 devices.

mikee47:
@slaff Yep, works a treat. Tested against a Chromecast dongle and a Sony SRS-HG1 speaker, sample running on ESP8266 nodeMCU and ESP32 wroom32. Sample could probably do with some polishing; for example, switching playback to a different device doesn't work (requires a reset) but does the job fine :-)

slaff:
> and ESP32 wroom32

Wow! 

> switching playback to a different device doesn't work (requires a reset) but does the job fine :-)

This can be part of another PR. 

@mikee47 Thanks again for your work and testing!

mikee47:
I'm particularly interested in chrome/googlecast because it's (relatively) open and well supported. One of my plans for Sming is a remote controller for chromecast/DLNA devices with physical buttons, volume control, etc. Touch screens are all very well but they can be a real PITA sometimes. This and the UPnP stuff will all be essential for that!

slaff:
@mikee47 How about adding also your audio stuff  https://github.com/mikee47/SmingAudio as a separate library to Sming?
Also what is the state of the Json Streaming Parser: https://github.com/mikee47/JsonStreamingParser/tree/develop ? Shall we merge it too?

mikee47:
@slaff I'll review both of those, will need a bit more work before I'm happy to submit them. The MP3 stuff is a bit of a gimmick on the ESP8266 (not enough RAM or processing power) but the sort of thing the ESP32 excels at; it'll need I2S support though so bit of work to be done there.

PR #2228, opened 2021-02-23
Add web test support to HostTests CI application

mikee47:
Enable loopback on host network interfaces and add CI web test module


mikee47:
@slaff This seems to work OK, if you're happy with this approach I'll rebase after #2226 ?

slaff:
> if you're happy with this approach I'll rebase after #2226 ?

@mikee47 I am more than happy :). Please rebase and let's merge it.

PR #2229, opened 2021-02-23
Replace `toString(uint8_t)` with template to avoid unintentional truncation of values

mikee47:
e.g. uint32_t -> uint8_t gets truncated without warning


slaff:
Please, check the failing CI tests.

PR #2237, opened 2021-03-02
Update MDNS library with query support

mikee47:
Add `Basic_Mdns` sample to demonstrate usage

TODO:

* [x] Why are queries not visible in Wireshark? AP must be disabled.
* [x] Add support for list of queries - at present only one question may be posed
* [x] Add support for queries so that callback can provide responder functionality
* [x] Move into separate library submodule (it's optional and will likely grow)


mikee47:
Like the SSDP server the Finder requires two UDP ports, one to listen and another to send. Few things to note here:

* Callback gets all incoming messages, not just the requested ones
* The responder will almost certainly not work whilst a Finder instances is active

The responder part should be rewritten along the same lines as the finder, sharing common code, etc. That will provide a common code base for all architectures and provide a more flexiblility. (Another PR)

slaff:
@mikee47 Can you check the failing CI build?

mikee47:
Yep, just reverting some changes which don't work on ancient compiler.
Also testing on ESP8266... looks OK.

mikee47:
OK, so running on ESP8266 responses come in OK but cannot see the outgoing queries in Wireshark. I'm using the SSDP server code as a reference and failing to spot the problem!

mikee47:
@slaff OK ready for testing.

mikee47:
@slaff This has grown a little, perhaps it should be in a separate library?

samples/Basic_Mdns/app/application.cpp
```diff
@@ -193,7 +192,7 @@ void parseFile(const String& name, const String& data)
 
 void test()
 {
-	debug_i("sizeof(mDNS::Finder) = %u", sizeof(mDNS::Finder));
+	debug_i("sizeof(mDNS::Finder) = %u", sizeof(mDNS::Server));
```
slaff, line 203: May be `"sizeof(mDNS::Finder) = %u"` should be replaced with `"sizeof(mDNS::Server) = %u"` ?

mikee47:
I've been unable to get discovery working in host emulator under Linux. Likely I need to adjust the bridge configuration - any ideas?

Sming/Components/mdns/src/Packet.h
```diff
@@ -0,0 +1,127 @@
...
+		write16(value & 0xffff);
+	}
+
+	void write(const void* s, uint16_t len)
+	{
+		memcpy(ptr(), s, len);
+		pos += len;
+	}
+
+	void writeName(const String& name)
```
slaff, line 104: The Packet stuff without the specific `writeName(...)` method could be moved to a more general location like Core/Data/ so that it can be used also from other places.

mikee47:
NB. Working in WSL2, so perhaps related. Here's my network info:

```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: bond0: <BROADCAST,MULTICAST,MASTER> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether c6:fe:c6:79:8a:2e brd ff:ff:ff:ff:ff:ff
3: dummy0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 5e:37:df:54:df:d3 brd ff:ff:ff:ff:ff:ff
4: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:15:5d:8f:24:53 brd ff:ff:ff:ff:ff:ff
    inet 172.29.167.48/20 brd 172.29.175.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::215:5dff:fe8f:2453/64 scope link
       valid_lft forever preferred_lft forever
5: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN group default qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
6: tap0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel state DOWN group default qlen 1000
    link/ether ba:a8:68:07:08:fe brd ff:ff:ff:ff:ff:ff
    inet 192.168.13.1/24 scope global tap0
       valid_lft forever preferred_lft forever
    inet6 fe80::b8a8:68ff:fe07:8fe/64 scope link
       valid_lft forever preferred_lft forever
```

Sming/Components/mdns/src/Packet.h
mikee47: Good idea. `Packet` is a pretty crap name though, any suggestions?

slaff: > any suggestions?

Go with `Packet` for now. 

At some later PR(s) we should try to unify a bit the DNS and mDNS parsing

mikee47: It should probably go into `/Core/Network/WebHelpers`

slaff: Possible but if we start using it in other non-network related code then it should move to `Core/Data/`.

mikee47: It's specifically because it deals with network byte order.

mikee47: If we use it for more general use then we might expect it to be able to write in both LSB-first (host order) and MSB-first (Network order) so renaming methods such as `write16` -> `write16n`, or adding `write16h`, etc.

mikee47: Happy to that, I'm sure it would be useful.

mikee47: Or, on reflection, two separate versions called `NetworkPacket` and `HostPacket` would be better I think.

slaff:
@mikee47 This PR looks good to me. Do you plan to add more changes or I should merge it as it is?

mikee47:
@slaff All done thanks

mikee47:
@slaff Spotted one minor bug in the code, further updates will arrive in responder

slaff:
@mikee47 Can you rebase please?

slaff:
> Spotted one minor bug in the code, further updates will arrive in responder

Ping me when ready.

mikee47:
@slaff OK rebased/tidied, plus relevant commits from the responder branch.

PR #2240, opened 2021-03-04
Updated  bearssl to its latest version.

slaff:


Sming/Components/ssl/component.mk
```diff
@@ -52,7 +52,7 @@ include/ssl/private_key.h:
 	$(info Generating unique certificate and key. This may take some time...)
 	$(Q) mkdir -p $(SSL_INCLUDE_DIR) $(SSL_CERT_DIR)
 	$(Q) chmod a+x $(SSL_TOOLS_PATH)/make_certs.sh
-	cd $(SSL_CERT_DIR); SSL_INCLUDE_DIR=$(SSL_INCLUDE_DIR) $(SSL_TOOLS_PATH)/make_certs.sh
+	$(Q) cd $(SSL_CERT_DIR); SSL_INCLUDE_DIR=$(SSL_INCLUDE_DIR) $(SSL_TOOLS_PATH)/make_certs.sh; rm -rf $(SSL_CERT_DIR)
```
mikee47, line 55: Some of those files may need to be kept; How about using `out/ssl/cert` instead?

mikee47 (approved):
Looks good. I've moved the openssl fix from #2224 as it occurs running openssl in all architectures.

mikee47:
Just one other thing. I get `rm: cannot remove 'srl': No such file or directory` when certs are generated. In https://github.com/SmingHub/Sming/blob/develop/Sming/Components/ssl/Tools/make_certs.sh#L151 is this cleanup actually necessary?

slaff:
> Just one other thing. I get rm: cannot remove 'srl': No such file or directory when certs are generated.

I will change this to `rm *.srl`. Openssl keeps the serial number per certificate in x.srl. 

> is this cleanup actually necessary?

Well... these files are intermediate so they can be removed.

PR #2241, opened 2021-03-04
IDE & install fixes

mikee47:
Add missing Host debug command line parameters to vscode config
Fix installer (legacy esp8266 toolchain)
Fix memory leak in OwnedLinkedObjectListTemplate


mikee47:
Pulled out of #2237

PR #2248, opened 2021-03-08
Rewrite MDNS responder

mikee47:
I was never really happy with #2031. Plus, it doesn't work on Esp32. Now it does :-)


samples/UdpServer_mDNS/app/application.cpp
```diff
@@ -90,6 +189,12 @@ void init()
 	WifiStation.enable(true);
 	WifiAccessPoint.enable(false);
 
+#ifdef ARCH_ESP8266
+	MacAddress mac({0x12, 0x34, 0x56, 0x78, 0x90, 0xab});
```
slaff, line 193: Why is that needed for mDNS?

mikee47: I was testing this using the Ubiquiti WifiManager android App and showing the device as `Espressif`. Changing the MAC address conceals this fact, which may be useful. I can either add a comment to this effect or just remove it...

mikee47: Either way we should have an abstraction for getting/setting the WiFi MAC address. I can add that to this PR or leave it for another?

slaff: > I can either add a comment to this effect or just remove it...

Remove it from here. 

> I can add that to this PR or leave it for another?

Separate PR please.

PR #2257, opened 2021-03-10
IFS useability improvements

mikee47:
Move spiffs_mount, etc. into Core FileSystem unit, add `fwfs_mount` and `hyfs_mount` functions


slaff:
@mikee47 The tests are failing.

Issue #2254, opened 2021-03-09
rBoot OTA and firmware filesystem


mostorer:
Hi all. New to Sming, I'm trying to enable OTA for a 1MB ESP8266 (ESP-03). 
Started from Basic_rBoot example, customized the HWCONFIG and set parameter for RBOOT_TWO_ROMS in component.mk.

Rom OTA is now working fine (to be honest I had to change Rom1 start address to 0x082000 as with 0x080000 the second rom was always corrupted), so I added the OTA update for an FWFS partition (next step is to create an Hybrid FileSystem).

As said, ROM OTA is working fine but, FWFS OTA get the partition corrupted and the mount fails.
So, at first boot after flashing I stain this printout:

```
********************************************************
Hello

133698 Found 'fwfs0v1' at 0x0f0000, length 0x010000
133845 Trying to mount 'fwfs0v1' at 0x000f0000, length 0x00010000
139392 Firmware filesystem created
143280 mount() returned 0 (Success)
145690 File system initialised
148454 =FWFS=====================
155305  -       2842 		 index.html
155681  -      15568 		 bootstrap.css
157994  -       1769 		 wifi-sprites.png
161639  -        971 		 style.css
164674  -       6252 		 settings.html
168058  -          0 		 .lastModified
171441  -      30218 		 jquery.js
174477  -          0 		 config
177353 8 file(s) found

179424 ==========================
182564 
Currently running rom 1.

185670 Type 'help' and press enter for instructions.
```

but after giving the OTA command, this is the result:

```
18866876 Updating fwfs partition 'fwfs0v1' at address 0x0f0000, size 0x010000
23713835 Firmware updated, rebooting to rom 1...

state: 5 -> 0 (0)
rm 0
pm close 7
del if0
usl

 ets Jan  8 2013,rst cause:2, boot mode:(3,7)

load 0x40100000, len 1104, room 16 
tail 0
chksum 0xe3
load 0x3ffe8000, len 572, room 8 
tail 4
chksum 0xeb
csum 0xeb

rBoot v1.4.2 - richardaburton@gmail.com
Flash Size:   8 Mbit
Flash Mode:   DIO
Flash Speed:  40 MHz

Booting rom 1 at 82000, load addr c62d0.
system param error, use last saved param!
~?
  oN?䒃? ~?c
            ?Nn?䛂?l?bl`r?o$
                            rn?`ph???N?|r??n|?
                                              ?lB?
                                                  ?|s?$?o??N?
[...]

Issue #2251, opened 2021-03-09
OTA upgrade to Sming 4.3


mikee47:
I've opened this issue to discuss how OTA (over-the-air) upgrading of existing ESP8266 devices might work when migrating to the new partition-table structure.

With partition tables the ESP8266 layout has necessarily been altered. See https://sming.readthedocs.io/en/latest/upgrading/4.2-4.3.html.

Re-flashing a device to 4.3 should be as simple as:

```
make flashinit
make flash
```

I've used the Microsoft (TM) Notepad+Pencil App to illustrate some scenarios. On the left are the flash sector numbers (easier than writing addresses).

(1) This is a typical layout for Sming 4.2 devices:

![slice1](https://user-images.githubusercontent.com/1982774/110454786-957c6b00-80bf-11eb-846a-ea5e54f09480.png)

(2) This is the new default layout for Sming 4.3:

![slice2](https://user-images.githubusercontent.com/1982774/110454788-96150180-80bf-11eb-9285-5af2fe439fc1.png)

The closest we can to the previous layout is shown in (3). There is a further development that could be made to the framework by allowing the partition table to share the boot sector, shown in (4).

![slice3](https://user-images.githubusercontent.com/1982774/110454791-96ad9800-80bf-11eb-9ef2-b12f42351018.png)

slaff:
For now:
=======
I would suggest the following layout:
- Sector 1: rBoot bootloader. 
- Sector 2: starts with the rBoot config, as before, and adds the Partition Tables (PT) in this same sector.
- Sector 3: Application starts here

Such a setup will make it possible to do remote upgrade OTA with minimal changes and risk.

Old apps will continue to work as expected.
For the new apps to work we will need a component which does the appropriate adjustments. The component itself will do very little. It will check if there is a partition table in sector 2 at offset X, X to be determined. if not - copies the current rBoot config and adds the new partition table in sector 2 with offset X. This is the most critical operation as there can be power failure after the second sector is cleared and before the new content is written. After powering on rBoot will generate a standard rBoot config which should work for most of the applications. So such a risk should be limited to small amount of devices.
This component can be enabled by default for all ESP8266 applications. And it should be possible to disable it, if request.
The execution of the component should happen before `init()`.

For the future
===========
rBoot and PT contain certain information that can be found in both. It will be better:

- To make rBoot to work with the partition table directly #2258.
[...]

PR #2258, opened 2021-03-10
Add rBoot partition support

mikee47:
This PR adds partition table support to the ESP8266 bootloader (rBoot). See #2254 #2251.

The ESP8266 flash layout has been reverted to the original Sming 4.2 layout, with the addition of the partition table in the sector before the RF calibration data at end of flash.

*Update rBoot*

Update bootloader to read ROM addresses from partition table on boot. These overwrite whatever values have been set in its own configuration, thus avoiding inconsistencies. rBoot has been forked to simplify management.

An additional partition subtype has been added for RF calibration data, so existing applications running off the develop branch will need re-building and re-flashing.

The partition table may now be freely relocated for the Esp8266 (+ Host) to allow compatibility with existing devices which must be upgrade OTA.

Methods have been added to OTA classes to allow update regions to be set using partitions; this is safer than working with raw flash addresses but the underlying mechanism hasn't changed.

TODO:

* [x] Look at OTA support
* [x] Update documentation


mikee47:
OK, I've checked this out as follows on a nodeMCU with 4MByte flash memory:

1. Build and flash `Basic_rBoot` with Sming 4.2 LTS develop branch
2. Check it's all working - 'info' is a handy command
3. Check out this PR and go to the `Basic_rBoot` project directory.
4. Run:

```
make HWCONFIG=legacy/spiffs
make map
```

Check the map looks sensible. The partition table is at the end of flash; anywhere else and it could collide with SPIFFS, etc.

5. Flash the partition map, then read it back again to verify it hasn't been overwritten. (This happened for me when I put the partition table at 0x100000 - that's where SPIFFS is so got written over.)

```
make flashmap readmap
```

6. Finally, flash the new application:

```
make flashapp terminal
```

And we should now be running with Sming 4.3.

mikee47:
The OTA stuff I haven't yet tested. Volunteers?!

mikee47:
NB. This PR is on top of #2263 so I'll rebase once that's merged.

slaff:
> so I'll rebase once that's merged.

PR  #2263 is merged. Feel free to rebase when you have time.

mikee47:
Minor fix to #2263 required...

Sming/Components/Storage/src/include/Storage/PartitionTable.h
```diff
@@ -45,7 +45,7 @@ class PartitionTable
 
 	/**
 	 * @brief Find partition by name
-	 * @param Name to search for, case-sensitive
+	 * @param Name Name to search for, case-sensitive
```
slaff, line 48: `Name Name to search` should be `Name to search`.

samples/Basic_Storage/basic_storage.hw
```diff
@@ -4,6 +4,7 @@
 	"devices": {
 		// Override default (conservative) flash settings for maximum performance
 		"spiFlash": {
+			// "mode": "qio", // This doesn't always work, depends on the flash chip
```
slaff, line 7: Replace the comment `// This doesn't always work, depends on the flash chip` with `// Make sure to set the mode according to your flash chip. Allowed SPI modes are listed here: https://sming.readthedocs.io/en/latest/_inc/Sming/Components/esptool/index.html#envvar-SPI_MODE`.

Sming/Components/rboot/include/Network/RbootHttpUpdater.h
```diff
@@ -18,32 +18,101 @@
...
+/**
+ * @brief Magic value for ROM slot indicating slot won't change after successful OTA
+ */
+constexpr uint8_t NO_ROM_SWITCH{0xff};
 
 class RbootHttpUpdater;
 
-typedef Delegate<void(RbootHttpUpdater& client, bool result)> OtaUpdateDelegate;
-
-struct RbootHttpUpdaterItem {
```
slaff, line 27: Since `RbootHttpUpdaterItem` this is gone and replace with `Item`  we should add it to the upgrade document with hints what needs to be done.

Sming/Arch/Esp8266/options.json
```diff
@@ -6,5 +6,24 @@
                 "filename": "$(FLASH_INIT_DATA_VCC)"
             }
         }
+    },
+    "legacy": {
```
slaff, line 10: @mikee47 Why we need a new memory layout for ESP8266 given the fact that now the partition table can be moved to a different location? Why not keep the current layout and not introduce a legacy layout? 
- OTA will function out of the box for ESP8266 applications
- The firmware ROMs for big flashes can be as big as possible AND use the same entry point address

Sming/Components/Storage/src/include/Storage/PartitionTable.h
mikee47: Here's what appears in the documentation:

image

Sming/Components/rboot/include/Network/RbootHttpUpdater.h
mikee47: AFAIK it's an internal variable, but as it's returned  from `getItem` I'll add it back as a deprecated alias.

Sming/Components/Storage/src/include/Storage/PartitionTable.h
slaff: > Here's what appears in the documentation:

Thank for clarifying. Ignore my comment.

Sming/Arch/Esp8266/standard.hw
```diff
@@ -13,27 +13,27 @@
...
+			"address": "self.device.size - 0x5000",
 			"size": "4K",
 			"type": "data",
 			"subtype": "rfcal"
 		},
 		"rom0": {
-			"address": "0x008000",
-			"size": "992K",
+			"address": "0x002000",
+			"size": "0x0F8000",
```
slaff, line 36: @mikee47 Thanks a lot for the changes. Would it be possible to change the value here back to a human readable size in K? At least for me it is easier to see the max size without going to another tool to calculate the size in K.

PR #2263, opened 2021-03-12
Storage improvements

mikee47:
Basic_rBoot

- Show ROM addresses in Hex
- Basic_rBoot use console by default instead of telnet emulation in Host

Add `Storage::Device::getID()` method

Add `spi_flash_get_id()` implementation for ESP32

Update Basic_Storage sample

- Reduce default esp32 app partition size, bit big
- QIO mode doesn't always work, remove it
- List storage devices, with ID

Add `flashconfig` target for Host

Fix inconsistent HWCONFIG path separation in Windows

- Using GNU make / mingw conventions

Remove unused header
Fix use of deprecated functions

### DIFF
--- a/Sming/Components/Storage/src/SpiFlash.cpp
+++ b/Sming/Components/Storage/src/SpiFlash.cpp
@@ -22,7 +22,7 @@ String SpiFlash::getName() const
 	return FS_SPIFLASH;
 }
 
-uint32_t SpiFlash::getID() const
+uint32_t SpiFlash::getId() const
 {
 	return spi_flash_get_id();
 }
--- a/Sming/Components/Storage/src/SpiFlash.cpp
+++ b/Sming/Components/Storage/src/SpiFlash.cpp
@@ -22,6 +22,11 @@ String SpiFlash::getName() const
 	return FS_SPIFLASH;
 }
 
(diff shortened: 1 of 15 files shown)


Sming/Components/Storage/src/SpiFlash.cpp
```diff
@@ -22,6 +22,11 @@ String SpiFlash::getName() const
 	return FS_SPIFLASH;
 }
 
+uint32_t SpiFlash::getID() const
```
slaff, line 25: Please rename the method to `getId()`

slaff, line 25: Please rename the method to `getId()`

Sming/Components/Storage/src/include/Storage/Device.h
```diff
@@ -88,6 +88,15 @@ class Device : public LinkedObjectTemplate<Device>
 	 */
 	virtual String getName() const = 0;
 
+	/**
+	 * @brief Obtain device ID
+	 * @retval uint32_t typically flash chip ID
+	 */
+	virtual uint32_t getID() const
```
slaff, line 95: Rename to `getId()`

PR #2265, opened 2021-03-14
Fix Basic_Storage sample

mikee47:


slaff:
@mikee47 I am having the following problem on the Host architecture. I have a hardware configuration that has this line in it:
```
    "partition_table_offset": "0x003fa000",
```
On Esp8266 the table offset is working as expected. But on the Host it does not seem to work. I get the following error back:
```
Warning! Note: No App partitions found
```
It seems that the Host emulator expects the partition table to be at offset `0x2000` and ignores the setting in the hardware configuration. Is it a lot of effort to make the Host emulator respect the partition table offset?

mikee47:
@slaff The Host 'bootloader' is baked into the `rboot` Component, so that needs re-building, as does `Storage` (i.e. `make rboot-rebuild Storage-rebuild`). There isn't any defined dependency at present but that's an easy fix - I'll push it shortly (with some other updates).

Note also that the esp8266 bootloader isn't rebuilt automatically if the partition table offset changes, so that needs a bit more thought.

mikee47:
@slaff OK #2302 should resolve your issues, although the ESP8266 bootloader has to be manually rebuilt/flashed if the partition table offset (or any of the other parameters) are changed. Not sure how best to address that ATM, could build it using variants perhaps?

PR #2267, opened 2021-03-15
Fix handling of `HWCONFIG_OPTS`

mikee47:
These are always applied **after** everything else has loaded.
This differs from `options` used within a profile, which are parsed after resolving `base_config` but before any other properties.


slaff:
It looks better now. @mikee47 Would it be possible to add also sorting of the partitions by their address when displaying them with `make hwconfig` ?

Issue #2246, opened 2021-03-07
 /opt/sming/Tools/install.sh fails on clang-format-6.0 \


frankdownunder:
ubuntu 20.04 doesnt have  clang-format-6.0 but it does have clang-format.

mikee47:
Strange. Here's what I get:

```
mike@ATHLONME:~/sandboxes/sming/Sming/Libraries/google-cast/samples/basic$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.2 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

mike@ATHLONME:~/sandboxes/sming/Sming/Libraries/google-cast/samples/basic$ apt search clang-format
Sorting... Done
Full Text Search... Done
arcanist-clang-format-linter/focal 0.git20161021-2 all
  clang-format linter for Arcanist

clang-format/focal 1:10.0-50~exp1 amd64
  Tool to format C/C++/Obj-C code

clang-format-10/focal 1:10.0.0-4ubuntu1 amd64
  Tool to format C/C++/Obj-C code

clang-format-11/focal-updates 1:11.0.0-2~ubuntu20.04.1 amd64
  Tool to format C/C++/Obj-C code

clang-format-6.0/focal,now 1:6.0.1-14 amd64 [installed]
  Tool to format C/C++/Obj-C code

clang-format-7/focal 1:7.0.1-12 amd64
  Tool to format C/C++/Obj-C code

clang-format-8/focal 1:8.0.1-9 amd64
  Tool to format C/C++/Obj-C code

clang-format-9/focal 1:9.0.1-12 amd64
  Tool to format C/C++/Obj-C code

mike@ATHLONME:~/sandboxes/sming/Sming/Libraries/google-cast/samples/basic$
```

slaff:
> ubuntu 20.04 doesn't have clang-format-6.0

@frankdownunder Try the following command: `apt install clang-format-6.0`

[...]

PR #2268, opened 2021-03-15
Upgrade to clang-format 8

mikee47:
This PR upgrades the standard clang-format tool to version 8.0.1 (final). It produces (slightly) different results to version 6.

Version 6 can no longer be used to submit core changes. Users should upgrade to verison 8 first: `sudo apt install clang-format-8`

Note that cf9 produces yet different results, however at present version 8 is available from Ubuntu 16.04 up to the latest 20.10 (groovy gorilla) release. (Note: cf6 no longer provided in the latter - see #2246).

Existing settings produce different results with version 8.
This seems to be related to the `AlignAfterOpenBracket: Align` setting which appears to use different break rules.
There may be another setting which fixes this but I've not found it!

Setting `AlignAfterOpenBracket: AlwaysBreak` produces consistent results with either version, however the code changes are significant and this is not desirable.

### DIFF
--- a/Sming/Core/Network/NtpClient.cpp
+++ b/Sming/Core/Network/NtpClient.cpp
@@ -46,13 +46,14 @@ void NtpClient::requestTime()
 	}
 
 	ip_addr_t resolvedIp;
-	int result = dns_gethostbyname(server.c_str(), &resolvedIp,
-								   [](const char* name, LWIP_IP_ADDR_T* ip, void* arg) {
-									   // We do a new request since the last one was never done.
-									   if(ip)
-										   reinterpret_cast<NtpClient*>(arg)->internalRequestTime(*ip);
-								   },
-								   this);
+	int result = dns_gethostbyname(
+		server.c_str(), &resolvedIp,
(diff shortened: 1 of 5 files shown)


mikee47:
@slaff Any thoughts on this? Moving to clang-format-8 is probably the easiest way forward.

slaff:
> Moving to clang-format-8 is probably the easiest way forward.

Definitely, but I am not keen on the new style changes. I will have to check what other values are available for `AlignAfterOpenBracket` and experiment a bit.

mikee47:
I've reverted the style changes so this is just what happens using cf8.

mikee47:
Note that this may affect code in submodules but shouldn't affect CI

slaff:
> so this is just what happens using cf8. I've failed to produce the same results with cf6 so it's a one-way change.

I like the impact and the changes. Only three files and a better visual style. This is the right way to go. Can you also add to this PR the following changes ?
- updates to the CI to use clang-8
- update the documentation to use clang-8

PR #2271, opened 2021-03-17
Update directory enumeration and FTP server

mikee47:
This PR deals with directory listings in the framework.
Using a Vector<String> is inefficient, `Directory` object is almost as simple and more flexible.

The FTP framework and sample have been updated accordingly, with improvements to demonstrate use of
IFS file permissions.

Fully deprecate `fileList` and update framework using `Directory` object instead

Fix command handler in Host mode - requires different EOL character

Various minor updates and fixes to IFS component.

**Revise FTP Server**

Fix bad TcpConnection write calls
Simplify operation - some logic involving `canTransfer` doesn't seem to be necessary
Re-structure command handling, add login and file access checks.
More commands available if user isn't logged in, permissions handled by filesystem
Note: Default file permissions for new files written to SPIFFS are read/write for Admin only.
Extended directory listings: displayed 'w' value combines READ attribute with user permissions
Add HELP, NLST, CDUP, XCWD and XPWD commands
Base `FtpServer` on `CustomFtpServer` class, user may choose to manage users differently
FTP server can use separate filesystem if required

**Update FtpServer_Files sample**

Add FWFS and mount together with SPIFFS as a hybrid volume (makes things a bit more interesting)
Add another two user accounts with different roles

### DIFF
--- a/samples/FtpServer_Files/app/application.cpp
+++ b/samples/FtpServer_Files/app/application.cpp
@@ -14,7 +14,10 @@ void gotIP(IpAddress ip, IpAddress netmask, IpAddress gateway)
 	Serial.println(ip);
 	// Start FTP server
 	ftp.listen(21);
-	ftp.addUser("me", "123"); // FTP account
+	// Add user accounts
+	ftp.addUser("guest", nullptr, IFS::UserRole::Guest);
+	ftp.addUser("me", "123", IFS::UserRole::User);
+	ftp.addUser("admin", "1234", IFS::UserRole::Admin);
 	// You can also use special FTP comand: "fsformat" for clearing file system (for example from TotalCMD)
 }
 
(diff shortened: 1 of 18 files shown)


samples/FtpServer_Files/app/application.cpp
```diff
@@ -28,11 +31,14 @@ void connectFail(const String& ssid, MacAddress bssid, WifiDisconnectReason reas
 
 void init()
 {
-	spiffs_mount(); // Mount file system, in order to work with files
-
 	Serial.begin(SERIAL_BAUD_RATE); // 115200 by default
```
slaff, line 34: @mikee47 I know that this is not something that you have added but would it be possible to remove also the unnecessary comment from line 26 ?
```
// .. some you code for configuration ..
```

samples/FtpServer_Files/app/application.cpp
```diff
@@ -28,11 +31,14 @@ void connectFail(const String& ssid, MacAddress bssid, WifiDisconnectReason reas
...
 	Serial.systemDebugOutput(true); // Enable debug output to serial
 
+	// Mount file system, in order to work with files
```
slaff, line 37: Maybe you can add a bit more text explaining the difference between the three mount commands.

PR #2273, opened 2021-03-20
Improve SPIFFS directory handling

mikee47:
The SPIFFS IFS wrapper has limited directory emulation, so that `opendir`, `readdir`, etc. work consistently across filesystems.

This PR fixes `readdir` so that 'directories' are included in the output. (This requires de-duplication so if there are multiple files such as `/dir/file.txt`, `/dir/file2.txt` then 'dir' is reported only once.)

Also add a `rewinddir` implementation.

Note: If there is a file called `/dir` in addition to the above files, then readdir() will report two entries, one as a directory and the other as the file.


slaff:
@mikee47 are you interested in adding also LittleFS support to IFS so that we can demo the usage of non-Spiffs file systems and better compare and benchmark them?

mikee47:
@slaff Yes, I'll get cracking on LittleFS...

PR #2274, opened 2021-03-21
Add graphical hardware profile editor

mikee47:
I went with TK for this as it's bundled with python3. Basic but hopefully useful.

Enter `make hwconfig-edit` from project directory.


slaff:
@mikee47 The editor looks great! Would it be possible to add also delimiter/borders marking the end of a MB? For example delimite/border after 1MB, 2MB and so on.

mikee47:
@slaff Adding markers, etc. isn't practical/possible with the treeview widget, but it was the simplest way to get a useable interface.

Further improvement could be made using a fully custom-drawn area, showing a more graphical map with some scaling, shading, etc. Something for the future...

slaff:
> Something for the future...

Ok, sounds fare. Are you ready with this PR or would like to add some more changes before I merge it?

mikee47:
@slaff All done for now thanks.

mikee47:
@slaff #2277 adds custom drawn map, is this the kind of thing you were after?

slaff:
>  is this the kind of thing you were after?

I like the map. I was thinking about something simple but this is much better. If you haven't done so take a look for inspiration at qtparted or gparted in Linux. Their interface is powerful and easy to use.

Issue #2276, opened 2021-03-22
Latest develop: Exception SMTP sample


bolzenwolf:
Hi together,

I'm currently facing some problems running Sming SMTP sample.
It compiles okay, but always runs into exception after sending the test mail.
`client.send(mail)`

Please see attachement: [stack-trace.txt](https://github.com/SmingHub/Sming/files/6184775/stack-trace.txt)

Does anybody has an idea what may cause this exception?

slaff:
Paste here also the decoded stacktrace `make decode-stacktrace`.

bolzenwolf:
Sure, forgot to do so.

Attached decoded stacktrace:
```
0x4022fba0: system_partition_get_item at ??:?
0x40101096: malloc at ??:?
0x40101053: pvPortRealloc at ??:?
0x402399a2: MemoryDataStream::write(unsigned char const*, unsigned int) at /opt/sming/Sming/Core/Data/Stream/MemoryDataStream.cpp:57
 (inlined by) MemoryDataStream::write(unsigned char const*, unsigned int) at /opt/sming/Sming/Core/Data/Stream/MemoryDataStream.cpp:48
0x402384f6: TcpClient::send(char const*, unsigned short, bool) at /opt/sming/Sming/Core/Network/TcpClient.cpp:68
0x402348fe: String::reserve(unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:199
 (inlined by) String::reserve(unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:173
0x4022fba0: system_partition_get_item at ??:?
0x40234a2a: String::concat(char const*, unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:383
0x402348fe: String::reserve(unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:199
 (inlined by) String::reserve(unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:173
0x40234e78: String::concat(char const*, unsigned int) at /opt/sming/Sming/Wiring/WString.cpp:389
0x40236bdc: TcpClient::sendString(String const&, bool) at /opt/sming/Sming/Core/Network/TcpClient.h:108
0x402381c3: String::~String() at /opt/sming/Sming/Wiring/WString.h:207
 (inlined by) StringSumHelper::~StringSumHelper() at /opt/sming/Sming/Wiring/WString.h:878
[...]

PR #2280, opened 2021-03-23
Fix SMTP issues

mikee47:
This PR should fix the exception raised in #2276, where a read is attempted after the SSL connection is closed. This creates an assertion, which has been replaced with a debug statement. The SMTP connection closure is now handled with a timeout to allow the connection to close gracefully.

Note that other problems were found during investigation and addressed via #2283, #2284. The latter caused corruption in attachments because the source file position got out of sync.

Also fixes bug introduced in #1505 25/10/2018 which messes up SMTP state machine.

Note: Behaviour is clearly observable by compiling with `DEBUG_VERBOSE_LEVEL=3`.

### DIFF
--- a/Sming/Core/Network/SmtpClient.cpp
+++ b/Sming/Core/Network/SmtpClient.cpp
@@ -35,7 +35,7 @@
 	}
 #define ADVANCE_UNTIL_EOL                                                                                              \
 	do {                                                                                                               \
-		if(*(buffer - 1) == '\r' && *buffer == '\n') {                                                                 \
+		if(buffer[-1] == '\r' && buffer[0] == '\n') {                                                                  \
 			ADVANCE_AND_BREAK;                                                                                         \
 		}                                                                                                              \
 		ADVANCE;                                                                                                       \
@@ -44,7 +44,7 @@
 #define ADVANCE_UNTIL_EOL_OR_BREAK                                                                                     \
 	{                                                                                                                  \
 		ADVANCE_UNTIL_EOL;                                                                                             \
-		if(*(buffer - 1) != '\n') {                                                                                    \
+		if(buffer[-1] != '\n') {                                                                                       \
 			break;                                                                                                     \
 		}                                                                                                              \
 	}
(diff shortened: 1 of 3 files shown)


slaff:
@mikee47 Can you move the stream changes to another PR? I will try to merge first the Stream PR and then this one.

mikee47:
@slaff For info, I've rebased this branch on develop and re-tested, works without #2288.

slaff:
> ..  re-tested, works without #2288.

Ok, I will review it and will merge it without waiting for #2288.

Sming/Core/Network/SmtpClient.cpp
```diff
@@ -143,45 +139,39 @@ void SmtpClient::onReadyToSendData(TcpConnectionEvent sourceEvent)
...
+			authPlain();
+		}
 
 		if(state == eSMTP_SendAuth) {
 			state = eSMTP_Ready;
 		}
 
 		break;
-
-	default:; // Do nothing
```
mikee47, line 184: It's a bit weird having this in the middle of the switch statement - I moved it to the normal place. In fact, This appears to be an error as it would handled all other states as if they were `eSMTP_SendAuthResponse`. I've re-tested develop branch and it fails the transfer with various messages, ending in `UNSUPPORTED_VERSION`. With this change it all works as expected.

slaff: Interesting. I have to see how the `default` ended in the wrong place. Thanks for pointing this out.

mikee47:
#1505 crikey

slaff:
image

Yep, I guess this is the result of some merging/diffing gone wrong. I will review the whole `SmtpClient::onReadyToSendData` method and see if something else went south.

slaff:
> I will review the whole SmtpClient::onReadyToSendData method and see if something else went south.

Ok, at least the logic of the `SmtpClient::onReadyToSendData` did not change, apart from the ill-fated `default: ...`. As far as I can see the changes in this PR should merge without any side-effects also in the LTS version. Please, correct me if I am wrong.

PR #2282, opened 2021-03-24
Fix `buildpart` target

mikee47:
Incorrectly reports `No partitions have build targets`

### DIFF
--- a/Sming/Components/Storage/component.mk
+++ b/Sming/Components/Storage/component.mk
@@ -145,7 +145,7 @@ endef
 
 .PHONY: buildpart
 buildpart: ##Rebuild all partition targets
-	@if [ "$(PARTITION_BUILD_TARGETS)" -eq "" ]; then \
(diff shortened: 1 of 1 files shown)


Sming/Components/Storage/component.mk
```diff
@@ -145,12 +145,12 @@ endef
 
 .PHONY: buildpart
 buildpart: ##Rebuild all partition targets
-ifeq (,$(PARTITION_BUILD_TARGETS))
-	@echo "No partitions have build targets"
-else
-	$(Q) rm -f $(PARTITION_BUILD_TARGETS)
-	$(MAKE) $(PARTITION_BUILD_TARGETS)
-endif
+	@if [ "$(PARTITION_BUILD_TARGETS)" -eq "" ]; then \
```
slaff, line 148: Please use $(Q) instead of @. Example: 
```
$(Q) if [
```

PR #2286, opened 2021-03-24
Base64 size calculations wrong [LTS]

mikee47:


slaff:
@mikee47 There is no need for a separate PR. I will merge it in the LTS just by cherry-picking #2283

PR #2290, opened 2021-03-26
Add support for repeated HTTP header fields

slaff:
Some fields can appear multiple times in one HTTP response or request.

The following headers are supported:
- Set-Cookie
- WWW-Authenticate
- Proxy-Authenticate

Updated HttpResponse::setCookie to support adding multiple cookies in one response.


mikee47 (commented):
@slaff I've pushed some suggestions to your repo.

PR #2304, opened 2021-04-10
Add LittleFS support

mikee47:
This PR adds an IFS implementation for [LittleFS](https://github.com/littlefs-project/littlefs).

The `Basic_IFS` sample has been updated to demonstrate use of littlefs volumes. Initial image files can be created using the `lfs-build` are defined using an FWFS build configuration file, which is then copied into the target littlefs image file using an `fscopy` tool.

**TODO**

These items are not critical to filesystem operation and can be added in future updates.

The following two methods are not fully implemented - they require modifications to the littlefs library.

```
FileHandle IFileSystem::fopen(const Stat& stat, OpenFlags flags);
int IFileSystem::fremove(FileHandle file);
```

Note: The `fopen()` call is an optimisation to avoid the need to re-parse the filesystem for a file which has already been discovered during a directory enumeration. The `Stat::id` field is intended to provide a way for the implementation to do this, but for littlefs it will probably need the filename as well. Also, there are a few internal optimisations which can be made to avoid the need to store filenames and paths separately.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -219,6 +219,10 @@
 	path = Sming/Libraries/libsodium/libsodium
 	url = https://github.com/jedisct1/libsodium.git
 	ignore = dirty
+[submodule "Libraries.LittleFS"]
+	path = Sming/Libraries/LittleFS
+	url = https://github.com/mikee47/Sming-LittleFS
+	ignore = dirty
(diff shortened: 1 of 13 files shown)


slaff:
@mikee47 are you ready with this PR?

mikee47:
@slaff yes

PR #2305, opened 2021-04-16
[WIP] Feature: component hosted

slaff:
The hosted component allows Sming's host emulator to run parts of the commands on an actual microcontroller.
The communication is done via [simplePRC](https://simplerpc.readthedocs.io)  and the microcontroller has to be flashed with a special application.

Overview
--------
Sming's host emulator allows easier debugging and development of embedded applications. This component named "Hosted" extends the host emulator and facilitates testing functionality that only a real microcontroller can provide as for example digital I/O operations or SPI operations.

For example in order to run the Basic_Blink application under the host emulator and run the actual blinking of a LED on a microcontroller
we can compile the application using the following directives::
```
make SMING_ARCH=Host ENABLE_HOSTED=tcp HOSTED_SERVER_IP=192.168.4.1
```
`SMING_ARCH=Host` instructs the build system to build the application for the Host architecture.
`ENABLE_HOSTED=tcp` instructs the host emulator to communication with the real microcontroller using TCP
`HOSTED_SERVER_IP=192.168.4.1` instructs the host emulator to connect to IP `192.168.4.1`.

In the sub-directory ``samples`` inside this component you will find the sample applications that will turn your microcontroller into
a remote RCP server.

The compilation and flashing for ESP32, for example, can be done using the following commands:

```
cd samples/tcp
make SMING_ARCH=Esp32 WIFI_SSID=YourSSID WIFI_PWD=YourPassword
make flash
```

If you replace ``SMING_ARCH=Esp32`` with ``SMING_ARCH=Esp8266`` then the hosted application will be compiled and flashed on a ESP8266 microcontroller.
Make sure to replace the values of  WIFI_SSID and WIFI_PWD with the actual name and password for the Access Point (AP).


Sming/Arch/Host/app.mk
```diff
@@ -12,6 +12,11 @@ LDFLAGS += \
 # Executable
 TARGET_OUT_0			:= $(FW_BASE)/$(APP_NAME)$(TOOL_EXT)
 
+# Hosted Settings
+ifneq ($(ENABLE_HOSTED),)
+	COMPONENTS_AR := $(USER_LIBDIR)/$(CLIB_PREFIX)Hosted-$(CMP_Hosted_LIBHASH).a $(USER_LIBDIR)/$(CLIB_PREFIX)Hosted-Lib-$(CMP_Hosted-Lib_LIBHASH).a $(COMPONENTS_AR)
```
slaff, line 17: @mikee47 Is it possible with the existing build system for a component to put its library at the front of the list of compiled libraries?

Sming/Components/Hosted/include/Hosted/Client.h
```diff
@@ -0,0 +1,178 @@
...
+	template <typename R> R wait()
+	{
+		size_t neededBytes = sizeof(R);
+
+		while(stream.available() < int(neededBytes)) {
+			stream.flush();
+			host_main_loop();
+		}
+
+		std::shared_ptr<uint8_t> data(new uint8_t[neededBytes]);
```
slaff, line 73: @mikee47 I would like to get a stream of bytes and cast them to the expected return type without also having a memory leak. So far I came up with this code here but is there a better way to do it in C++?

mikee47: How about this:

```
R result;
stream.readBytes(&result, sizeof(result));
return result;
```

slaff: Wow! Cannot be that simple :)

Sming/Arch/Host/app.mk
mikee47: This is probably the only way. I'm guessing the reason for this is to get the linkage with weak functions working properly? Note that weak functions won't work at all in Windows! Function wrappers are more reliable - the `malloc_count` Component uses these a lot so a good reference.

slaff: > Function wrappers are more reliable - the malloc_count Component uses these a lot so a good reference.

Ok, thanks for pointing this out.  The `host_init` weak function is replaced with wrapper.

Sming/Arch/Host/Core/SPI.h
```diff
@@ -30,24 +30,18 @@
...
+	void transfer(uint8_t* buffer, size_t numberBytes) override;
 
 protected:
-	void prepare(SPISettings& settings) override
-	{
-	}
+	void prepare(SPISettings& settings) override;
 };
 
 /** @brief  Global instance of SPI class */
```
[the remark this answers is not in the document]
slaff: I have split them into header and implementation to make it easier for the hosted component to replace the `real` implementation with its own implementation. I would prefer to leave it as it is and add the SPI hosted implementation in a separate PR if that is ok for you.

PR #2312, opened 2021-04-20
Feature: hosted serial

slaff:
Adds easier way on Linux and Windows to connect to a real serial port and communicate with the device over it.


slaff:
@mikee47 Sometimes I am testing applications that have to communicate with another microcontroller via serial. It would be great if the Sming Host Emulator can be instructed to use `serialib` (part of this PR) for UART0 communication instead of the Telnet server that is created. UART1 till N should continue to use the Telnet server.  An example compilation can be 

```
make SMING_ARCH=Host UART0_PORT=/dev/ttyUSB0 UART0_SPEED=9600  
```

Is that doable?

mikee47:
> Is that doable?

Of course :-)

mikee47:
@slaff See if #2326 does what you're after...

PR #2316, opened 2021-04-25
Pull Core/Network out into separate Component

mikee47:
The `DISABLE_WIFI` setting was introduced as an experimental feature for Esp8266 to reduce image sizes. This PR moves the networking support code into a separate library so it doesn't need to be built for these applications.

This also improves built times for tools using the Host architecture, for example the LittleFS `fscopy` tool.

The related code for each architecture has also been moved, including related units from `Platform`.

Documentation has been revised and a few minor fixes included.


Sming/Components/Network/src/Network/Http/HttpRequestAuth.cpp
```diff
@@ -12,7 +12,7 @@
 
 #include "HttpRequestAuth.h"
 #include "HttpRequest.h"
-#include "Network/WebHelpers/base64.h"
+#include <Data/WebHelpers/base64.h>
```
slaff, line 15: @mikee47 Please start an upgrade document for this release and add information that `Network/WebHelpers/base64.h` is now moved to `Data/WebHelpers/base64.h`.

mikee47: You happy with this change? We could leave the path as-is... or provide an alias so it doesn't break anything...

slaff: > or provide an alias so it doesn't break anything

I am ok with the change. It would be great if you can add the old header file pointing to the new one with a deprecation notice for smoother transition. Similar to what we have here: https://github.com/SmingHub/Sming/blob/d375ef74fd11f0796aa4483c0bdaacb4fb755655/Sming/SmingCore/SmingCore.h

PR #2325, opened 2021-05-01
ESP32/network fixes

mikee47:
- Basic_WebSkeletonApp goes pop with ESP32 due to `ESP_ERROR_CHECK`.

* [x] Don't start webserver until ~~gotIP~~ onReady

Don't use ESP_ERROR_CHECK so much...
I guess calling these when interface isn't initialised is a genuine logic error.. but could be annoying

- Fix `Ssl::Options::toString()`

- Change `Timer` to `SimpleTimer` for HttpClient
Uses a little less RAM.

- Increase ESP32 timer task stack size
Bugchecks during HostTests


avr39-ripe:
Don't think that we can start web server AFTER got IP, we need this also for station to configure wifi.

mikee47:
> 
> 
> Don't think that we can start web server AFTER got IP, we need this also for station to configure wifi.

Thanks for pointing this out. The problem in this (and other samples) appears to be the call to `getIP()` in `startWebServer` which clearly must fail! I'll take another look at this.

mikee47:
Actually other sames start webservers after GotIP() so the problem is more widespread.

I've done a brief test on the Esp8266, and the IP is allocated **after** leaving `init()`, then a couple of mode change notifications (to Wifi event handler) before our ready handler is called - that looks like the best place to startup web servers.

```
129858 mount() returned 0 (Success)
129996 File system initialised
sleep disable
142933 open('.therm.conf'): NOT_FOUND (-11002)
143110 File '.therm.conf' open error: NOT_FOUND
143853 Station configuration is: PleaseEnterSSID
145334 WifiAccessPoint.enable()
mode : sta(dc:4f:22:1c:26:66) + softAP(de:4f:22:1c:26:66)
add if0
add if1
dhcp server start:(ip:192.168.4.1,mask:255.255.255.0,gw:192.168.4.1)
bcn 100
162955 event 8

163099 Old mode: 2, new mode: 3
165172 event 8

166479 Old mode: 3, new mode: 3
169415 onReady()
scandone
no PleaseEnterSSID found, reconnect after 1s
3004981 event 1

3005127 disconnect from ssid PleaseEnterSSID, reason 201

3005360 DISCONNECT - SSID: PleaseEnterSSID, REASON: No AP found

reconnect
```

slaff:
@mikee47 Ready with this PR? Did you test the changes on a real Esp32?

mikee47:
Just re-tested with Basic_IFS sample (which is where all this started) and all good. Moved the `startWebServer` call into `init()` and it doesn't go pop :-)

mikee47:
This is on ESP32. All done!

PR #2326, opened 2021-05-04
Enable Host access to physical COM ports

mikee47:
This PR extends the Host serial port emulation to include local physical devices using the `--device` command-line option. For example:

```
make run HOST_UART_OPTIONS="--uart=0 --device=/dev/ttyUSB0"
```

The `--device` option must follow the `--uart` option. Another example:

```
make run HOST_UART_OPTIONS="--uart=0 --device=/dev/ttyUSB0 --uart=1 --device=/dev/ttyUSB1"
```

The port is opened when `uart_init` gets called. The default baud rate is whatever the application has requested. This can be overridden as follows:

```
make run HOST_UART_OPTIONS="--uart=0 --device=/dev/ttyUSB0 --baud=921600"
```


.gitmodules
```diff
@@ -65,6 +65,10 @@
 	path = Sming/Components/rboot/rboot
 	url = https://github.com/mikee47/rboot
 	ignore = dirty
+[submodule "seriallib"]
```
slaff, line 68: May be it will be better to move serialib to Arch/Host/Components. It is only Host component and it was my mistake making it available for all architectures.

mikee47:
@slaff I've added [TEMPORARY] commit for a way to test this more easily. Should remove that before merge :-)

Sming/Arch/Host/Components/hostlib/options.h
```diff
@@ -31,6 +31,10 @@
 	XX(help, no_argument, "Show help", nullptr, nullptr, nullptr)                                                      \
 	XX(uart, required_argument, "Enable UART server", "PORT", "Which UART number to enable",                           \
 	   "e.g. --uart=0 --uart=1 enable servers for UART0, UART1\0")                                                     \
```
slaff, line 33: There is one side effect when calling this command  `make run HOST_UART_OPTIONS="--uart=0 --device=/dev/ttyUSB0"`. On Linux a new terminal is still opened although it is not needed. Can you adjust also the `run` makefile target? If it is too much of hassle you can also leave it for another PR.

mikee47: Tools use this variable separately to easiest fix is to just set `ENABLE_HOST_UARTID=` on the command line.

slaff:
@mikee47 Are you planning to add something more here?

mikee47:
@slaff  Just remove temporary commit and done.

Issue #2328, opened 2021-05-12
ESP32 Support


gbaranski:
Hello, I'm creator of [Houseflow](https://github.com/gbaranski/houseflow), it's an IoT platform for embedded devices like ESP8266 and also willing to support ESP32. I've been using [ESP8266_RTOS_SDK](https://github.com/espressif/ESP8266_RTOS_SDK), but for me it is too close to metal, and toolchain is quite complicated, no support for C++, and as far as I know no support for ESP32. I've recently tried Arduino Framework, but it has weird API, it encourages way of programming with `void loop()`. And now I found your project, it looks really great, and I have few questions about it,

- How compatible ESP8266 code is with ESP32? Is it like with Arduino framework, so you can copy-paste most things and they will work?
- How stable ESP32 is with Sming framework? 
- Do libraries such as websockets, HTTP,  OTA, and etc work on ESP32?

Thanks in advance.

slaff:
@gbaranski We do have some initial support for the ESP32 architecture but it is not complete. For example there is no SPI or I2S support yet.  ​

>  Do libraries such as websockets, HTTP, OTA, and etc work on ESP32?


All network protocols, clients and server that are supported in ESP8266 are supported also here. This means that the HTTP client and server will work, the SSL support is there, the MQTT client will work, FTP and so on. 

OTA on ESP32 is not working yet. For Esp8266 we use rBoot as the bootloader but ESP32 has another bootloader. 
With a bit of work the OTA component could support also ESP32. 

> How stable ESP32 is with Sming framework?

Hard to say for an architecture that is marked as experimental and incomplete. The network is stable and working. For the other features you might find a lot of rough edges.

> How compatible ESP8266 code is with ESP32? 

I guess you wanted to ask how portable is Sming code between the different supported architectures: Esp8266, Esp32, Host. If that is the question then all features coming from Sming that are supported for Esp8266 and Esp32 should work without any porting/re-writing. Sming itself takes care to hide the differences and do the heavy lifting.

> Is it like with Arduino framework, so you can copy-paste most things and they will work?

Yes and no. Take a look at the multiple [samples](https://github.com/SmingHub/Sming/tree/develop/samples) that we have to get a better understanding. And don't forget to read the [documentation](https://sming.readthedocs.io/).

slaff:
@gbaranski  OTA for Esp32 has landed with the latest commits in our `develop` branch.

gbaranski:
[...]

PR #2332, opened 2021-05-21
Feature: unified Over-The-Air (OTA) upgrades

slaff:
This feature adds architecture-agnostic OTA support. Some of the highlights here:

- Added component `Ota` that encapsulates the most important Ota functions and works with storage partitions.
- Added component `OtaNetwork` to add network support for the Ota functions.
- Sample Basic_rBoot is replaced with Basic_Ota. Direct use or rboot component inside your application is discouraged.
- HttpServer_FirmwareUpdate is updated to use the new Ota component.
- Sample OtaUpgradeMqtt::Upgrade is updated to use new Ota component.

> With a bit of work the OTA component could support also ESP32.

Related to #2328


Sming/Arch/Esp32/spiffs-two-roms.hw
```diff
@@ -0,0 +1,34 @@
+{
+	"name": "Two ROM slots with single SPIFFS",
+	"base_config": "spiffs",
+	"partitions": {
+	    "factory": {
+	       "address": "0x10000",
```
mikee47, line 6: Only

Sming/Libraries/OtaUpgradeMqtt/samples/Upgrade/app/application.cpp
```diff
@@ -113,15 +97,20 @@ void showInfo()
...
-	debug_d("Count: %d", conf.count);
-	debug_d("ROM 0: 0x%08x", conf.roms[0]);
-	debug_d("ROM 1: 0x%08x", conf.roms[1]);
-	debug_d("ROM 2: 0x%08x", conf.roms[2]);
-	debug_d("GPIO ROM: %d", conf.gpio_rom);
+	auto part = OtaManager.getRunningPartition();
 
-	Serial.printf(_F("\r\nCurrently running rom %d. Application version: %s\r\n"), conf.current_rom, APP_VERSION);
+	Serial.printf(_F("\r\nCurrently running %s: 0x%08x. Application version: %s\r\n"), part.name(), part.address(),
+				  APP_VERSION);
```
mikee47, line 113: `part.name.c_str()`

mikee47 (commented):
Much, much better!

slaff:
@mikee47 The OtaUpgrade library uses the RBOOT_xxx variables to read their location and address. But in ESP32 they are not present. 
```bash
	$(Q) $(OTATOOL) mkfile \
		$(OTA_CRYPTO_FEATURES_IMAGE) \
		$(if $(OTA_CRYPTO_FEATURES_IMAGE),--key=$(OTA_KEY_IMAGE)) \
		--rom=$(RBOOT_ROM_0_BIN)@$(RBOOT_ROM0_ADDR) \
		$(if $(RBOOT_ROM_1_BIN),--rom=$(RBOOT_ROM_1_BIN)@$(RBOOT_ROM1_ADDR)) \
```
Is there an easy to get them from the hardware configuration using one of the python scripts that come with the partitioning?  Can we somehow query the JSON hardware configuration without having to install aditional python library as https://github.com/h2non/jsonpath-ng for example?

mikee47:
@slaff Yes, there is an easy way! If you look in `out\Esp32\debug\hwconfig.mk` there are corresponding definitions which can be used without resorting to JSON. For example:

`RBOOT_ROM0_BIN` -> `PARTITION_rom0_FILENAME`

PR #2346, opened 2021-07-08
Added method to return a list of stations connected to an Access Point(AP)

slaff:


mikee47:
@slaff Have pushed some changes to your repo. for consideration

mikee47:
@slaff Problem with the new approach to StationList is lack of iterator support, so I've had another go at it. Main focus is to avoid those extra copies and allocations. Something similar for BssInfo would be good but that would break existing code which I think warrants leaving it well alone!

slaff:
@mikee47 Is the following the recommended way to iterate over the results?
```cpp
IpAddress ip; 
// ...
auto stations = WifiAccessPoint.getStations();
for(auto station = stations->begin(); station != stations->end(); station++) {
	if(station->ip() == ip) {
		// TODO: do something useful...
		break;
	}
}
```

mikee47:
I like auto:

```cpp
IpAddress ip; 
// ...
auto stations = WifiAccessPoint.getStations();
for(auto& station: *stations) {
	if(station.ip() == ip) {
		// TODO: do something useful...
		break;
	}
}
```

PR #2354, opened 2021-07-21
Feature: HttpServer plugins

slaff:
This PR adds plugin support to the HttpServer. 
The plugins allow implementation of things like Http Basic Authentication, IP restrictions and content decoding.
See `samples/HttpServer_Plugins/app/application.cpp` and `samples/HttpServer_Plugins/include/ContentDecoderPlugin.h` to get an idea how the plugins are working.


Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.h
```diff
@@ -0,0 +1,54 @@
...
+
+	bool addEvent(EventType type, EventCallback callback, int priority = 0);
+
+	const PriorityList<EventCallback>* getEvents(EventType type) const
+	{
+		return events[type];
+	}
+
+private:
+	const HttpResource* delegate = nullptr; // << once set the delegate is owned by this class
```
mikee47, line 52: `std::unique_ptr<HttpResource>`. Also `delegate` is perhaps confusing as it implies use of `Delegate` class instance.

samples/HttpServer_Plugins/include/ContentDecoderPlugin.h
```diff
@@ -0,0 +1,82 @@
...
+#pragma once
+
+#include <Network/Http/Resource/HttpResourcePlugin.h>
+#include <Data/WebHelpers/base64.h>
+
+namespace
+{
+constexpr char ENCODING_NAME[] = "test";
+}
+
```
mikee47, line 21: Don't use anonymous namespaces in headers!

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf21-dont-use-an-unnamed-anonymous-namespace-in-a-header

mikee47 (commented):
I've pushed some suggested changes to your repo.

Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+static constexpr char SKIP_HEADER[]{2, 1, 0};
+
+HttpEventedResource::HttpEventedResource(HttpResource* resource)
+{
+	originalResource.reset(resource);
+
+	// [ Register the main events with priority 0 ]
+	addEvent(
+		EventType::EVENT_URL, [this](HttpServerConnection& connection, char** at = nullptr, int* length = nullptr) {
```
mikee47, line 12: optional param values not necessary here (`= nullptr`)

Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.cpp
```diff
@@ -0,0 +1,126 @@
...
+				if(hasError) {
+					return false;
+				}
+			}
+
+			return true;
+		});
+
+	addEvent(EventType::EVENT_HEADERS,
+			 [this](HttpServerConnection& connection, char** at = nullptr, int* length = nullptr) {
```
mikee47, line 25: ... or here

Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.cpp
```diff
@@ -0,0 +1,126 @@
...
+					 if(hasError) {
+						 return false;
+					 }
+				 }
+
+				 return true;
+			 });
+
+	addEvent(EventType::EVENT_UPGRADE,
+			 [this](HttpServerConnection& connection, char** at = nullptr, int* length = nullptr) {
```
mikee47, line 38: ... here

Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.cpp
```diff
@@ -0,0 +1,126 @@
...
+					 if(hasError) {
+						 return false;
+					 }
+				 }
+
+				 return true;
+			 });
+
+	addEvent(EventType::EVENT_BODY,
+			 [this](HttpServerConnection& connection, char** at = nullptr, int* length = nullptr) {
```
mikee47, line 50: .. you get the idea ;-)

Sming/Components/Network/src/Network/Http/Resource/HttpEventedResource.cpp
slaff: @mikee47 Thank you Mikee!

Sming/Core/Data/PriorityList.h
```diff
@@ -0,0 +1,63 @@
...
+ * Created 2015 by Skurydin Alexey
+ * http://github.com/SmingHub/Sming
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ *
+ * @author: 2021 - Slavey Karadzhov <slaff@attachix.com>
+ *
+ ****/
+
+#include <Data/LinkedObjectList.h>
```
mikee47, line 12: `#include "LinkedObjectList.h"

mikee47 (commented):
I'm a little concerned about the over-use of delegates in this PR: a virtual event handling class would be simpler and consume less RAM. This class would contain methods corresponding to `EventType`. Possibly also replace `int` return value with a strong `enum class` instead - unless of course the actual value is important.

I'm happy to take a look at this approach if you agree?

slaff:
> I'm happy to take a look at this approach if you agree?

With pleasure. From this PR I will extract the changes related to the linked list classes and move  them to a separate PR that should be merged before this one.

slaff:
>> I'm happy to take a look at this approach if you agree?
> .. From this PR I will extract the changes related  ...

This PR was updated and ready for your changes :)

mikee47:
@slaff To help me better understand, can you give examples of possible uses where the plugin is called *after* the resource callback?

slaff:
> called after the resource callback?

Yes, a [HTML proxy plugin](https://httpd.apache.org/docs/2.4/mod/mod_proxy_html.html) would be a good candidate. Once the response is ready to be sent a Sming HTML proxy plugin will be attached after the resource.onRequestComplete call and will try to rewrite the URL links.

PR #2357, opened 2021-08-11
Asynchronous Graphics Library

mikee47:
An asynchronous graphics management library intended for SPI-based displays.

On first use several required python modules must be installed. Try building a sample application:

```
cd $SMING_HOME
make fetch Graphics
cd Libraries/Graphics/samples/Basic_Graphics
make python-requirements
make
```

Currently only the ILI9341 display has a driver, but you can use the virtual screen for testing.
This is a display device for use by the Host Emulator. The 'display' is a python application which communicates
with the Sming :cpp:class:`Graphics::Display::Virtual` display driver using a TCP socket.
Graphics processing is handled using SDL2.

To start the virtual screen server type ``make virtual-screen`` from your project directory.
The default TCP port is 7780. If you need to change this, use::

    make virtual-screen VSPORT=7780

The screen server's IP address is shown in the caption bar.
Build and run your project in host mode as follows::

    make SMING_ARCH=Host
    make run VSADDR=192.1.2.3


slaff:
@mikee47 Look at the failing CI tests. Some compilers require small code changes.

PR #2358, opened 2021-08-13
Update ESP32 to build with IDF v4.3

mikee47:
Currently ESP32 builds with IDF v4.1 which, amongst other things, only has beta support for ESP32S2, etc.
Version 4.3 is the latest current release build with support for esp32, esp32s2, esp32c3 and esp32s3 variants.

Note: Existing development systems must be updated to v4.3 as v4.1 will no longer work. The toolchains have also been updated.

TODO:

* [x] Fix final link stage issue - error is from esptool, updated to latest
* [x] Update installer scripts to use IDF v4.3. Consider how to handle patches
* [x] Get `Basic_Blink` to build on all variants
* [x] Get `HostTests` to build on all variants
* [x] Run `HostTests`, `Basic_IFS`, `Basic_Ssl`, `Basic_Serial` for esp32 and confirm no problems

## Installing v4.3 manually

Remove existing installation first.

```
rm -rf /opt/esp-idf /opt/esp32
```

Note that the updated installer scripts (when present in develop) can do this, though the removal step should probably be done manually.

```
export IDF_PATH=/opt/esp-idf
git clone -b release/v4.3 https://github.com/espressif/esp-idf.git $IDF_PATH
cd $IDF_PATH
git apply -v --ignore-whitespace --whitespace=nowarn $SMING_HOME/Arch/Esp32/Tools/idf.patch
git submodule update --recursive
python3 tools/idf_tools.py install
python3 -m pip install -r requirements.txt
```

Minor patches are applied. I've forked esp-idf to better track these changes - https://github.com/mikee47/esp-idf/tree/sming/4.3.

## Python package issues

During installation of IDF python requirements I encountered an error building `gwheel`. This was resolved by `sudo apt install python3.9-dev`.


slaff:
> Patch #1: Issue with portGET_ARGUMENT_COUNT() macro

@mikee47 Is this issue happening due to different compiler versions? Can we suggest this patch to the ESP-IDF team?

Is patch 2 due to windows only? Can we suggest it too to the ESP-IDF team?

mikee47:
> 
> > Patch #1: Issue with portGET_ARGUMENT_COUNT() macro
> 
> @mikee47 Is this issue happening due to different compiler versions? Can we suggest this patch to the ESP-IDF team?

@slaff Hi, this is a known issue - see https://github.com/espressif/esp-idf/pull/6692.

> 
> Is patch 2 due to windows only? Can we suggest it too to the ESP-IDF team?

I haven't tried Windows (Linux WSL2) but having re-checked this it seems OK. My repo may have been out of sync. If you haven't had any issues with this then we can dispense with it I think.

mikee47:
Current issues to resolve:

* If `ESP_VARIANT` value is changed (e.g. from default `esp32` to `esp32s2`) it is necessary to run `make sdk-menuconfig-clean` to recreate the default settings. This will overwrite any changes made via `make sdk-menuconfig` so isn't ideal.
* The 'legacy' timer hardware used by Sming is not supported on all variants. An alternative implementation is required (without using freeRTOS code, if at all possible).
* `flashmem_get_addrress` needs revising for unicore variants, code fails to compile as `DPORT_APP_FLASH_MMU_TABLE` not defined.

mikee47:
@slaff OK tested Basic_Ssl and SPI on regular ESP32. Builds on other variants but not tested yet.

mikee47:
Just tried Basic_IFS and getting `LoadProhibited` exception. It's probably the flash memory routines...

mikee47:
I've changed status back to `WIP` as HostTests shows polled timers are broken for esp32. This uses the hw_timer API so needs fixing. Also, basic stuff like `noInterrupts()` needs revising to use HAL. This can be verified by ensuring `HostTests` builds on all variants.

slaff:
> I've added an internal `SUBARCH_xxx` value for code use. Perhaps `ESP_VARIANT` should be renamed accordingly, i.e. `SMING_SUBARCH` ?

SUBARCH_xxx is ok.

mikee47:
@slaff OK, ready for review

Sming/Arch/Esp32/Components/esp32/component.mk
```diff
@@ -199,44 +225,73 @@ SDK_ESP_WIFI_LIBS := \
...
 
+ifeq ($(ESP_VARIANT),esp32)
+SDK_ESP_WIFI_LIBS += rtc
+endif
+
 SDK_NEWLIB_LIBS := \
 	c \
 	m  \
-	stdc++
+	stdc++ \
```
slaff, line 237: Is this `\` needed here?

Sming/Arch/Esp32/Tools/install.sh
```diff
@@ -31,18 +31,20 @@ $PKG_INSTALL ${PACKAGES[*]}
 if [ -d "$IDF_PATH" ]; then
     printf "\n\n** Skipping ESP-IDF clone: '$IDF_PATH' exists\n\n"
 else
-    git clone -b release/v4.1 https://github.com/espressif/esp-idf.git $IDF_PATH
+    git clone -b release/v4.3 https://github.com/espressif/esp-idf.git "$IDF_PATH"
```
slaff, line 34: @mikee47 How about cloning the content to `/opt/esp-idf-X` where X is the current version. Example `/opt/esp-idf-4.3` and after that making a symbolic link to `/opt/esp-idf` ?

mikee47: Good idea. I'll take a look at that, should simplify things as removal of v4.1 then becomes optional :-)
The toolchains aren't compatible either, so v4.3 needs the newer toolchains and v4,1 won't build with those. However, I've just spotted a `toolchain_versions.mk` file in the idf directory though so can use that to ensure the correct tools are used.

Sming/Arch/Esp32/Components/esp32/component.mk
mikee47: Nope, well spotted!

Sming/Arch/Esp32/Tools/install.sh
```diff
@@ -31,18 +31,20 @@ $PKG_INSTALL ${PACKAGES[*]}
...
 
-# Espressif downloads very slow, fetch from SmingTools
-mkdir -p $IDF_TOOLS_PATH
-ESPTOOLS=esp32-tools-linux-4.1.zip
-$WGET $SMINGTOOLS/$ESPTOOLS -O $DOWNLOADS/$ESPTOOLS
-unzip $DOWNLOADS/$ESPTOOLS -d $IDF_TOOLS_PATH/dist
+# Apply IDF patches
+IDF_PATCH="$(dirname $BASH_SOURCE)/idf.patch"
+pushd "$IDF_PATH"
+git apply -v --ignore-whitespace --whitespace=nowarn "$IDF_PATCH"
```
mikee47, line 41: @slaff We have a choice of either applying patches or cloning a forked repo. Forking would give more control, but if we go that route it should live in SmingHub.

mikee47: For example, cloning a `sming-4.3` branch means fixes can be applied using `git pull` rather than a re-install.

Sming/Arch/Esp32/build.mk
```diff
@@ -37,7 +44,8 @@ endef
 DEBUG_VARS			+= ESP32_COMPILER_PATH ESP32_ULP_PATH ESP32_OPENOCD_PATH ESP32_PYTHON_PATH
 
 ifndef ESP32_COMPILER_PATH
-ESP32_COMPILER_PATH	:= $(call FindTool,tools/$(ESP32_COMPILER_PREFIX),/$(ESP32_COMPILER_PREFIX))
+include $(IDF_PATH)/tools/toolchain_versions.mk
+ESP32_COMPILER_PATH	:= $(IDF_TOOLS_PATH)/tools/$(ESP32_COMPILER_PREFIX)/$(CURRENT_TOOLCHAIN_COMMIT_DESC)-$(CURRENT_TOOLCHAIN_GCC_VERSION)/$(ESP32_COMPILER_PREFIX)
```
mikee47, line 48: Note: Installing IDF 4.3 will break previous Sming builds. This patch would probably fix that.

mikee47:
> 
> 
> Note: Windows tests failing because it's just taking too long!

Do we need to build every sample for Windows? Perhaps a smaller subset is sufficient. (WSL2 offers a much better experience anyway.)

mikee47:
Patching the IDF is going to cause too many problems down the line so the installer clones from my fork. We can always change the repo at a later date. I've set up the following branches:

* **sming/dev/v4.3** Internal 'work in progress' branch
* **sming/release/v4.3** Used by installers and CI
* **sming/dev/master** If we want to try building against the latest ESP-IDF

Issue #1712, opened 2019-06-01
Configuration management


mikee47:
## Options, options, everywhere!

Sming is highly configurable with a lot of settings. You can see some of them by running `make list-config`, though that list is actually incomplete (e.g. SPI_* settings aren't in there).

I've opened this issue to discuss the problem in general terms so we can come up with a solid solution and make life easier for both end-users and developers. My proposal is that we adopt KConfig for this, with a graphical menu system to allow editing and adjustment to all relevant settings. Further, settings are fully defined and documented inside each code module which will reduce the need for additional documentation.

ESP-IDF uses a [kconfiglib fork](https://github.com/ulfalizer/Kconfiglib) to manage configuration data. Unlike the original kconfig it's cross-platform.

## Example

At the moment I edit `Makefile-Windows.mk` to set defaults like this:

```
SPI_SPEED = 40
SPI_MODE = dio
SPI_SIZE = 4M

COM_SPEED_SERIAL = 74880
COM_SPEED_ESPTOOL = 921600
```

So they apply globally to any projects. Not an ideal solution, but neither is it a good idea to set them all as global environment variables. In fact, the `SPI_*` settings above should really be stored as a group, identified by the target hardware (in this case, nodeMCU).

We need a structured way to manage all these settings, both to allow changing of global defaults and on a per-project basis. In addition, the system should accommodate pre-defined lists of configuration values, so we can allow easy selection of target hardware.

We could start by adding `make menuconfig`, which gets triggered after a fresh install of Sming so the user can set up their required default settings. Also, some settings (like the target hardware) won't have defaults so will trigger menuconfig also.

aemseemann:
Considering the recent discussion over at https://github.com/SmingHub/Sming/issues/1684, wouldn't  cmake-gui be a start?

![sming-cmake-gui](https://user-images.githubusercontent.com/51268362/58754093-de09fc00-84c9-11e9-9707-c8bb2019d857.png)

mikee47:
Thanks, that is helpful to see, and certainly a lot simpler than adopting a whole separate system for it.

Perhaps here we can discuss how Sming presents these options to the user, rather than how we actually implement them? It's great that CMake has an integrated way to deal with these, but for the time being at least we should be mindful of keeping both build systems broadly in sync.

mikee47:
For PR #2171 I propose that the information provided by some build variables (see below) is placed instead in a hardware configuration file attached to the project (and under revision control).
[...]

PR #2374, opened 2021-09-26
Add Kconfig support

mikee47:
This PR introduces the `menuconfig` build target as per Slaff's idea in https://github.com/SmingHub/Sming/issues/1712#issuecomment-869804339.

Sming configuration does not fully comply with Kconfig spec. For example, boolean values in Sming are typically '1', '0' or undefined but Kconfig requires them to be 'y' or 'n'. A small python program handles this conversion and merging Kconfig settings into the cached configuration data.

The current set of Kconfig files aren't complete but these can be added as required.


mikee47:
Got the rough bones of this working, but choices need sorting out. For example, cannot set the DEBUG_VERBOSE_LEVEL value directly but need to work out which of the appropriate `SELECT_DEBUG_n` values corresponds to the value.

mikee47:
OK, seems to be working.

mikee47:
@slaff Tried and failed to resolve "No whitespaces allowed surrounding docstring text (D210)". Also added fix so this works on Windows (paths issues). That should be all for now.

PR #2378, opened 2021-09-29
Fix Hosted operation for Windows

mikee47:
This PR updates the Hosted Component so it builds and runs for Windows and also Esp32 architecture.

I've done some basic testing with Hosted running on an ESP32-S2 target, which pretty much worked straight away (apart from `pulseIn` signature). However, as I use WSL if using a serial interface the Host application must be run in a Windows build. That needed a few fixes.

This also seems like a good time to get the Kconfig settings for this Component updated!

* Move `simpleRPC` into Components (Hosted is a Component so should not depend on a Library)
* Fix compilation problems for Windows
* Update environment variable handling & Kconfig settings
* Fix `pulseIn` signature for Esp32


slaff:
> I've left this WIP in case you'd like to add//modify this PR?

So far it looks good to me. Test it on your Windows and if it works for you then let's merge it.

PR #2385, opened 2021-10-09
Fix hwconfig option handling, add 1m, 2m hwconfig options and spiffs-2m.hw

mikee47:
Fix allows inherited configurations to re-apply options.

Current behaviour is to apply options only once, but expected behaviour is that the most recent application takes effect.

For example, 2m - 4m - 2m. Otherwise the second 2m is ignored and uses 4m instead.


slaff:
4.2.X LTS doesn't have hardware configuration features.

PR #2388, opened 2021-10-10
Rationalise compiler flags and esp32 improvements

mikee47:
Fix build inconsistencies and move common settings from arch. into main Sming to reduce code duplication.

- Rationalise compiler flags, add COMPONENT_CPPFLAGS
- Add SMING_C_STD to debug vars
- Get rid of obsolete Esp32 build settings (these related to pre-CMake builds)
- Add `Undef`, `UndefWrap`, `Wrap`, `DefSym` linker function
- Split Esp32 linker settings out. Use one file per esp32 component to simplify tracking settings and changes with IDF.
- Simplify meminfo, rename as `fwMeminfo.txt` and remove old/saved
- Wrap esp32 printf functions so they don't get pulled in from C library. Saves about 60K flash (and a little RAM).

### DIFF
--- a/Sming/Arch/Esp32/Components/esp32/sdk/newlib.mk
+++ b/Sming/Arch/Esp32/Components/esp32/sdk/newlib.mk
@@ -8,3 +8,6 @@ SDK_UNDEF_SYMBOLS += \
 	newlib_include_heap_impl \
 	newlib_include_syscalls_impl \
 	newlib_include_pthread_impl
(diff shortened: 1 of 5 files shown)


mikee47:
The codacy warning is about `strlen` - there are other instances of this in the framework. We could rewrite using `strnlen` instead but that increases code complexity and not sure it's warranted...

PR #2393, opened 2021-10-13
Fix some codacy advisories

mikee47:
Been looking through the logs trying to fixup some of these warnings.
Not sure about messing with some of those script files though - difficult to test!

Sure there's more to add here.


mikee47:
@slaff No, done for now. If you're happy with the changes to the shell scripts, not my area of expertise!

PR #2409, opened 2021-10-29
Update documentation for CLion, WSL and environment variable configuration

mikee47:
- Add note in docs re. CLion support
- Use `$(AWK)` not `awk` in makefile
- Add details for environment variable configuration
- Revise WSL notes


mikee47:
@slaff I'll have a few things to add here so changed to WIP

PR #2411, opened 2021-10-31
Fix minor warnings

mikee47:
- Fix Esp32 runtime warning calling `pinMode(INPUT)` where GPIO is input-only
- Fix build warnings
- Delete `OwnedLinkedObjectListTemplate` assignment operator (as for constructor)

### DIFF
--- a/Sming/Libraries/OtaUpgradeMqtt/src/StandardPayloadParser.cpp
+++ b/Sming/Libraries/OtaUpgradeMqtt/src/StandardPayloadParser.cpp
@@ -20,7 +20,9 @@ namespace Mqtt
 {
 bool StandardPayloadParser::switchRom(const UpdateState& updateState)
 {
+#if DEBUG_VERBOSE_LEVEL >= DBG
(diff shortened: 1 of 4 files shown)


Sming/Libraries/OtaUpgradeMqtt/src/StandardPayloadParser.cpp
```diff
@@ -20,7 +20,9 @@ namespace Mqtt
 {
 bool StandardPayloadParser::switchRom(const UpdateState& updateState)
 {
+#if DEBUG_VERBOSE_LEVEL >= DBG
```
slaff, line 23: This change makes the code fragile. May I suggest the following change instead:
```c++
auto after = OtaManager.getNextBootPartition();
#if DEBUG_VERBOSE_LEVEL >= DBG
   auto before = OtaManager.getBootPartition();
   debug_d("Swapping from %s @ 0x%08x to %s @ 0x%08x.\r\n", before.name().c_str(), before.address(),
			after.name().c_str(), after.address());
#endif
```

PR #2413, opened 2021-11-03
Feature: multipart upload passing filename and allowing to check file size.

slaff:
Make it possible to store an uploaded file with the provided in the form filename and limit the upload size to a predefined size.
To achieve this the developer needs to make a small change to its code and wrap the file stream in a `LimitedWriteStream`. The rest will be done behind the scenes from the MultipartParser component.
```c++
files["firmware"] = new LimitedWriteStream(MAX_FILE_SIZE, new FileStream());
```

In addition it is possible to add a more complex check using PartCheckerStream:
```c++
files["firmware"] = new PartCheckerStream(allowPart, new FileStream());
```

and  `allowPart` gives the control to the developer to decide whether a file should be saved or discarded:
```C++
/**
 * @brief This function is used to check if an upload file can be stored.
 *
 * @param headers - HTTP headers for the specific part
 * @param source - the original stream used to store data
 * @param part
 *
 * @retval bool false to reject the saving of the content
 */
bool allowPart(const HttpHeaders& headers, ReadWriteStream* stream, const PartCheckerStream::FilePart& part)
{
	// below is an example how to check for the filename length before storing it
	if(part.fileName.length() > 32) {
		uploadError = "Filename too long!";
		return false;
	}

	// If needed it is possible also to check if the file type is as expected
	// The code below requires the uploaded file to be plain text.
	if(part.mime != toString(MIME_TEXT)) {
		uploadError = "Only text files allowed!";
		return false;
	}

	// here is an example how to set the provided file name
	if(part.fileName.length() > 0) {
		auto fileStream = static_cast<FileStream*>(stream);
		fileStream->open(part.fileName, File::CreateNewAlways | File::WriteOnly);
	}

	return true;
}
```


Sming/Libraries/MultipartParser/src/PartCheckerStream.h
```diff
@@ -0,0 +1,79 @@
...
+	{
+		save = callback(headers, source, part);
+		return save;
+	}
+
+	bool isSuccess()
+	{
+		return save;
+	}
+
```
mikee47, line 50: Do we need another method here? Streams have an `isValid` virtual method which seems appropriate.

Sming/Wiring/WString.cpp
```diff
@@ -951,6 +951,28 @@ void String::trim(void)
...
+	while(*end == ch && end >= begin) {
+		end--;
+	}
+	len = end + 1 - begin;
+	if(begin > buf) {
+		memmove(buf, begin, len);
+	}
+	setlen(len);
+}
+
```
mikee47, line 975: I've suggested an alternative implementation which avoids duplication and is more flexible, operating on a set of characters.

mikee47 (commented):
The addition of more Stream types seems more complicated than I'd expect but cannot readily think of an alternative implementation... will chew it over! I've pushed some suggestions to your repo.

PR #2414, opened 2021-11-05
Allow applications to set the list of supported architectures.

slaff:
If not set all architectures are supported.

### DIFF
--- a/Sming/building.rst
+++ b/Sming/building.rst
@@ -61,6 +61,10 @@ These are the main variables you need to be aware of:
 
 These variables are available for application use:
 
+.. envvar:: PROJECT_ARCH
+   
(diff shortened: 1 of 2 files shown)


slaff:
@mikee47 how shall I proceed with this PR? Do you want to merge it in your `feature/board-config` branch or you already provide there a functionality that will replace this one?

mikee47:
Got a replacement in the works. Components need this setting also, plus Arch is too coarse. For example, a library may require USB hardware so esp32s2 OK but not esp32. Probably call it COMPONENT_SOC and build system will resolve dependencies, perform appropriate checks.

slaff:
> Got a replacement in the works.

Ok, I will close this PR and wait for your improvements.

PR #2418, opened 2021-11-09
Cs5460 fix

kmihaylov:
The user's app should take care of the SPI configuration. When the library does some very basic setting of the MISO pin, it might brake previous SPI configuration set in the user's app.
This PR extends the patch so that the library won't make any changes to the SPI/gpio's peripheral.


slaff:
@mikee47 are ok this PR?

mikee47:
@slaff Yep.

PR #2419, opened 2021-11-10
Feature: rbpf containers

slaff:
This PR adds support for rBPF in Sming. 
The rBPF subsystem provides a minimal virtual machine for microcontrollers.
rBPF is a virtual machine based on the popular Linux BPF virtual machine. 
It is small and fast enough to host multiple instances on architectures targeted by Sming.


mikee47:
Looks interesting... Definitely should move this into submodule though.

slaff:
> Looks interesting... Definitely should move this into submodule though.

I will :) Just need you to review it. And your reviews are sooooo good :)

mikee47:
I've having difficulty tracking down the origin of this source code. Can you provide a link?

slaff:
> Can you provide a link?

Sure, here it is: https://github.com/future-proof-iot/Femto-Container_tutorials . Take a look also at the component docs/ directory.

slaff:
... And here is a video presentation: https://www.youtube.com/watch?v=N6yVnuJlwIY

mikee47:
@slaff Pushed some suggestions to your repo to improve the build. You've called the library `Rbpf` but perhaps `rbpf` or preferably `rBPF` would be less confusing?

mikee47:
Don't think the build system is quite there yet. Should add a dependency for all container source files so if they're edited get rebuilt automatically.

mikee47:
Can you perhaps suggest some real-world practical application for rBPF? We could then focus on working up a more comprehensive sample.

slaff:
> We could then focus on working up a more comprehensive sample.

A more comprehensive example for Rbpf can be one which demonstrates the three possible ways to exchange information between the host and the guest application. So far we have example of using context to do so. The other two are using a key value store and binding. The latter being basically a function inside the VM that is communicating with the host.

>  Can you perhaps suggest some real-world practical application for rBPF

For the watch I would like to allow the use of multiple "applications" running in a sandbox. 
- One way to do it would be in a JavaScript VM (PR coming soon). JavaScript is widespread and has multiple advantages that will allow users to develop, debug and test their applications easily even in a browser before running them on a Sming based application.
-  Another way which boats to be  more memory and storage efficient, plus secure, is to use Rbpf as micro-controller VM.

slaff:
> You've called the library Rbpf but perhaps rbpf or preferably rBPF would be less confusing?

Ok, I will rename it to `rbpf`. Time to put it in a separate submodule...

PR #2420, opened 2021-11-10
OTA Deployer should exit when there is an error.

slaff:
Noticed when trying to do automatic releases for [SmingSwitch](https://github.com/slaff/SmingSwitch)


slaff:
The issue was that the deployer was not exiting. And the workflow was handing for ever. 
With this PR it is working as expected. These finished runs [1](https://github.com/slaff/SmingSwitch/runs/4164172003?check_suite_focus=true) and [2](https://github.com/slaff/SmingSwitch/runs/4163995148?check_suite_focus=true) confirm it.

PR #2421, opened 2021-11-10
Add SoC configurations and SMING_SOC setting

mikee47:
Each architecture supports one or more SoC variants. These can be seen with `make list-soc`.
This mainly affects Esp32 builds as there are currently three different SoCs.
The `SMING_SOC` setting replaces use of `ESP_VARIANT`.

In addition, all Components (including applications) may identify which SoCs are supported.
Attempting to build a sample for an unsupported SoC will get a meaingful error rather than build failure.

*Add SoC definitions*

- Each architecture provides a .json file for each SoC. Currently contains just the variant (e.g. esp32s2) and name (e.g. ESP32-S2).

*Add `COMPONENT_SOC` setting*

- Values are all lowercase (esp8266, esp32s2, hoist, etc.)
- Components list which SoCs are supported.
- Can use wildcard, e.g. `esp32*`. Default is `*` (all SoCs)
- Build system will check (with recursive dependencies) and fail with error if unsupported

*Add `SMING_SOC` setting*

- Used to identify SOC for build, finer granularity than SMING_ARCH
- Replaces use of `ESP_VARIANT`, although that is still used internally for Esp32 builds, no particular reason to remove it.
- Builds can use `SMING_ARCH` as usual, or `SMING_SOC` which also configures the appropriate ARCH.
- Allow `SMING_ARCH=Host` to override `SMING_SOC` setting for building utilities.

*CI revisions*

- Maximise coverage by building all supported samples for each SoC, one job per SoC.
- Added internal `sample` build target which aborts build without error if SoC is unsupported. Invoked by `make samples`
- Sample and test building revised so that any samples already built are skipped. This allows much faster re-testing if a sample fails during `make samples`.

*Code revisions*

- Set `DISABLE_NETWORK=1` for samples which don't use networking at all.
- Fixes and revisions as required to enable more samples to build on all architectures.
- Main restriction for Rp2040 is current lack of SPI and Networking layer (in progress).
- Move `fatfs` into Libraries

*Build revisions*

- Add sdk reconfigure step to re-run CMake when building `esp32` library for reliable builds when switching between library variants.

Todo:

* [x] Review/update documentation
* [x] Add some auto-generated information for each sample/Component page indicating supported SoCs.


mikee47:
@slaff Ready for review!

Sming/Components/Storage/src/include/Storage/Partition.h
```diff
@@ -267,7 +267,7 @@ class Partition
 	 * @brief Obtain partition size
 	 * @retval uint32_t Size in bytes
```
slaff, line 268: Update also the comment here.

slaff (commented):
Looks great to me! Only one small change needed.

mikee47:
@slaff Just noticed an issue with this. Running `make list-soc` doesn't give the right result because when it resolves dependencies there are arch-specific Components which restrict the values. I'll need to have a think about that one!

PR #2428, opened 2021-11-16
Fix `make list-soc`, rename `SUBARCH_xxx` as `SOC_xxx`

mikee47:
The only reliable way to do this is to run a make for each available SoC.

Change to build behaviour is that build type is not cached on clean.
This is consistent with other flags.

Build subtype was removed in #2421

Rename `SUBARCH_xxx` value to `SOC_xxx`, consistent with how `SMING_ARCH` becomes `ARCH_xxx` in code.


slaff:
@mikee47 maybe the detection should not fail when a toolchain for an arch is missing and this arch is not the selected SMING_ARCH. Here is what I am getting for `Basic_Blink`

```
Basic_Blink: Invoking 'list-soc' for Esp8266 (debug) architecture
SAVE esp8266
Available SoCs: esp32 esp32c3 esp32s2 esp32s3 esp8266 host rp2040
Project support:
- YES: esp32
- YES: esp32c3
- YES: esp32s2
- YES: esp32s3
- YES: esp8266
- YES: host
/tmp/Sming/Sming/Arch/Rp2040/build.mk:26: *** PICO_TOOLCHAIN_PATH not set correctly: /opt/esp-idf.  Stop.
make: *** [/tmp/Sming/Sming/project.mk:740: list-soc] Error 2
```

mikee47:
> @mikee47 maybe the detection should not fail when a toolchain for an arch is missing and this arch is not the selected SMING_ARCH. Here is what I am getting for `Basic_Blink`
> 
> ```
> Basic_Blink: Invoking 'list-soc' for Esp8266 (debug) architecture
> SAVE esp8266
> Available SoCs: esp32 esp32c3 esp32s2 esp32s3 esp8266 host rp2040
> Project support:
> - YES: esp32
> - YES: esp32c3
> - YES: esp32s2
> - YES: esp32s3
> - YES: esp8266
> - YES: host
> /tmp/Sming/Sming/Arch/Rp2040/build.mk:26: *** PICO_TOOLCHAIN_PATH not set correctly: /opt/esp-idf.  Stop.
> make: *** [/tmp/Sming/Sming/project.mk:740: list-soc] Error 2
> ```

@slaff Found a solution for this. Support is determined by running `make checksoc-print` for each SoC, using `MAKE_CLEAN=1` to bypass various redundant checks. Since we don't need SDKs at all for this we shouldn't need to include any architecture-specific `build.mk` files.

slaff (approved):

slaff:
@mikee47 I am still having problems with the latest SOC changes. Using the latest `develop` I am trying to compile Basic_Blink for Esp8266, then for Esp32 and finally for Host. Here are the steps to reproduce the problem: 
```
cd $SMING_HOME/../samples/Basic_Blink/$ 

# Compiles as expected
make SMING_ARCH=Esp8266 DEBUG_VERBOSE_LEVEL=3

# Fails
make SMING_ARCH=Esp32 DEBUG_VERBOSE_LEVEL=3
make: /opt/esp32/tools/xtensa-esp8266-elf/esp-2021r1-8.4.0/xtensa-esp8266-elf/bin/xtensa-esp8266-elf-gcc: Command not found

# Works...
make SMING_ARCH=Host DEBUG_VERBOSE_LEVEL=3 ENABLE_GDB=1
```

When I delete the `Basic_Blink/out` directory I can compile the sample for Esp32 but then trying to compile for Esp8266 fails with an error message that `esp8226` is not supported?!

```
cd $SMING_HOME/../samples/Basic_Blink/
rm -rf out/

# works 
make SMING_ARCH=Esp32 DEBUG_VERBOSE_LEVEL=3

# fails
make SMING_ARCH=Esp8266 DEBUG_VERBOSE_LEVEL=3
...
$SMING_HOME/project.mk:371: *** Project only supports: esp8266 host.  Stop. ?!

```
Can you try it on PC?

mikee47:
@slaff If you use `SMING_SOC=esp32` instead it will work.

The problem arises because the build system gets both `SMING_ARCH=Esp32` (from command-line) and `SMING_SOC=esp8266` (cached), so has to decide which one is valid. The logic for this (at the top of `build.mk`) uses SMING_SOC to determine SMING_ARCH.

What *should* happen is that SMING_SOC overrides SMING_ARCH. However, it should be possible to distinguish between a command-line parameter and a cached value so that needs to be added into the mix. I'll try to improve it!

PR #2432, opened 2021-11-23
Feature: JavaScript VM

slaff:
These libraries allow running JavaScript in a sandbox on all architectures supported by Sming. The actual JavaScript engine is [JerryScript](https://github.com/jerryscript-project/jerryscript).

This is port from the initial work done for the [U:Kit project](https://github.com/attachix/ukit).

Tasks:
--------
- [x] Copyright headers
- [x] Add documentation
- [x] Move to submodule

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -225,6 +225,10 @@
 	path = Sming/Libraries/HueEmulator
 	url = https://github.com/mikee47/HueEmulator
 	ignore = dirty
+[submodule "Libraries.jerryscript"]
+	path = Sming/Libraries/jerryscript
+	url = https://github.com/slaff/Sming-jerryscript.git
+	ignore = dirty
 [submodule "Libraries.IR"]
 	path = Sming/Libraries/IR
 	url = https://github.com/markszabo/IRremoteESP8266.git


mikee47 (commented):
I've had an initial play with this, pushed some changest to your repo.
Looks great! Need to move `jerryscript` and `jsvm` into submodules though.
Maybe jerryscript could be contained within jsvm?

mikee47:
I see there's a docker image to work with, but is there an easy way to compile .js into compatible snaps which might be integrated into the build system somehow?

slaff:
> but is there an easy way to compile .js into compatible snaps which might be integrated into the build system somehow?

Yes, there is. I will add further documentation and also improve the build system. 
In short: JerryScript has JavaScript parser, compiler to byte code (called snapshots) and vm that executes the snapshots. In order to save space the JerryScript library here is compiled ONLY to execute the snapshots. Parsing and compiling the snapshots can be done via a compiled C binary running on Host or directly in the browser (using Emscripten). This allows for example users to change the JavaScript code directly in their browsers, compile it again in the browser and upload it to the embedded device. In addition since the modern browsers understand JavaScript it should be doable to run and debug the code in a sort of simulator directly in the browser. I guess I have to add another sample where the JavaScript code can be edited and compiled in the browser.

mikee47:
> > but is there an easy way to compile .js into compatible snaps which might be integrated into the build system somehow?
> 
> Yes, there is....

@slaff Thanks for the clarification. I've had a play we can just use the standard jerryscript build tool to create the `jerry-snapshot` tool (it works for both Linux and Windows). If you like I'll try integrating that so the sample can use it to build the snapshot automatically.

slaff:
> If you like I'll try integrating that so the sample can use it to build the snapshot automatically.

This would be awesome :)

Sming/Libraries/jerryscript/JerryScriptUpdate.rst
```diff
@@ -0,0 +1,47 @@
+JerryScript Update
+==================
+
```
mikee47, line 3: `.. highlight:: bash`

mikee47:
@slaff Can you create a submodule for this? I'll review then.

PR #2442, opened 2021-12-07
CI installer and build improvements

mikee47:
This PR adds documentation and support for Hyper-V BYOC installation.

**Install changes**

- Windows `install.cmd` moved to `Tools/ci` consistent with `install.sh`.
- Reduce console noise during installs
    - Messages fill up logs and make diagnosing problems impossible
    - Don't show download progress in choco install
    - Use non-interactive mode for idf installs
    - wget not verbose
- Document Hyper-V / Docker setup for BYOC
- Update esp-idf submodules during byoc image setup
- Return script error if install fails

**CI changes**

Add SmingTest CI integration

- Test status/result submitted to appveyor (Note: Tests listed unsorted, see https://github.com/appveyor/ci/labels/tests)
- Allow all tests to run before checking for failure

Build documentation in separate job

Can't rely on APPVEYOR_BUILD_WORKER_CLOUD being set, as BYOC can be enabled just by remapping images.

- Allow failure during setup of TAP network for testing (yes, would like to get this working for BYOC!)
- Check for presence of rp2040 toolchain - if that's installed we can safely assume everything else is

- Ensure FlashString integration tests run on all architectures (i.e. rp2040 w/o networking).

**Build changes**

- Improve error if unknown SOC specified. e.g. `make SMING_SOC=smart-banana` says 'SOC unknown' rather than '' bad ARCH
- Fix `tests-clean` (typo)


mikee47:
Now does full build without error https://ci.appveyor.com/project/mikee47/sming/builds/41801613
A good test for the installers since the source Windows/Linux images are pretty bare, typically just git and python installed.

Sming/Arch/Host/Tools/ci/build.setup.sh
```diff
@@ -21,12 +21,9 @@ if [ -n "$SMING_SECRET" ]; then
...
-fi
-
-# Build documentation
-make -C "$SMING_HOME" docs
+# Setup networking
+set +e
+sudo ip tuntap add dev tap0 mode tap user $(whoami)
+sudo ip a a dev tap0 192.168.13.1/24
+sudo ip link set tap0 up
+sset -e
```
slaff, line 29: Did you mean `set ...` instead of `sset ...` ?

PR #2453, opened 2021-12-21
Fix documentation, including library sample links

mikee47:
This PR addresses some issues raised in the documentation build log, such as missing references, ill-formed tables, etc.

Some documentation sections have been extended.

`:sample:` links in libraries have been fixed. It was previously assumed that all samples were located in the main 'samples' directory, so links within library documentation to contained samples did not work. They're now a bit smarter so will work logically when used within Components. They can also direct to a specific Component such as ``:sample:`CS5460/generic` ``.

`:source:` links have also been fixed so they can be used within Components. Links starting with '/' refer to Sming tree, otherwise relative to Component root.

See https://smingdev.readthedocs.io/en/fix-documentation/ for test documentation build.

todo

* [x] Fix incorrect use of doxygen `@name` tag


slaff:
@mikee47 Wow, so far this PR looks really good! Thank you Mikee!

Issue #1428, opened 2018-08-10
SDCard only with software SPI?


tius2000:
There might be problem with hardware SPI and SDCard. The SDCard sample compiles fine, but does only work (slow) when I replace 

`SDCardSPI = new SPIClass();`

with

`SDCardSPI = new SPISoft(PIN_CARD_DO, PIN_CARD_DI, PIN_CARD_CK, 0);`

The error message (with some additional debugging enabled) is:

```
SPIClass::setClock(prediv 10, cntdiv 2) for target 4000000
disk_initialize (send 80 0xFF cycles)
disk_initialize (send n send_cmd(CMD0, 0)
disk_initialize (until n = 5 && ret != 1
SDCard ERROR: ff
SDCard init FAIL
```

Could that be hardware related? I checked it with two different cards and cardreaders.

slaff:
@ADiea you are our SDCard expert. Can you help here?

ADiea:
Hi sorry for late reply I think software sp[i differs from hardware arduino class.
If I remember correctly the main difference is related to who sets the chip select i guess hardware spi relies on the user to select the chip.  The SDCard class needs to to chip select by itself you should investigate if the chip select is done correctly using the hardware spi class.

harry-boe:
Try using lower speeds for HW SPI.

We used the samples in a training class and figured out that only 4 out of 10 boards did work.
The issue was that it became unstable with higher speeds than approx. 20 MHz.

Change the initializer (in the Sample) and try again

```
// overwrite default SPI Speed
	SDCardSPI->beginTransaction(SPISettings(40000000, MSBFIRST, SPI_MODE0));

// Change to 

	SDCardSPI->beginTransaction(SPISettings(20000000, MSBFIRST, SPI_MODE0));
```

About the comment above. True you cant use the HW chip select. This is due to the fact that the initializer sequence of the SD card hat to have that signal set for some time - without sending date.

Unfortunately, you can't do that when using HW SPI

mopac:
I was trying to get SDCard sample running on Wemos D1 mini with an added SD card.
Same fault as the original post.
When I got down to analysing the SPI packets send to the SDcard, I discovered the byte order was incorrect.
CMD0 is meant to be 40 00 00 00 00 95.
What was being sent was 00 00 00 40 95 00.
Seems the SPI was shifting the bytes in the wrong order on a 32 bit boundary.
[...]

PR #2462, opened 2021-12-31
Overhaul SPI classes, replacing `byteOrder` parameter with `bitOrder`

mikee47:
This PR fixes the `SPISettings` parameters to be consistent with Arduino and related libraries, which all expect a `bitOrder` parameter. Byte order is always LSB first, corresponding with system endianness.

ESP8266 and ESP32 code simplified and refactored.

An SPI test module has also been added to HostTests to perform basic checks in loopback mode. Fixes arising:

- Fix Esp32 + Esp8266 misaligned reads
- Fix Esp32 + Esp8266 `set_mode`
- Fix Esp8266 clock register setting
- Fix `transfer32` method so it works correctly with all bit lengths in both LS/MS bit order
- Revise SPISoft to support modes and bit order

See https://github.com/SmingHub/Sming/issues/1428#issuecomment-1003334483

Verification tests:

For ESP8266, ESP32, ESP32S2, ESP32C3 and Software SPI:

* [x] Check bit ordering is correct using 'scope
* [x] Check SPI mode produces expected output
* [x] Test `ScreenTFT_ILI9340-ILI9341` sample
* [x] Test `SDCard` sample


mikee47:
@slaff OK, ready for review.

PR #2470, opened 2022-01-20
Upgrade HardwareSPI library with Rp2040 support

mikee47:
This PR adds support for Rp2040 to the HardwareSPI and Graphics libraries.

Host support is also improved to allow integration testing.
This includes revisions to thread/interrupt handling to more closely emulate real hardware behaviour.

## HardwareSPI

- Refactor using architecture-specific `ControllerBase` class to simplify/clarify Controller.h.

- Replace `maxTransactionSize` with `sizeAlign`

    Required only where requests have to be split up due to size.

    For example, transferring 4K of display pixel data for ESP8266 will require splitting into transactions of 64 bytes or less.
    The data must be transferred strictly in groups of 3 bytes (R, G, B) so `sizeAlign=3` ensures this happens.

    ESP32 has ~4K DMA capability, so still must be considered.
    RP2040 has no DMA size restriction so can be simplified.

- Fix Host Controller sequencing

    A complex application (e.g. Basic_Graphics) should run without stalling, race conditions or exceptions.

### Improvements

- Add version header

- Add loopback method to Controller

- Copy pin setup from SPI library

    Provides some consistency between the two libraries.

- Supported device IO mode also depends on controller capability

    e.g. Device may support quad modes but controller may not,
    so `Device::isSupported` method returns intersection of device/controller capabilities.

- Add Rp2040 implementation

    DMA transaction sizes are unrestricted which simplifies code, offset by lack of cmd/address/dummy
    phase support which complicates it.

    Hardware CS behaviour is weird. It gets asserted for every word in mode 0 or 2 (but not 1 or 3).
    This breaks operation with some devices, e.g. ILI9341.
    Also, there's no 'request complete' interrupt available.

    This is solved by using standard GPIO for CS control and completing request on RX DMA interrupt.

- Add test application

    Provides a vehicle for testing basic operation and performing integration testing.

    Improved Host interrupt support in framework allows inclusion in CI testing,
    inserting calculated delay (based on number of bits being transferred) before firing interrupt.

## Graphics

- Add Rp2040 support

- Remove extraneous file added in error - duplicate of file in resource/fonts/Linux/

- Fix issue with `showFonts` in Basic_Graphics sample

    Gets stuck second time around the loop

- Update Basic_Graphics sample, use for integration tests

    Simplify font functions.
    Add build variables to configure application to run more quickly so it can be used in CI testing.

    Host HardwareSPI may not do much but it allows the control flow to be verified and checked with valgrind, etc.

- Fix access violation rendering fonts

    Detected in Windows CI run.

- Don't assume location of Graphics library

    Running CI on standalone library means it won't be in the Sming repo.

## Framework

- Fix flashmem read

    Must claim DMA channel so it doesn't get given to HardwareSPI.

- Improve CThread interrupt handling

    Make `host_printf`, etc. threadsafe using `write` instead of buffered `printf` calls.

    Honour interrupt levels, masking out lower-level threads during interrupt

    Main thread synchronisation fixed. Mitigate signal deadlock using timer.

    Note: Running in valgrind may encounter 'undefined syscall' errors (407, 409)
    due to missing wrappers for 64-bit time functions.
    Valgrind version 3.17 has the wrappers (available in Ubuntu 21.04).

- Add CI test notification support

    Graphics library test uses a sample application rather than SmingTest,
    so notifications are handled in a makefile.

- Update Graphics & HardwareSPI libraries

    RP2040 support added.

- Use 64-bit python

    Freetype python library doesn't work on Windows as DLL isn't installed for 32-bit python.
    This only happens with standard appveyor system as `C:\Python39` contains 32-bit python installation.

### DIFF
--- a/Sming/Arch/Host/Components/driver/hw_timer.cpp
+++ b/Sming/Arch/Host/Components/driver/hw_timer.cpp
@@ -13,11 +13,12 @@
 #include <driver/hw_timer.h>
 #include <hostlib/threads.h>
 #include <sys/time.h>
-#include <errno.h>
+#include <cerrno>
 #include <hostlib/hostmsg.h>
 #include <cassert>
 #include <muldiv.h>
 #include <esp_system.h>
+#include <memory>
 
 /* Timer 1 */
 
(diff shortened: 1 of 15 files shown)


slaff:
@mikee47 are you ready with this PR or would like to add something more? I have tested only the host emulator and so far looks good to me. Were you able to test the changes on real Rp2040?

mikee47:
@slaff Yes, all done. Tested on real Rp2040, runs great.

PR #2472, opened 2022-01-24
Reduce Host emulator CPU load

mikee47:
This PR improves Host performance by using efficient waiting for tasks & timers rather than just looping.
Running the `HttpServer_ConfigNetwork` example sees a drop from 100% CPU usage to < 2% running in linux. Difference in Windows is less pronounced as the lwip layer incorporates a 1ms wait.

LWIP is now serviced using an event timer. The interval is adjusted dynamically which allows good file transfer performance and reduced CPU usage when idle.


Sming/Arch/Host/Components/esp_hal/tasks.cpp
```diff
@@ -80,7 +85,13 @@ bool system_os_post(uint8_t prio, os_signal_t sig, os_param_t par)
 		return false;
 	}
 
-	return task_queues[prio]->post(sig, par);
+	if(!task_queues[prio]->post(sig, par)) {
+		return false;
+	}
+
+	extern CSemaphore host_main_loop_semaphore;
```
slaff, line 92: @mikee47  Can you move this line to the beginning of the file?

Sming/Arch/Host/Components/hostlib/startup.cpp
```diff
@@ -271,6 +271,7 @@ int main(int argc, char* argv[])
 		return 1;
 	}
 
+	extern void host_init_bootloader();
```
slaff, line 274: @mikee47  Can you move this line to the beginning of the file or a separate header file?

Sming/Arch/Host/Components/hostlib/startup.cpp
```diff
@@ -291,8 +292,8 @@ int main(int argc, char* argv[])
 
 #ifndef DISABLE_NETWORK
 		if(config.enable_network) {
-			lwip_initialised = host_lwip_init(&config.lwip);
-			if(lwip_initialised) {
+			if(host_lwip_init(config.lwip)) {
+				extern void host_wifi_lwip_init_complete();
```
slaff, line 296: @mikee47  Can you move this line to the beginning of the file or a separate header file?

Sming/Arch/Host/Components/esp_hal/tasks.cpp
mikee47: I'll need to rewrite this using functions anyway.
As you know from CI the Windows run keeps stalling. It happens because the Windows pthreads library uses a mutex to guard semaphores. If that mutex is locked when the main thread is suspended, the next call to sem_post() results in a deadlock. Oh, the joys of multithreading.

slaff:
@mikee47 this PR looks much better now. Shall I merge it or you would like to add something?

mikee47:
@slaff All done thanks

PR #2474, opened 2022-01-26
Implement Smart Config for ESP32

### LINKED ISSUE #2473: Esp32 SmartConfig
I forgot that Esp32 lacks support for SmartConfig. Any directions would be helpful! 
The current compiler messages are only related to ```Components/Network/Arch/Esp32/Platform/StationImpl.cpp``` and ```StationImpl.h```. I'm starting digging up.

mikee47:
Re. #2473 here are the bones of a smart config implementation for ESP32.

TODO:

* [x] Fix missing `__wrap_putchar` error (`libsc_get_ssid_passwd.a`)
* [x] Test

Closes #2473


kmihaylov:
Ah I missed it :)
https://github.com/kmihaylov/Sming/commit/94dc8faa06805e63d82c857053372026dc0ca919
I got lost with std::unique_ptr so I took a rest and here you already made it... 

Anyway, I'll be glad to tell me how Sming includes the esp-idf header files?

mikee47:
The `Basic_SmartConfig` sample kind of works, with a couple of provisos:

Smart config won't work if station has already been configured *and* set to auto-connect on startup.
Error is `STA is connecting, scan are not allowed!`.
This is perfectly logical and probably something the applicatdion should address rather than the framework.
i.e. need to call `WifiStation.disconnect()` first.

Attempting to configure with wrong credentials isn't really handled. Again, that's something the application needs to consider. I took a quick look at ESP8266 behaviour and same deal.

mikee47:
Note: The `__wrap_putchar` doesn't seem to get called, no idea what it's for...

kmihaylov:
Works well for me. Both the example and in my project.
Great work @mikee47 !

mikee47:
Refinement to the event handling now blocks auto-connect if WPS or SmartConfig are active, so the `Basic_SmartConfig` sample works even if already configured.

PR #2475, opened 2022-02-02
Fix issue with coding style fixer failing when list of files is too long

slaff:
At least fails for me on a stock Linux system. 

Excerpt from `xargs --show-limits`
```
POSIX upper limit on argument length (this system): 16772719
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 16770270
```

### DIFF
--- a/Sming/Makefile
+++ b/Sming/Makefile
@@ -215,7 +215,7 @@ CS_SEARCH_DIRS	?= $(call ListAllSubDirs,$(CS_ROOT_DIRS))
 # Resultant set of directories whose contents to apply coding style to
 CS_DIRS			?= $(patsubst %/.cs,%,$(wildcard $(foreach d,$(CS_SEARCH_DIRS),$d/.cs)))
 # Files to apply coding style to
-CS_FILES		= $(if $(CS_DIRS),$(call rwildcard,$(CS_DIRS:=/*),%.cpp %.hpp %.h %.c),)
+CS_FILES		= $(if $(CS_DIRS),$(sort $(call rwildcard,$(CS_DIRS:=/*),%.cpp %.hpp %.h %.c)))
 
 .PHONY: cs
 cs: ##Apply coding style to all core files
--- a/Sming/build.mk
+++ b/Sming/build.mk
@@ -281,9 +281,7 @@ endef
 # $1 -> List of files
 define ClangFormat
 	$(if $(V),$(info Applying coding style to $(words $1) files ...))
-	@for FILE in $1; do \
-		$(CLANG_FORMAT) -i -style=file $$FILE; \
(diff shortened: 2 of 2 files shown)


mikee47:
Is there a performance hit with this?

slaff:
> Is there a performance hit with this?

Yes, for every file in the list a new shell will be opened. But the delay is less than 30 seconds (at least on my Ubuntu box). Other possibility would be to split the big list of files and to pass 20 files at a time to the `clang-format` app. I just don't have enough makefile `fu` skills to do it.

mikee47:
> > Is there a performance hit with this?
> 
> Yes, for every file in the list a new shell will be opened. But the delay is less than 30 seconds (at least on my Ubuntu box). Other possibility would be to split the big list of files and to pass 20 files at a time to the `clang-format` app. I just don't have enough makefile `fu` skills to do it.

Try this... should be a teeny bit quicker :-)

slaff:
> > > Is there a performance hit with this?
> > 
> > 
> > Yes, for every file in the list a new shell will be opened. But the delay is less than 30 seconds (at least on my Ubuntu box). Other possibility would be to split the big list of files and to pass 20 files at a time to the `clang-format` app. I just don't have enough makefile `fu` skills to do it.
> 
> Try this... should be a teeny bit quicker :-)

Wonderful. It is indeed much faster. Here are the results on my machine: 

Before:

```sh
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs
make: Nothing to be done for 'cs'.

real	0m24,658s
user	0m12,166s
sys	0m8,524s
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs
make: Nothing to be done for 'cs'.

real	0m20,467s
user	0m11,879s
sys	0m8,286s
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs
make: Nothing to be done for 'cs'.

real	0m20,843s
user	0m12,052s
sys	0m8,366s
slaff@computer:/media/sla....box/dev/Sming/Sming/$
```

After:

```
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs

real	0m5,611s
user	0m4,748s
sys	0m0,757s
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs

real	0m5,609s
user	0m4,775s
sys	0m0,746s
slaff@computer:/media/sla....box/dev/Sming/Sming/$ time make cs

real	0m5,597s
user	0m4,669s
sys	0m0,844s
```

@mikee47 I will merge this PR if you don't have other optimizations in mind.

mikee47:
@slaff OK, all done now.

PR #2493, opened 2022-02-18
Docs: Fix issue with updated MarkupSafe.

slaff:


mikee47:
I've just updated jinja2=3.0.3 and markupsafe=2.1.0 and docs build locally without issue... Taking a deeper look...

docs/requirements.txt
```diff
@@ -3,6 +3,7 @@
 sphinx==4.2.0
 sphinx-rtd-theme==1.0.0
 m2r2==0.3.1
+MarkupSafe==2.0.1 # Version 2.1.0 is not yet compatible with jinja2
```
mikee47, line 6: I think setting `jinja2>=3.03` instead is preferable - that version's been out since november last year and 2.11 is quite old :-)

mikee47:
NB. I cancelled appveyor as it's not necessary and wasteful!

docs/requirements.txt
mikee47: Works https://smingdev.readthedocs.io/en/fix-doc-generation/

Issue #2494, opened 2022-02-22
Running Websocket_Client sample causes `HTTP parser error: HPE_CB_headers_complete`


gbaranski:
I'm using sample from https://github.com/SmingHub/Sming/tree/develop/samples/Websocket_Client.
Running on ESP8266 Wemos D1 Mini Board.

slaff:
Ok, I see the problem. The websocket server that we use in the sample `echo.websocket.org` is no longer publicly available. Take a look at application.cpp:

https://github.com/SmingHub/Sming/blob/develop/samples/Websocket_Client/app/application.cpp#L39

and replace the websocket server URL with your own websocket server.

gbaranski:
alright that works now. Don't you think that those errors should be more readable to the user?

mikee47:
> alright that works now. Don't you think that those errors should be more readable to the user?

Agreed. I've had a dig into exactly what's going on here using the _**Simple Websocket Client**_ plugin for firefox. These are the response headers:

```
HTTP/1.1 200 OK
x-amz-id-2: aDP+2nX5soleb7+lPTk6c8QMuXDvTCWpToz6IMrI9rgsw6VyWJUmQ0+kmcS9NdkbqH4OB8zZeoY=
x-amz-request-id: 2C8VTRTYZA541D4K
Date: Sun, 27 Feb 2022 13:35:37 GMT
Last-Modified: Thu, 26 Aug 2021 03:35:42 GMT
ETag: "6d5a28f4c5e297e95a29988884c9e896"
Content-Type: text/html
Server: AmazonS3
Content-Length: 4413
Connection: close
```

So no error, just that the `Sec-WebSocket-Accept` header isn't present. Running the sample application in host mode with debugger locates the point of failure and I've opened a PR so we can have  a look at improving this.

In this particular instance the exact cause is only obvious by visiting `http://echo.websocket.org`...

PR #2496, opened 2022-02-27
[WIP] Add websocket failure debug messages

mikee47:
If a remote server sends a normal HTTP response without appropriate websocket headers then websocket open requests fail with `HPE_CB_headers_complete` error. This PR is intended to look at this, initially by adding an additional debug message at the point of failure to clarify what's happened.

How applications should handle this situation is perhaps more complicated.

See #2494 for discussion.

### DIFF
--- a/Sming/Components/Network/src/Network/WebsocketClient.cpp
+++ b/Sming/Components/Network/src/Network/WebsocketClient.cpp
@@ -77,6 +77,7 @@ bool WebsocketClient::connect(const Url& url)
 int WebsocketClient::verifyKey(HttpConnection& connection, HttpResponse& response)
 {
 	if(!response.headers.contains(HTTP_HEADER_SEC_WEBSOCKET_ACCEPT)) {
+		debug_e("[WS] Websocket Accept missing from headers");
 		state = eWSCS_Closed;
 		return -2; // we don't have response.
 	}


slaff:
@mikee47 Do you want to add something more to this PR or we can leave it for another PR?

mikee47:
The debug message is helpful for development, but I think the issue here is that the call to `wsClient.connect()` never gets a response. We need a failure callback, so related to #1939.

slaff:
Let's do the following
- Merge this PR as is

And then
- Create a second PR which adds a second optional parameter to the `bool connect(const Url& url);`  -> `bool connect(const Url& url, TcpClientConnectCallback callback = nullptr);` methods for all [TCP clients](https://sming.readthedocs.io/en/latest/api/group__tcpclient.html).
- A `TcpClientConnectCallback` has one parameter which is the id of the connection error.
- A connection error can be:
     -  Invalid parameters - invalid URL scheme
     -  DNS error - domain not resolvable
     - TCP connection error - no IP and port to connect to
     - Protocol error - when the server side does not provide the expected response. For example when a websocket client tries to connect to a server and the server does not speak HTTP or speaks HTTP but does not provide the required headers

What do you say?

mikee47:
I see what you mean, though only need callback for 'deferred' errors since invalid parameters, etc. will result in the `connect()` call failing immediately with a `false` return value.

The sample code is generally written with a call to `connect()` (without checking the return value) then immediately calling `send()`. If the connect fails, then it's not clear what happens with the queued request(s). A more logical approach might be:

```
auto connectionCallback = [&](bool success) {
  if(succcess) {
    client.send(...);
  }
};

if(!client.connect(url, connectionCallback)) {
  debug_e("Connect failed!");
}
```

So your approach of a per-connect callback is good but we can take the opportunity to provide a more logical and consistent way to use the framework.

In general, the framework doesn't both with error codes, just success/failure since the resolution is typically just to retry later and knowing the exact cause probably won't help much. Protocol clients could provide additional methods if it's useful.

NB. I've done some work on the websocket stuff and updated #1939 so may be of use.

slaff:
Closing in favor of #1939.

PR #2499, opened 2022-04-13
Feature: Bluetooth Low Energy (BLE) support for Esp32 

slaff:


mikee47:
@slaff Can reproduce failing CI locally by removing `$SMING_HOME/Arch/Esp32/Components/esp32/projects` then rebuilding.
Because they're inter-related the bluetooth stuff needs to go into the Esp32 Component itself, probably controlled with `ENABLE_BLUETOOTH`. I'm working on this ATM and will report later today!

Issue #2500, opened 2022-04-13
Properly System.restart() esp32


kmihaylov:
Hello. When ESP utilizes WiFi it couldn't properly disconnect or restart the device.
```
E (18626) task_wdt: Task watchdog got triggered. The following tasks did not reset the watchdog in time:
E (18626) task_wdt:  - Sming (CPU 1)
E (18626) task_wdt: Tasks currently running:
E (18626) task_wdt: CPU 0: IDLE
E (18626) task_wdt: CPU 1: IDLE
E (18626) task_wdt: Aborting.

abort() was called at PC 0x400de9fc on core 0

Backtrace:0x40087b6a:0x3ffb0580 0x400882b5:0x3ffb05a0 0x4008e466:0x3ffb05c0 0x400de9fc:0x3ffb0630 0x40081d49:0x3ffb0650 0x4012fae7:0x3ffbaee0 0x400de80a:0x3ffbaf00 0x40089408:0x3ffbaf20


ELF file SHA256: 5e395847006b60a9

Rebooting...
```

I use the Basic_WiFi example and I've added ```System.restart(10000);``` at the end of the ```init()``` function.

Also the same applies if I try to use ```Wifi.disconnect();``` or ```WifiStation.enable(false);```.

Asanga-Viraj:
I saw the same issue in ESP32. `System.restart()` stuck in infinite loop and do Watchdog reset. Watchdog reset doesn't start the ESP as fresh start. It makes a core synching issue and keeps restarting. The problem origin is [idf-v4.3](https://github.com/espressif/esp-idf/tree/v4.3) from expressif. I hope new [idf-4.4.1](https://github.com/espressif/esp-idf/compare/v4.3...v4.4.1) version may not have issue like this. As temporarily, I connected a wire from digital output to EN to make reset (ugliest way to do it).

Methods of the WiFi are not working. It is a huge problem. I tried things in many ways. But they kept ESP32 making Watchdog reset.

kmihaylov:
Hm, some answers here are missing.

Anyway, I'm watching esp-idf's function ```esp_wifi_stop()```. How do I properly call it from within Sming?

kmihaylov:
I'm answering myself. the ```esp_wifi_stop()``` and ```esp_wifi_deinit()``` functions can easily be accessed once the user adds the ```#include <esp_wifi.h>``` header to the application.cpp file.

[...]

PR #2543, opened 2022-08-24
Fix ESP32 system restart problems

mikee47:
See #2500 for discussion.


mikee47:
The cause of the WDT is because the call to `esp_wifi_stop()` hangs waiting on a queued message to be handled. Because the call originates from the same task which handles the message queue, a deadlock occurs. That means a regular IDF application would behave the same way if `esp_wifi_stop()` were called from within an event handler (not yet confirmed, but easy enough to test).

The problem can be avoided by calling `esp_restart()` in the context of another task:

```
esp_ipc_call(0, esp_ipc_func_t(esp_restart), nullptr);
```

The following demonstrates that the cause is indeed in `esp_wifi_stop` as it too succeeds:

```
esp_ipc_call(0, esp_ipc_func_t(esp_wifi_stop), nullptr);
esp_restart();
```

slaff:
@kmihaylov can you test if this change is fixing the restart issue for you?

kmihaylov:
Sure, just need to do some daily tasks and I'll try it in the afternoon.

kmihaylov:
It looks like Mike's patch is working!

I'm updating Sming this way:
```
git fetch origin pull/2543/head:pr/2543
git checkout pr/2543
make dist-clean
make Sming-clean
```

Here is the code:
```
#include <SmingCore.h>
#include "esp_wifi.h"
#include <esp_ipc.h>

Timer testTimer;

void wifiInit() {
	esp_wifi_deinit();
	esp_netif_create_default_wifi_sta();
	wifi_init_config_t cfg = WIFI_INIT_CONFIG_DEFAULT();
	ESP_ERROR_CHECK(esp_wifi_init(&cfg));
	esp_wifi_set_mode(WIFI_MODE_STA);
	wifi_config_t wifiCfg;
	esp_wifi_set_config(WIFI_IF_STA, &wifiCfg);
	esp_wifi_start();
}

void shutdown() {
	//esp_ipc_call(0, esp_ipc_func_t(esp_wifi_stop), nullptr);
	//esp_wifi_stop();
	System.restart();
}

void init()
{
	wifiInit();
	testTimer.initializeMs(5000, shutdown).startOnce();
}
```

This code successfully restarts the app. If I use the ```esp_wifi_stop()``` then WDT occurs. Calling it in the context of another task (alone; w/o ```System.restart()```) doesn't trigger the WDT. Calling both ```esp_wifi_stop()``` (in another task) and then ```System.restart()``` triggers WDT, but I think this might be expected...?

mikee47:
Note that with IDF 4.4 IPC is no longer available for single-core SOCs, so instead the call can be handled by the `esp_timer` task (see #2371).

```
const esp_timer_create_args_t create_args = {
  .callback = esp_timer_cb_t(esp_restart),
  .dispatch_method = ESP_TIMER_TASK,
};
esp_timer_handle_t handle{nullptr};
esp_timer_create(&create_args, &handle);
esp_timer_start_once(handle, 100);
```

PR #2551, opened 2022-09-09
Add basic C++ output streaming support

mikee47:
Add some improvements to the `Print` class, with support from new `String` methods.

A full STL stream implementation would require considerable work, with iostream wrappers, etc.

This PR adds basic support to the `Print` classs for the `<<` insertion streaming operator. Avoiding `printf` is generally a good idea, and strong C++ types can do a much better job.

I've started going updating the sample projects using this capability to improve code flow and readability,  but not yet complete. Let's see what people think before going any further!

## C++ streaming operation <<
   Building output is commonly done like this::

      Serial.print("Temperature: ");
      Serial.print(temperature);
      Serial.print(" °C, humidity: ");
      Serial.print(humidity);
      Serial.println("%");

   We can now do this::

      Serial << "Temperature" << temperature << " °C, humidity: " << humidity << "%" << endl;

## Number Printing
   Examples::

      Serial.print(12, HEX);         // "c"
      Serial.print(12, HEX, 4);      // "000c"
      Serial.print(12, HEX, 4, '.'); // "...c"
      Serial.print(12);              // "12"
      Serial.print(12, DEC, 4);      // "0012"

   Similar extensions are provided for `String` construction::

      Serial << "0x" << String(12, HEX, 8);       // "0x0000000c"
      Serial << String(12, DEC, 4, '.');          // "..12"

## Field-width control
   Supported via String methods::

      Serial << String(12).padLeft(4);          // "  12"
      Serial << String(12).padLeft(4, '0');     // "0012"
      Serial << String(12).padRight(4);         // "12  "
      Serial << String(12).pad(-4, '0');        // "0012"
      Serial << String(12).pad(4);              // "12  "

## Strongly-typed enumerations
   Use of ``enum class`` is good practice as it produces strongly-typed and scoped values.
   Most of these are also provided with a standard ``toString(E)`` function overload.

   This allows string equivalents to be printed very easily::

      auto status = HTTP_STATUS_HTTP_VERSION_NOT_SUPPORTED;
      auto type = MIME_HTML;
      Serial.print(type); // "text/html"
      // "Status: HTTP Version Not Supported, type: text/html"
      Serial << "Status: " << status << ", type: " << type << endl;
      // Status: 505, type: 0
      Serial << "Status: " << int(status) << ", type: " << int(type) << endl;

### DIFF
--- a/Sming/Core/DateTime.cpp
+++ b/Sming/Core/DateTime.cpp
@@ -271,13 +271,6 @@ String DateTime::format(const char* sFormat)
 
 	String sReturn;
 
-	// Append a number to the return buffer, padding to a fixed number of digits
-	auto appendNumber = [&sReturn](unsigned number, unsigned digits, char padChar = '0') {
-		char buf[8];
-		ultoa_wp(number, buf, 10, digits, padChar);
-		sReturn.concat(buf, digits);
-	};
-
 	char c;
 	while((c = *sFormat++) != '\0') {
 		if(c != '%') {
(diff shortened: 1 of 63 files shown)


mikee47:
@slaff I didn't revise all the samples; what are your thoughts on that? Should we complete the job?

slaff:
> what are your thoughts on that? Should we complete the job?

@mikee47 sure, but only if you have time for it.

PR #2553, opened 2022-09-13
Apply printing improvements to remaining samples

mikee47:
This PR continues from #2551 to apply C++ streaming style to remaining samples, where appropriate.

Additional improvements:

- `MacAddress` no longer inherits from `Printable` so is no longer a virtual class
- The ESP8266 Network WifiSniffer now returns data using `MacAddress` instead of uint8_t[]. The corresponding sample application has also been updated to check for invalid characters in the received SSID.
- Update `makeHexString()` to accept `void*` parameter (instead of `uint8_t*`) so it can be more easily used with any type of data.


slaff:
@mikee47 please check the failing tests.

PR #2556, opened 2022-09-20
Improve HashMap memory usage, add sorting capability

mikee47:
The `HashMap` class template currently stores both keys and values using double-indirection. This is necessary for object types (e.g. String) but for simple (scalar) types it is very wasteful of memory.

This PR delegates key/value storage to separate list class optimised for either scalar or pointer (class) types.
Also, the allocation algorithm has been adjusted to expand storage by 4 or 16 entries, instead of only one.

The `stdc++` library has also been added so that `std::map` and `std::vector` can be used to compare memory usage.
The `Wiring` module in `HostTests` has been updated to show this.

A `HashMap<MimeType, uint16_t>` containing 13 items now uses 96 bytes instead of 182. `std::map` uses 312 bytes.
NB. enum declarations default to 32-bits, so reducing storage class for `MimeType` to uint8_t reduces map further to 48 bytes, whilst `std::map` reduces to 260 bytes.

One reason for the reduced memory usage is that HashMap uses two lists: one for keys, and a second for values, whilst `std::map` uses a single list of `std::pair` structures. These structures must be aligned so with small data types there is unused space.

Note, however, that a `std::map` will likely out-perform a simple HashMap for larger data sets as they use more sophisticated storage mechanisms https://en.cppreference.com/w/cpp/container/map. HashMap is much simpler, does not sort the keys and does lookups by iterating from the start of the map.

This PR also adds a basic `sort()` method so entries can be re-ordered, perhaps as part of post-processing or display.


slaff:
> TODO: Add methods so HashMap can be sorted by key or value

@mikee47 ping me when ready with this PR.

mikee47:
@slaff OK, happy with this PR. Sorting isn't the fastest, but it is flexible.

PR #2569, opened 2022-10-14
Fix Windows Host github CI workflow, tidy

mikee47:
This PR makes a few improvements to the CI scripts:

- Re-enable Windows Host builds. This was failing because of line endings: system default is core.autocrlf=true, requires override before checkout
- Use `checkout@v3`: version 2 is deprecated
- Move ninja and python setup code into scripts


mikee47:
> @mikee47 I guess we don't need [appveyor CI](https://github.com/SmingHub/Sming/blob/develop/appveyor.yml) any longer, right?

Agreed. I suggest we restore `appveyor.yml` from pre- da1ff08 then it's available for use should we desire.

slaff:
> Agreed. I suggest we restore appveyor.yml from pre- https://github.com/SmingHub/Sming/commit/da1ff089d29dfa00e6183722656221ae67ca4e9d then it's available for use should we desire.

Ok, can you remove `appveyor.yml` as  part of this PR? Git will keep a version of it and we can restored it if needed/desired.

mikee47:
> > Agreed. I suggest we restore appveyor.yml from pre- [da1ff08](https://github.com/SmingHub/Sming/commit/da1ff089d29dfa00e6183722656221ae67ca4e9d) then it's available for use should we desire.
> 
> Ok, can you remove `appveyor.yml` as part of this PR? Git will keep a version of it and we can restored it if needed/desired.

So do we remove travis.yml as well?

My point is that if I wished to run Sming through appveyor I'd just need to enable it in my appveyor account. Removing appveyor.yml and it's more involved.

slaff:
> So do we remove travis.yml as well?

I would say yes. Let's do cleaning now. Otherwise old/deprecated/unused things might start to pile up...

> My point is that if I wished to run Sming through appveyor I'd just need to enable it in my appveyor account. Removing appveyor.yml and it's more involved.

I agree with you, but are you using or planning to use appveyor for Sming (not for the libraries)? If not then I promise I will take care to restore appveyor.yml and travis.yml if the need arises ;)

mikee47:
@slaff That's fine. I've updated documentation, can look at supporting libraries separately so for the time being that remains.

PR #2571, opened 2022-10-22
Fix SPIFFS error translation

mikee47:


slaff:
@mikee47 please rebase.

PR #2572, opened 2022-10-24
Remove CustomDevice, use `editablePartitions()` method

mikee47:
This PR tidies up how device in-memory partition tables are updated.

The Device `partitions()` method returns a read-only (const) reference to the device's `PartitionTable` suitable for general application use.
A new `editablePartitions()`  method has been added which provides a non-const version so the table can be updated.
This removes the need for the `CustomDevice` class, which has been removed.

For example:

    part = device->createPartition("archive", Storage::Partition::SubType::Data::fwfs, startOffset, size);

becomes:

    part = device->editablePartitions().add("archive", Storage::Partition::SubType::Data::fwfs, startOffset, size);


Sming/Components/Storage/src/include/Storage/Device.h
```diff
@@ -64,6 +64,14 @@ class Device : public LinkedObjectTemplate<Device>
 		return mPartitions;
 	}
 
+	/**
+	 * @brief Provide full access to partition table
+	 */
+	PartitionTable& editable_partitions()
```
slaff, line 70: Rename the method name to use camelCase

Sming/Components/Storage/src/include/Storage/SysMem.h
```diff
@@ -70,20 +70,20 @@ class SysMem : public CustomDevice
...
 		{
-			return CustomPartitionTable::add(name, type, reinterpret_cast<uint32_t>(fstr.data()), fstr.size(),
-											 Partition::Flag::readOnly);
+			return PartitionTable::add(name, type, reinterpret_cast<uint32_t>(fstr.data()), fstr.size(),
+									   Partition::Flag::readOnly);
 		}
 	};
 
-	SysMemPartitionTable& partitions()
+	SysMemPartitionTable& editable_partitions()
```
slaff, line 86: camelCase

PR #2577, opened 2022-10-29
Add Github actions support for library CI testing

mikee47:
This PR adds a re-usable `library` workflow for Github actions so that Sming libraries can be independently tested.
It also  tidies up the main ``ci.yml`` file and fixes a couple of minor related issues.

The ``library.yml`` re-useable workflow is provided, which takes care of these tasks:

- Checking in the library to test
- Checking in the Sming framework
- Installing build tools
- Builds all applications within the library's ``samples`` directory, for all supported architectures
- If a test application is provided then that should be located in a ``test`` directory.
  This is built for all architectures, and also executed for Host.

Builds are handled using :source:`Tools/ci/library/Makefile`.

See also https://docs.github.com/en/actions/using-workflows/reusing-workflows.

To use this in a project, add a suitable workflow to the ``.github/workflows`` directory. Templates are provided in the ``.github/workflows/library`` directory.

Here is the basic ``push`` scenario:

```
name: CI Push
on: [push]
jobs:
  build:
  uses: SmingHub/Sming/.github/workflows/library.yml@develop
  # Inputs are all optional, defaults are shown
  with:
    # Repository to fetch Sming from
    sming_repo: 'https://github.com/SmingHub/Sming'
    # Sming branch to run against
    sming_branch: 'develop'
    # Library alias
    alias: ''
```

The ``sming_repo`` and ``sming_branch`` inputs are provided if your library requires modifications to Sming which are not (yet) in the main repository.

The ``alias`` input is required where the library repository name does not correspond with the working title. For example, the ``jerryscript`` library is in a repository called ``Sming-jerryscript``, so must be checked out using a different name. If Sming contains a library (or Component) with the same name then it will be overridden,
with a warning ``Multiple matches found for Component 'jerryscript'`` in the build log.

The ``ci-dispatch.yml`` example demonstrates manual triggering, which allows these inputs to be easily changed. See https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow.

Note that the workflow must be available in the library's default branch, or it will not appear in the github web page.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,22 +11,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: [Esp8266, Host, Esp32, Rp2040]
-        variant: [""]
         os: [ubuntu-latest, windows-latest]
+        variant: [esp8266, host, esp32, esp32s2, esp32c3, rp2040]
         include:
-        - arch: Esp32
-          variant: esp32s2
-          os: ubuntu-latest
-        - arch: Esp32
-          variant: esp32s2
-          os: windows-latest
-        - arch: Esp32
-          variant: esp32c3
-          os: ubuntu-latest
-        - arch: Esp32
-          variant: esp32c3
-          os: windows-latest
+          - variant: esp8266
+            arch: Esp8266
+          - variant: host
+            arch: Host
+          - variant: esp32
+            arch: Esp32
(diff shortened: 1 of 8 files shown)


mikee47:
NB. This won't work of course until added to Sming repo. See https://github.com/mikee47/Sming-jerryscript/actions/runs/3350862770 for test run.

PR #2586, opened 2022-12-01
Newer version of GCC complain about usage of printf(string-or-var) without …

slaff:
…thout formatting arguments.

Example:
```
error: format not a string literal and no format arguments [-Werror=format-security]
  320 |         printf("Error: Invalid base64\n"); TTY_FLUSH();
```

This PR fixes the compilation issues.


slaff:
Tested and working on Esp8266 and Host archs.

PR #2599, opened 2023-01-01
initial functioning version of ESP32 HardwarePWM.

pljakobs:
a few weeks ago I said I would take a swing at implementing HardwarePWM for the ESP32 Platform and here it is.
This is very preliminary and does not extend the api a lot (in fact, the only public function is curently HardwarePWM::getFrequency(uint8_t pin) which I used in debugging).
I would go ahead and extend the interface to expose the more advanced functions of ledc_ bt would appreciate your input on how to do so. My current thinking is that I put conditionals into the master HardwarePWM.h - but conditionals can make code very hard to read if the number of platforms gets too large.

### DIFF
--- a/Sming/Arch/Esp32/Core/HardwarePWM.cpp
+++ b/Sming/Arch/Esp32/Core/HardwarePWM.cpp
@@ -1,0 +1,230 @@
+/****
+ * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
+ * Created 2015 by Skurydin Alexey
+ * http://github.com/SmingHub/Sming
+ * All files of the Sming Core are provided under the LGPL v3 license.
+ *
+ * HardwarePWM.cpp
+ *
+ * Original Author: https://github.com/hrsavla
+ *
+ * This HardwarePWM library enables Sming framework user to use ESP SDK PWM API
+ * Period of PWM is fixed to 1000us / Frequency = 1khz
+ * Duty at 100% = 22222. Duty at 0% = 0
+ * You can use function setPeriod() to change frequency/period.
+ * Calculate the max duty as per the formulae give in ESP8266 SDK
+ * Max Duty = (Period * 1000) / 45
+ *
+ * PWM can be generated on up to 8 pins (ie All pins except pin 16)
+ * Created on August 17, 2015, 2:27 PM
+ *
+ * See also ESP8266 Technical Reference, Chapter 12:
+ * http://espressif.com/sites/default/files/documentation/esp8266-technical_reference_en.pdf
+ *
+ */
+
+// orig #include <Clock.h>
+// orig #include "ESP8266EX.h"
+#include <debug_progmem.h>
+//#include <soc/soc_caps.h>
+#include "driver/ledc.h"
+#include "esp_err.h"
+#include <HardwarePWM.h>
+
+
+// orig #include <HardwarePWM.h>
+
+// #define SOC_LEDC_TIMER_BIT_WIDE_NUM 9 
+// #define SOC_LEDC_CHANNEL_NUM 6
+
+
+ // default period in µs -> 5kHz
+
+
+// define default resolution
+
+/*#if static_cast<ledc_timer_bit_t>SOC_LEDC_TIMER_BIT_WIDE_NUM >= DESIRED_DEFAULT_RESOLUTION
+        #define DEFAULT_RESOLUTION DESIRED_DEFAULT_RESOLUTION
+#else
+        #define DEFAULT_RESOLUTION static_cast<ledc_timer_bit_t>SOC_LEDC_TIMER_BIT_WIDE_NUM // if 10 bit is not available, set to max
+#endif
+*/
+#define DEFAULT_RESOLUTION static_cast<ledc_timer_bit_t>(10)
+#define DEFAULT_PERIOD 200
+
+HardwarePWM::HardwarePWM(uint8_t* pins, uint8_t no_of_pins) : channel_count(no_of_pins)
+{
+	ledc_timer_config_t ledc_timer;
+	ledc_channel_config_t ledc_channel;
+	debugf("starting HardwarePWM init");
+	periph_module_enable(PERIPH_LEDC_MODULE);
+	if((no_of_pins > 0) && (no_of_pins < SOC_LEDC_CHANNEL_NUM))
+	{
+		uint32_t io_info[SOC_LEDC_CHANNEL_NUM][3];	// pin information
+		uint32_t pwm_duty_init[SOC_LEDC_CHANNEL_NUM]; // pwm duty
+		for(uint8_t i = 0; i < no_of_pins; i++) {
+			/* todo: update the digitalPins map
+			io_info[i][0] = EspDigitalPins[pins[i]].mux;
+			io_info[i][1] = EspDigitalPins[pins[i]].gpioFunc;
+			io_info[i][2] = EspDigitalPins[pins[i]].id;
+			*/
+			pwm_duty_init[i] = 0; // Start with zero output
+			channels[i] = pins[i];
+
+			/* 
+			/  Prepare and then apply the LEDC PWM timer configuration
+			/  this may cofigure the same timer more than once (in fact up to 8 times)
+			/  which should not be an issue, though, since the values should be the same for all timers
+			*/
+			ledc_timer.speed_mode      = pinToGroup(i);  // the two groups (if available) are operating in different speed modes, hence speed mode is an alias for group or vice versa
+			ledc_timer.timer_num       = pinToTimer(i);
+			ledc_timer.duty_resolution = LEDC_TIMER_10_BIT; // todo: make configurable later
+			ledc_timer.freq_hz         = periodToFrequency(DEFAULT_PERIOD); // todo: make configurable later
+			ledc_timer.clk_cfg         = LEDC_AUTO_CLK;
+
+			debugf("ledc_timer.\n\tspeed_mode: %i\n\ttimer_num: %i\n\tduty_resolution: %i\n\tfreq: %i\n\tclk_cfg: %i\n\n",
+				   (uint32_t)ledc_timer.speed_mode, (uint32_t)ledc_timer.timer_num,
+				   (uint32_t)ledc_timer.duty_resolution, (uint32_t)ledc_timer.freq_hz, (uint32_t)ledc_timer.clk_cfg);
+			ESP_ERROR_CHECK(ledc_timer_config(&ledc_timer));
+
+			/*
+			/   Prepare and then apply the LEDC PWM channel configuration
+			*/
+			ledc_channel.speed_mode     = pinToGroup(i);
+			ledc_channel.channel        = pinToChannel(i);
+			ledc_channel.timer_sel      = pinToTimer(i);
+			ledc_channel.intr_type      = LEDC_INTR_DISABLE;
+			ledc_channel.gpio_num       = pins[i];
+			ledc_channel.duty           = 0; // Set duty to 0%
+			ledc_channel.hpoint         = 0;
+			debugf("ledc_channel\n\tspeed_mode: %i\n\tchannel: %i\n\ttimer_sel %i\n\tinr_type: %i\n\tgpio_num: "
+				   "%i\n\tduty: %i\n\thpoint: %i\n\n",
+				   pinToGroup(i) , pinToChannel(i), pinToTimer(i), LEDC_INTR_DISABLE, pins[i], 0, 0);
+			ESP_ERROR_CHECK(ledc_channel_config(&ledc_channel));
+			ledc_bind_channel_timer(pinToGroup(i), pinToChannel(i), pinToTimer(i));
+		}
+		maxduty=maxDuty(DEFAULT_RESOLUTION);
+		const int initial_period=DEFAULT_PERIOD;
+	}
+}
+
+HardwarePWM::~HardwarePWM()
+{
+	// There is no function in the SDK to stop PWM output, yet.
+}
+
+/* Function Name: getChannel
+ * Description: This function is used to get channel number for given pin
+ * Parameters: pin - Esp8266 pin number
+ */
+uint8_t HardwarePWM::getChannel(uint8_t pin)
+{
+	for(uint8_t i = 0; i < channel_count; i++) {
+		if(channels[i] == pin) {
+			//debugf("getChannel %d is %d", pin, i);
+			return i;
+		}
+	}
+	return -1;
+}
+
+/* Function Name: getDutyChan
+ * Description: This function is used to get the duty cycle number for a given channel
+ * Parameters: chan -Esp8266 channel number
+ */
+uint32_t HardwarePWM::getDutyChan(uint8_t chan)
+{
+	if(chan == PWM_BAD_CHANNEL) {
+		return 0;
+	} else {
+		return ledc_get_duty(pinToGroup(chan),pinToChannel(chan));		
+	}
+		// esp32 defines the frequency / period per timer,  
+}
+
+/* Function Name: setDutyChan
+ * Description: This function is used to set the pwm duty cycle for a given channel. If parameter 'update' is false
+ *              then you have to call update() later to update duties.
+ * Parameters: chan - channel number
+ *             duty - duty cycle value
+ *             update - update PWM output
+ */
+bool HardwarePWM::setDutyChan(uint8_t chan, uint32_t duty, bool update)
+{
+	if(chan == PWM_BAD_CHANNEL) {
+		return false;
+	} else if(duty <= maxduty) {
+		ESP_ERROR_CHECK(ledc_set_duty(pinToGroup(chan), pinToChannel(chan), duty));
+		if(update) {
+			ESP_ERROR_CHECK(ledc_update_duty(pinToGroup(chan), pinToChannel(chan)));
+			//update();
+		}
+		return true;
+	} else {
+		debugf("Duty cycle value too high for current period.");
+		return false;
+	}
+}
+
+/* Function Name: getPeriod
+ * Description: This function is used to get Period of PWM.
+ *				Period / frequency will remain same for all pins.
+ *
+ */
+uint32_t HardwarePWM::getPeriod()
+{
+	// sming does not know how to handle different frequencies for channels, this will require an extended interface
+	// for now, just report the period for group 0 channel 0 
+	return frequencyToPeriod(ledc_get_freq(static_cast<ledc_mode_t>(0),static_cast<ledc_timer_t>(0)));
+}
+
+/* Function Name: setPeriod
+ * Description: This function is used to set Period of PWM.
+ *				Period / frequency will remain same for all pins.
+ */
+void HardwarePWM::setPeriod(uint32_t period)
+{
+	// setting the frequency globally, will add per timer functions later
+	// also, this can be done smarter 
+	for(uint8_t i = 0; i < channel_count; i++) {
+		ESP_ERROR_CHECK(ledc_set_freq(pinToGroup(i), pinToTimer(i), periodToFrequency(period)));
+	}
+	//sledc_update_duty();
+	update();
+}
+
+/* Function Name: update
+ * Description: This function is used to actually update the PWM.
+ */
+void HardwarePWM::update()
+{
+	//ledc_update_duty();	
+}
+
+ledc_channel_t HardwarePWM::pinToChannel(uint8_t pin){
+	return (ledc_channel_t)(pin % 8);
+}
+
+ledc_mode_t HardwarePWM::pinToGroup(uint8_t pin){
+	return (ledc_mode_t) (pin / 8);
+}
+
+ledc_timer_t HardwarePWM::pinToTimer(uint8_t pin){
+	return (ledc_timer_t) ((pin /2) % 4);
+}
+
+uint32_t HardwarePWM::periodToFrequency(uint32_t period){
+	return (1000000 / period);
+}
+
+uint32_t HardwarePWM::frequencyToPeriod(uint32_t freq){
+	return (1000000 / freq);
+}
+
+uint32_t HardwarePWM::maxDuty(ledc_timer_bit_t bits){
+	return (1<<(uint32_t)bits) - 1; 
+}
+
+uint32_t HardwarePWM::getFrequency(uint8_t pin){
+	return ledc_get_freq(pinToGroup(pin), pinToTimer(pin));
+}
--- a/Sming/Core/HardwarePWM.h
+++ b/Sming/Core/HardwarePWM.h
@@ -29,6 +29,9 @@
 
 #include <cstdint>
 #include <driver/pwm.h>
+#if SMING_SOC == Esp32
+	#include "hal/ledc_types.h"
+#endif
 
 #define PWM_BAD_CHANNEL 0xff ///< Invalid PWM channel
 
@@ -121,8 +124,21 @@ public:
 	/** @brief  This function is used to actually update the PWM.
 	 */
 	void update();
+	
+	#if SMING_ARCH == Esp32
+		uint32_t getFrequency(uint8_t pin);
+	#endif
 
 private:
+#if SMING_ARCH == Esp32
+	ledc_mode_t pinToGroup(uint8_t pin);
+	ledc_channel_t pinToChannel(uint8_t pin);
+	ledc_timer_t pinToTimer(uint8_t pin);
+	uint32_t periodToFrequency(uint32_t period);
+	uint32_t frequencyToPeriod(uint32_t freq);
+	uint32_t maxDuty(ledc_timer_bit_t bits);
+#endif
+
 	uint8_t channel_count;
 	uint8_t channels[PWM_CHANNEL_NUM_MAX];
 	uint32_t maxduty;


pljakobs:
I've added a number of toughts as a comment to the cpp file but don't want to open a new PR for just that, so I post it here for discussion 
```
` * the ESP32 PWM Hardware is much more powerful than the ESP8266, allowing wider PWM timers (up to 20 bit)
 * as well as much higher PWM frequencies (up to 40MHz for a 1 Bit wide PWM)
 * 
 * Overview:
 * +------------------------------------------------------------------------------------------------+
 * | LED_PWM                                                                                        |
 * |  +-------------------------------------------+   +-------------------------------------------+ |
 * |  | High_Speed_Channels¹                      |   | Low_Speed_Channels                        | |
 * |  |                   +-----+     +--------+  |   |                   +-----+     +--------+  | |
 * |  |                   |     | --> | h_ch 0 |  |   |                   |     | --> | l_ch 0 |  | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  | | h_timer 0 | --> |     |                 |   | | l_timer 0 | --> |     |                 | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 1 |  |   |                   |     | --> | l_ch 1 |  | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   |     |                 |   |                   |     |                 | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 2 |  |   |                   |     | --> | l_ch 2 |  | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  | | h_timer 1 | --> |     |                 |   | | l_timer 1 | --> |     |                 | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 3 |  |   |                   |     | --> | l_ch 3 |  | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   | MUX |                 |   |                   | MUX |                 | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 4 |  |   |                   |     | --> | l_ch 4 |  | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  | | h_timer 2 | --> |     |                 |   | | l_timer 2 | --> |     |                 | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 5 |  |   |                   |     | --> | l_ch 5 |  | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   |     |                 |   |                   |     |                 | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 6 |  |   |                   |     | --> | l_ch 6²|  | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  | | h_timer 3 | --> |     |                 |   | | l_timer 3 | --> |     |                 | |
 * |  | +-----------+     |     |     +--------+  |   | +-----------+     |     |     +--------+  | |
 * |  |                   |     | --> | h_ch 7 |  |   |                   |     | --> | l_ch 7²|  | |
 * |  |                   |     |     +--------+  |   |                   |     |     +--------+  | |
 * |  |                   +-----+                 |   |                   +-----+                 | |
 * |  +-------------------------------------------+   +-------------------------------------------+ |
 * +------------------------------------------------------------------------------------------------+
 * ¹ High speed channels are only available when SOC_LEDC_SUPPORT_HS_MODE is defined as 1
 * ² The ESP32C3 does only support six channels, so 6 and 7 are not available on that SoC
 * 
 * The nomenclature of timers in the high speed / low speed blocks is a bit misleading as the idf api 
 * speaks of "speed mode", which, to me, implies that this would be a mode configurable in a specific timer
 * while in reality, it does select a block of timers. I am considering re-naming that to "speed mode block"
 * in my interface impmenentation.
 * 
 * As an example, I would use
 * setTimerFrequency(speedModeBlock, timer, frequency);
 * 
 * ToDo: see, how this can be implemented to provide maximum overlap with the RP2040 pwm hardware so code does 
 * not become overly SoC specific.
 * 
 * Maximum Timer width for PWM:
 * ============================
 * esp32   SOC_LEDC_TIMER_BIT_WIDE_NUM  (20)
 * esp32c3 SOC_LEDC_TIMER_BIT_WIDE_NUM  (14)
 * esp32s2 SOC_LEDC_TIMER_BIT_WIDE_NUM  (14)
 * esp32s3 SOC_LEDC_TIMER_BIT_WIDE_NUM  (14)
 * 
 * Number of Channels:
 * ===================
 * esp32   SOC_LEDC_CHANNEL_NUM         (8)
 * esp32c3 SOC_LEDC_CHANNEL_NUM         (6)
 * esp32s2 SOC_LEDC_CHANNEL_NUM         (8)
 * esp32s3 SOC_LEDC_CHANNEL_NUM 		(8)
 * 
 * Some SoSs support a mode called HIGHSPEED_MODE which is essentially another full block of PWM hardware 
 * that adds SOC_LEDC_CHANNEL_NUM channels. 
 * Those Architectures have SOC_LEDC_SUPPORT_HS_MODE defined as 1.
 * In esp-idf-4.3 that's currently only the esp32 SOC 
 * 
 * Supports highspeed mode:
 * ========================
 * esp32 SOC_LEDC_SUPPORT_HS_MODE	(1)
 *
 * ToDo: implement awareness of hs mode availablility
 * ==================================================
 * currently, the code just uses a % 8 operation on the pin index to calculate whether to assign a pin to either
 * high speed or low speed pwm blocks. This doesn't make a whole lot of sense since it makes it impossible
 * for Sming devs to actually use the functionality behind it. 
 * Also, it currently does not reflect the fact that different SOCs have a different number of channels per block
 * (specifically, the esp32c3 only has six channels and no highspeed mode). 
 * I will continue in two ways: 
 * - implement the "vanilla" Sming HardwarePWM interface that will hide the underlying architecture but allow up to 16 
 *   channels on an ESP32 
 * - implement overloads for the relevant functions that allow selecting hs mode where applicable. 
 * 
 * ToDo: implement PWM bit width control
 * =====================================
 * the current HardwarePWM implementation does not care about the PWM timer bit width. To leverage the functionality 
 * of the ESP32 hardware, it's necessary to make this configurable. As the width is per timer and all the Sming defined
 * functions are basically per pin, this needs to be an extension of the overal model, exposing at least timers. 
 * This, too, will require a compatible "basic" interface and an advanced interface that allows assiging pins (channels) 
 * to timers and the configuration of the timers themselves. 
 * The esp_idf does not provide a way to read the bit width configured for a channel, but I think it'll be useful to be able
 * to read back this value, not least to find the value for getMaxDuty() for a channel. It will either have to be stored in the
 * module or maybe read from the hardware directly (LEDC_[HL]STIMERx_CONF_REG & 0x1f)
 * 
 * hardware technical reference: 
 * =============================
 * https://www.espressif.com/sites/default/files/documentation/esp32_technical_reference_manual_en.pdf#ledpwm
 * 
 * Overview of the whole ledc-system here: 
 * https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/ledc.html
 * 
 * the ledc interface also exposes some advanced functions such as fades that are then done in hardware. 
 * ToDo: implement a Sming interface for fades
 * `
```
Not sure what a good implementation for a more abstract interface is. Maybe a baseclass / SoC specific implementation class scheme to implement generic functions in software and provide SoC specific capabilities plus extensions where available? That's how the adafruit gfx library builds it for specific hardware.

Sming/Core/HardwarePWM.h
```diff
@@ -121,8 +124,21 @@ class HardwarePWM
 	/** @brief  This function is used to actually update the PWM.
 	 */
 	void update();
+	
+	#if SMING_ARCH == Esp32
+		uint32_t getFrequency(uint8_t pin);
+	#endif
```
mikee47, line 130: We should implement this for all architectures, so remove conditional.

NB. Use `#ifdef ARCH_ESP32`, there's no definition for 'Esp32'.

mikee47 (changes requested):
Great start, thank you! In general just a little tidying.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,345 @@
...
+			ledc_channel.duty           = 0; // Set duty to 0%
+			ledc_channel.hpoint         = 0;
+			debugf("ledc_channel\n\tspeed_mode: %i\n\tchannel: %i\n\ttimer_sel %i\n\tinr_type: %i\n\tgpio_num: "
+				   "%i\n\tduty: %i\n\thpoint: %i\n\n",
+				   pinToGroup(i) , pinToChannel(i), pinToTimer(i), LEDC_INTR_DISABLE, pins[i], 0, 0);
+			ESP_ERROR_CHECK(ledc_channel_config(&ledc_channel));
+			ledc_bind_channel_timer(pinToGroup(i), pinToChannel(i), pinToTimer(i));
+		}
+		maxduty=maxDuty(DEFAULT_RESOLUTION);
+		const int initial_period=DEFAULT_PERIOD;
```
mikee47, line 222: Codacy is warning about unused definition here...

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,345 @@
...
+	}
+}
+
+HardwarePWM::~HardwarePWM()
+{
+	// There is no function in the SDK to stop PWM output, yet.
```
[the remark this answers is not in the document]
pljakobs: this was original code that I have not changed yet :D

pljakobs:
(sorry for takign some more time, I was travelling and then sick for two weeks - will get to this soon)

Sming/Arch/Esp32/Core/HardwarePWM.cpp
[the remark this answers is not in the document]
pljakobs: I believe this should do the trick. will have to test if I have to de-associate the timers, too
```
	for(uint8_t i = 0; i < no_of_pins; i++) {
		//stop pwm for all pins and set idle level to 0.
		
	ledc_stop(pinToGroup(pins[i]), pinToChannel(pins[i]), uint32_t 0); 
	}
```

pljakobs:
ok, I'm sorry if this is now very muddled, but I believe I got lost in git here - for reasons I don't understand, both, the .h as well as the .cpp file were repeatedly lost by git and no longer tracked. I *think*, now it's all there

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,345 @@
...
+	//sledc_update_duty();
+	update();
+}
+
+/* Function Name: update
+ * Description: This function is used to actually update the PWM.
+ */
+void HardwarePWM::update()
+{
+	//ledc_update_duty();
```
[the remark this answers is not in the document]
mikee47: I believe the purpose of this method is so that multiple channels can be reconfigured, then updated all together. For some applications this may be important to ensure they are synchronised. Unfortunately the IDF HAL doesn't support this.

My personal preference here would be to strip away all the HAL code and implement using low-level calls (such as `ledc_ll_ls_channel_update`). For now, probably just add a note to the code.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,303 @@
...
+	for(uint8_t i = 0; i < channel_count; i++) {
+		ESP_ERROR_CHECK(ledc_set_freq(pinToGroup(i), pinToTimer(i), periodToFrequency(period)));
+	}
+	// ledc_update_duty();
+	update();
+}
+
+void HardwarePWM::update()
+{
+	// ledc_update_duty();
```
mikee47, line 299: Aside from optimisation, this method should allow multiple channels to be reconfigured, then updated simultaneously to keep them synchronised. Unfortunately the IDF HAL doesn't support this so requires further work.

My personal preference here would be to strip away all the HAL code and implement using low-level calls (such as `ledc_ll_ls_channel_update`). We'd also need to track which channels have changed and update them as a group. Something to consider as a future update, perhaps?

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,305 @@
...
+	if(chan == PWM_BAD_CHANNEL) {
+		return false;
+	}
+
+	if(duty <= maxduty) {
+		ESP_ERROR_CHECK(ledc_set_duty(pinToGroup(chan), pinToChannel(chan), duty));
+		if(update) {
+			ESP_ERROR_CHECK(ledc_update_duty(pinToGroup(chan), pinToChannel(chan)));
+			//update();
+		}
```
mikee47, line 271: If the `update` method isn't implemented then this needs changing so it ignores the `update` flag and (at least for now) *always* updates the channel.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+uint32_t periodToFrequency(uint32_t period)
+{
+	return (period == 0) ? -1 : (1000000 / period);
+}
+
+uint32_t frequencyToPeriod(uint32_t freq)
+{
+	return (freq == 0) ? -1 : (1000000 / freq);
+}
```
mikee47, line 162: Is it intentional to return `-1` for invalid values? Should it be `0`?

mikee47 (changes requested):
I've rebased and done some tidying but code functionality hasn't been changed.
Couple of points to consider regarding return values for invalid period/frequency and how update() works.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: you're right, -1 doesn't even make sense as it's uint32_t. I guess I was thinking that 0 could be an almost valid response but that isn't true either. I'll change it to 0

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: I see where you're coming from. I was trying to create new static structures holding state to keep things easy, but being able to synchronize a change does require that.
But then again, when I thought about all the channel/timer relationships, this could get a much richer thing anyway and would need more state to be handled. Maybe it's worth writing an intermediary abstraction to handle all that. I'll think about it

pljakobs: so, thinking a bit more about the interface, there's a few tings that might be of interest.

- depending o the SoC, there are four or eight timers that feed into six to 16 pwm channels. 
- the timers are configurable for duty cycle resolution and output frequency 
- high speed timers allow for glitch free change
- the same basic pwm system is also used for motor control, which might require different frequencies

My primary motivation to work on this is that I build led controllers that do thee to five channel (RGB to RGBWWCW) pwm. Longer term, my goal is to have a firmware that can use all available pwm channels of an SOC but break it out into multilple lights (maybe five times RGB, four times RGBW or three times RGBWWCW). As for the update function: that to me would mean that I want to be able to update specific groups of channels with one call and leave the rest alone. Also, I would want to have the ability to configure all channels that belong to one light to use one timer while respecting the fact, that some motor control might use an altogether different set of pwm parameters.

On the API level, this would require some notion of grouping channels / pwm pins, which would extend the current API and possibly break compatibility with existing code and/or other platforms, hence my thought of implementing an intermediary abstraction.

On the HardwarePWM level, I would then implement all channels to use the same timer (or two, if HS block is available), thus making sure they are all running at the same frequency and bit width.
I'm thinking of introducing a concept of channel groups that would configure channel assignment to timers and timer parameters. This way, the HardwarePWM interface would (mostly) still be concerned with channels.
There are a few functions in HardwarePWM that would, by this definition, not technically set channel parameters but group parameters, such as get/setPeriod/Frequency and update. I would overload them with a group parameter.
If no group is explicitly configured, all channels would end up in one default group (two if HS_ channels are being used) and those calls would change the values for that group.
I'll try to put that together in a new branch.

pljakobs:
what's that remaining change request? I can't seem to see it.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,315 @@
...
+	debug_d("starting HardwarePWM init");
+	periph_module_enable(PERIPH_LEDC_MODULE);
+	if((no_of_pins == 0) || (no_of_pins > SOC_LEDC_CHANNEL_NUM)) {
+		return;
+	}
+
+	uint32_t io_info[SOC_LEDC_CHANNEL_NUM][3];	// pin information
+	uint32_t pwm_duty_init[SOC_LEDC_CHANNEL_NUM]; // pwm duty
+	for(uint8_t i = 0; i < no_of_pins; i++) {
+		pwm_duty_init[i] = 0; // Start with zero output
```
mikee47, line 188: `io_info` and `pwm_duty_init` aren't used, please remove

mikee47 (changes requested):
Great, just remove those unused variables

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,327 @@
...
+		* if(update) {
+		*	ESP_ERROR_CHECK(ledc_update_duty(pinToGroup(chan), pinToChannel(chan)));
+		*	//update();
+		* }
+		*/
+		ESP_ERROR_CHECK(ledc_update_duty(pinToGroup(chan), pinToChannel(chan)));
+		return true;
+	} else {
+		debug_d("Duty cycle value too high for current period.");
+		return false;
```
mikee47, line 280: Remove (Merge issue ?)

mikee47 (approved):

slaff:
@pljakobs Please run the coding style fixer. The following file has an issue:
```
--- a/Sming/Arch/Esp32/Core/HardwarePWM.cpp
+++ b/Sming/Arch/Esp32/Core/HardwarePWM.cpp
@@ -310,4 +310,3 @@ uint32_t HardwarePWM::getFrequency(uint8_t pin)
 {
 	return ledc_get_freq(pinToGroup(pin), pinToTimer(pin));
 }
-' ']'
++ echo '!!! Coding Style issues Found!!!'
++ echo '    Run: '\''make cs'\'' to fix them. '
++ echo 'diff --git a/Sming/Arch/Esp32/Core/HardwarePWM.cpp b/Sming/Arch/Esp32/Core/HardwarePWM.cpp
index f6aefe8..38ef1ad 1006[44](https://github.com/SmingHub/Sming/actions/runs/4582860432/jobs/8149817907?pr=2599#step:7:45)
--- a/Sming/Arch/Esp32/Core/HardwarePWM.cpp
+++ b/Sming/Arch/Esp32/Core/HardwarePWM.cpp
@@ -310,4 +310,3 @@ uint32_t HardwarePWM::getFrequency(uint8_t pin)
 {
 	return ledc_get_freq(pinToGroup(pin), pinToTimer(pin));
 }
-'
++ exit 1
```

slaff:
@pljakobs Also fix the typos reported here: https://github.com/SmingHub/Sming/pull/2599/files#diff-647d65279a6f999683c4cd320288f28dd4f901ca8a4eba5a75f1c9e2ad4ff451

Sming/Core/HardwarePWM.h
```diff
@@ -122,6 +122,11 @@ class HardwarePWM
 	 */
 	void update();
 
+	/** @brief Get PWM Frequency
```
slaff, line 125: Add also 
@param pin GPIO to get frequency for

pljakobs:
you guys are thorrough! hope I caught it all by now.

Also - can I get some feedback on the other implementation that I've started (#2624)?

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,312 @@
...
+ * The esp_idf does not provide a way to read the bit width configured for a channel, but I think it'll be useful to be able
+ * to read back this value, not least to find the value for getMaxDuty() for a channel. It will either have to be stored in the
+ * module or maybe read from the hardware directly (LEDC_[HL]STIMERx_CONF_REG & 0x1f)
+ * 
+ * ToDo: implement an abstraction layer
+ * ====================================
+ * as it stands now, this implementation does not provide a function to synchronize all the PWM channels (HardwarePWM::update())
+ * It might be a good idea to provide an intermediary abstraction that handles all the low level PWM functions (such as flexible
+ * timer to channel assignments, hs/ls mode awareness, pwm bit width etc) and implements the update() function on that level.
+ *
```
mikee47, line 121: Should probably remove this, perhaps put it in the new driver?

pljakobs: done

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+ledc_timer_t pinToTimer(uint8_t pin)
+{
+	return ledc_timer_t((pin / 2) % 4);
+}
+
+uint32_t periodToFrequency(uint32_t period)
+{
+	return (period == 0) ? -1 : (1000000 / period);
```
slaff, line 126: Please, change the return value to be a valid uint32_t. -1 is not a valid unsigned int.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+uint32_t frequencyToPeriod(uint32_t freq)
+{
+	return (freq == 0) ? -1 : (1000000 / freq);
```
slaff, line 131: Please, change the return value to be a valid uint32_t. -1 is not a valid unsigned int.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+uint32_t maxDuty(ledc_timer_bit_t bits)
+{
+	return (1U << bits) - 1;
```
slaff, line 136: Please, check the return value to be valid uint32_t.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,276 @@
...
+	for(uint8_t i = 0; i < channel_count; i++) {
+		ledc_stop(pinToGroup(i), pinToChannel(i), 0);
+	}
+}
+
+uint8_t HardwarePWM::getChannel(uint8_t pin)
+{
+	for(uint8_t i = 0; i < channel_count; i++) {
+		if(channels[i] == pin) {
+			// debug_d("getChannel %d is %d", pin, i);
```
slaff, line 214: Either remove the commented line or uncomment it. I would prefer to have it removed if it is not improving the debug output.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
```diff
@@ -23,115 +23,95 @@
...
+#include <HardwarePWM.h>
 #include <Clock.h>
 #include "ESP8266EX.h"
 #include <debug_progmem.h>
 
-#include <HardwarePWM.h>
-
 #define PERIOD_TO_MAX_DUTY(x) (x * 25)
 
 HardwarePWM::HardwarePWM(uint8_t* pins, uint8_t no_of_pins) : channel_count(no_of_pins)
```
slaff, line 33: Please, keep close to Sming's coding style: https://sming.readthedocs.io/en/latest/information/develop/coding-style.html?highlight=Variables#naming. 
```
Local variables, instance variables, and class variables must also be written in lowerCamelCase. Variable names must not start with, end with or contain underscore (_) or dollar sign ($) characters. 
```

For example in the code here `uint8_t no_of_pins` should be renamed to `uint8_t numberOfPins`, `uint32_t io_info[PWM_CHANNEL_NUM_MAX][3];` -> `uint32_t ioInfo[PWM_CHANNEL_NUM_MAX][3];`,  `uint32_t pwm_duty_init[PWM_CHANNEL_NUM_MAX]` -> `uint32_t pwmDutyInit[PWM_CHANNEL_NUM_MAX]` and so on.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
```diff
@@ -23,115 +23,95 @@
...
-/* Function Name: getChannel
- * Description: This function is used to get channel number for given pin
- * Parameters: pin - Esp8266 pin number
- */
 uint8_t HardwarePWM::getChannel(uint8_t pin)
 {
 	for(uint8_t i = 0; i < channel_count; i++) {
 		if(channels[i] == pin) {
-			//debugf("getChannel %d is %d", pin, i);
+			// debug_d("getChannel %d is %d", pin, i);
```
slaff, line 63: Remove commented debug statements.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
```diff
@@ -23,115 +23,95 @@
...
 			return i;
 		}
 	}
-	//debugf("getChannel: can't find pin %d", pin);
+
+	// debug_d("getChannel: can't find pin %d", pin);
```
slaff, line 68: Remove commented debug statements.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
```diff
@@ -23,115 +23,95 @@
...
 		if(update) {
 			this->update();
 		}
 		return true;
-	} else {
-		debugf("Duty cycle value too high for current period.");
-		return false;
 	}
+
+	debugf("Duty cycle value too high for current period.");
```
slaff, line 91: Replace `debugf` with `debug_w`. And include also the duty cycle value to make the debug statement more useful.

slaff (changes requested):
Overall looks good to me. There are still few coding style and return value issues that have to be fixed before merging this PR.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
slaff: > should this be part of this PR or may be a PR of it's own

Leave it for another PR. This one waited already for too long to be merged. 

> I believe, all other occurences of underscored names are defined in the esp32 ledc interface.

Please, rename all local variables in the code that you have introduced to use lowerCamelCase . Example:

```C++
	uint32_t io_info[PWM_CHANNEL_NUM_MAX][3];	// pin information
	uint32_t pwm_duty_init[PWM_CHANNEL_NUM_MAX]; // pwm duty
  // ...
	const int initial_period = 1000;

```

and so on.

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: hmm.. I can't see how it would not (I know, famous last words) - the only error condition here would be bits>32 - so I would tend to check for bits <= DEFAULT_RESOLUTION (which is currently set to 10).

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: done

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: I don't use those, they are part of the original ESP8266 version and are used both there and in the RP2040 implementation. I'll go ahead and rename them all and send a separate PR.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: is this really part of my PR? I have not touched the 8266 code

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
pljakobs: so, after looking at this a bit more, I believe what has happened.
The only files I have changed for this should be Arch/Esp32/Core/HardwarePWM.cpp and Core/HardwarePWM.h 
At some point, I did run make cs which has touched a ton of files throughout the whole source tree. I believe that's how the Arch/Esp8266 files got into this pr. They should not be there. 
At the risk that I'm now being more work than it was to write the original pwm implementation, I guess it would be best to just rebase the changes I did and leave the rest as it is in the upstream develop branch. I'm most happy to implement the changes suggested there, but they really should not be in this PR.

As a side note: I am very impressed with the level of professionality and support you are applying here, and I really appreciate your help a lot, even though it makes me feel very much like a total noob.

Sming/Arch/Esp8266/Core/HardwarePWM.cpp
mikee47: That was me! See commit eaf77f91 - there's a new `getFrequency` method which is required.

pljakobs: ah, I was really confused for a moment now. I have renamed the local vars to comply with camelCase now, not yet removed the debug calls. I think some of them are worthy of being there and functional, so I will uncomment them

Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
```diff
@@ -1,65 +1,3 @@
...
- * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
- * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
- * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
- *
- */
-
-#ifndef __PWM_H__
-#define __PWM_H__
-
 #pragma once
```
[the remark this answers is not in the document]
pljakobs: done, and also made it correctly use SOC_LEDC_CHANNEL_NUM as the max.

pljakobs:
sice I've deleted my earlier, confused comment, I'll say it again:
I really am impressed and I appreciate the diligence, professionality and support you are showing here. As I am sure you could tell, I don't do any of my coding at that level so it's an interesting experience for me, albeit I fear I've created more work for you than I actually did.

slaff:
> I really am impressed and I appreciate the diligence, professionality and support you are showing here.

@pljakobs Thank you for your contribution and congratulations for your first merged PR :)

nidefawl:
I tried building the Basic_HwPWM sample for ESP32 getting this error:
`HardwarePWM::~HardwarePWM()`
I tried reinstalling/rebuilding everything without luck

PR #2612, opened 2023-02-03
Add ESP32 support for IDF v4.4

mikee47:
This PR adds support for ESP IDF version 4.4 and changes the default installed version accordingly.
Integration tests include both 4.3 and 4.4.


mikee47:
The 4.4 branch adds some missing support for USB integration (in development). IDF version 5.0 was released in December but has some more major changes. I've got a PR for this in the works.

PR #2614, opened 2023-02-07
Add Esp32 IDF v5.0 support

mikee47:
IDF version 5 has a number of breaking changes. See https://docs.espressif.com/projects/esp-idf/en/v5.0/esp32/migration-guides/release-5.x/index.html.

Change summary:

- hw_timer driver has been reworked to replace timer_hal_xxx calls with  timer_ll_xx. IDF 5 replaces the timer HAL with a gptimer driver.
- The ADC interface has changed so new code has been added when compiled with IDF 5. Calibration has not been implemented.
- Minor API changes for uart, Ethernet
- Bluetooth support updated using latest esp-nimble-cpp master branch
- Separate CI workflow added for ESP32
- Add esp32s3 and esp32c2 variants (both untested - hardware not available)
- Update esptool to v4.4, required for esp32c2
- Gdbstub panic handler enabled by default only for debug builds
- The RMT driver interface has changed. This is used by the Adafruit_Neopixel library which has not yet been updated. The previous driver interface has been enabled but is deprecated - a warning is visible in the build logs.


mikee47:
I've left the default installed version at 4.4 since 5.0 is going to have more rough edges...

Issue #2616, opened 2023-02-09
Identifying code quality issues e.g. unnecessary null pointer checks


elfring:
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [AccessPointImpl::config](https://github.com/SmingHub/Sming/blob/112da39f9d718a37e937de8825f6faf4a1c45cc1/Sming/Components/Network/Arch/Esp8266/Platform/AccessPointImpl.cpp#L85 "AccessPointImpl::config function")
- [doUpgrade](https://github.com/SmingHub/Sming/blob/2345c3e8aacd94af917441f93f55e9a60dd7cee3/samples/Basic_Ota/app/application.cpp#L51 "doUpgrade function")
- [YeelightBulb](https://github.com/SmingHub/Sming/blob/07ecd3bdf33f5b56ec1c32c9587df456ffaa4c07/Sming/Libraries/Yeelight/YeelightBulb.cpp#L14 "Destructor for YeelightBulb")

slaff:
@elfring can you prepare a PR with the suggested changes?

slaff:
> Some of those should be using std::unique_ptr as well :-)

@mikee47 do you want to create a PR with the changes?

elfring:
:thought_balloon: Would you become interested to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability/delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

slaff:
Fixed in #2643

mikee47:
I'm reopening this issue as it is not yet fully resolved. clang-tidy is undoubtedly useful, especially if it can be integrated into the build system in some way. Whether automated code changes are appropriate is another matter.... 

Are there any other offline code checking tools we might consider?

elfring:
[Another update candidate was left over](https://github.com/SmingHub/Sming/blob/02835d8f4e539931c46b16191a6b38f7c0a4b04e/Sming/Components/Network/src/Network/TelnetServer.cpp#L34 "Update candidate: TelnetServer::enableCommand()").

mikee47:
@elfring

PR #2648, opened 2023-05-28
Add `clang-tidy` support for static code analysis

mikee47:
This PR allows static code analysis using clang-tidy, suggested by #2616.

Clang has more limited `constexpr` support than GCC so cannot parse some of the FlashString and NanoTime code without modification. However, these changes are only made when `__clang__` is defined and do not affect regular builds with GCC.

This PR also includes some basic code fixes which clang identifies. There are a lot more to look at.

To try it out, build a project with:

```
make CLANG_TIDY=clang-tidy
```

Add additional parameters like this:

```
make CLANG_TIDY="clang-tidy --fix --fix-errors"
```

Notes:

- Don't use `-j` option as clang-tidy output and fixes don't get serialised correctly.
- Settings are in the main `.clang-tidy` file. A custom version can be used and passed on the command line.
- I've been using clang 17.0.6 https://releases.llvm.org/17.0.1/tools/clang/tools/extra/docs/clang-tidy/index.html
- Only source files which haven't been built are inspected. So, to restrict which code gets processed built the entire application then 'clean' the relevant modules before proceeding with clang-tidy.
- No object code is generated by clang.


mikee47:
I'm starting to get a feel for the power of this tool! For example:

```
make CLANG_TIDY="clang-tidy --checks='-*,modernize-use-equals-default' --fix"
```

Will apply a single class of fix across the entire codebase. And the results are consistent and work very well.

Same is not true for all checks but I think this is probably the way to go when we decide to apply these sorts of changes.

What I'm currently not seeing is an easy way to track changes made into submodules...

PR #2668, opened 2023-09-23
Add CStringArray::join() method

mikee47:
Available in other languages (python, javascript) this provides an efficient way to concatenate array members.

### DIFF
--- a/Sming/Core/Data/CStringArray.h
+++ b/Sming/Core/Data/CStringArray.h
@@ -247,6 +247,15 @@ public:
 	unsigned count() const;
 
 	/**
+	 * @brief Get contents of array as delimited string
+	 * @param separator What to join elements with
+	 * @retval String
+	 *
+	 * e.g. CStringArray(F("a\0b\0c")).join() returns "a,b,c"
+	 */
+	String join(const String& separator = ",") const;
+
+	/**
 	 * @name Iterator support (forward only)
 	 * @{
 	 */
--- a/Sming/Core/Data/CStringArray.cpp
+++ b/Sming/Core/Data/CStringArray.cpp
@@ -168,3 +168,25 @@ unsigned CStringArray::count() const
 	}
 	return stringCount;
 }
+
+String CStringArray::join(const String& separator) const
+{
+	if(isNull()) {
+		return nullptr;
+	}
+	unsigned len = length();
+	if(len == 0) {
+		return "";
+	}
+	len -= count(); // NUL separators
+	len += separator.length() * (stringCount - 1);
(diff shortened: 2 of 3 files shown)


Sming/Core/Data/CStringArray.h
```diff
@@ -246,6 +246,15 @@ class CStringArray : private String
...
 	unsigned count() const;
 
+	/**
+	 * @brief Get contents of array as delimited string
+	 * @param separator What to join elements with
+	 * @retval String
+	 *
+	 * e.g. CStringArray(F("a\0b\0c")).join() returns "a,b,c"
+	 */
+	String join(const String& separator = ",") const;
```
slaff, line 256: @mikee47 Is there a reason for the separator parameter to be optional? In JavaScript, PHP and Python as far as I can say this parameter is mandatory.

mikee47: I hadn't considered PHP but default is same as javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join.

Python would be `",".join(cs)` which is an interesting construction - clearly there cannot be any 'default' in this case. However, to actually implement this efficiently for Sming would require that the array itself does the join. We could add this as a template method to String easily enough:

```
class String
{
  ...

  template <class Array> String join(const Array& array)
  {
    return array.join(*this);
  }

  ...
};
```

mikee47: I don't know of any C++ STL equivalent for this, though I see Boost has https://www.geeksforgeeks.org/boostalgorithmjoin-in-cpp-library/ https://www.boost.org/doc/libs/1_83_0/doc/html/boost/algorithm/join.html

mikee47: > to actually implement this efficiently for Sming would require that the array itself does the join

The efficiency here is avoiding reallocations for the returned String as in the general case:

```
template <class Array> String join(const Array& array, const String& separator)
{
  String s;
  for(auto& e: array) {
    if(s) {
      s += separator;
    }
    s += e;
  }
  return s;
}
```

This should work for any `Vector` or other iterable objects with an implicit String() operator. Something that works with HashMaps would also be useful (returning delimited 'name=value' pairs). However the implementation as it stands does not perform any string escaping, for example:

```
CStringArray cs;
cs.add("first");
cs.add("second,plus,stuff");
```
Then `cs.join()` returns "first,second,plus,stuff" as requested. If "first,\"second,plus,stuff\"" is required that'll have to be done manually. This is same behaviour as python.

Issue #2659, opened 2023-08-12
dangling pointer warning in Components/Storage/src/Debug.cpp ?


pljakobs:
this might be a self-created issue, but I don't see how to fix it.
I have successfully built my firmware on sming 5.0 this morning, but that fails now and I cannot say why that is.
The compiler warns me of a dangling pointer here:
```
Building /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/out/Host/debug/lib/clib-Storage-e2fa65b3ebab4603f1bf91105423d574.a
C+ /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/Debug.cpp
C+ /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/Device.cpp
C+ /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/Partition.cpp
In file included from /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/include/Storage/Debug.h:3,
                 from /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/Debug.cpp:1:
In member function 'size_t Print::print(const char*)',
    inlined from 'typename std::enable_if<(! std::is_array< <template-parameter-1-1> >::value), Print&>::type operator<<(Print&, const T&) [with T = char*]' at /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Wiring/Print.h:272:9,
    inlined from 'void Storage::Debug::listPartitions(Print&, const Storage::Device&)' at /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/Storage/src/Debug.cpp:21:29:
/home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Wiring/Print.h:108:29: error: using a dangling pointer to 'buf' [-Werror=dangling-pointer=]
  108 |                 return write(str);
      |                        ~~~~~^~~~~
In file included from /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/FlashString/src/include/FlashString/config.hpp:24,
                 from /home/pjakobs/devel/esp_rgbww_firmware/Sming/Sming/Components/FlashString/src/include/FlashString/Utility.hpp:24,
[...]

PR #2671, opened 2023-10-11
Provide workaround for 'dangling pointer' error/warning

mikee47:
Sorts warnings for host builds as discussed in #2659.


slaff:
@mikee47 run the coding style fixer and add the changes to this PR.

PR #2687, opened 2023-12-05
Refactor command processor

slaff:
Simplify the command handler. Remove it as dependency in the core components. Move it to a separate component. Add samples with Websocket, Serial and Telnet server to the component.

TODO
--------
- [x] Add migration document


mikee47 (commented):
@slaff I've pushed some suggestions as a PR to your repo.

PR #2118, opened 2020-10-19
Update appveyor build to use latest MinGW

mikee47:
Let's have another try to get appveyor working again.
For reference, appveyor builds on my account are without issue.

PR #2694, opened 2024-01-03
Add support for ESP32 IDF version 5.2 (beta)

mikee47:
TODO:

* [x] No quick fix for Windows build issues, so probably just avoid compiling the problem code.
* [x] GDB doesn't work for esp32 with IDF 5.2. Toolset directory layout has changed.


mikee47:
CI failures for IDF 5.2 are a recurrence of the issues addressed in #2086, #2118. Early builds were failing for other SDK versions too, this was fixed by restricting number of include directories.

PR #2697, opened 2024-01-08
Fix esp32 timer issues

mikee47:
HostTests fails on esp32 due to following issues:

**Flawed `hw_timer2_read()` implementation**

Without setting update bit counter value won't change. Probably worked previously because some other code did this.

**WDT with callback timers**

Checked against IDF implementations and there's a missing flag which keeps interrupts initially disabled during setup.

**Callbacks missing during timer testing**

Offending code looks like this:

```
Timer timer;
String s;
s += timer; //<< Not what I expected
```

The above line actually calls the `Timer` copy constructor. Delete this and the compiler catches the error. The corrected line now reads `s += String(timer)`.

**HostTests needs more program space for ESP32**

Builds are getting chunky!

### DIFF
--- a/Sming/Arch/Esp32/Components/driver/hw_timer.cpp
+++ b/Sming/Arch/Esp32/Components/driver/hw_timer.cpp
@@ -62,7 +62,8 @@ public:
 #else
 		int source = timer_group_periph_signals.groups[group].timer_irq_id[index];
 #endif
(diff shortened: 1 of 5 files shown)


mikee47:
@slaff Some of the what-the-diff comments are pretty unhelpful so I tend to delete them. (Others are surprisingingly accurate, though!)

PR #2700, opened 2024-01-09
Fix esp32c3 FPU issues

mikee47:
The HostTests application crashes on the esp32c3 because it doesn't have hardware floating-point support. By default code is compiled with FPU instructions which produce 'illegal instruction' exceptions.

The `-march=rv32imc` setting is required for these processors so emulation code is used instead. The IDF 5.2 compilers have additional `zicsr` and `zifencei` extensions. Haven't looked into these, these settings are just copied from the IDF makefiles.

Similiarly, a scan of the IDF makefiles shows a few missing definitions. Don't know whether these are 100% necessary or not but don't seem to do any harm.

### DIFF
--- a/Sming/Arch/Esp32/Components/esp32/sdk/soc.mk
+++ b/Sming/Arch/Esp32/Components/esp32/sdk/soc.mk
@@ -1,0 +1,10 @@
+#
+# soc
+#
+
+ifeq (esp32,$(SMING_SOC))
+    # esp_dport_access_reg_read is added as an undefined symbol because otherwise
+    # the linker can ignore dport_access.c as it would no other files depending on any symbols in it.
+	SDK_UNDEF_SYMBOLS += \
+		esp_dport_access_reg_read
+endif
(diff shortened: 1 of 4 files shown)


Sming/Arch/Esp32/Components/esp32/sdk/soc.mk
```diff
@@ -0,0 +1,10 @@
+#
+# soc
+#
+
+ifeq (esp32,$(SMING_SOC))
+    # esp_dport_access_reg_read is added as an undefined symbol because otherwise
+    # the linker can ignore dport_access.c as it would no other files depending on any symbols in it.
```
slaff, line 7: > as it would no other files depending on any symbols in it.

Maybe you meant `as there are no other files depending on any symbols in it` ?

mikee47: Maybe. Just cut'n'pasted from the IDF :-)
I'll remove the comment, it's superfluous.

PR #2704, opened 2024-01-10
Update requirements.txt

slaff:
In our CI system sometimes the installed protoc version on the system is incompatible with the installed python-protobuf version. This fix, recommended here: https://groups.google.com/g/nanopb/c/YJn3DlY20_g, should take care of this.


mikee47:
@slaff I've addressed this in #2706

PR #2709, opened 2024-01-13
Fix esp8266 timer1 testing

mikee47:
This PR fixes an issue with `HostTests` which hangs consistently when testing Timer1 (used by HardwareTimer) on the Esp8266.

The problem occurs whilst checking each timer for consistency against system time.
This happens because the test code doesn't set a callback interrupt routine.
It is only a problem on the Esp8266 because the callback routine is set directly on the interrupt vector.
Other architectures use indirection and so do nothing if the callback isn't set.
NB. Doing this with the Esp8266 would increase interrupt latency which would affect PWM timing, etc.

Normal code uses the `HardwareTimer` class which contains logic to ensure this doesn't happen during setup.

Timer1 has only a 23-bit counter so fails test with /16 divisor as it wraps at around 1.7s (test duration is 2 seconds).
Fixed by restricting duration to timer maximum (less 1ms).

Elapsed tick calculation check also improved by first checking whether timer is an UP or DOWN counter,
and using timer maximum tick value to handle wraps correctly.

This PR also reduces the number of test loops on the Host (from 2000 to 50).

Note: Bug was introduced in #2456. Timer1 needed to be correctly configured (but inactive) to produce correct result for Esp32.

### DIFF
--- a/Sming/Core/HardwareTimer.h
+++ b/Sming/Core/HardwareTimer.h
@@ -141,11 +141,10 @@ private:
 template <hw_timer_clkdiv_t clkdiv, HardwareTimerMode mode> uint8_t Timer1Api<clkdiv, mode>::state;
 template <hw_timer_clkdiv_t clkdiv, HardwareTimerMode mode> uint32_t Timer1Api<clkdiv, mode>::interval;
 
-template <hw_timer_clkdiv_t clkdiv = TIMER_CLKDIV_16, HardwareTimerMode mode = eHWT_NonMaskable>
-
 /**
  * @brief Hardware Timer class template with selectable divider and interrupt mode
  */
+template <hw_timer_clkdiv_t clkdiv = TIMER_CLKDIV_16, HardwareTimerMode mode = eHWT_NonMaskable>
 using HardwareTimer1 = CallbackTimer<Timer1Api<clkdiv, mode>>;
 
 /**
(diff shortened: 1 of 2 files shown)


mikee47:
@slaff Must remember to run HostTests on all chips (esp8266, rp2040, esp32, esp32c3, esp32s2 and esp32s3 if available) to catch stuff like this! Certainly before a major/minor release I think?

mikee47:
Looking again at this today I've found an issue with Timer1 on the esp8266, for exactly the reason I gave above.
This code (modified `Basic_Blink` sample) doesn't work as expected:

```
#include <SmingCore.h>
#include <HardwareTimer.h>

#define LED_PIN 2

HardwareTimer procTimer;

bool state = true;

void IRAM_ATTR blink()
{
	digitalWrite(LED_PIN, state);
	state = !state;
}

void init()
{
	Serial.begin(COM_SPEED_SERIAL);
	Serial.systemDebugOutput(true);

	debug_i("Welcome");

	pinMode(LED_PIN, OUTPUT);
	procTimer.initializeMs<250>(blink).startOnce();

	auto timer = new SimpleTimer;
	timer->initializeMs<1000>([]() { debug_i("Hello"); });
	timer->start();
}
```

We expect the LED to toggle once then stop, but instead it keeps flashing at ~1.7s intervals. This is because the timer is still enabled and just keeps wrapping. Confirmed using `HardwareTimer1<TIMER_CLKDIV_256>` and we get ~27s intervals.

So timer needs to be explicitly disabled in the callback.

Adding this to the driver code is going to add latency, so will need to take a closer look at this. Suggestions welcome!

slaff:
@mikee47 ping me when ready.

mikee47:
@slaff OK, done with this PR.

PR #2712, opened 2024-01-15
Fix Adafruit_GFX `writeColor` implementation for ESP32

### LINKED ISSUE #2711: AdafruitGFX  `fillscreen()`, `drawFastVLine()`, `drawFastHLine()` working only partially (ESP32)
Sample: ScreenTFT_ILI9340-ILI9341
In that sample, fillscreen only works with color 0, all other colors draw only a very short line of the color chosen.
This issue happens also with functions drawFastVLine(...) drawFastHLine(...)
Drawing rectangles is also affected.
Is it my setup ? Any body able to reproduce this behavior ?

mikee47:
Fixes #2711

The problem occurs with a code path for ESP32 which buffers up to 32 pixels at a time for transfer. However, the first call to 'writePixels' clears the buffer to black (0 - empty response data). If the fill colour is black no abnormal operation is observed, but any other colour and at most 32 pixels get filled with the requested colour.

The benefit of these opimisations is minimal so the library is patched to simplify code path for all architectures.

### DIFF
--- a/Sming/Libraries/.patches/Adafruit_GFX.patch
+++ b/Sming/Libraries/.patches/Adafruit_GFX.patch
@@ -1,5 +1,5 @@
 diff --git a/Adafruit_SPITFT.cpp b/Adafruit_SPITFT.cpp
-index b78d5ce..d2d02b1 100644
+index b78d5ce..f919595 100644
 --- a/Adafruit_SPITFT.cpp
 +++ b/Adafruit_SPITFT.cpp
 @@ -35,6 +35,10 @@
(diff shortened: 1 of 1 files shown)


mikee47:
The problem occurs with a code path for ESP32 which buffers up to 32 pixels at a time for transfer. However, the first call to 'writePixels' clears the buffer to black (0 - empty response data). If the fill colour is black no abnormal operation is observed, but any other colour and at most 32 pixels get filled with the requested colour.

The benefit of these opimisations is minimal so the library is patched to simplify code path for all architectures.

Issue #2720, opened 2024-03-01
vPortFree 517 line assert fail


pljakobs:
I'm experiencing the above error when running an OTA on my firware. The OTA seems to not update the spiffs partition, but I'm not sure if this is a result of the above Error.
What I could find about the error is a very old issue on NONOS 3.0.0 here: https://github.com/espressif/ESP8266_NONOS_SDK/issues/243

Is that a patch that needs to be cherry-picked into Sming, is it something unavoidable like I'm using too much heap or is it something else? 

I know I have some memory pressure, running with a free heap of about 16kB, I could terminate all tcp connections before attempting the OTA but I'm wondering if that's really the only way.

pj

mikee47:
The SDK issue you mention should not affect Sming code. It seems to boil down to whether IRAM can be used by the allocator. According to the SDK documentation this feature was added in version 3.0 but requires applications to define a function `user_iram_memory_is_enabled()` which returns 1. This function isn't defined in Sming so allocations should always come from regular DRAM.

The `OsMessageInterceptor` class was added to deal with this sort of problem by catching the relevant SDK messages and breaking into the debugger or printing a stack dump. Its use is demonstrated in the `LiveDebug sample`.

Below are the bits you'd need to add to your application.

```
#include <Platform/OsMessageInterceptor.h>

static OsMessageInterceptor osMessageInterceptor;

/**
 * @brief See if the OS debug message is something we're interested in.
 * @param msg
 * @retval bool true if we want to report this
 */
static bool __noinline parseOsMessage(OsMessage& msg)
{
	m_printf(_F("[OS] %s\r\n"), msg.getBuffer());
	if(msg.startsWith(_F("E:M "))) {
		Serial.println(_F("** OS Memory Error **"));
		return true;
	}
	if(msg.contains(_F(" assert "))) {
		Serial.println(_F("** OS Assert **"));
		return true;
	}
	if(msg.contains(_F("vPortFree"))) {
		Serial.println(_F("** vPortFree **"));
		return true;
	}
	return false;
}

/**
 * @brief Called when the OS outputs a debug message using os_printf, etc.
 * @param msg The message
 */
static void onOsMessage(OsMessage& msg)
{
	// Note: We do the check in a separate function to avoid messing up the stack pointer
	if(parseOsMessage(msg)) {
		if(gdb_present() == eGDB_Attached) {
			gdb_do_break();
		} else {
#ifdef ARCH_ESP8266
[...]

PR #2643, opened 2023-05-26
Improve code with `std::unique_ptr`

mikee47:
This PR addresses #2616 using the appropriate C++ mechanisms.
Also some tidying.

PR #2722, opened 2024-03-06
Fix HttpUpgrader duplicate stream de-allocation

mikee47:
This PR fixes the issue raised in #2720, where a stream object gets free'd twice resulting in unpredictable behaviour.

Looks like this bug was introduced in #2643.


slaff:
> Looks like RbootHttpUpdater is obsolete and duplicate code. Should we remove it?

Yes, but I would prefer to do it in a major version. Maybe I should open a ticket for the 6.0.0 version and add this one ...

PR #2729, opened 2024-03-11
PartitionStream requires blockErase, other OTA-related improvements

mikee47:
Leaving `blockErase` parameter at `false` when constructing a `PartitionStream` will result in corruption unless partition has already been erased. This is not always immediately apparent so can be difficult to diagnose.

- **Breaking change** Replacing `blockErase` with an enumerated `mode` parameter enforces read-only behaviour by default (new feature and a good thing), with write access more explicitly defined. It's a simple change to existing code and will ensure it gets manually checked to ensure correct behaviour.

- Update the `Basic_Ota` sample to check that an OTA update is not being attempted on the active running partition. This can happen with esp8266, for example, when running in temporary boot mode. (NB. Other checks may include active filing system partitions but that is application-specific.)

- Update notes on migrating from Sming 4.2 to include updating boot sector. (Learning points froim #2727.)

- Remove redundant flashsize calculation from rboot and update version string to `Sming v1.5` (from v1.4.2). Note this should have been done in #2258.

- Fix rboot README regarding use of slot 2, should correspond with subtype `ota_2` not `ota_1`.

### DIFF
--- a/Sming/Arch/Esp8266/Components/esp8266/startup.cpp
+++ b/Sming/Arch/Esp8266/Components/esp8266/startup.cpp
@@ -18,6 +18,7 @@
 extern void init();
 extern void cpp_core_initialize();
 
+// Normal entry point for user application code from SDK
 extern "C" void user_init(void)
 {
 	// Initialise hardware timers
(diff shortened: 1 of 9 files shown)


slaff:
@mikee47 Can you add the braking change to an upgrade document: https://sming.readthedocs.io/en/latest/upgrading/index.html ?

PR #2730, opened 2024-03-12
[WIP]rboot missing `RBOOT_BIG_FLASH` definition

mikee47:
Bootloader code never gets `BOOT_BIG_FLASH` definition. This is evident by unexpected absence of `rBoot Option: Big flash` during (non-silent) boot. Looks like this crept in with #1817.

Prior to Sming 4.3 this is quite important as it affects how rboot determines the active flash memory size. If it needs to reset the boot configuration it won't be correct.

With 4.3 the boot configuration is validated against the partition table so largely unaffected. However, if performing OTA update on a device running an earlier bootloader then the flash layout may not be as expected and addresses could be wrong.


mikee47:
Reviewing the current `rboot.c` we could probably remove `flashsize` calculation altogether...

PR #2736, opened 2024-03-17
Use `std::make_unique`

mikee47:
This PR improves use of `std::unique_ptr` by using `std::make_unique` where appropriate instead of `reset()`, and also updating some class constructors to use direct assignment.

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r23-use-make_unique-to-make-unique_ptrs

### DIFF
--- a/Sming/Arch/Esp32/Services/Profiling/TaskStat.cpp
+++ b/Sming/Arch/Esp32/Services/Profiling/TaskStat.cpp
@@ -27,7 +27,7 @@ struct TaskStat::Info {
 TaskStat::TaskStat(Print& out) : out(out)
 {
 #if CONFIG_FREERTOS_USE_TRACE_FACILITY && CONFIG_FREERTOS_GENERATE_RUN_TIME_STATS
-	taskInfo.reset(new Info[2]);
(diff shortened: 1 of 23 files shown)


slaff:

mikee47:
Looking at the Codacy issues:

**Member variable 'Decompressor::dict' is not initialized in the constructor.**

Don't care. block of memory gets overwritten. No advantage really in spending time zeroing it out.

**Condition '!decompressor' is always false**

That implies `std::make_unique` can throw exceptions. I've tested it running on Esp8266 and can confirm there is a difference. For example:

```
	auto mem = new uint8_t[128000];
```

Gives `mem = nullptr`, whereas this:

```
	auto mem = std::make_unique<uint8_t[]>(128000);
```

Raises a fatal `STORE_PROHIBITED` exception.

I'm changing this PR to WIP as we need to consider this further.

mikee47:
OK, so the reason why `new uint8_t[]` and `std::make_unique` differ is purely because the second one default-initialises the allocated data. If the allocation fails, then attempting to write to location 0 (null) raises the exception. I was aware of this distinction whilst making these changes so removing the WIP tag.

PR #2739, opened 2024-03-21
Provide consistent serial console interface for samples

mikee47:
This PR provides a consistent mechanism for a basic serial console interface using the `LineBuffer` class.

**Refactor LineBuffer**

Implement LineBufferBase so we take non-trivial methods out of header.

**Add `bool`, `String` operators, `printTo` method&**

Simplify usage

**Add `process`, `processKey` methods and update samples**

Use LineBuffer for LiveDebug, Basic_Ota and Basic_FlashIP samples.
These new methods are used to simplify application code.

Note that the `CommandProcessing::Handler` class hasn't been updated here,
mainly because it requires customisation of EOL handling (setCommandEOL, getCommandEOL).
Presumably this is because of the variation in EOL sequences (CR, LF, CRLF or LFCR combinations).
The `processKey` method treats any of these the same way.


slaff:
@mikee47 can you rebase?

PR #2742, opened 2024-03-24
Address CI build warnings

mikee47:
This PR addresses warnings which have crept into the CI build logs.
Worst case is probably esp32 with IDF 5.2 (GCC 13.2) producing 350 warnings, now 65.

**General fixes**

- Remove deprecated calls to command processing library from samples
- Fix initialisation order and type names: still some instances of e.g. `uint8` which should be `uint8_t`.
- Discovered component samples aren't built for Host in CI! Fixed this - note build time extended accordingly.
- Simplify nested conditionals https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f56-avoid-unnecessary-condition-nesting

**Fix compiler warnings/errors**

Replaced a few more instances of printf with streaming operators, simpler and type-safe.
Fixed `Overloaded virtual` warnings.

**Note**: See remaining warnings for NimBLE (submodule) which need attention.

**Graphics library fixes**

- Fix control rendering: Should use their own bounds for positioning, set button font
- Fix const parameters
- use unique_ptr for getGlyph calls
- Use `std::make_unique` and assignment instead of reset()
- Avoid printf
- Demote `reinterpret_cast` to `static_cast` where appropriate
- Apply coding style to samples
- Fix compiler warnings

**Disable esp32 warning about RWX segments**

Check introduced with linker (LD, binutils 2.39) from IDF 5.2 onwards
This specifically refers to IRAM, which we do actually need to be writeable and executable.
(IDF link code disables this warning also.)

### DIFF
--- a/Sming/Arch/Esp32/app.mk
+++ b/Sming/Arch/Esp32/app.mk
@@ -9,6 +9,11 @@ LDFLAGS	+= \
 	-nostdlib \
 	-Wl,-static
 
+ifeq ($(IDF_VERSION),v5.2)
+LDFLAGS += \
+	-Wl,--no-warn-rwx-segments
+endif
+
 ifdef IDF_TARGET_ARCH_RISCV
 	LDFLAGS += \
 		-nostartfiles \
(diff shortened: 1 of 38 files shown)


slaff:
> Shall we do that here or in a separate PR? ... Also readthedocs can be updated.

@mikee47 I would prefer separate PRs. Thank you for your hard work.

mikee47:
@slaff Yes, all done thanks.

Issue #2751, opened 2024-03-31
LwIP seems to not correctly handle packages larger than MTU


pljakobs:
Due to an ever growing json structure, my client sends an ip package larger than the MTU of 1500 bytes (1960 bytes to be precise). Due to path MTU being on by default, my client ip stack happily sends this package out to the wire, expecting to be told if that's not what is expected. 
On Sming, when I check the body length, I get 1497 bytes, so the rest is basically just cut off.

Expected behaviour:
the lwip stack should quietly discard the received package and respond to the sender with an ICMP "fragmentation required" package, allowing the sender to configure the best package size for this receiver.

I have not yet dug into the lwip debug for this, but the observed behaviour clearly points toward lwip not helping with path discovery.

pljakobs:
(testing lwip options)

pljakobs:
further checking with ENABLE_LWIPDEBUG=! I see
```
CONNECTION DROPPED
	(free heap: 7560)


15829943 tcp_abandon: sending RST

15830431 tcp_rst: seqno 6515 ackno -1212338775.

15830887 

CONNECTION DROPPED
	(free heap: 7848)


15831068 tcp_abandon: sending RST

15831469 tcp_rst: seqno 6516 ackno 1612911029.

```
but the sequence numbers don't match with what I see in wireshark (which might be because lwip debug outputs them as signed integers :o)

mikee47:
Maybe unrelated to your issue, but there's clearly memory pressure. Assuming this is your `esp_rgbww_firmware`, take a look in your `.rodata` segment: `objdump -s -j .rodata out/Esp8266/debug/build/app_0.out`. I'd guess there's 3K+ of RAM you can reclaim by moving strings into flash.

pljakobs:
Good point, also, the original design keeps the whole configuration in a large struct which alone is at least around 1.5kB and is not really used that much, so I should probably refractor this as well. 4.5kB more RAM is 25% more than now, so that should certainly be helpful.

pljakobs:
I have moved almost all string literals into F() and now have mostly >20kB of heap free, so that was well worth while the effort. (as a side note: the F macro doesn't work for debug_ - why would that be?)
lwip still doesn't deal well with my oversized packets, though. 
I am actually confused about this, in my network analysis days, senders would stick to their local MTU but it seems that modern stacks try path MTU whenever possible and negotiate the largest possible frame size.

mikee47:
[...]

PR #2754, opened 2024-04-02
Fix building `lwip2` with `ENABLE_LWIPDEBUG=1`

mikee47:
Build fails with errors due to missing definitions for debug output.


mikee47:
As discussed in #2751 the esp8266 LWIP2 stack needs updating but this PR can at least allow debugging to be used with the current version.

PR #2756, opened 2024-04-04
Improve sample applications

mikee47:
This PR updates all the main Sming samples to try and demonstrate best practice when coding Sming applications. Specifically:

- Initialise timers using templated methods where possible as this provides static range check on values
- Using `SimpleTimer` is sufficient in most cases. This is both more efficient and avoids complication where compiler cannot distinguish between delegates and callback functions.
- Use timer `startOnce()` method rather than `start(false)`
- Use C++ iterators in preference to C-style loops `for(int i=0; i<list.count(); ++i) {...}`
- Ensure consistent use of `WIFI_SSID` and `WIFI_PASSWORD`
- Use `Serial` methods in preference to `debug_x` where intent is not actually for debugging
- Use anonymous namespaces and remove redundant static declarations
- Avoid nested `if` statements
- Place `init()` function at end of main application source file
- Reduce variable scope (declare at point of first use)
- static/global variables do not require explicit initialisation to 0

Sample-specific changes:

- Revise `HttpServer_AJAX` so that GPIO numbers are consistent with web code
- Update Basic_Ssl keys, requires Bearssl
- Update Basic_DateTime sample
  - Simplify code using LineBuffer
  - Accept either numeric timestamp or HTTP date string. Note: ISO8601 string conversion not currently supported by `DateTime`.
  - Don't emulate console via telnet, use default CLI
- Rewrite Basic_Neopixel sample
  - Enumerate color values
  - Reduce global variables and duplicated code
  - Verify logic flow using Host
- Rewrite Display_TM1637 as state machine (coroutine)
- Fix Basic_Storage, didn't actually use RAM as indicated
- Add Host `Print` stream support, update `Basic_Utility` sample

Other changes

- Update `jerryscript` library to latest. CI improvements and removes deprecation warnings.
- Fix bad memory casts in si4432 library. Also builds for other architectures than esp8266, so remove SOC restrictions.
- Fix unused variable warning in `CommandProcessing` library


slaff:
Awesome work @mikee47 ! I will be more active next work and start reviewing the other PRs. Please, be patient with me as I am away from a proper computer at the moment :)

mikee47:
@slaff No worries, we all have lives to lead!

PR #2757, opened 2024-04-04
Allow pillow upgrade to latest

mikee47:
This PR changes documentation build to use graphviz instead of seqdiag, allowing pillow upgrade to latest and satisfying #2755.


slaff:
@mikee47 The old sequence diagram generated with `sphinxcontrib.seqdiag` looks like a proper sequence diagram. The new one is nothing close to a sequence diagram. I would prefer to have problematic version of pillow than having a bad sequence diagram. Therefore I am not keen on merging this PR. Maybe we have to search in internet if someone has already attempted to migrate `sphinxcontrib.seqdiag` to work with latest `pillow`

slaff:
@mikee47 I have another suggestion - let's just add the generated sequence diagram from `sphinxcontrib.seqdiag` to the docs and then get rid of `sphinxcontrib.seqdiag` or any replacements. Plus upgrade pillow to non-vulnerable version.  If this sounds reasonable to you then, please, change this PR to reflect these changes.

mikee47:
@slaff OK, can always revisit if seqdiag gets fixed

slaff (approved):

PR #2764, opened 2024-04-12
esp8266: NMI not disabled when timer1 ISR cleared

### LINKED ISSUE #2763: HostTests stalling on esp8266 `Waiting for timer1 callback test to complete`
This problem has been tracked back to PR #2712. Prior to that, the test code didn't actually wait for the timer callback to fire before continuing, but it still doesn't fire there.

Simplest way to verify this issue is to remove all tests from `include/modules.h` except `Timers`. Can also build with `DISABLE_NETWORK=1` - setting makes no difference to the issue but reduces code size.

With the following changes and the problem goes away:

```
diff --git a/tests/HostTests/modules/Timers.cpp b/tests/HostTests/modules/Timers.cpp
 void REGISTER_TEST(Timers)
 {
-	registerGroup<CallbackTimerApiTest<Timer1TestApi>>();
+	// registerGroup<CallbackTimerApiTest<Timer1TestApi>>();
 	registerGroup<CallbackTimerApiTest<OsTimerApi>>();
 	registerGroup<CallbackTimerApiTest<OsTimer64Api<Timer>>>();
 
-	registerGroup<CallbackTimerSpeedTest<HardwareTimerTest>>();
+	// registerGroup<CallbackTimerSpeedTest<HardwareTimerTest>>();
 	registerGroup<CallbackTimerSpeedTest<SimpleTimer>>();
 	registerGroup<CallbackTimerSpeedTest<Timer>>();
 
 	registerGroup<CallbackTimerTest>();
 }
``` 

The issue appears to occur when the hardware timer is disarmed, then rearmed multiple times.
I don't know whether this is a hardware problem with my specific device, but unlikely.

Investigations continue...

mikee47:
Fixes #2763. 

SDK doesn't actually disable NMI interrupts when interrupt callback cleared. This blocks regular (FRC) timer1 interrupts if subsequently enabled.

The HardwareTimer (timer1) appears to stop working if applications switch from using NMI interrupts to regular timer interrupts.

A more subtle manifestation of this bug is that the original NMI callback continues to be called when a different FRC handler is set.

TODO:

* [x] Check nothing else gets broken (e.g. Wifi)
* [x] Currently using SDK 3.0.3; does anything change in 3.0.4 or 3.0.5 ?

### DIFF
--- a/Sming/Arch/Esp8266/Components/driver/hw_timer.cpp
+++ b/Sming/Arch/Esp8266/Components/driver/hw_timer.cpp
@@ -36,6 +36,28 @@ static void IRAM_ATTR nmi_handler()
 	nmi_callback.func(nmi_callback.arg);
 }
 
+/*
+ * The `ETS_FRC_TIMER1_NMI_INTR_ATTACH` macro calls SDK `NmiTimSetFunc` which
+ * doesn't actually disable NMI (a known bug).
+ * If we subsequently enable FRC interrupts, the timer won't work so we need to properly
+ * disable NMI manually.
+ *
+ * The NmiTimSetFunc code looks like this:
+ *
+ * 		uint32_t value = REG_READ(NMI_INT_ENABLE_REG);
+ * 		value &= ~0x1f;
+ * 		value |= 0x0f;
+ * 		REG_WRITE(NMI_INT_ENABLE_REG, value);
+ *
+ * Note that there is no published documentation for this register.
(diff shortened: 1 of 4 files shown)


mikee47:
Tested this against various updated SDK versions, all of them have the same `NmiTimSetFunc` implementation.

3.0.4 b1c14cdb : no change
3.0.5 7b5b35da : no change
3.0.5 38fda5f9 tagged release/v3.0.5 : bootloop crashing
master : doesn't boot

Issue #2758, opened 2024-04-08
`time_t` size is inconsistent


mikee47:
## Background

I'm working on some improvements to the `DateTime` class and one issue is with using signed 32-bit values for `time_t`.

First, I'm not sure what size `time_t` actually is, so did a couple of CI builds with a `static_assert(sizeof(time_t) == 8)` statement in `DateTime.cpp`.

So a successful run ✓indicates time_t is 64-bits, and failed run X indicates 32-bits.

(1) Current develop branch
(2) With `_TIME_BITS=64`

Note that all esp32 SOCs have the same result.

(NB. These results are different from my original post, since corrected.)

| Platform     |  (1) | (2)  | GCC      | Newlib version |
| --------     | ---  |---   | -------- | -------------- |
| **Ubuntu**                                             |
| host         |   X  | ✓    |  13.2    |  glibc 2.38    |
| esp8266      |   ✓  | ✓    |  10.2.0  |  4.0.0         |
| rp2040       |   ✓  | ✓    |  10.3.1  |  4.1.0         |
| esp32, 4.3   |   X  | X    |  8.4.0   |  3.3.0         |
| esp32, 4.4   |   X  | X    |  8.4.0   |  3.3.0         |
| esp32, 5.0   |   ✓  | ✓    |  11.2.0  |  4.1.0         |
| esp32, 5.2   |   ✓  | ✓    |  13.2.0  |  4.1.0         |
| **Windows**                                            |
| host         |  X   |  X   |  9.2.0   |   mingw-32      |
| esp8266      |  ✓   |  ✓   | 10.2.0   |  4.0.0         |
| rp2040       |  ✓   |  ✓   | 10.3.1   |                |
| esp32, 4.3   |  X   |  X   |  8.4.0   |                |
| esp32, 4.4   |  X   |  X   |  8.4.0   |                |
| esp32, 5.0   |  ✓   |  ✓   | 11.2.0   |                |
| esp32, 5.2   |  ✓   |  ✓   | 13.2.0   |                |

- For esp8266, rp2040 and esp32 (IDF 5.0+) `time_t` is fixed at 64-bits.
- For esp32 (IDF 4.x) it's fixed at 32-bits.
- For Ubuntu host build, it's configurable via `_TIME_BITS` define.
- For Windows host build, using mingw, fixed at 32-bits.

## Library versions

`time_t` is C-library dependent.
[...]

PR #2768, opened 2024-04-20
Improvements to HostEd

mikee47:
This PR proposes some changes to the Hosted classes

**Update simpleRPC to master**

With clang-tidy #2648 some failures occur in the `simpleRPC` library. Updating this to current master solves the issues.
I've also put all the code into the `simpleRPC` namespace which sorts out the conflict with `Vector` but makes it available for use if required.

**Tidy HostTests module**

Also decode the anonymous 'packet' blob so we can see what's in it and compare with spec. should we wish to update it.

**Use abstract base class for callbacks**

Simplifies implementation since usually all callbacks are required.
Compiler ensures all methods have implementations.
`clang-tidy` gave potential memory leak indication because of Delegates. Not sure if that's real but this fixes it.


slaff:
@mikee47 I am trying to compile the Basic_Blink using the command below and I get compilation error?!

```
cd $SMING_HOME/../samples/Basic_Blink
make SMING_ARCH=Host ENABLE_HOSTED=tcp HOSTED_SERVER_IP=192.168.4.1
```

The error is: 
```
static_assert(sizeof(time_t) == 8, "Expecting 64-bit time_t");
```

Any idea how to work around this issue?

mikee47:
> ... Expecting 64-bit time_t

I'm going to guess your version of GCC is old; I'm with Fedora on GCC 13.2.1 - what version are you using?

slaff:
>  what version are you using?

Gcc version 9.4.0. I will check if I can upgrade on my Ubuntu 20.04 LTS ...

slaff:
Gcc version 10.5.0 is the latest GCC that I can use ... Hm.. have to think about OS upgrade ...

mikee47:
> Gcc version 10.5.0 is the latest GCC that I can use ... Hm.. have to think about OS upgrade ...

That should be fine https://github.com/SmingHub/Sming/issues/2758

You can just comment out that assert - I put it there deliberately so we'd know about this sort of thing!

mikee47:
@slaff You'll doubtless have run into the other issue attempting to build `Basic_Blink` for Hosted using a `tcp` connection: `DISABLE_NETWORK` is hard-coded in `component.mk` and has to be changed. I'm thinking that's probably just something to bear in mind using Hosted.

slaff (approved):

PR #2769, opened 2024-04-22
Make `time_t` size check message more helpful

mikee47:
Further to #2758 host builds should generally match code run on hardware. As mentioned this is not straightforward for Windows, but for Linux we should expect 64-bits as the default.

GCC 10+ can make use of `_TIME_BITS` to ensure we get this behaviour. More specifically, this is a feature of `glibc` shipped with those versions.

For earlier versions, even though the compiler is 64-bit we are building in 32-bit mode and `time_t` stays stuck in 32-bits. There are various discussions about this but the short answer is that prior to GCC 10 (libc 4) 32-bit applications get 32-bit `time_t`.

Instead of generating a `static_assert` (which stops compilation) we get a compiler warning - this still requires building in STRICT mode.

### DIFF
--- a/Sming/Core/DateTime.cpp
+++ b/Sming/Core/DateTime.cpp
@@ -18,7 +18,7 @@
 static_assert(sizeof(time_t) != 8, "Great! Now supports 64-bit - please update code");
 #warning "**Y2038** time_t is only 32-bits in this build configuration"
 #else
-static_assert(sizeof(time_t) == 8, "Expecting 64-bit time_t");
+static_assert(sizeof(time_t) == 8, "Expecting 64-bit time_t - please use GCC 10 or later");
 #endif
 
 namespace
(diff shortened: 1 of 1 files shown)


Sming/Core/DateTime.cpp
```diff
@@ -18,7 +18,7 @@
 static_assert(sizeof(time_t) != 8, "Great! Now supports 64-bit - please update code");
 #warning "**Y2038** time_t is only 32-bits in this build configuration"
 #else
```
slaff, line 20: If GCC v10 is ok then wouldn't it be better to change this line to

```
#elif defined(__GNUC__) && __GNUG__ < 10
```

slaff:
> Are you OK with making this a hard assert? There may be a way to generate a deprecation message or warning.

@mikee47 I would prefer a warning.

Sming/Core/DateTime.cpp
mikee47: Perhaps... but as this is technically a glibc issue the more correct way would be to check `__USE_TIME_BITS64` - see https://lwn.net/Articles/664800/.

PR #2796, opened 2024-06-10
Improve FlashString portability

mikee47:
This PR removes copy support from the FlashString library. Doing so simplifies the code, improves performance and portability to other compilers (e.g. clang).

The FlashString library previously supported copies (references) like this::

```
 FlashString emptyString;
 FlashString stringCopy(FS("Inline string"));

 DEFINE_FSTR_DATA_LOCAL(flashHelloData, "Hello");
 auto myCopy = flashHelloData;
```

These will now fail to compile.
Copy construction and assignment has been explicitly deleted so avoid unintentional side-effects.
Objects should always be passed by reference.

This change has the additional benefit of catching pass-by-copy errors. These function/method templates have been fixed:

- `fileSetContent`
- `IFS::File::setAttribute`
- `IFS::FileSystem::setUserAttribute`
- `IFS::ArchiveStream::setUserAttribute`

### DIFF
--- a/Sming/Core/FileSystem.h
+++ b/Sming/Core/FileSystem.h
@@ -235,7 +235,8 @@ template <typename TFileName> inline int fileSetContent(const TFileName& fileNam
 	return fileSystem->setContent(fileName, content, length);
 }
 
-template <typename TFileName, typename TContent> inline int fileSetContent(const TFileName& fileName, TContent content)
+template <typename TFileName, typename TContent>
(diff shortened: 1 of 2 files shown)


slaff:
@mikee47 can you add this information also to the UPGRADING document?

mikee47:
> @mikee47 can you add this information also to the UPGRADING document?

? Already there...

PR #2802, opened 2024-06-12
SMTP Client improvements.

slaff:
- Continue plain-text communication if SSL is not compiled
- Fix memory leaks in MailMessage
- Expose `TcpConnection::enableSsl(hostName)` which allows a plain-text TCP connection to be changed to SSL one after the tcp connection is established.

### DIFF
--- a/Sming/Components/Network/src/Network/MailMessage.h
+++ b/Sming/Components/Network/src/Network/MailMessage.h
@@ -38,6 +38,15 @@ public:
 	String subject;
 	String cc;
 
+	~MailMessage()
+	{
+		delete stream;
+		for(auto attachment: attachments) {
+			delete attachment.headers;
+			delete attachment.stream;
+		}
+	}
+
 	/**
 	 * @brief Set a header value
 	 * @param name
--- a/Sming/Components/Network/src/Network/MailMessage.h
+++ b/Sming/Components/Network/src/Network/MailMessage.h
@@ -40,8 +40,7 @@ public:
 
 	~MailMessage()
 	{
-		delete stream;
-		for(auto attachment : attachments) {
+		for(auto& attachment : attachments) {
 			delete attachment.headers;
 			delete attachment.stream;
 		}
@@ -115,7 +114,7 @@ public:
 	MailMessage& addAttachment(IDataSourceStream* stream, const String& mime, const String& filename = "");
 
 private:
-	IDataSourceStream* stream = nullptr;
+	std::unique_ptr<IDataSourceStream> stream = nullptr;
 	HttpHeaders headers;
 	Vector<MultipartStream::BodyPart> attachments;
 };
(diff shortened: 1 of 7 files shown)


Sming/Components/Network/src/Network/MailMessage.h
```diff
@@ -38,6 +38,15 @@ class MailMessage
...
 
+	~MailMessage()
+	{
+		delete stream;
+		for(auto attachment : attachments) {
+			delete attachment.headers;
+			delete attachment.stream;
+		}
+	}
+
```
mikee47, line 49: Looks like `stream` should be `unique_ptr`. Also not sure it's necessary to destroy attachments, should happen automatically when vector is destroyed.

Sming/Components/Network/src/Network/TcpConnection.cpp
```diff
@@ -423,6 +423,29 @@ err_t TcpConnection::internalOnConnected(err_t err)
...
+		return false;
+	}
+
+	if(!sslCreateSession()) {
+		return false;
+	}
+
+	if(hostName) {
+		ssl->hostName = hostName;
+	}
```
mikee47, line 442: Is check necessary? Or is that in case there's an `sslInit` callback registered which might do it. Maybe document to clarify? Or Just assign `ssl->hostName = hostName`

Sming/Components/Network/src/Network/TcpConnection.cpp
```diff
@@ -423,6 +423,29 @@ err_t TcpConnection::internalOnConnected(err_t err)
...
+
+	useSsl = true;
+	useSsl = (internalOnConnected(ERR_OK) == ERR_OK);
```
mikee47, line 445: Remove redundant `useSsl = true;`

slaff: I know that the code looks a bit wierd but in `internalOnConnected()`  useSsl has to be true to kick start the SSL connection.

```C++
err_t TcpConnection::internalOnConnected(err_t err)
{
	debug_tcp_d("connected: useSSL: %d, Error: %d", useSsl, err);

	if(useSsl && err == ERR_OK) {
```

mikee47: May I suggest a comment and/or slight rewrite:

```
useSsl = true;
if (internalOnConnected(ERR_OK) != ERR_OK) {
  useSsl = false;
}
```

slaff: > May I suggest a comment and/or slight rewrite:

Sure. I will apply your change in a minute.

Sming/Components/Network/src/Network/MailMessage.h
mikee47: Just done some checking and of course attachment has no destructor...

slaff:
@mikee47 shall I merge this PR with the applied review changes?

PR #2827, opened 2024-06-22
Minor websocket client issues

mikee47:
This PR adds a few fixes for websockets I seem to have kicking about.

- Tidy, remove unused websocket strings
- `wsserver` tool missing port argument handling, default is 8000 not 9999

**Check return value from http `connect()`**

 I can force this to fail by calling 'connect' twice in succession, though I do notice a disparity here. If the http connection is 'processing' then `WebsocketClient::connect` returns `false`; however, `HttpClientConnection::connect` returns `true` in this situation.

TODO:

* [x] ~~Document all the various `connect` methods and clarify meaning of return value.~~ Outside scope of this PR. See #2829

### DIFF
--- a/Sming/Components/Network/component.mk
+++ b/Sming/Components/Network/component.mk
@@ -98,7 +98,7 @@ endif
 
 # Websocket Server
 CACHE_VARS			+= WSSERVER_PORT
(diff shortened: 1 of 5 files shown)


slaff:

PR #2850, opened 2024-07-02
Add IDF caching for libraries

mikee47:
Cache esp-idf and tools for library builds. Be nice if we could use the cache from `SmingHub/Sming` directly but that sort of cross-repo sharing isn't supported.

Scanlog tool:

- Fix scanlog tool windows path normalisation - didn't account for change from c:/tools/ to d:/opt/, and that Windows idf path contains version.
- Print warnings for each job by default, add 'm' flag to merge them

Also fix a few more warnings.

### DIFF
--- a/.github/workflows/library.yml
+++ b/.github/workflows/library.yml
@@ -71,7 +71,7 @@ jobs:
     env:
       SMING_ARCH: ${{ matrix.arch }}
       SMING_SOC: ${{ matrix.variant }}
-      INSTALL_IDF_VER: ${{ matrix.idf_version }}
+      INSTALL_IDF_VER: ${{ matrix.idf_version || '5.2' }}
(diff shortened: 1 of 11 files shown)


mikee47:
I'm actually impressed with that summary from 'what-the-diff' - it's actually useful :-)

mikee47:
Changing to WIP, library caching not quite working as expected for Ubuntu.

mikee47:
Yep, turning build debugging on it's using relative paths when restoring the cache. Looks like the cache for IDF was saved in Windows and ended up being restored to a sub-folder somewhere. I'll have a dig about and see if the damn thing can use absolute paths like it's been given.

mikee47:
Hang on, what's `varibale` ? I thought AI could spell...

Issue #2514, opened 2022-05-24
Some idf issues


Asanga-Viraj:
@mikee47 
I am working on ESP32 using Sming framework. I saw few things (Win 10).

- Sming `choco install -y sming.esp32` doesn't install [idf-v4.3](https://github.com/mikee47/esp-idf/tree/sming/release/v4.3) from your repository. It installs idf-v4.1. 
- Sming re-installation doesn't clean and reinstall ESP32 components.
- Saw many issues with current version such as System.restart, WiFi.Disconnect, etc.

Also just a suggestion, it is better if you can make Sming support with latest release idf (like idf-v4.4.1).

Thanks

Asanga-Viraj:
@slaff 
I found a issues in ESP32. `WifiStation.connect()` is not working. I found `esp_wifi_connect()` is not calling on this method. My suggestions are below.

- https://github.com/SmingHub/Sming/blob/4.5.0/Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp#L120  `return esp_wifi_start() == ESP_OK;`
- https://github.com/SmingHub/Sming/blob/4.5.0/Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp#L125-L126
 `if(isConnected())
disconnect();`
`ESP_ERROR_CHECK(esp_wifi_connect());`
`return true;`


Thanks

slaff:
@Asanga-Viraj thank you for your remarks.

> Sming choco install -y sming.esp32 doesn't install [idf-v4.3](https://github.com/mikee47/esp-idf/tree/sming/release/v4.3) from your repository. It installs idf-v4.1.

This should be fixes today with [this commit](https://github.com/slaff/chocolatey-packages/commit/40ebba32be0ad4ea9fe393b4d4d79fa4b8577d74).

> Sming re-installation doesn't clean and reinstall ESP32 components.

sming.esp32 should be the only package responsible for this. 

>  it is better if you can make Sming support with latest release idf (like idf-v4.4.1).

We usually skip a minor version and jump to the next one. But if you have the time and desire you are free to submit a PR with the needed changes.

kmihaylov:
@Asanga-Viraj do you have some progress on System.restart() ?

On my ESP8266 board I have to disconnect from the wifi network before calling System.restart() in order not to trigger (some) WD timer. I have to try wifi-less app with Esp32 to verify this.

PR #2859, opened 2024-07-13
Update windows installer

mikee47:
The Windows installer makes use of custom chocolatey packages, which are out of date. See issue #2514.
These packages are separate from the main Sming repository and represent an un-necessary maintenance overhead.

This PR updates the installation scripts so these custom chocolatey packages are no longer required.
The scripts were previously only used for CI testing so have been reworked appropriately.
The installation procedure is now similar to that for linux/MacOS.

The initial bootstrap installation remains unchanged: The ``choco-install.cmd`` script is fetched from the Sming ``develop`` branch and executed.

Note: Contrary to other installs and the documentation, the current sming choco script installs the `master` branch.

The user runs a second install script `tools\install` to install required toolchains.
This is separate because:

- Administrative priviledges are not required
- The user gets more control over which toolchains get installed
- The script doesn't change anything outside of the `C:\tools` directory (other than python packages)

The directory can also be customised by setting `SMING_TOOLS_DIR`.

The documentation at https://sming.readthedocs.io/en/latest/getting-started/windows/index.html has been updated to reflect these changes.

**CI Compiler Cache**

I've moved caching to just before the build starts, so that configuration and zeroing ccache is handled by the build script. This is more logical than doing it as part of the install phase.

**Esp32 install/build**

A standard IDF tools installation includes the following packages:

- cmake
- ninja
- ccache

As Sming installs these system-wide, these are not required.
Also, as we now have Ninja in the system path for all platforms some logic in build.mk can be removed.

These additional items are not used:

- idf-exe
- dfu-util

The Sming Esp32 installer omits all the above items, for linux/MacOS also.
Turns out we can just pass the filter `"*elf*"` to the installer to get all the stuff we **do** actually need.

**Python**

This defaulted to `python3`, which failed in testing as current installers don't provide it, so change to `python`.

The `PYTHON` variable now defaults to the full path determined by `which python`. In Windows, the makefile resolves this to something like `/c/Python312/python.exe`.

**MinGW**

I've also added `MinGW-2021-04-12.7z` to the [SmingTools](https://github.com/SmingHub/SmingTools/releases).
This differs from `MinGW-2020-10-19.7z` which was previously used (in both choco package and CI install) in the following ways:

- MinGW version changed from 5.4.1 to 5.4.2
- Files unpack into `./` instead of `MinGw/` so makes it easier to relocate during installation if required
- Contents of `var/cache` removed to reduce size of archive (48MB vs 133MB)

No changes to GCC, still 9.2.0.

**Docs build**

Documentation build fails in Windows, problems with libcairo-2.dll.
Initially, the error was about not finding the DLL.
Some reserach suggests installing gtk+ runtime should fix this.
Still fails but with error "can't find .so".

Leaving this as a known issue, just build in Linux or WSL.

Todo:
* [x] Check library CI builds still work

### DIFF
--- a/.github/workflows/ci-esp32.yml
+++ b/.github/workflows/ci-esp32.yml
@@ -83,8 +83,8 @@ jobs:
     - name: Install build tools for Windows
       if: matrix.os == 'windows-latest'
       run: |
-         . Tools/ci/setenv.ps1
-         Tools/ci/install.cmd
+        . Tools/ci/setenv.ps1
+        Tools/ci/install.cmd
 
     - name: Compiler Cache
       uses: actions/cache@v4
@@ -95,7 +95,7 @@ jobs:
     - name: Build and test for ${{matrix.variant}} with IDF v${{matrix.idf_version}} on Ubuntu / MacOS
       if: matrix.os != 'windows-latest'
       run: |
-        source $SMING_HOME/../Tools/export.sh
+        . Tools/export.sh
         Tools/ci/build.sh
 
     - name: Build and test for ${{matrix.variant}} with IDF v${{matrix.idf_version}} on Windows
--- a/.github/workflows/ci-esp32.yml
+++ b/.github/workflows/ci-esp32.yml
@@ -75,12 +75,6 @@ jobs:
           /opt/esp32
         key: ${{ matrix.os }}-idf-${{ matrix.idf_version }}
 
-    - name: Compiler Cache
-      uses: actions/cache@v4
-      with:
-        path: .ccache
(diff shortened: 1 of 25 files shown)


mikee47:
Note: This PR invalidates ccache information because the timestamp on mingw GCC compiler has changed (even though content hasn't). Also, Esp32 IDF on Windows is now installed in a versioned directory (e.g. `C:\tools\esp-idf-5.2`) with junction (link) `C:\tools\esp-idf` as for linux/MacOS.

Before merging to develop, should wipe all the caches so they get rebuilt.

mikee47:
@slaff I broke the installer for esp32 IDF (all platforms) as only the IDF 5.2 supports wildcards. Will have a fix later today.

PR #2864, opened 2024-07-16
Some more documentation fixes

mikee47:
Sorry, more documentation fixes.

Two stragglers in source/arch and we can remove that directory.

- Move esp-quick-toolchain info to esp8266 page and update.
- arch/host/index is redundant, remove it

Move arch-specific features off index page

Fix broken GCC upgrade URLs - if compiler check fails direct to getting started page.


mikee47:
Test build here https://smingdev.readthedocs.io/en/fix-doc-fixes-minor/

PR #2865, opened 2024-07-17
Update CI ccache on every push to develop

mikee47:
Build times are increasing faster than I expected, so it's clear that to be effective the ccache needs to be kept up to date.
This needs to happen automatically so we can forget about it.

Github caching behaviour is to fail the save operation if a hit occurred on primary key.
In other words, a cache entry will never be overwritten (updated).
There is no provision for a cache update as such: https://github.com/actions/cache/issues/342

The workaround is fairly easy though: break into discrete restore/delete/save steps.

Only the cache for the current branch is deleted, using the `--branch` option to `gh action-cache delete`.
This prevents pull requests from contributors with repo write access from deleting the main `develop` cache.

This table shows the intended progression of cache updates, starting from an empty cache:

| Event   | Branch       | Restore      | Delete  | Save         |
| ------- | ------------ | --------     | ------- | ------------ |
| PR      | feature/new  | -            | -       | feature/new  |
| PR      | feature/new  | feature/new  | -       | -            |
| Merge   | develop      | -            | -       | develop      |
| PR      | fix/lwip     | develop      | -       | -            |
| Merge   | develop      | develop      | develop | develop      |

So always save if (event.branch == develop) or no cache hit

### DIFF
--- a/.github/workflows/ci-esp32.yml
+++ b/.github/workflows/ci-esp32.yml
@@ -109,7 +109,7 @@ jobs:
       run: ccache -sv
 
     - name: Delete Previous Compiler Cache
-      if: ${{ steps.ccache.outputs.cache-hit }}
+      if: github.ref_name == github.event.repository.default_branch && steps.ccache.outputs.cache-hit
       continue-on-error: true
       run: |
         gh extension install actions/gh-actions-cache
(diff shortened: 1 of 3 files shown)


mikee47:
@slaff Still issues with caching:

1. Library caching broken because I moved them out of installers in #2859
2. With 42 jobs, ccache file size limit of 500MB means they'd eventually grow to 21GB.

I'm working on fixes but not worth delaying release 5.2 for them.

slaff:
> I'm working on fixes but not worth delaying release 5.2 for them.

Don't worry, I'll wait.

PR #2870, opened 2024-07-24
Possible DnsServer memory leak reported by coverity

slaff:


mikee47 (approved):

Sming/Components/Network/src/Network/DnsServer.cpp
```diff
@@ -85,6 +85,7 @@ void DnsServer::onReceive(pbuf* buf, IpAddress remoteIP, uint16_t remotePort)
 
 	unsigned requestLen = pbuf_copy_partial(buf, buffer, buf->tot_len, 0);
 	if(requestLen != buf->tot_len) {
+		delete[] buffer;
 		return;
 	}
```
mikee47, line 90: I was chewing on whether this check is actually necessary since by inspection we pass `buf->tot_len` so the call is guaranteed to succeed. Hence this is really just shutting up the static analyser.

If we just remove lines 87-89 the code becomes much simpler.

We should also remove line 101 as I cannot see any use case for the user-provided callback here.

mikee47 (changes requested):

Sming/Components/Network/src/Network/DnsServer.cpp
slaff: Something like these lines below ?
```C++
void DnsServer::onReceive(pbuf* buf, IpAddress remoteIP, uint16_t remotePort)
{
	debug_d("DNS REQ from %s:%d", remoteIP.toString().c_str(), remotePort);

	// Allocated buffer with additional room for answer
	char* buffer = new char[buf->tot_len + MAX_ANSWER_LENGTH];
	if(buffer == nullptr) {
		return;
	}

	unsigned requestLen = pbuf_copy_partial(buf, buffer, buf->tot_len, 0);
	
	debug_hex(DBG, "< DNS", buffer, requestLen);

	auto responseLen = processQuestion(buffer, requestLen);
	if(responseLen != 0) {
		debug_hex(DBG, "> DNS", buffer, responseLen);
		sendTo(remoteIP, remotePort, buffer, responseLen);
	}

	delete[] buffer;
}
```

mikee47: Yes, that's much cleaner!

PR #2887, opened 2024-09-22
Improved checking for uninstalled tools for unsupported linux distributions.

berhoel:
- Support for linux platforms not providing apt or dnf. Installation script searches for required tools and list missing.

- Better support for installing outside "/opt". esp and rp2040 tools are not installed in "/opt" if Sming isn't.

I tried to install Sming on my OpenSuse Tumbleseed system. For one, I did not like the forced usage of "sudo" without knowing what was going on (I had a wrapper installed that provided "apt", using zypper in the background).

I would have liked code better that checks for required tools and asks the user to install them to install them if missing.

Second installing the tools to "/opt" did not work for me with user install. I don't want "/opt" to be user writable, and even installing the tools using "sudo" lead to fails later, when I tried to compile a project for ESP32.

My patch simply looks for required tools if neither apt nor dnf are found and lets the tools install in the Base directory Sming is installed.


berhoel:
I will rework my pull request taking into account the failed CI run.

Tools/install.sh
```diff
@@ -99,12 +99,34 @@ elif [ -n "$(command -v dnf)" ]; then
     DIST=fedora
     PKG_INSTALL="sudo dnf install -y"
 else
+    _OK=1
```
mikee47, line 102: `OK` is a bit terse. `TOOLS_MISSING` would be better.
Please don't use _ idenfiers - follow the style of existing script.

Tools/install.sh
```diff
@@ -99,12 +99,34 @@ elif [ -n "$(command -v dnf)" ]; then
...
+            python3 \
+            pip3 \
+	    wget \
+	    )
+    for _TOOL in "${_REQUIRED_TOOLS[@]}"; do
+	if ! [ -x $(command -v "${_TOOL}") ]; then
+	    _OK=0
+	    echo "Install required tool ${_TOOL}"
+	fi
+    done
```
mikee47, line 122: This is a good idea. I'm wondering if this could be added as a general step earlier in the script as a `pre-check` for all platforms?

Sming/Arch/Esp32/Tools/install.sh
```diff
@@ -31,6 +31,29 @@ case $DIST in
...
+		echo "Install development package providing ${_INCLUDE}"
+	    fi
+	done
+	if [ $_OK != 1 ]; then
+	    echo "ABORTING"
+	    exit 1
+	fi
+	PACKAGES=()
+        ;;
+
```
mikee47, line 56: Seems reasonable, though is an awful lot of script for an edge case. How does Espressif handle non-standard installs of their IDF?

Tools/export.sh
```diff
@@ -33,14 +34,14 @@ fi
...
-export IDF_TOOLS_PATH=${IDF_TOOLS_PATH:=/opt/esp32}
+export IDF_PATH=${IDF_PATH:=${_SMIG_BASE}/esp-idf}
+export IDF_TOOLS_PATH=${IDF_TOOLS_PATH:=${_SMIG_BASE}/esp32}
 
 # Rp2040
-export PICO_TOOLCHAIN_PATH=${PICO_TOOLCHAIN_PATH:=/opt/rp2040}
+export PICO_TOOLCHAIN_PATH=${PICO_TOOLCHAIN_PATH:=${_SMIG_BASE}/rp2040}
 
 # Provide non-apple CLANG (e.g. for rbpf library)
 if [ -n "$GITHUB_ACTIONS" ] && [ "$(uname)" = "Darwin" ]; then
```
mikee47, line 47: No. Using the `Sming` install location is worse than using `/opt`. Whilst there are permissions issues with `/opt` I do believe it is the most sensible default. See https://www.baeldung.com/linux/opt-directory and https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard.

Incidentally I never install `Sming` in /opt but use sandboxes on a separate ZFS volume. I specifically *never* put bloatey tool installations (or the IDF) there. Thus the locations of toolchains should be decoupled from the location for Sming.

The procedure for customising install directories is outlined here https://sming.readthedocs.io/en/stable/getting-started/config.html. That should meet your needs. In particular, the ESP32 stuff has separate `IDF_PATH` for the framework (which needs to be writeable) and `IDF_TOOLS_PATH` for the toolchains.

.gitignore
```diff
@@ -30,3 +30,9 @@ GPATH
 .submodule
 *.built
 *.completed
+
+/esp-idf*
+/esp-quick-toolchain/
+/esp32/
+/rp2040/
```
mikee47, line 37: No. See below for `export.sh`

.gitignore
```diff
@@ -30,3 +30,9 @@ GPATH
...
+/samples/Basic_Serial/files/
```
mikee47, line 38: If it bothers you. But this should be in `/samples/Basic_Serial/.gitignore`.

Tools/install.sh
mikee47: AND provide some generic script functions which can be called from the esp32 install.sh with a list.

mikee47 (changes requested):
Thank you for this. I do have some reservations though...

mikee47:
> I tried to install Sming on my OpenSuse Tumbleseed system. For one, I did not like the forced usage of "sudo" without knowing what was going on (I had a wrapper installed that provided "apt", using zypper in the background).

Well, it's not really forced as you get a prompt to enter password. But yes I agree this aspect could be improved. Maybe just add some more messages to let the user know what's going on?

Tools/install.sh
berhoel: I will rework this.

Tools/install.sh
berhoel: I will look into implementing the `pre-check` function.

Tools/export.sh
berhoel: I reverted these changes. 

I handle the non-standart tools installation using `direnv` now.

With Sming checked out to my home directory, and adding a small `.envrc` to this directory (and every directory with projects using Sming):
```
layout Sming
```
and adding
```
layout_Sming() {

    _SMIG_BASE=${HOME}/Sming
    _SMIG_TOOLS=${_SMIG_BASE}/.tools
    export ESP_HOME=${_SMIG_TOOLS}/esp-quick-toolchain
    export IDF_PATH=${_SMIG_TOOLS}/esp-idf
    export IDF_TOOLS_PATH=${_SMIG_TOOLS}/esp32
    export PICO_TOOLCHAIN_PATH=${_SMIG_TOOLS}/rp2040

    . ${_SMIG_BASE}/Tools/export.sh
}
```
to `~/.config/direnv/direnvrc` which handles the required path settings.

.gitignore
berhoel: reverted

.gitignore
berhoel: Moved

berhoel:
> Also, I note there are no additional checks for rp2040 / esp8266 ? Does this mean they install as-is or that you don't use them?

Both rp2040 and esp8266 install as they are when ESP_HOME and PICO_TOOLCHAIN_PATH are set to writable paths.

Sming/Arch/Esp32/Tools/install.sh
berhoel: The environment variables `ESP_HOME`, `IDF_PATH`, and `IDF_TOOLS_PATH` should be used for non-standard installs of Expressif IDF (ESP32 and ESP8266).

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
     fonts-droid-fallback \
     )
 
+check_for_installed_tools () {
```
mikee47, line 25: I'm not keen on the location of these two functions as they split the declaration of `FONT_PACKAGES` from the `EXTRA_PACKAGES` and `OPTIONAL_PACKAGES`. Maybe move them to the end of the script?

Tools/install.sh
```diff
@@ -99,12 +139,20 @@ elif [ -n "$(command -v dnf)" ]; then
     DIST=fedora
     PKG_INSTALL="sudo dnf install -y"
 else
+    TOOLS_MISSING=0
```
mikee47, line 142: Remove

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+            echo "Install required FILE ${FILE}"
+        fi
+    done
+    if [ $FILES_MISSING != 0 ]; then
+    echo "ABORTING"
+    if [ $sourced = 1 ]; then
+        return 1
+    else
+        exit 1
+    fi
```
mikee47, line 61: Duplicate code, move this into helper function, perhaps `abort()` ?  (Unless that's a reserved word? I'm not a bash expert!)

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+            echo "Install required tool ${TOOL}"
+        fi
+    done
+    if [ $TOOLS_MISSING != 0 ]; then
+    echo "ABORTING"
+    if [ $sourced = 1 ]; then
+        return 1
+    else
+        exit 1
+    fi
```
mikee47, line 41: Duplicate code - see line 56

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
     fonts-droid-fallback \
     )
 
+check_for_installed_tools () {
+    REQUIRED_TOOLS="${@}"
+    echo -e "Checking for installed tools.\nrequirded tools: ${REQUIRED_TOOLS}"
```
mikee47, line 27: `requirded` -> `Required`

Sming/Arch/Esp32/Tools/install.sh
```diff
@@ -31,6 +31,13 @@ case $DIST in
     darwin)
         ;;
 
+    *)
+        check_for_installed_tools dfu-util bison flex gperf
+
```
mikee47, line 36: Blank line a bit weird

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+        return 1
+    else
+        exit 1
+    fi
+    fi
+}
+
+check_for_installed_files () {
+    REQUIRED_FILES="${@}"
+    echo -e "Checking for installed files.\nrequirded files: ${REQUIRED_FILES}"
```
mikee47, line 47: `requirded` -> `Required`

Tools/install.sh
mikee47, line 25: `check_for_installed_tools () {` -> `check_for_installed_tools() {` (no space before paretheses)

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+    echo "ABORTING"
+    if [ $sourced = 1 ]; then
+        return 1
+    else
+        exit 1
+    fi
+    fi
+}
+
+check_for_installed_files () {
```
mikee47, line 45: `check_for_installed_files () {` -> `check_for_installed_files() {` (no space before paretheses)

Tools/install.sh
mikee47: This is just a style thing but does `${REQUIRED_TOOLS}` have any advantages over `$REQUIRED_TOOLS`? I'm looking at consistency here.

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+check_for_installed_tools () {
+    REQUIRED_TOOLS="${@}"
+    echo -e "Checking for installed tools.\nrequirded tools: ${REQUIRED_TOOLS}"
+    TOOLS_MISSING=0
+    for TOOL in ${REQUIRED_TOOLS}; do
+        if ! command -v "${TOOL}" > /dev/null ; then
+            TOOLS_MISSING=1
+            echo "Install required tool ${TOOL}"
+        fi
+    done
```
mikee47, line 34: `${REQUIRED_TOOLS}`, `${TOOL}` vs `$REQUIRED_TOOLS`, `$TOOL` ?

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+check_for_installed_files () {
+    REQUIRED_FILES="${@}"
+    echo -e "Checking for installed files.\nrequirded files: ${REQUIRED_FILES}"
+    FILES_MISSING=0
+    for FILE in ${REQUIRED_FILES}; do
+        if ! [ -f "${FILE}" ]; then
+            FILES_MISSING=1
+            echo "Install required FILE ${FILE}"
+        fi
+    done
```
mikee47, line 54: `${REQUIRED_FILES}`, `${FILE}` vs `$REQUIRED_FILES`, `$FILE` ?

Tools/install.sh
```diff
@@ -183,9 +231,11 @@ case $DIST in
 
 esac
 
+if [ $(/usr/bin/python -c "import sys;print(sys.version_info[0])") != 3 ]; then
```
mikee47, line 234: shellcheck: `SC2046 (warning): Quote this to prevent word splitting.`

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
     fonts-droid-fallback \
     )
 
+check_for_installed_tools () {
+    REQUIRED_TOOLS="${@}"
```
mikee47, line 26: shellcheck: `SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.`

Tools/install.sh
```diff
@@ -22,6 +22,46 @@ FONT_PACKAGES=(\
...
+    if [ $sourced = 1 ]; then
+        return 1
+    else
+        exit 1
+    fi
+    fi
+}
+
+check_for_installed_files () {
+    REQUIRED_FILES="${@}"
```
mikee47, line 46: shellcheck: `SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.`

mikee47 (changes requested):
Sorry for being finicky, but this is good work. Thank you.

Tools/install.sh
berhoel: End of script does not work, but I moved the code down in the file.

Tools/install.sh
berhoel: Curly braces are not making a difference for the above use case, but are required in other cases:

- accessing arrays
- using substitution functions
- combining variable content with fixed value ("${foo}bar").

I am used to using them everywhere for consistency :-). But yes here it's general use is inconsistent.

berhoel:
> Sorry for being finicky, 

That's what code reviews are for :+1: 

> but this is good work. Thank you.

mikee47 (approved):

PR #2893, opened 2024-10-22
Allow installation on Ubuntu 24.04 

slaff:
Ubuntu 24.04 is not allowing the overwriting of system python packages.

This PR creates a virtual environment and activates it to avoid conflicts with the python packages coming from the OS.

### DIFF
--- a/Tools/install.sh
+++ b/Tools/install.sh
@@ -254,7 +254,7 @@ fi
 
 PYTHON_INSTALL_CMD="python3 -m pip install --upgrade pip protobuf -r \"$SMING_HOME/../Tools/requirements.txt\""
 
-eval "$PYTHON_INSTALL_CMD" || install_venv "$PYTHON_INSTALL_CMD"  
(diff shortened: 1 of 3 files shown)


Tools/install.sh
```diff
@@ -243,7 +252,9 @@ if [ -f "/usr/bin/clang-format-8" ]; then
     sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-8 100
 fi
 
-python3 -m pip install --upgrade pip protobuf -r "$SMING_HOME/../Tools/requirements.txt"
+PYTHON_INSTALL_CMD="python3 -m pip install --upgrade pip protobuf -r \"$SMING_HOME/../Tools/requirements.txt\""
```
mikee47, line 255: Should `protobuf` be added to `requirements.txt` ?

slaff: Yep, seems reasonable to me. I will commit the change shortly.

PR #2897, opened 2024-10-25
update mqttClient to catch empty clientNames

pljakobs:
catch situations where mqttClient::connect is called with an empty clientName that would otherwise lead to a failing malloc

### DIFF
--- a/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
+++ b/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
@@ -206,7 +206,7 @@ bool MqttClient::connect(const Url& url, const String& clientName)
 		return false;
 	}
   if(clientName==""){
-    debug_e("no client name configured");
+    debug_e("clientName cannot be empty");
     return false;
   }
     
--- a/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
+++ b/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
@@ -199,11 +199,17 @@ bool MqttClient::setWill(const String& topic, const String& message, uint8_t fla
 bool MqttClient::connect(const Url& url, const String& clientName)
 {
 	this->url = url;
+
 	bool useSsl{url.Scheme == URI_SCHEME_MQTT_SECURE};
 	if(!useSsl && url.Scheme != URI_SCHEME_MQTT) {
 		debug_e("Only mqtt and mqtts protocols are allowed");
 		return false;
 	}
+  if(clientName==""){
+    debug_e("no client name configured");
+    return false;
+  }
+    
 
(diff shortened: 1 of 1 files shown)


slaff:
@pljakobs make sure to fix the coding style.

pljakobs:
> @pljakobs make sure to fix the coding style.

when I run "make cs" in $SMING_HOME, I get a huge changeset with all sorts of files that I didn't touch but my four lines look unchanged - I think the way cs is set up right now, it wants to change the tab size across the whole project. I would rather not submit that large a change.

```
diff --git a/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp b/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
index be7e8896..0d9bd1bc 100644
--- a/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
+++ b/Sming/Components/Network/src/Network/Mqtt/MqttClient.cpp
@@ -199,11 +199,16 @@ bool MqttClient::setWill(const String& topic, const String& message, uint8_t fla
 bool MqttClient::connect(const Url& url, const String& clientName)
 {
        this->url = url;
+
        bool useSsl{url.Scheme == URI_SCHEME_MQTT_SECURE};
        if(!useSsl && url.Scheme != URI_SCHEME_MQTT) {
                debug_e("Only mqtt and mqtts protocols are allowed");
                return false;
        }
+       if(clientName == "") {
+               debug_e("clientName cannot be empty");
+               return false;
+       }
 
        if(getConnectionState() != eTCS_Ready) {
                close();
```
If I use eight character indents on those four lines alone they stick out like a sore thumb.

mikee47:
Needs to be `clang-format` version 8, I suspect you're running a more recent version...

mikee47:
It's also a good idea to create a new branch for PR's (i.e. don't use develop directly). See https://sming.readthedocs.io/en/latest/_inc/CONTRIBUTING.html

pljakobs:
> Needs to be `clang-format` version 8, I suspect you're running a more recent version...
```
clang-format --version
clang-format version 18.1.8 (Fedora 18.1.8-1.fc40)
```
that's a lot more more than 8 than I would have guessed - am I missing something?

> It's also a good idea to create a new branch for PR's (i.e. don't use develop directly). See https://sming.readthedocs.io/en/latest/_inc/CONTRIBUTING.html

I know, and usually I try to.

mikee47:
The issues with differing versions of clang-format is a widely discussed issue. You can find a standalone version here https://github.com/muttleyxd/clang-tools-static-binaries

mikee47:
`CLANG_FORMAT=clang-format-8 make cs`

pljakobs:
> `CLANG_FORMAT=clang-format-8 make cs`

really? it would have been that easy? :o 

would that not be something to add the the makefile then?

PR #2904, opened 2024-10-30
HardwarePWM Esp32 - added assert for max pwm channels

pljakobs:
if called with more than the supported channels (depending on the SoC), the constructor would silently fail, yielding a misleading error message.

### DIFF
--- a/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
+++ b/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
@@ -7,7 +7,7 @@
  * pwm.h
  *
  ****/
-#include <soc/soc_caps.h>
+
 
 #pragma once
 
--- a/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
+++ b/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
@@ -1,16 +1,15 @@
-/****
- * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
- * Created 2015 by Skurydin Alexey
- * http://github.com/SmingHub/Sming
- * All files of the Sming Core are provided under the LGPL v3 license.
- *
- * pwm.h
- *
- ****/
+/****  
+ * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.  
+ * Created 2015 by Skurydin Alexey  
+ * http://github.com/SmingHub/Sming  
+ * All files of the Sming Core are provided under the LGPL v3 license.  
+ *  
+ * pwm.h  
+ *  
+ ****/  
 
-#include <soc/soc_caps.h>
+#pragma once  
 
+#include <soc/soc_caps.h>  
 
-#pragma once
-
-#define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
+#define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM  
--- a/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
+++ b/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
@@ -8,12 +8,9 @@
  *
  ****/
 
+#include <soc/soc_caps.h>
+
 
 #pragma once
 
-#ifdef SOC_LEDC_CHANNEL_NUM
 #define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
-#else
-// this should not happen if the correct esp32 includes are used, just to be absolutely sure
-#define PWM_CHANNEL_NUM_MAX 8
-#endif
--- a/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
+++ b/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
@@ -7,6 +7,7 @@
  * pwm.h
  *
  ****/
+#include <soc/soc_caps.h>
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


Sming/Arch/Esp32/Core/HardwarePWM.cpp
```diff
@@ -140,12 +140,14 @@ uint32_t maxDuty(ledc_timer_bit_t bits)
...
 HardwarePWM::HardwarePWM(uint8_t* pins, uint8_t no_of_pins) : channel_count(no_of_pins)
 {
-	debug_d("starting HardwarePWM init");
-	periph_module_enable(PERIPH_LEDC_MODULE);
-	if((no_of_pins == 0) || (no_of_pins > SOC_LEDC_CHANNEL_NUM)) {
+	assert(no_of_pins <= SOC_LEDC_CHANNEL_NUM);
+
+	if(no_of_pins == 0) {
 		return;
 	}
```
mikee47, line 147: May I suggest:

```
assert(no_of_pins > 0 && no_of_pins <= SOC_LEDC_CHANNEL_NUM);
no_of_pins = std::min(uint8_t(SOC_LEDC_CHANNEL_NUM), no_of_pins);
```

This ensures release builds are safe (`assert` won't be compiled).

The for-loop will catch `no_of_pins == 0` so the `if` check is redundant.

mikee47 (changes requested):
The definition for `PWM_CHANNEL_NUM_MAX` in `Sming/Arch/Esp32/Components/driver/include/pwm.h` should *always* be obtained from the IDF. We cannot make assumptions, please amend to:

```
include <soc/soc_caps.h>

#define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
```

Sming/Arch/Esp32/Core/HardwarePWM.cpp
pljakobs: that's a good solution. thank you.

mikee47:
@pljakobs It would be nice to have the `Basic_HwPWM` sample working for esp32. Would you like to include that in this PR?

pljakobs:
> @pljakobs It would be nice to have the `Basic_HwPWM` sample working for esp32. Would you like to include that in this PR?

I have been using just that with minor changes. Will create a new PR

Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
```diff
@@ -7,6 +7,7 @@
  * pwm.h
  *
  ****/
+#include <soc/soc_caps.h>
```
mikee47, line 10: Please get rid of chuff, just require one line in this file:

```
#define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
```

pljakobs: without the `#ifdef` even? I mean, I agree, `SOC_LEDC_CHANNEL` *should* be defined from idf

mikee47: The entire file should read:

```
/****
 * Sming Framework Project - Open Source framework for high efficiency native ESP8266 development.
 * Created 2015 by Skurydin Alexey
 * http://github.com/SmingHub/Sming
 * All files of the Sming Core are provided under the LGPL v3 license.
 *
 * pwm.h
 *
 ****/

#pragma once

#include <soc/soc_caps.h>

#define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
```

mikee47: Please make sure `#pragma once` is the first line in the file.

mikee47: You can just paste in the above text.

pljakobs: dang, hope I got it now.

mikee47: That is very weird: exactly 2 spaces added to the end of *every* line. What editor are you using?

pljakobs: that's vs-code, albeit for quick edits, I'll use vim for quick edits.

pljakobs: sometimes I feel sorry that I take so much of your time for such simple changes. 
but then I think: we're getting bugs fixed, so it's not for nothing.

mikee47: Simple changes are often the best place to spend time learning about how a particular software project works. I'm always looking to simplify and clarify wherever possible so forgive me if I seem pedantic. Something I always do is run a visual difference on a PR to check that I haven't touched unrelated lines of code, even if it's only coding style.

pljakobs: re vs-code and auto-formatter - could that be integrated into the workspace definition? It would be really helpful to have coding style enforced right there.

mikee47: Configuring vscode for clang-format-8 is very much a Sming development-only thing and users may have other coding styles to consider for their projects which we don't want to mess up.

I'll get the guidance at https://sming.readthedocs.io/en/latest/information/develop/clang-tools.html updated, but for vscode, we need three things:

1. Install `clang-format-8`. A bit tricker now as it's no longer generally available in distros. Have to say the standalone builds are a god-send!
2. Install `clang-format` extension in vscode
3. Configure extension with location of clang-format.

I think this is beyond the scope of `make ide-vscode`.

mikee47 (approved):
I've taken the liberty of fixing pwm.h. All good, thank you for the fix!

pljakobs:
> I've taken the liberty of fixing pwm.h. All good, thank you for the fix!

oh heck, was I still missing somehing? I thought I had it fixed. Thank you.

Issue #2905, opened 2024-10-30
network scanning does not find any networks on ESP32


pljakobs:
on arch Esp32, network scan does not return any networks.
Tested with Basic_WiFi, the result is ok for Esp8266 but 0 for Esp32 devices.
I have added some debug code 
```
void StationImpl::staticScanCompleted(wifi_event_sta_scan_done_t* event, uint8_t status)
{
    BssList list;
    if(status == OK) {
        debug_i("Scan completed successfully");
        if(station.scanCompletedCallback) {
            uint16_t number = event->number;
            debug_i("Number of networks found: %u", number);
            wifi_ap_record_t ap_info[number];
            uint16_t ap_count{0};
            memset(ap_info, 0, sizeof(ap_info));
            esp_err_t err;

            err = esp_wifi_scan_get_ap_records(&number, ap_info);
            if (err != ESP_OK) {
                debug_e("Failed to get AP records: %s", esp_err_to_name(err));
            }

            err = esp_wifi_scan_get_ap_num(&ap_count);
            if (err != ESP_OK) {
                debug_e("Failed to get AP count: %s", esp_err_to_name(err));
            }

            debug_i("AP count: %u", ap_count);

            // TODO: Handle hidden APs
            for(unsigned i = 0; (i < event->number) && (i < ap_count); i++) {
                debug_i("AP %u: SSID: %s, RSSI: %d", i, ap_info[i].ssid, ap_info[i].rssi);
                list.addElement(new BssInfoImpl(&ap_info[i]));
            }
            station.scanCompletedCallback(true, list);
        } else {
            debug_w("No scanCompletedCallback set");
        }

        debug_i("scan completed: %u found", list.count());
    } else {
        debug_e("scan failed %u", status);
        if(station.scanCompletedCallback) {
            station.scanCompletedCallback(false, list);
        } else {
            debug_w("No scanCompletedCallback set");
        }
    }
}
```
the resulting output is
```
5903497 Scan completed successfully
5903530 Number of networks found: 4
5904023 AP count: 0
5904108 scan completed: 0 found
```

also, on the Esp32, the scan seems to run only once while on the 8266, it runs continuously.

pljakobs:
it seems that `esp_wifi_scan_get_ap_num()` returns 0 even though there are valid networks (and I suppose APs for them)

outputing the elements of `ap_info[]` show the expected five networks, but since `ap_count` is the number used to return valid networks, zero are being returned.

[...]

PR #2906, opened 2024-10-31
fixed network scan

pljakobs:
`esp_wifi_scan_get_ap_records()` clears the ap data structure, so calling `esp_wifi_scan_get_ap_num()` *after* it will return zero. Switching the order fixes the error of Sming `WifiStation.scan()` not returning any networks on the esp32 platform

### DIFF
--- a/Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp
+++ b/Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp
@@ -420,8 +420,8 @@ void StationImpl::staticScanCompleted(wifi_event_sta_scan_done_t* event, uint8_t
 			wifi_ap_record_t ap_info[number];
 			uint16_t ap_count{0};
 			memset(ap_info, 0, sizeof(ap_info));
-			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
 			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_num(&ap_count));
+			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
 			// TODO: Handle hidden APs
 			for(unsigned i = 0; (i < event->number) && (i < ap_count); i++) {
 				list.addElement(new BssInfoImpl(&ap_info[i]));


pljakobs:
this should fix #2905

Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp
```diff
@@ -420,8 +420,8 @@ void StationImpl::staticScanCompleted(wifi_event_sta_scan_done_t* event, uint8_t
 			wifi_ap_record_t ap_info[number];
 			uint16_t ap_count{0};
 			memset(ap_info, 0, sizeof(ap_info));
-			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
 			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_num(&ap_count));
+			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
```
mikee47, line 424: The call to `esp_wifi_scan_get_ap_num` is actually redundant since it's returned from `esp_wifi_scan_get_ap_records`:

```
memset(ap_info, 0, sizeof(ap_info));
ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
// TODO: Handle hidden APs
for(unsigned i = 0; i < number; i++) {
list.addElement(new BssInfoImpl(&ap_info[i]));
}
```

pljakobs: i was wondering about this and under which conditions the number of detected SSIDs and APs are different, especially, when the number of APs would be less than the number of SSIDs, but since this is the code that the ESP_IDF suggests, I thought it was "the right way" 
I can drop the `esp_wifi_scan_get_ap_num()` call and replace `ap_count` with `number` where necessary (or actually use a variable that is a bit more descriptive)

Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp
```diff
@@ -418,12 +418,12 @@ void StationImpl::staticScanCompleted(wifi_event_sta_scan_done_t* event, uint8_t
...
 			wifi_ap_record_t ap_info[number];
-			uint16_t ap_count{0};
+
 			memset(ap_info, 0, sizeof(ap_info));
 			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
-			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_num(&ap_count));
+
 			// TODO: Handle hidden APs
-			for(unsigned i = 0; (i < event->number) && (i < ap_count); i++) {
+			for(unsigned i = 0; (i < event->number); i++) {
```
mikee47, line 426: I think this loop should use `number` instead of `event->number` as it's the more recently aquired value. Presumably these values should always be the same, so could include an `assert(number == event->number)` but I think that's overkill.

pljakobs: ack, especially as it's unclear if `esp_wifi_scan_get_ap_records` changes `number` and what it would mean if those diverge

Sming/Components/Network/Arch/Esp32/Platform/StationImpl.cpp
```diff
@@ -418,12 +418,12 @@ void StationImpl::staticScanCompleted(wifi_event_sta_scan_done_t* event, uint8_t
...
 			wifi_ap_record_t ap_info[number];
-			uint16_t ap_count{0};
+
 			memset(ap_info, 0, sizeof(ap_info));
 			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_records(&number, ap_info));
-			ESP_ERROR_CHECK(esp_wifi_scan_get_ap_num(&ap_count));
+
 			// TODO: Handle hidden APs
-			for(unsigned i = 0; (i < event->number) && (i < ap_count); i++) {
+			for(unsigned i = 0; (i < number); i++) {
```
mikee47, line 426: Please remove redundant parenthesis `(i < number)`

mikee47 (approved):
Lovely, thank you for the fix!

pljakobs:
> Lovely, thank you for the fix!

thank you for your support

PR #2908, opened 2024-10-31
Implement HardwarePWM class for Rp2040 and update Basic PWM sample

mikee47:
The `Basic_HwPWM` sample should work for all architectures. Esp32 is fixed but Rp2040 requires an implementation so may as well do that here.

Note that as with the Esp32, the Rp2040 PWM hardware is much more capable than the `HardwarePWM` class allows, but it is an easy way to port existing code and get started.

TODO:

* [x] Come up with suitable list of pins for testing. Probably needs a different list for each chip: LED pin will only be best-guess as massive variation with dev-boards.
* [x] Provide Rp2040 hardware PWM implementation and verify on hardware


mikee47:
@pljakobs This PR updates the Basic_HwPWM sample which I know you were intending to do, but wanted to get Rp2040 support in there first. Would you have a look and let me know if you'd like any changes?

Sming/Arch/Rp2040/Core/HardwarePWM.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+	maxduty = top;
+}
+
+void HardwarePWM::update()
+{
+	// Not implemented
+}
+
+uint32_t HardwarePWM::getFrequency(uint8_t pin) const
```
mikee47, line 129: I note we can get the frequency for an individual pin, added for esp32 support I believe, but no corresponding `setFrequency` or `setPeriod` for a pin. We should either do that or get rid of the `pin` parameter I think.

pljakobs: the way it's currently implemented for the esp32 is that the frequency is global, not per pin.
While the esp32 platform can set the frequency per timer (of which it has up to eight, depending on the actual Soc), timers are mapped to Pins and there can be more pins than timers.
Getting the frequency for a sigle pin, with the current implementation, makes indeed no sense.
I think this is something that should be part of an extended implementation that exposes the better hardware of the esp32 and rp2040 platforms.

On the BasicHWPwm, I have something that works a bit better for me. I'll clean it up and push it in the next days (I'm travelling right now)

slaff:

mikee47:
Yes, I'm all done thanks.

Issue #2912, opened 2024-11-19
esp32 network code runs in different task


mikee47:
I'm getting unpredictable hangs and crashes with an esp32-s2 project which communicates with an SPI touch display.
The details aren't relevant as this issue applies to any resource which isn't thread-aware, so that's pretty much everything in the framework.

The problem is because Sming code is split across two tasks, not just one.
Referring to PR #2371, all code runs in the **Sming** task *except* for stuff coming in over the network, which runs in the **tiT** (tcpip) task.
The FreeRTOS scheduler runs both these tasks in a round-robin fashion.
If an SPI request is executing inside the **tiT** task (as a result of, say, an incoming websocket command request) it can get pre-empted by the higher-priority **Sming** task, resulting in a crash/hang.

The whole point of Sming is that it avoids these multitasking issues, but as previously discussed the IDF black-boxes the WiFi stack which itself depends on a FreeRTOS environment.

Possible solutions include:

1. Could we implement a context switch to the **Sming** task at a suitable point in the **tiT** task? Not sure.
2. Use a sempahore/mutex which would be used in the network stack somewhere and in the main Sming task loop.
3. Find a way to get Sming to run in the **tiT** task rather than creating its own.
4. Modify the tcpip stuff so it's polled from the main Sming loop rather than using its own task.

I modified the debug output to include the current task handle. In combination with `TaskStat` this shows the issue.

```
@@ -14,11 +14,17 @@
  *
  ****/
 
 #pragma once
 
-#include "FakePgmSpace.h"
+#include <FakePgmSpace.h>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+#pragma GCC diagnostic pop
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
@@ -74,12 +80,12 @@ extern uint32_t system_get_time();
 		m_printf(fmtbuf, __LINE__, ##__VA_ARGS__);                                                                     \
 	}))
 #else
 #define debug_e(fmt, ...)                                                                                              \
 	(__extension__({                                                                                                   \
-		PSTR_ARRAY(fmtbuf, "%u " fmt "\r\n");                                                                          \
-		m_printf(fmtbuf, system_get_time(), ##__VA_ARGS__);                                                            \
+		PSTR_ARRAY(fmtbuf, "%u %x " fmt "\r\n");                                                                          \
[...]

PR #2913, opened 2024-11-20
Fix tcpip task event handling

mikee47:
This PR addresses the issue raised in #2912 by pushing events into the tcpip thread for handling. Network applications therefore 'live' in the tcpip task and the **Sming** task is used for initialisation and the task queue.

- Add current task name to debug_x output.
- Push task callbacks to tcpip thread in network builds
- Call `init` via task queue so it gets run in correct task
- Fix stack sizes. Network builds just use default size for event task stack
- Use heap for initial partition table load. Not enough room on event task stack for this.
- Add a semaphore to `smg_uart_write` so debug output doesn't get garbled between tasks (e.g. wifi).

TODO:

- [ ] Deal with events from wifi task


mikee47:
So message handling is, I think, much improved. I've discarded the customised (hacked) event loop and reverted to the standard event loop since it's now just used for IDF stuff. Sming now uses a separate queue for `queueCallback` stuff and includes a watchdog reset in there.

I'm now using the RTC watchdog which is normally disabled after boot, but it's more convenient than a task watchdog as it can be reset from any task.

However, in idle situations the watchdog expires as there's nothing being posted to the message queue. 

If anyone has thoughts on this please comment! Here's the IDF page for reference https://docs.espressif.com/projects/esp-idf/en/v5.2.3/esp32/api-reference/system/wdts.html.

mikee47:
Also the RTC watchdog is only available for esp32 and esp32s2. I'm inclined to ditch the wdt stuff entirely...

slaff:
> Also the RTC watchdog is only available for esp32 and esp32s2. I'm inclined to ditch the wdt stuff entirely...

Just for my understanding? Are you adding the wdt reset to make it more convenient for the developers or because there is some requirement? If there is no requirement maybe we should ditch the wdt and explain in a document that it is users responsibility to use the watchdog for time-consuming tasks.

mikee47:
> > Also the RTC watchdog is only available for esp32 and esp32s2. I'm inclined to ditch the wdt stuff entirely...
> 
> Just for my understanding? Are you adding the wdt reset to make it more convenient for the developers or because there is some requirement? If there is no requirement maybe we should ditch the wdt and explain in a document that it is users responsibility to use the watchdog for time-consuming tasks.

The WDT is there to make behaviour consistent with esp8266 (and rp2040). I've gone back to the default IDF behaviour which enables the IDLE task watchdog by default. All I've done is change the timeout from 5s to 8s. When applications  call `system_soft_wdt_feed()` we yield the task rather than resets the watchdog hardware directly.

PR #2915, opened 2024-11-25
Add Message-ID header in email when not set.

slaff:
Google Mail and others started requesting Message-ID header to be present, otherwise messages are blocked/getting marked as spam.

### DIFF
--- a/Sming/Components/Network/src/Network/SmtpClient.cpp
+++ b/Sming/Components/Network/src/Network/SmtpClient.cpp
@@ -300,7 +300,7 @@ void SmtpClient::sendMailHeaders(MailMessage* mail)
 	if(!mail->headers.contains(F("Message-ID"))) {
 		Uuid uuid;
 		uuid.generate();
-		mail->headers[F("Message-ID")] =  "<" + uuid.toString() + "@" + url.Host + ">";
+		mail->headers[F("Message-ID")] = "<" + uuid.toString() + "@" + url.Host + ">";
 	}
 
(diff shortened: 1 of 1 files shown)


Sming/Components/Network/src/Network/SmtpClient.cpp
```diff
@@ -296,6 +297,12 @@ void SmtpClient::sendMailHeaders(MailMessage* mail)
 		mail->stream = std::make_unique<QuotedPrintableOutputStream>(mail->stream.release());
 	}
 
+	if(!mail->headers.contains(F("Message-ID"))) {
+		auto uuid = Uuid();
```
mikee47, line 301: `Uuid uuid;`

Sming/Components/Network/src/Network/SmtpClient.cpp
```diff
@@ -296,6 +297,12 @@ void SmtpClient::sendMailHeaders(MailMessage* mail)
 		mail->stream = std::make_unique<QuotedPrintableOutputStream>(mail->stream.release());
 	}
 
+	if(!mail->headers.contains(F("Message-ID"))) {
```
mikee47, line 300: Should we add `Message-ID` to `HTTP_HEADER_FIELDNAME_MAP` ?

slaff: https://en.wikipedia.org/wiki/Message-ID is not really related or used in HTTP.

mikee47:
Guess not. Something's broken with IDF python...

PR #2933, opened 2025-01-04
add a base hardware config without 'factory' 

pljakobs:
partition to allow two large OTA partitions

when working with OTA on the esp32c3, it became clear that `factory` would never be overwritten by OTA, making it difficult to fully upgrade the device over OTA when running an OTA version already. 
ESP-IDF suggests to drop the `factory` partition altogether to be able to configure two large OTA.
```
    factory (0x00) is the default app partition. The bootloader will execute the factory app unless there it sees a partition of type data/ota, in which case it reads this partition to determine which OTA image to boot.

        OTA never updates the factory partition.

        If you want to conserve flash usage in an OTA project, you can remove the factory partition and use ota_0 instead.
```
this hardware definition does provide a basis for a `factory` less OTA setup including the OTADATA partition the bootloader needs. For size considerations, I have moved it to the end of the 4M flash area to avoid moving around the first partition.

### DIFF
--- a/Sming/Arch/Esp32/single-rom-for-ota.hw
+++ b/Sming/Arch/Esp32/single-rom-for-ota.hw
@@ -1,0 +1,42 @@
+{
+	"name": "OTA base config with single ROM",
+	"arch": "Esp32",
+	"comment": "ota base config to eliminate the factory partition and instead build two rom partitions.",
+	"bootloader_size": "0x8000",
+	"partition_table_offset": "0x8000",
+	"devices": {
+		"spiFlash": {
+			"type": "flash",
+			"size": "4M",
+			"mode": "dio",
+			"speed": "60 if SMING_SOC=='esp32c2' else 40"
+		}
+	},
+	"partitions": {
+		"phy_init": {
+			"address": "0x00f000",
+			"size": "0x1000",
+			"type": "data",
+			"subtype": "phy"
+		},
+		"nvs": {
+			"address": "0x009000",
+			"size": "0x6000",
+			"type": "data",
+			"subtype": "nvs"
+		},
+		"rom0": {
+			"address": "0x010000",
+			"size": "0x0f0000",
+			"type": "app",
+			"subtype": "ota_0",
+			"filename": "$(TARGET_BIN)"
+		},
+		"otadata":{
+			"address":"0x3fe000",
+			"size":"8k",
+			"type":"data",
+			"subtype":"ota"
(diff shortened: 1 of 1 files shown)


mikee47:
Here's a patch you can try:

```
diff --git a/Sming/Components/Storage/Tools/hwconfig/partition.py b/Sming/Components/Storage/Tools/hwconfig/partition.py
index f7b5e312..187ab9ad 100644
--- a/Sming/Components/Storage/Tools/hwconfig/partition.py
+++ b/Sming/Components/Storage/Tools/hwconfig/partition.py
@@ -146,7 +146,11 @@ class Table(list):
             if part is None:
                 part = Entry(devices[0], name)
                 self.append(part)
-            part.parse_dict(entry, devices)
+                part.parse_dict(entry, devices)
+            else:
+                part.parse_dict(entry, devices)
+                if part.size == 0:
+                    self.remove(part)
 
     def sort(self):
         def get_key(p):
@@ -498,7 +502,7 @@ class Entry(object):
         align = self.alignment(arch)
         if self.address % align:
             raise ValidationError(self, "Offset 0x%x is not aligned to 0x%x" % (self.address, align))
-        if self.size is None or self.size == 0:
+        if self.size is None:
             raise ValidationError(self, "Size field is not set")
         if self.size % align and secure:
             raise ValidationError(self, "Size 0x%x is not aligned to 0x%x" % (self.size, align))
```

pljakobs:
sounds like a good solution.

pljakobs:
quick test shows expected result. Thank you

pljakobs:
one thought: 
setting the size to `0` is certainly good shorthand for dropping the partition within the given schema.
I would, however, prefer to see a more explicit definition here, such as
```
"partitions":{
  "factory":{
    "drop":true
  }
}
```
although I could see this to be more change than the use case justifies, especially if there is no other similar "action" that might get it's own explicit statement.

mikee47:
As there's not really any other useful interpretation of a zero-length partition I think it's probably the most appropriate choice.

The main purpose of these layered configurations is so that small changes or additions can be accommodated without having to replicate the entire schema. It also allows new architectures to be supported.

The programmer is always free to write a complete config from scratch of course.

I'll raise a PR with the change.

pljakobs:
> I'll raise a PR with the change.

I will go ahead and close this, then

mikee47:
Thank you for raising this issue!

PR #2934, opened 2025-01-07
Support dropping named partitions in hardware configs

mikee47:
This PR adds the ability to 'drop' a previously defined partition from a hardware configuration by setting `"size": 0` in an inherited config. This is silently ignored if the partition is undefined.

Issue #2933 highlighted the need to be able to eliminate the `factory` partition and use OTA partitions instead. The Esp32 `spiffs-two-roms` configuration has been modified to remove the factory partition. This was previously set to 1MByte in size, but is not required when using OTA mode.

A couple of sample applications contained redundant `subtype` overrides since these are already defined in `spiffs-two-roms`, so this has been tidied.


slaff:
@mikee47 ready with this one?

mikee47:
@slaff Yes, all done thanks.

PR #2937, opened 2025-01-24
Backported igmp_start_timer 

pljakobs:
backport from upstream lwip to fix possible division by zero error when igmp max_time == 1

### DIFF
--- a/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip.patch
+++ b/Sming/Arch/Esp8266/Components/esp-open-lwip/esp-open-lwip.patch
@@ -167,7 +167,7 @@ index eaa8dd6..6568657 100644
  #endif
  
  /**
-@@ -1992,7 +1990,7 @@
+@@ -1992,7 +1990,7 @@igmp_start_timer
   * TCP_RST_DEBUG: Enable debugging for TCP with the RST message.
   */
  #ifndef TCP_RST_DEBUG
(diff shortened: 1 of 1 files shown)


pljakobs:
note: the upstream patch has a conditional on LWIP_RAND - as far as I could see, that should always be defined in Sming.
I still decided to keep the conditionals in to not deviate from the upstream function.

mikee47:
> note: the upstream patch has a conditional on LWIP_RAND - as far as I could see, that should always be defined in Sming.
> I still decided to keep the conditionals in to not deviate from the upstream function.

Is there any reason why we wouldn't want the timer randomised? I suspect it's part of the IGMP spec. so building with `LWIP_RAND` undefined would be an error. If that is the case I'd vote for removing the conditional so if `LWIP_RAND` is undefined the compiler tells us.

mikee47:
There's a catch for this in `core/init.c` line 178 so the conditional code is effectively redundant.

pljakobs:
updated the patch, removing the conditional compilation. My intention was really to not deviate from the upstream, but thinking about it, this is  Esp8266 only code, so there is probably no situation where a new platform would change anything so fundamental as not having a random number generator defined.

mikee47 (approved):
Thank you, that's great.

slaff:
Backport from upstream lwip to fix possible division by zero error when igmp max_time == 1

PR #2944, opened 2025-02-07
Fix Formatter UTF8 encoding

mikee47:
PR #2880 added a generic `Format::escapeControls()` function to work towards providing an efficient and standard way for the framework to support escaping textual data in various ways.

This PR fixes the ASCII -> UTF8 conversion (via `Format::Option::utf8`) which was too simplistic and did not correctly handle all codepoints from 0x80 to 0xff as intended.

It also reverts the unicode output for `Format::Json::escape` as this produces incorrect output when the text is already encoded, such as with UTF8. Only the mandatory escaping is performed. Non-standard control characters below 0x20 are escaped using the unicode syntax. For example, "one\0two" is escaped as "one\u0000two".

This means applications are responsible for validating, encoding and decoding text formats as required.

Note: Prior to PR #2880 the `Format::Json::escape` call checked for invalid UTF8 and corrected it. This code was never properly validated and may not be required. If so, the issue of validating/fixing UTF8 can be revisited. It's potentially complex so I don't intend to address that here.

### DIFF
--- a/Sming/Core/Data/Format/Formatter.cpp
+++ b/Sming/Core/Data/Format/Formatter.cpp
@@ -58,7 +58,7 @@ unsigned escapeControls(String& value, Options options)
 		if(escapeChar(c, options)) {
 			extra += 1; // "\"
 		} else if(options[Option::unicode]) {
-			if(uint8_t(c) < 0x20 || (c & 0x80)) {
+			if(uint8_t(c) < 0x20) {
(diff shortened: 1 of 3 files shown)


mikee47:
I have a feeling that firefox handles invalid UTF8 more gracefully these days. I've tried some invalid JSON and cannot get it to fall over!

Issue #2943, opened 2025-02-05
Timer on ESP32 hangs the chip if used with WiFi and a too fast rate for the Timer (ex 15mS), not happening on ESP8266.


profjmer:
On a ESP8266 I have an app, that is running an FSM by a Timer each 100uSec. (running fine with: Wifi, webserver, i2c, spi ...)

On an ESP32 a single bit toggle by a Timer can work at the same rate in Basic_Blink, but in Basic_Wifi sample, a rate faster than 15mS will hang the system.

Here is an extract of the test code:

	int i;
	IRAM_ATTR void systick()
	{
		if(i++>9){
			digitalWrite(22,!digitalRead(22));  // this is an LED on my ESP32 Board TTGO V1.3
			i=0;
		}
	}

in void init()
	pinMode(22,OUTPUT);
	pulse_timer.initializeMs<15>(systick).start();

ESP8266 has only one core and is performing better with Timer and WiFi
Esp32 has two cores, but Timer seems to interfere with WiFi at a rate too fast.
Is it something that can be improved in ESP32 ? (by not using the same core that WiFi uses ? or else)

mikee47:
Fast timers are very inefficient on the esp32 due to the interrupt and software servicing overhead. Perhaps look for a way to use the hardware directly, e.g. LEDC or other modules.

Sming builds with `CONFIG_FREERTOS_UNICORE=y` by default. See #2916 et. al.

mikee47:
NB. Are you using `HardwareTimer`? This has a lower software overhead as the callback is invoked in interrupt context. Software timers use the task queue which slows things down.

mikee47:
See also https://github.com/SmingHub/Sming/issues/2445#issuecomment-2485307349

mikee47:
@profjmer Be interested to hear about any solution you come up with for this.

profjmer:
I tested different ways with no great result. As Sming runs on unicore, I
guess controlling hardware with core 1 would do a good job, but right now I
can't figure out how to compile code for core 1 and start running it from a
Sming project. I'll let you know if I find a way to achieve that.

profjmer:
I'd like to test two unicore apps but on different cores.Is it possible to
specify core on which Sming will run ? Then cores would run separately but
simultaneously.

profjmer:
Concerning hardware timers, when interrupt handler is service to often,
freertos tasker kill the app

mikee47:
[...]

PR #2949, opened 2025-03-18
Add sample to demonstrate basic dual-core operation

mikee47:
This PR adds a sample application to show how to run code on the second CPU of an ESP32, RP2040 and RP2350 devices.

Issue #2943 highlights the sort of problem which occurs with the ESP32 when attempting to use timers to run code very frequently, where the CPU time required to service an interrupt and get the code to run is too long. The problem is the overhead impose by FreeRTOS and the nature of multitasking.

A solution to this is to run time-sensitive code on the second CPU. Polled timers are a good choice here as they don't use interrupts at all.


mikee47:
@profjmer Hopefuly this sample will be of use. Is there anything in particular you would like to see covered by this sample application?

slaff:
@mikee47 do you plan to add something more to this PR?

mikee47:
@slaff No, can merge it anytime thanks.

PR #2954, opened 2025-05-09
Remove the cstdbool include for newer C++ compilers

pljakobs:
newer c++ compilers don't need to include cstdbool anymore and compiling will fail with it still in. 
This PR removes the include from host_lwip.h for compilers newer than c++17

### DIFF
--- a/Sming/Components/lwip/src/Arch/Host/include/host_lwip.h
+++ b/Sming/Components/lwip/src/Arch/Host/include/host_lwip.h
@@ -20,7 +20,10 @@
 #pragma once
 
 #include <cstdint>
+#if __cplusplus < 201703L
+// Only include for older C++ standards
 #include <cstdbool>
+#endif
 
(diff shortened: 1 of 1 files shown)


pljakobs:
I don't quite understand those fails. Are they related to the change I introduced?

mikee47:
Thank you for spotting this. We can just remove the `#include` line since C++17 is the minimum standard (https://sming.readthedocs.io/en/latest/_inc/Sming/building.html#envvar-SMING_CPP_STD). This header was originally committed as C/C++ but it only gets used internally from C++ code so C support was removed in #2472, should have been removed then!

mikee47:
> I don't quite understand those fails. Are they related to the change I introduced?

Unrelated. The `DiskStorage` library tests are causing issues with the CI system, I'll investigate that.

mikee47 (approved):

PR #2956, opened 2025-05-13
[WIP] Refactor/i hardware pwm

pljakobs:
after the discussion in #2955, I started thinking a bit more about how I could implement a richer HardwarePWM interface without breaking compatibility and while maintaining the core API in place, too.

My goal was to:

- provide a richer set of options when instantiating HardwarePWM (currently used to initialize phase shifting, spread spectrum will come later, but also configuring channel groups to timers - which gets rather close to the Esp32 ledc API)
- provide extensibility to abstract functions provided by hardware and / or the platform API and allow them to be used as part of the SMING HardwarePWM interface - those will often be hardware specific but might be implemented as hardware independent, too

I took the suggestion of extending the C header at `pwm.h` to provide the configuration option structures necessary at HardwarePWM instantiation. I'm not perfectly happy with that as it creates quite a bit of opportunity to mess up memory - and no good way to test for correct parameters. But I guess that's what we get for using C.

I have also moved the pre-existing `HardwarePWM` class to an interface only implementation `IHardwarePWM` that all hardware specific classes now inherit from. This allows me to extend the interface in the the child classes while maintaining a coherent set of base functions for any older code to work with. Code that wants to use the extended functionality will have to be platform specific.

This PR currently only implements the Hardware specific classes for Esp32 and Esp8266, CI failures for Host and RP2040 are expected. 

The main goal is to get feedback.

### DIFF
--- a/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
+++ b/Sming/Arch/Esp32/Components/driver/include/driver/pwm.h
@@ -13,3 +13,106 @@
 #include <soc/soc_caps.h>
 
 #define PWM_CHANNEL_NUM_MAX SOC_LEDC_CHANNEL_NUM
+
+/**
+ * @file pwm.h
+ * @brief PWM (Pulse Width Modulation) configuration structures and enumerations for ESP32.
+ */
+
+/**
+ * @enum PWM_phase_shift_mode
+ * @brief Defines the modes for phase shifting in PWM.
+ * 
+ * @var PWM_PHASE_OFF
+ * No phase shifting.
+ * 
+ * @var PWM_PHASE_CUSTOM
+ * Use custom per-channel phase delay values.
+ * 
+ * @var PWM_PHASE_AUTO
+ * Evenly distribute phases across the PWM period.
+ */
+enum PWM_phase_shift_mode {
+    PWM_PHASE_OFF,           // No phase shifting
+    PWM_PHASE_CUSTOM,        // Use custom per-channel values
+    PWM_PHASE_AUTO    // Evenly distribute phases across period
+};
+
+/**
+ * @enum PWM_spread_spectrum_mode
+ * @brief Defines the modes for spread spectrum in PWM.
+ * 
+ * @var PWM_SPREAD_OFF
+ * No spread spectrum.
+ * 
+ * @var PWM_SPREAD_CUSTOM
+ * Use custom per-channel frequency offset values.
+ * 
+ * @var PWM_SPREAD_AUTO
+ * Automatically apply spread spectrum with a specified percentage.
+ */
+enum PWM_spread_spectrum_mode {
+    PWM_SPREAD_OFF,
+    PWM_SPREAD_CUSTOM,
+    PWM_SPREAD_AUTO
+};
+
+/**
+ * @struct PWM_phase_shift
+ * @brief Configuration structure for PWM phase shifting.
+ * 
+ * @var PWM_phase_shift::mode
+ * Specifies the phase shift mode. See @ref PWM_phase_shift_mode.
+ * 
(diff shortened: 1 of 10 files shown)


mikee47:
My thoughts on the code as it stands:

- Be careful with over-commenting. For instance, only `IHardwarePWM` interface needs to be documented, adding comments to the implementations will result in increased maintenance burden, risk of inconsistency, etc.
- Now I'm looking at it, I think `driver/pwm.h` is the wrong place to add C++ interface definitions. I'd only put stuff there if it was available in the lower-level driver, but for the Esp32 that's part of the IDF. Just put these definitions directly into `Arch/Esp32/Core/HardwarePWM.h`.

Given that you are currently focused on the Esp32 capability, I think it would be better to create a separate library first. There are a number of advantages to this:

- Trying to develop new capabilities *and* integrated it in one go is a bit much
- Only need to consider Esp32 capabilities, just mark it as Esp32-only (`COMPONENT_SOC=esp32*`)
- Start by copying the existing Esp32 code
- Get the code running as *you* require rather than trying to do everything all-at-once

Once the library is stable and works as you like, we then have something concrete to consider migrating into the main Sming codebase. The `HardwareSPI` library may serve as an example of how it could be structured. The object structure is a bit convoluted (with a `Controller` class inheriting from the architecture-specific `ControllerBase` class) but the main goal was to avoid code duplication without virtual classes.

pljakobs:
I have finally gotten around to implement the first bits of the new HardwarePWM class. 
PWM, Phase Shift and Spread Spectrum now work on both the ESP32 and the ESP32c3 
However, I noticed, if I put `component-soc:=esp32*` in my component build, the whole app won't build anymore for non Esp32 SOCs (specifically the Esp8266).
I am sure I'm missing something, I've placed this component into the ./Components folder and was expecting the build process would only enable it for Esp32* SOCs, instead it extends the Esp32 requirement to the whole project.

mikee47:
Ref.  https://github.com/pljakobs/Esp32HardwarePwm

> I have finally gotten around to implement the first bits of the new HardwarePWM class. PWM, Phase Shift and Spread Spectrum now work on both the ESP32 and the ESP32c3 However, I noticed, if I put `component-soc:=esp32*` in my component build, the whole app won't build anymore for non Esp32 SOCs (specifically the Esp8266). I am sure I'm missing something, I've placed this component into the ./Components folder and was expecting the build process would only enable it for Esp32* SOCs, instead it extends the Esp32 requirement to the whole project.

Yes, this is by design. You'll need some logic in your project's `component.mk` to conditionally import the library:

```make
ifeq ($(SMING_ARCH),Esp32)
COMPONENT_DEPENDS += Esp32HardwarePwm
endif
```

Your project code will also need to select the appropriate PWM class and conditional compilation when calling those supplementary methods, e.g.:

```c++
#ifdef ARCH_ESP32
#include <Esp32HardwarePwm.h>
Esp32HardwarePwm pwm;
#else
HardwarePWM pwm;
#endif
```

I omitted to document the `COMPONENT_SOC` setting anywhere - sorry, will fix that. Was added in #2421. There are several reasons for this setting:

1. If the Component doesn't support an architecture then the error message is specific, rather than just having random build failures which leave the user none the wiser.
2. To support integration testing, the special build target `sample` prints a message and completes *without* error if the current project is unsupported for a given SOC. For example, the `samples/WifiSniffer` is an Esp8266-only sample so `make SMING_SOC=rp2040` fails with an error, but `make SMING_SOC=rp2040 sample` does not.
3. Supports the informational `make list-soc` build target.

pljakobs:
thanks @mikee47 - I finally figured out that I needed a conditional in the makefile itself.

on the library itself - I have it working with phase shift and interrupt driven spread spectrum now, will look into implementing fades and make it a bit more pretty, then I'll submit it to your srutiny ;-)

Issue #2971, opened 2025-09-13
Services/Profiling/CpuUsage doesn't work on esp32c3


pljakobs:
It seems that, on the 32c3, cpuUsage::init(onReady) never calls back to onReady().
I have tested samples/Basic_Tasks on both the 8266 and the 32c3, and while the 8266 works as expected, 

esp8266:
```
Available heap: 42440
mode : null
Calibrating CPU usage...mode : sta(84:f3:eb:65:bf:95)
add if0
sampleIntervalTicks = 500
14143, 1006, 172, 240, 334, 2551 - 117 missed
13749, 1083, 186, 304, 411, 2445 - 114 missed
13435, 1087, 170, 204, 316, 2217 - 115 missed
14135, 1054, 200, 328, 792, 2416 - 114 missed
13816, 1061, 165, 223, 353, 2421 - 115 missed
13326, 1070, 186, 262, 327, 2321 - 114 missed
count = 4002, hwTimerCount = 4002, av. interval = 0, maxTasks = 3, CPU usage = 89.56%
13699, 1039, 446, 506, 382, 2429 - 118 missed
13827, 1018, 192, 293, 381, 2744 - 114 missed
13356, 1066, 195, 216, 320, 2191 - 115 missed
Loop Times: count=48345, total=435ms, min=0ms, max=0ms, average=0ms
13871, 927, 241, 317, 650, 2733 - 116 missed
13859, 1014, 211, 254, 354, 2511 - 115 missed
13270, 1088, 195, 243, 336, 2375 - 113 missed
13621, 1042, 209, 248, 351, 2543 - 112 missed
```

esp32c3:
```
I (230) cpu_start: Min chip rev:     v0.3
I (235) cpu_start: Max chip rev:     v1.99 
I (240) cpu_start: Chip rev:         v0.3
I (245) heap_init: Initializing. RAM available for dynamic allocation:
I (252) heap_init: At 3FC8EF20 len 000310E0 (196 KiB): RAM
I (258) heap_init: At 3FCC0000 len 0001C710 (113 KiB): Retention RAM
I (265) heap_init: At 3FCDC710 len 00002950 (10 KiB): Retention RAM
I (272) heap_init: At 50000014 len 00001FD4 (7 KiB): RTCRAM
I (279) spi_flash: detected chip: generic
I (283) spi_flash: flash io: dio
I (288) sleep: Configure to isolate all GPIO pins in sleep state
[...]

PR #2972, opened 2025-09-13
Fix Esp32 issue with Sming tasks not getting handled at startup

mikee47:
Issue #2971 revealed a problem with the task queue where the `onReady` callback never gets called.
The callback gets queued but the event loop is created afterwards and lacked a 'kick' to the tcpip thread to start message processing.

This bug has been present since PR #2913.


mikee47:
@slaff The CI failures in jerryscript really need it updating to a more recent version. I'm prepping v3 for that, will have a PR for you shortly.

PR #2974, opened 2025-09-15
Fix build errors in MacOS (jerryscript)

mikee47:
This PR provides some maintenance fixes for the follow issues with jerryscript:

- Patch CMake minimum version < 3.10: CI build is failing on MacOS
- Fix missing IMPORT_FSTR data building with GCC 15.2
- `day_names_p` and `month_names_p` ill-defined

MacOS CI runner latest now at version 15 (Sequoia) which uses clang 18 (up from 15).

TODO:

* [x] Find out why `rbpf` library is failing. Looks probable that we need to use a newer version of llvm, version 15 isn't found in MacOS runner. `macos-latest` changed from version 14 to version 15 3 weeks ago.

### DIFF
--- a/Tools/export.sh
+++ b/Tools/export.sh
@@ -53,6 +53,6 @@ export PICO_TOOLCHAIN_PATH=${PICO_TOOLCHAIN_PATH:=$SMING_TOOLCHAINS/rp2040}
 
 # Provide non-apple CLANG (e.g. for rbpf library)
 if [ -n "$GITHUB_ACTIONS" ] && [ "$(uname)" = "Darwin" ]; then
-CLANG="$(brew --prefix llvm@15)/bin/clang"
+CLANG="$(brew --prefix llvm@18)/bin/clang"
 export CLANG
 fi


slaff:
> Switch back to the “macos-14” label by changing workflow YAML to use runs-on: macos-14. We support the two latest stable macOS versions plus latest beta, so macOS 14 will be maintained for the next year.

@mikee47 We can switch this build to macos-14 instead of the new latest.

mikee47:
~~Looks like `rbpf` library needs updating...~~ Nope, just llvm version or MacOS runner version.

```
** Building Libraries/rbpf/samples/Basic_VM **
Using configuration 'standard'
Installing Python requirements...
make[4]: /opt/homebrew/opt/llvm@15/bin/clang: Command not found
```

mikee47:
> > Switch back to the “macos-14” label by changing workflow YAML to use runs-on: macos-14. We support the two latest stable macOS versions plus latest beta, so macOS 14 will be maintained for the next year.
> 
> @mikee47 We can switch this build to macos-14 instead of the new latest.

Missed this! I'll let the current CI run to see what happens using llvm ~~16~~ 18.

PR #2986, opened 2026-01-16
Add python `Evaluator` module

pljakobs:
Introduce an 'Evaluator' Class to Tools/Python/evaluator.py designed to evaluate logical and mathematical expressions in various build-time configurations like ConfigDB

Key features:
- implements a set of arithmetic functions (+,-,*,/, pow2, log2)
- implements logical operations (&&, ||, ! and !=)
- implements min/max and align_up functions to calculate hardware- specific values such as "how much flash do I need to store x bytes if flash block size is 4096


mikee47 (approved):
@pljakobs Are you happy with the changes I've made? The two main changes are:

- Allow custom variable resolution using `get_variable` callback
- Propagate exceptions rather than returning a string. That might be OK in a CLI but not for use in other code. We cannot know that legitimate return values don't start with 'Error: ' for example.

We may need other features in the future but I think this covers the basics.

pljakobs:
looks good to me. Thanks for your work!

PR #2992, opened 2026-01-31
Update littlefs to v2.11.2

mikee47:
This September 2025 update includes multiple bug fixes, since v2.9.3 was released in April 2024.

A forked version of littlefs is still used, but has been revised to reduce the number of changes and improve maintainability. The changes are all internal and should not affect applications.

This PR also includes a minor update to IFS so that directory timestamps are copied when building a filesystem image.
Previously they defaulted to the build time.


slaff:
@mikee47 when you are ready with this PR change it from draft to normal PR so that I can merge it.

PR #2995, opened 2026-02-03
Fix: call esp_netif_dhcps_stop before destrying the interface 

pljakobs:
when stopping the AccessPoint

originally, stopping the Access Point just called `esp_netif_destroy()`. If a dhcp package was queued at that point, the dhcp server would still process it, leading to a use after free condition in lwip.

### DIFF
--- a/Sming/Components/Network/Arch/Esp32/Platform/AccessPointImpl.cpp
+++ b/Sming/Components/Network/Arch/Esp32/Platform/AccessPointImpl.cpp
@@ -52,6 +52,7 @@ void AccessPointImpl::enable(bool enabled, bool save)
 			break;
 		}
 		if(apNetworkInterface) {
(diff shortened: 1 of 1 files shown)


slaff:
> Fix: call esp_netif_dhcps_stop before destrying the interface

Please, submit in this PR ONLY the changes related to this fix.

pljakobs:
sorry, my fault, I had pushed a working branch.
Btw, thinking about this fix: my issues came from the dhcp code procesing a dhcp package after the interface had been killed - that's what this fix addresses.
I've been wondering if this should clean up more than that, after all, there might also be cases where http, ws, ntp or any other sort of packets are lingering - but I believe that that should be the task of the application to clean up - the dhcp server however is started by the framework, thus it makes sense for the framework to clean it up.

PR #2997, opened 2026-02-05
fix request method being 0 or undef

pljakobs:
moved request.setMethod from `onMessageBegin` (before parsing the request line supposedly) to `onHeadersComplete`


Sming/Components/Network/src/Network/Http/HttpServerConnection.cpp
```diff
@@ -29,7 +29,7 @@ int HttpServerConnection::onMessageBegin(http_parser* parser)
 
 	// ... and Request
 	request.reset();
-	request.setMethod(HttpMethod(parser->method));
+	//request.setMethod(HttpMethod(parser->method));
```
slaff: Please, remove completely the commented line above.

PR #3003, opened 2026-02-08
Evaluator improvements

mikee47:
- Fix handling expressions such as "!value". Substitution converts this to " not value" and space at beginning needs stripping
- Support lists and sets
- Pull evaluator README into main documentation tree and reference directly

### DIFF
--- a/Tools/Python/evaluator.py
+++ b/Tools/Python/evaluator.py
@@ -142,6 +142,9 @@ class Evaluator:
         if isinstance(node, ast.List):
             return [self._eval(v) for v in node.elts]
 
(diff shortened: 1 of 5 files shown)


mikee47:
@pljakobs Anything else you'd find useful in the evaluator?

pljakobs:
not at this time, it does anything I need in ConfigDB.
If it will be enabled in .hw files, hex/dec conversion would probably be useful but beyond that, I currently lack the imagination what else could be useful.

mikee47:
> If it will be enabled in .hw files, hex/dec conversion would probably be useful...

I'll take a look at that separately

PR #3004, opened 2026-02-09
Abstracted some complexity related to WebcamStream.

slaff:
Every camera should be able to report its frames per second.


Sming/Libraries/WebCam/src/WebcamStream.h
```diff
@@ -12,68 +12,55 @@
...
+ * @brief Webcam stream producer for multipart streaming
+ */
+class WebcamStream : public MultipartStream
 {
 public:
-	/**
-	 * @param camera pointer to the camera object.
-	 */
-	WebcamStream(CameraInterface* camera, size_t blockSize = 512): camera(camera), blockSize(blockSize)
+	WebcamStream(CameraInterface* camera) : MultipartStream(std::bind(&WebcamStream::produce, this)), camera(camera)
```
mikee47: `CameraInterface& camera`

Sming/Libraries/WebCam/src/WebcamPictureStream.h
```diff
@@ -0,0 +1,73 @@
...
+#include <Data/Stream/DataSourceStream.h>
+#include "Camera/CameraInterface.h"
+
+class WebcamPictureStream : public IDataSourceStream
+{
+public:
+	/**
+	 * @param camera pointer to the camera object.
+	 */
+	WebcamPictureStream(CameraInterface* camera, size_t blockSize = 512) : camera(camera), blockSize(blockSize)
```
mikee47: If `camera` cannot be null, and we don't own it, then make it a `CameraInterface&` reference

Sming/Libraries/WebCam/src/Camera/FakeCamera.h
mikee47: Should be:

```
FakeCamera(const Vector<String>& files) : files(files)
{
}
```

This is also inefficent as the entire vector is copied. How about adding/replacing it with a default constructor and adding:

```
void addImage(const String& filename)
{
  files.add(filename);
}
```

Then you don't need to use `FakeCamera*` int the WebcamServer app, just instantiate it.

Sming/Libraries/WebCam/src/Camera/FakeCamera.h
```diff
@@ -15,10 +15,9 @@
 #include "CameraInterface.h"
 #include <Data/Stream/FileStream.h>
 
-class FakeCamera: public CameraInterface
+class FakeCamera : public CameraInterface
 {
 public:
-
 	/**
 	 * Sets the list of all images that should be used to roll over.
```
mikee47: `@brief`

Sming/Libraries/WebCam/src/Camera/CameraInterface.h
mikee47: `@brief`

Sming/Libraries/WebCam/src/Camera/FakeCamera.h
mikee47: Yikes. Doesn't belong here. What if the application wants to use a different filesystem arrangement?

mikee47: If there are no images in the `files` vector this will go bang.

samples/WebcamServer/app/application.cpp
mikee47, line 85: I needed this as well:

```
```

mikee47 (commented):
I'm not getting any images through on the web interface. Tried in in host and on esp8266...

samples/WebcamServer/app/application.cpp
slaff: > I'm not getting any images through on the web interface. Tried in in host and on esp8266...

Strange. I am able to get images using Host and Esp8266 MCU. Without making changes to the WIFI setup.

mikee47: I tried with my phone, and it works. But not with firefox on Fedora desktop.

mikee47: So with it running on the phone, as soon as I try to view on desktop the phone one stops.

mikee47: So I installed chromium and that works. But can only stream to one client at a time, even with framerate reduced to 5.

slaff: > But not with firefox

I was able to reproduce the issue with Firefox. Have to see what is needed by Firefox...

>  But can only stream to one client at a time,

Yes, this is a known problem. The camera has only one stream... I can probably rework it so that one camera can produce multiple streams. Basically a camera has only one image active at a time. Multiple streams linked to this image represent the reading progress from the different browser connections.

slaff: >> But not with firefox

> I was able to reproduce the issue with Firefox. Have to see what is needed by Firefox...

The Firefox issue should be fixed now.

mikee47: Yes, that's working now.

mikee47: Without the above WiFi setup, the esp8266 only works in AP mode. (Try `make flashinit && make flash`)

slaff: > But can only stream to one client at a time,

Now you should be able to have multiple clients running at the same time...

mikee47: The frame rate regulation doesn't seem to work for me. Either I get full frame rate or 1 frame per second. Also seems like a use for timer classes - I've been playing around with this https://github.com/mikee47/Sming/tree/feature/webcam-fps.

slaff: Thank you @mikee47. I have merged your changes into this PR. Will test it later today.

slaff: > The frame rate regulation doesn't seem to work for me.

Tried with your timer code which is integrated into this PR. The visible frame rate still does not match the expected :(.

PR #3008, opened 2026-02-13
Fix very slow `IFS::Debug::listDirectory` operation on LittleFS

mikee47:
Retrieving free space involves filesystem traversal which can be very slow. We only want the filesystem type for directory listings so add a `basicOnly` option for `FileSystem::getinfo()`.

This defaults to false so no change to default behaviour.


slaff:

PR #3009, opened 2026-02-13
Fix nanopb build issues

mikee47:
Fix random CI build failures associated with nanopb.

Updates the `nanopb` library to export `NANOPB_GENERATE` for use by e.g. `Sming-GoogleCast` library.


mikee47:
@slaff Pushed a small fix to GoogleCast so library builds work as expected

PR #3015, opened 2026-04-17
New module: Esp32HardwarePwm

pljakobs:
## Summary
This PR adds Esp32HardwarePwm, a comprehensive C++ wrapper for the ESP32 LEDC PWM peripheral, to Arch/Esp32/Components

## Motivation
The ESP32 LEDC hardware is significantly more capable than the ESP8266 PWM — it supports up to 20-bit resolution, frequencies up to 40 MHz, hardware-accelerated linear fading, and phase shifting across multiple independent timer+channel groups. The pre-existing HardwarePWM API surface does not expose these capabilities on Esp32. This component provides a native Esp32-specific driver that takes full advantage of the hardware.

## Features

- Multiple independent instances — each with its own timer configuration (frequency, resolution, speed mode, timer index); multiple instances coexist by using non-overlapping channel ranges
- Flexible duty cycle control — both channel-indexed and legacy pin-indexed analogWrite API; configurable 1–20-bit resolution
- Phase shifting — OFF, AUTO (evenly distributed), or MANUAL (per-pin hpoint values) to reduce inrush current spikes and EMI
- Spread spectrum — optional frequency dithering around the base frequency for EMI reduction
- Hardware fade — hardware-accelerated linear transitions via the LEDC fade engine; ISR installed lazily on first use
- Per-channel fade queue — chain multiple fades per channel without application polling:
  - FIFO mode: plays entries once in order, auto-starts on first enqueue, fires onQueueEmpty callback when drained
CYCLIC mode: loops the entry set endlessly, fires onCyclicWrap each loop, requires explicit startQueue() after seeding
  - Per-channel configurable queue depth (default 10, uint16_t); runtime resize while queue is empty
  - Three callback slots: onFadeDone, onQueueEmpty, onCyclicWrap — dispatched via System.queueCallback (task context, ISR-safe)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -433,6 +433,10 @@
 	path = Sming/Libraries/BLEKeyboard/ESP32-BLE-Keyboard
 	url = https://github.com/T-vK/ESP32-BLE-Keyboard.git
 	ignore = dirty
+[submodule "Libraries.Esp32HardwarePwm"]
+	path = Sming/Libraries/Esp32HardwarePwm
+	url = https://github.com/pljakobs/Esp32HardwarePwm
+	ignore = dirty
 [submodule "Libraries.esp-nimble-cpp"]
 	path = Sming/Libraries/NimBLE/esp-nimble-cpp
 	url = https://github.com/h2zero/esp-nimble-cpp.git
@@ -451,7 +455,3 @@
 # END OF .gitmodules
 #
 #################################################################################################
-[submodule "Sming/Arch/Esp32/Components/Esp32HardwarePwm"]
-	path = Sming/Arch/Esp32/Components/Esp32HardwarePwm
-	url = https://github.com/pljakobs/Esp32HardwarePwm
-	branch = devel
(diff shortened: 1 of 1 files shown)


pljakobs:
This has become pretty massive but it's also now very functionally complete. 
It covers the minimal functionality of configuring a set of channels for PWM and sentting both percentage and absolute value.
It also allows to leverage the Esp32 ledc's functionality to set a variety of PWM frequencies and bit widths (tested to 64kHz and 8Bit).
Additional features:
- automatic or manual hpoint setting to allow phase shifting channels
- spread spectrum to reduce the peak spectral load and "smear" it out
- CIE 1931 curve to allow for linear perception brightness fading
- hardware fade support
- queued fade support with Sming native callbacks
- experimental timing correction support with companion application to calculate SOC specific timing parameters

pljakobs:
the build errors are changed function names in the sample app.  fixed

mikee47:
I've pushed some changes to your repo, and within Sming this library should go into `Libraries` rather than `Components` as the latter is intended for core dependencies rather than optional libraries.

slaff:
@mikee47 are you ready with the review or you will need more time?

mikee47 (changes requested):
Just need to review/merge recommended changes in submodule, then update here.

pljakobs:
I was unsure if putting this into Libraries (which is arch-neutral) was the right way since this is ESP32 only. 
hope everything else is now fine.
There is a number of issues still open on the Library, but none seem functional/critical right now, I'll fix them asap

mikee47 (approved):

PR #3021, opened 2026-05-14
Fix HttpClient death loop using `llhttp`

mikee47:
llhttp parser requires resetting before connection attempt. Closes #3010.

To check behaviour the `HttpClient` sample has been updated to perform its post operation every 30 seconds.

Note: This doesn't actually do much to test this particular problem because the SSL session/connection appears to be persistent...

Note: The SSL stack locks up so will raise an issue for that. Behaviour is consistent with both parsers.


slaff:
Thank you @mikee47 !

PR #3030, opened 2026-07-04
Update ConfigDB library to v1.2.0

mikee47:
Many bugfixes and improvements


mikee47:
@slaff Somethings changed in MacOS jerryscript builds, I've pushed a fix https://github.com/slaff/Sming-jerryscript/pull/23

PR #3031, opened 2026-07-07
Fix jerryscript build errors for MacOS

mikee47:
This PR updates the jerryscript library


mikee47:
CI `macos-latest` label is migrating to  macos-26, completing by July 15th 2026
https://github.com/actions/runner-images/issues/14167
For MacOS CLANG version changes (from llvm@18 to llvm@20) so might be spotty until migration is complete.

